Issue #87, opened 2022-01-31
Give notifications their own icon


huumn:
Right now, it's 2 clicks to get to notifications. It should probably be just one.

https://stacker.news/items/9404

PR #131, opened 2022-04-02
Turn notifications box into a link

secondl1ght:
As mentioned in issue #87, I turned the red notifications box into a link that takes you to the notifications page. This allows the users to access notifications in only 1 click instead of 2 if they choose to.

Note I have not tested this locally yet because I am still working on getting login working in my dev environment. Its a small change and it should work, but if you could please test that would be great. I did test adding it to the dom in devtools and it does work as expected there.

image

Thanks!


huumn:
I went and made notifications its own icon. The little box is too small to be reliable on mobile and wouldn't be obvious either. 

I really hate to not use your work though and do really appreciate the effort! In the future it might be good to discuss any changes you plan to make so we can debate the approach a bit together.

PR #173, opened 2022-07-12
Update search-fill.svg

kerooke:
This image more closely resembles a magnifying glass.


huumn:
Are you making these yourself? The radius on the edges is quite round and I feel like it's be more consistent with other icons (e.g. the pin) if the edges didn't have radius. Not a huge deal. Still overall an upgrad.

PR #200, opened 2022-09-29
change drop down scroll to auto

suhailsaqan:
I changed the overflow to `auto` so that it's not always showing the scroll when it's not needed.

@huumn we discussed the issue here: https://stacker.news/items/73723

image


ekzyis:
Haha, I have the same issue but for some reason never thought about reporting or fixing it. Thank you! Now that I am super aware of it, it's really annoying, haha

PR #205, opened 2022-10-15
Use Array.from(value).length for title length validation

### LINKED ISSUE #204: Treat emoji as single character when calculating length for a title
SN has implemented an 80 character limit for the Title of a post. If I use an emoji in a title, that counts as two characters for the length.

For example, this title would trigger the hint message "1 too many":

*1234567890123456789012345678901234567890123456789012345678901234567890123456789📺*

An emoji renders wider than an ASCII character, yes, so maybe that is the reasoning for an emoji counting as two characters.  But so do other unicode characters that render wider than an ASCII character  (e.g., this alternative comma+space alternative thingy "、"), and those are counted only as a single character (i.e., not penalized).

So I don't know if the emoji counted as two characters was intentional or not, but my request is to count it as a single character.  Or if they count as two to prevent titles with many emojis rendering wider, maybe if there is only a single emoji, then that one emoji gets treated as a single character, but when the title has multiple emojis then the current behavior (treating as two characters each) is warranted.

It might seem much ado about nothing (the difference of one single character), but when posting a podcast or video where the content's original title is too long for SN, then every character counts.  So fitting in one extra character will sometimes makes all the difference. Thus I ask that this request be considered.

ekzyis:
I created a custom validation function to use a different length than `str.length` during validation to fix #204.

I also used a hacky way to still access the `length` property used during validation for the error message by passing in an object with a property `length` set to the "new" length instead of the actual string value.

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -924,7 +924,7 @@ export const updateItem = async (parent, { id, data: { title, url, text, boost, 
     throw new UserInputError(`boost must be at least ${BOOST_MIN}`, { argumentName: 'boost' })
   }
 
-  if (!old.parentId && title.length > MAX_TITLE_LENGTH) {
+  if (!old.parentId && Array.from(title ?? '').length > MAX_TITLE_LENGTH) {
     throw new UserInputError('title too long')
   }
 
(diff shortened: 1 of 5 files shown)


ekzyis:
Just to be clear: I don't mind if this does not get merged. I just wanted to provide a potential fix. I also think this helps to see what changes would be required to fix #204. It's up to you to decide what gets into SN and what not, of course :smile:

huumn:
I think having a small penalty for emojis is a good thing. I don't even like seeing them in titles personally and would (almost) prefer to prevent them. 

Thanks for the work. Going to close this one and merge your other PRs shortly.

PR #210, opened 2022-11-02
Updated Filled Lightning Bolt

kerooke:


huumn:
While it's in the repo, I don't use that svg on the site. I use bolt.svg for our lightning bolt. 

These svgs are modified/filled/colored in code, so while changing the fill properties in the svg file impacts the way the file looks in isolation, it has no carry over to the actual rendering on the site.

PR #215, opened 2022-11-16
Ignore subdomains during dupes detection

### LINKED ISSUE #169: Dup detection with www.domain.com doesn't check for domain.com and vice-versa
If a website is accessible via either `www.` subdomain or without a subdomain, SN could help lessen the chance that different users will try to post the same content by checking for `www.` subdomain as part of its dup detection.

For example, currently, the first of the following would be a duplicate of the second:
1. https://podpage.com/citadeldispatch/cd68-a-sunday-afternoon-chat-with-clarkmoody/
2. https://www.podpage.com/citadeldispatch/cd68-a-sunday-afternoon-chat-with-clarkmoody/

But posting either when the other already exists on SN will, currently, not cause SN to show the warning about it being a duplicate. 

In the above example, the first will actually get an HTTP 301 redirect which includes the `www.` subdomain.  So detecting HTTP redirects might be one way of checking that it might be a duplicate.  But some web servers are (poorly) configured to let the response be either -- i.e., without any redirect, but www.domain.com serves the same content as domain.com 

And this can happen with other subdomains as well.   For instance, a URL might see redirect to a specific language, such as what Wikipedia does: 

- https://wikipedia.org/wiki/Bitcoin
- https://en.wikipedia.org/wiki/Bitcoin

That's a particularly gnarly one.  The first will get an HTTP 301 redirect to add the `www.` subdomain, ... and the request for that then gets an HTTP 301 redirect where the language (en.) subdomain replaces the `www.` subdomain (i.e., en.wikipedia.org/... )
  
So without doing further research and analysis, I can't say what the best solution is offhand.  I do think simply stripping any `www.` prefix before doing dup detection (and indexing the URL of existing posts without the `www.` and using that for the dup detection) takes care of nearly every instance that I could envision occurring.

ekzyis:
Fixes #169

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -4,6 +4,7 @@ import serialize from './serial'
 import { decodeCursor, LIMIT, nextCursorEncoded } from '../../lib/cursor'
 import { getMetadata, metadataRuleSets } from 'page-metadata-parser'
 import domino from 'domino'
+import { parseDomain, ParseResultType } from 'parse-domain';
 import {
   BOOST_MIN, ITEM_SPAM_INTERVAL, MAX_POLL_NUM_CHOICES,
   MAX_TITLE_LENGTH, ITEM_FILTER_THRESHOLD, DONT_LIKE_THIS_COST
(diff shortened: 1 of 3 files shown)


huumn:
While I think the issue is worth fixing, I don't think we need to introduce a dependency to solve it. 

Also, there's probably 10 issues dedicated to dupe problems and it's probably best to consider them all at the same time to try to find the most concise solution. Going to close and take a stab at this.

huumn:
Aw man you already deleted it. 

This was a harder problem than I realized, requiring a list of all valid tlds lol.

ekzyis:
I pushed it again haha

ekzyis:
There is also the package [`haraka-tld`](https://github.com/haraka/haraka-tld) which contains all valid tlds.

Maybe it's better suited but it doesn't parse the domain into subdomains.

huumn:
Man you're fast. I ended up using this https://www.npmjs.com/package/tldts because it has fewer dependencies.

I also realized it needed another branch in the if statement to handle a url without a subdomain for matching against urls *with* subdomains.

```
      const parseResult = parse(urlObj.hostname)
      if (parseResult?.subdomain?.length) {
        const { subdomain } = parseResult
        console.log(subdomain)
        uri = uri.replace(subdomain, '(%)?')
      } else {
        uri = `(%.)?${uri}`
      }
```

I'll send you some sats tonight for the trouble! I appreciate it.

PR #219, opened 2022-12-11
Updated copy on reward notifications highlighting the new SN rewards page

kerooke:


components/login.js
```diff
@@ -61,7 +61,7 @@ export default function Login ({ providers, callbackUrl, error, Header }) {
       <div className={styles.login}>
         {Header && <Header />}
         <div className='text-center font-weight-bold text-muted pb-4'>
-          Not registered? Just login, we'll automatically create an account.
```
huumn, line 64: What are you thoughts on the hazard of removing this? 

This was communicating something that isn't communicated in the new copy.

Issue #223, opened 2023-01-02
Improve privacy by not accessing Coinbase API directly


cointastical:
It appears Coinbase is getting my IP address every time I access Stacker News.

Specifically a GET request is occurring for the "spot" price:
https://api.coinbase.com/v2/prices/BTC-USD/spot

I think this could be handled by proxy where the API request from the browser goes to a Stacker.News endpoint, and then Stacker.News sends an API request to Coinbase.

huumn:
True. 

I think Coinbase has API rate limits by IP.  If we want to proxy them, we'll need to cache the price on our servers to not go over their limits.

ekzyis:
Another post related to this: https://stacker.news/items/119700

Following ideas were mentioned:
1. only fetch price when required (the user is not using moscow time)
2. add "disabled" to fiat currency dropdown in settings which makes only moscow time available

This is unrelated to using a proxy. Using a proxy would still be useful, I think.

huumn:
Proxying is my preferred approach. This prevents IPs leaking for people who also want the price. It's more work though. 

For doing satistics exports for tax purposes, we also need a minute to minute price, so storing prices is something we need to do for that anyway.

I think the way this should work, is we setup a job in the job queue that runs every minute and records the price in a DB table. 

Then on the client for the price widget, we just poll the backend which retrieves the latest price from this DB table.

Given we support multiple currency pairs, we probably want to figure out how to grab all BTC-fiat pairs in a single query and store them all every minute.

huumn:
I think this query basically gives us all their exchange rates `https://api.coinbase.com/v2/exchange-rates?currency=BTC` but it includes shitcoins and not just fiat.

PR #226, opened 2023-01-12
Add proxy with cache for coinbase API requests

ekzyis:
Fixes https://github.com/stackernews/stacker.news/issues/223

This MR only adds a proxy with a cache (to not get rate-limited).

Everything else that was discussed in #223 was ignored for sake of simplicity.

These other things can be added (later) on top of this.


components/price.js
```diff
@@ -20,7 +20,7 @@ export const CURRENCY_SYMBOLS = {
   ZAR: 'R '
 }
 
-const endpoint = (fiat) => `https://api.coinbase.com/v2/prices/BTC-${fiat ?? 'USD'}/spot`
+const endpoint = (fiat) => `https://stacker.news/api/price?fiat=${fiat ?? 'USD'}`
```
ekzyis, line 23: Next.js seems to only support absolute URLs.

This means that as it is now, this will fetch data from the production API even if the API is run locally for development.

huumn:
Do you think you can roll this without a dependency? I'm trying to be increasingly cautious about deps. This introduces quite a few.

In this case, for an in-memory cache (that'll only ever have ~6 items in it), we could simply use a global `Map` object, where the currency pair is the key, and the value is an object with `{ createdAt, price }` and we use price until `createdAt` is 30 seconds old?

This could also (optionally) be converted into a graphql query for consistency with the rest of the API. In either case, I can merge and make the required changes if you'd like.

ekzyis:
> Do you think you can roll this without a dependency? I'm trying to be increasingly cautious about deps. This introduces quite a few.
> 
> In this case, for an in-memory cache (that'll only ever have ~6 items in it), we could simply use a global `Map` object, where the currency pair is the key, and the value is an object with `{ createdAt, price }` and we use price until `createdAt` is 30 seconds old?

Yes, no worries. I actually did start to replace the dependency but then I thought, why not show you with the dependency first and see what you think

> This could also (optionally) be converted into a graphql query for consistency with the rest of the API. In either case, I can merge and make the required changes if you'd like.

Ah, yes, I've noticed the regular GraphQL queries. I think I'll be able to do that.

components/price.js
ekzyis: Outdated because GraphQL API is now used

api/resolvers/price.js
```diff
@@ -0,0 +1,32 @@
...
+async function getPrice(fiat) {
+  fiat ??= 'USD';
+  if (cache.has(fiat)) {
+    const { price, createdAt } = cache.get(fiat)
+    const expired = createdAt + expiresIn < Date.now()
+    if (!expired) {
+      return price;
+    }
+  }
+  const url = `https://api.coinbase.com/v2/prices/BTC-${fiat}/spot`;
```
ekzyis, line 13: Without a lock here, in the worst case, every visitor would cause a coinbase API call if every request notices that the cache is expired at the same time and then goes on to do a API call.

@huumn, do you think a lock is thus required here?

api/resolvers/price.js
```diff
@@ -0,0 +1,32 @@
...
+  const price = await fetch(url)
+    .then((res) => res.json())
+    .then((body) => parseFloat(body.data.amount))
+    .catch((err) => {
+      console.error(err);
```
ekzyis, line 18: Is this how you log errors in your app?

api/resolvers/price.js
ekzyis: Thinking deeper about this and if this is indeed possible since NodeJS event loop is single-threaded:

I still think it's still possible that this happens due to how async/await works.
While we are fetching (which uses `await fetch`), the scheduler continues with another event which could invoke `getPrice` and see that the cache entry is (still) expired. It will then go to `await fetch` and give up execution to the scheduler. Therefore, two API calls will happen. This could happen an "infinite amount of times"

api/resolvers/price.js
huumn: Yes, that works.

api/resolvers/price.js
huumn: Worst case here seems to be that we make so many calls to coinbase at the same time, we get rate limited and responses from coinbase fail temporarily. I'd guess the probability of that happening is near zero, and so long as we can gracefully handle coinbase failing (in the rare event that it would occur), there's no need for a lock.

PR #234, opened 2023-02-09
Mention slashtags auth in README

ekzyis:
This also makes `SLASHTAGS_SECRET` optional.
If none is set, random bytes will be generated. This is useful for local development.


huumn:
Nice!

PR #235, opened 2023-02-12
Implement bookmarking of posts and comments

### LINKED ISSUE #71: bookmarking
Not high priority but people keep asking and I'd probably use it.

ekzyis:
Closes #71 

Hey, you posted this as a [bounty on SN](https://stacker.news/items/134939) and I thought I take a shot at it.

I implemented bookmarking of posts and comments using a star icon. Existing bookmarks can be seen under a new tab called "bookmarks". I recorded a short demo:

https://user-images.githubusercontent.com/27162016/218290511-10ed262f-c9cf-4c19-920f-547436399db1.mp4

At the end, you can see there is still a bug where if you bookmark a comment within the comment section of your profile, it also navigates to the comment instead of only bookmarking. Guess I have to put a `event.preventDefault()` somewhere.

The star icon is also not properly aligned with the post title yet, I think.

So consider this to be a first draft where we can discuss if the current approach makes sense.

For example, maybe bookmarking comments should be a button inside the dropdown menu and not an icon as for posts? Or be aligned at the right side next to the share icon?

What is also unclear if bookmarks should even be stored on the server. I decided to store it on the server so people can access their bookmarks from any browser and device.


prisma/schema.prisma
```diff
@@ -522,3 +524,12 @@ model VerificationRequest {
...
   @@map(name: "verification_requests")
 }
+
+model Bookmark {
+  user      User     @relation(fields: [userId], references: [id])
+  userId    Int
+  item      Item     @relation(fields: [itemId], references: [id])
+  itemId    Int
+  createdAt DateTime @default(now()) @map(name: "created_at")
+  @@id([userId, itemId])
```
ekzyis, line 534: I think I don't need an additional index on `userId` to speed up querying user bookmarks since it's part of the primary index.

ekzyis:
> For example, maybe bookmarking comments should be a button inside the dropdown menu and not an icon as for posts? Or be aligned at the right side next to the share icon?

In 970d566, I aligned it to the share icon now so it doesn't clutter the UI as much. I don't think we need the bookmark button everywhere. It is only visible for top level content now. It also now fits much better into the existing design, imo:
2023-02-12-050541_888x120_scrot
_bookmarking a post_

2023-02-12-050552_879x101_scrot
_bookmarking a comment_



This also fixes the navigation issue when bookmarking from within the comment section since it is no longer possible to bookmark from within the comment section.

prisma/schema.prisma
huumn: Yeah afaik that index can be used without specifying an itemId. You probably want an index on createdAt though for sorting.

ekzyis: Ah, right. Added an index

ekzyis:
Replaced the star icon with a bookmark icon (https://stacker.news/items/135163)

New showcase:

https://user-images.githubusercontent.com/27162016/218319803-1115e348-3d13-4e2b-bf82-ee50103c02aa.mp4

huumn:
On first look, it looks pretty good. I'm going to pull, refine, and test.

PR #245, opened 2023-03-02
Render images without markdown and use image proxy

jowo-io:
Here's a video demo of the feature:

https://www.loom.com/share/7e6ed97981a64840b25d9e62507eb279
> (please note, the `IS_NEW` state has been removed since I recorded the loom)

This code is quite rough and ready, if we're happy with the basic direction I'm happy to refine it further.

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -8,12 +8,14 @@ import sub from '../lib/remark-sub'
 import remarkDirective from 'remark-directive'
 import { visit } from 'unist-util-visit'
 import reactStringReplace from 'react-string-replace'
-import React, { useEffect, useState } from 'react'
+import React, { useRef, useEffect, useState } from 'react'
 import GithubSlugger from 'github-slugger'
 import LinkIcon from '../svgs/link.svg'
 import Thumb from '../svgs/thumb-up-fill.svg'
 import { toString } from 'mdast-util-to-string'
 import copy from 'clipboard-copy'
+import { IMG_URL_REGEXP } from '../lib/url'
+import { extractUrls } from '../lib/md'
 
 function myRemarkPlugin () {
   return (tree) => {
@@ -60,11 +62,51 @@ function Heading ({ h, slugger, noFragments, topLevel, children, node, ...props 
   )
 }
 
+const CACHE_STATES = {
+  IS_LOADING: 'IS_LOADING',
+  IS_LOADED: 'IS_LOADED',
+  IS_ERROR: 'IS_ERROR'
+}
+
 export default function Text ({ topLevel, noFragments, nofollow, children }) {
   // all the reactStringReplace calls are to facilitate search highlighting
   const slugger = new GithubSlugger()
 
   const HeadingWrapper = (props) => Heading({ topLevel, slugger, noFragments, ...props })
+
+  const imgCache = useRef({})
+  const [urlCache, setUrlCache] = useState({})
+
+  useEffect(() => {
+    const urls = extractUrls(children)
+
+    urls.forEach((url) => {
+      if (IMG_URL_REGEXP.test(url)) {
+        setUrlCache((prev) => ({ ...prev, [url]: CACHE_STATES.IS_LOADED }))
+      } else {
+        const img = new Image()
+        imgCache.current[url] = img
+
+        setUrlCache((prev) => ({ ...prev, [url]: CACHE_STATES.IS_LOADING }))
+
+        const callback = (state) => {
+          setUrlCache((prev) => ({ ...prev, [url]: state }))
+          delete imgCache.current[url]
+        }
+        img.onload = () => callback(CACHE_STATES.IS_LOADED)
+        img.onerror = () => callback(CACHE_STATES.IS_ERROR)
+        img.src = url
+      }
+    })
+
+    return () => {
+      Object.values(imgCache.current).forEach((img) => {
+        img.onload = null
+        img.onerror = null
+        img.src = ''
+      })
+    }
+  }, [children])
 
   return (
     <div className={styles.text}>
@@ -101,6 +143,10 @@ export default function Text ({ topLevel, noFragments, nofollow, children }) {
           a: ({ node, href, children, ...props }) => {
             if (children?.some(e => e?.props?.node?.tagName === 'img')) {
               return <>{children}</>
+            }
+
+            if (urlCache[href] === CACHE_STATES.IS_LOADED) {
+              return <ZoomableImage topLevel={topLevel} {...props} src={href} />
             }
 
             // map: fix any highlighted links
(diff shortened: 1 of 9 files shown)


components/text.js
```diff
@@ -60,12 +62,52 @@ function Heading ({ h, slugger, noFragments, topLevel, children, node, ...props
...
   const HeadingWrapper = (props) => Heading({ topLevel, slugger, noFragments, ...props })
 
+  const imgCache = useRef({})
+  const [urlCache, setUrlCache] = useState({})
+
+  useEffect(() => {
+    const urls = extractUrls(children)
+
+    urls.forEach((url) => {
+      if (IMG_URL_REGEXP.test(url)) {
```
jowo-io, line 84: > note: we could remove this regex check and just let all urls be checked asynchronously

ekzyis:
Another thing I just noticed: This image fetching happens in the client since it's run inside `useEffect`, right?

Then this opens up the possibility to leak IP addresses of SN users.

One can just use a freely available IP logger service like [this]( https://iplogger.org/) with a URL shortener (to make the URL less suspicious).

Then the client will happily check if that URL is an image and leak the IP address of the user.

This can even be done by targeting specific users. For example, you can discuss something with someone in a thread and check if they are replying pretty fast. This means they will most likely be the first user who sees your comments.

Then you can just post that link and wait until the client fetches the image.

I assume that's why most clients (like discord and snort.social for example) don't fetch anything directly. They always go to a proxy.

So this means code like this has to happen in the backend.

ekzyis:
Another thing I just noticed: This image fetching happens in the client since it's run inside useEffect, right?

Then this opens up the possibility to leak IP addresses of SN users.

One can just use a freely available IP logger service like [this](https://iplogger.org/) with a [URL shortener](https://www.shorturl.at/) (to make the URL less suspicious).

Then the client will happily check if that URL is an image and leak the IP address of the user.

This can even be done by targeting specific users. For example, you can discuss something with someone in a thread and check if they are replying pretty fast. This means they will most likely be the first user who sees your comments.

Then you can just post that link and wait until the client fetches the image.

I assume that's why most clients (like discord and snort.social for example) don't fetch anything directly. They always go to a proxy.

So this means code like this has to happen in the backend.

jowo-io:
> e uploaded to a proxy service and then

yeah really good point. definitely a concern for this PR, but also more generally for the whole SN site? (everywhere else that images are displayed)

ekzyis:
> but also more generally for the whole SN site? (everywhere else that external images are displayed)

pssst :joy: I noticed that too but didn't want to publicly disclose it here. I DM'ed @huumn on telegram. If you delete your comment, I will delete mine, too. Haha

But I don't consider it that severe. But don't want people start exploiting this before it is fixed.

ekzyis (approved):

jowo-io:
@ekzyis do you have any update on this bounty? would be a big shame to let the code go to waste 😕

huumn:
It won’t go to waste! 

We just want to support an image proxy for privacy reasons first.

ekzyis:
Looking at the following subdomain and processing options of the following URL, https://snort.social/ also seems to use https://docs.imgproxy.net/ so I think it's a good choice:

https://imgproxy.snort.social/h_yDski4D7T74dzC6fwcBqwBoMOw3og33BmkNewTMjA/rs:fit:586:586/dpr:2/aHR0cHM6Ly9tZWRpYS50ZW5vci5jb20vVzY0SVhKQmZGcFlBQUFBQy9maXJpbmctZ3VuLWplbm5pZmVyLWNvb2xpZGdlLmdpZg

Current state:


https://github.com/stackernews/stacker.news/assets/27162016/b4821dc9-2766-4453-96ce-156816ec2032



TODO:
- [ ] Support GIF
- [ ] Only replace image links
- [ ] Also replace image links in markdown formatting with image proxy URLs

jowo-io:
have you decided on a third party proxy? does it introduce much risk in the form of links being taken down or the service going bankrupt, etc? it might be prudent to save all the image urls into a table that maps the original img url to the proxied url, just to have it for future reference.

ekzyis:
https://github.com/stackernews/stacker.news/assets/27162016/c823863b-434d-4d41-ac61-e809776747e8

ekzyis:
For `IMGPROXY_KEY` and `IMGPROXY_SALT`, you can use `openssl rand -hex 64`

ekzyis:
Added embedding of images on link posts where the link is an image:


https://github.com/stackernews/stacker.news/assets/27162016/b97045c4-6ea3-4d41-a4d8-4a5b9f9ed69e

huumn:
I'm testing locally and it looks solid. Nice work guys and sorry it took so long!

I just need to clean up some stuff and make minor ui changes. It's date night tonight so I'll probably ship it tomorrow.

lib/url.js
```diff
@@ -26,3 +26,7 @@ export const URL_REGEXP = /^((https?|ftp):\/\/)?(www.)?(((([a-z]|\d|-|\.|_|~|[\u
 
 // eslint-disable-next-line
 export const WS_REGEXP = /^(wss?:\/\/)([0-9]{1,3}(?:\.[0-9]{1,3}){3}|(?=[^\/]{1,254}(?![^\/]))(?:(?=[a-zA-Z0-9-]{1,63}\.)(?:xn--+)?[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*\.)+[a-zA-Z]{2,63})(:([0-9]{1,5}))?$/
+
+export const IMGPROXY_URL_REGEXP = new RegExp(`^${process.env.NEXT_PUBLIC_IMGPROXY_URL}.*$`)
```
ekzyis, line 30: @huumn 

make sure you add a `/` at the end of `NEXT_PUBLIC_IMGPROXY_URL`

huumn: Are you sure? I end up with broken links: `https://imgprxy.stacker.news//v8Ghq8SIKhPhZ9mbMgCN7TKSQz5ij8pSoVqWAiwnENY/rs:fit:600:500:0/g:no/aHR0cHM6Ly9zdGF0aWMubmV3cy5iaXRjb2luLmNvbS93cC1jb250ZW50L3VwbG9hZHMvMjAyMy8wNy9iYWxhamktYmFua3MtMzgweDIxNC5qcGc`

I think this is handled in `createImageProxyUrl`

ekzyis: Ah, you are right.

Then I should add a slash here so only URLs which start with `https://imgproxy.stacker.news/` are considered to be from our image proxy and not also `https://imgproxy.stacker.news.malicious.domain`

huumn: I'll make the change. I'll grep around and make sure I don't miss somewhere.

jowo-io:
🚀 🚀 🚀 🚀 🚀 🚀

PR #251, opened 2023-03-24
Only run `npm install` once during `docker compose up`; worker process waits for this to complete

mvpratt:
Current:
Both the app and worker containers mount the host volume containing the source code. Currently, both run `npm install` which results in concurrently building the same code.

Proposed changes:
Run `npm install` in the `app` container only; the worker process waits for the install to complete before starting up.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-FROM node:14.17.0
+FROM node:14.21.3-bullseye
 
 ENV NODE_ENV=development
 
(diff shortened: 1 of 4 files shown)


huumn:
Thank you! I agree with the idea but I was able to find the long form of [`depends_on`](https://docs.docker.com/compose/compose-file/05-services/#depends_on) can do this when paired with `service_healthy` and [`healthcheck`](https://docs.docker.com/compose/compose-file/05-services/#healthcheck) on `app`.

I'm going to merge and remove the script locally.

PR #263, opened 2023-05-05
Add PWA feature to stacker news

spraveenitpro:
Summary:

I am adding PWA feature to stacker news so that It can be installed on a phone/desktop like a native app.

Testing Instructions:

1. Once you have switched to this branch then open the site in a browser of a phone or on the desktop computer and you should get an option to install it as a PWA like this 
2. If any issues, feel free to leave a comment.


.env.sample
ekzyis, line 1: Why was this file deleted? This is file is required.

next.config.js
```diff
@@ -1,115 +1,124 @@
...
-    value: 'GET, HEAD, OPTIONS'
-  }
+    {
+        key: 'Access-Control-Allow-Origin',
+        value: '*'
+    },
+    {
+        key: 'Access-Control-Allow-Methods',
+        value: 'GET, HEAD, OPTIONS'
+    }
```
ekzyis, line 14: Please undo code style changes like this.

I assume it was just auto-formatting but changes like these make it hard to see what was actually changed and what the actually required changes are.

next.config.js
```diff
@@ -1,115 +1,124 @@
...
-      {
-        source: '/statistics',
-        destination: '/satistics?inc=invoice,withdrawal',
-        permanent: true
-      }
-    ]
-  }
-})
+    }
+}))
```
ekzyis, line 124: Didn't take a look at all these changes since I assume most are because of changed indentation.

Will take another look when this was undone

public/manifest.json
```diff
@@ -0,0 +1,32 @@
+{
+    "theme_color": "#101010",
+    "background_color": "#f8e71c",
+    "display": "standalone",
+    "scope": "/",
+    "start_url": "/",
+    "name": "Stacker News",
+    "short_name": "SN",
+    "description": "Your favourite news sources",
```
ekzyis, line 9: Maybe this should use the default SN description?

`It's like Hacker News, but we pay you Bitcoin`

ekzyis (changes requested):
Lot of code style changes. Will take another view when these were undone.

huumn:
Thanks for the solid effort! I'll send some sats to your SN account.

I decided to go with `display: minimal-ui` until email login and navigation are figured out on all browsers.

For SN archeologists:
1. `next-pwa` fails to build SN into a service worker (at least in dev mode but probably in prod too ... which means it probably needs a lot more config to actually do anything)
2. email based magic link login is incompatible with `display: standalone` on iOS Safari
3. `standalone` strips back button navigation, which the average user will want, so we need a solution to that too

ekzyis:
For reference: #263

Issue #264, opened 2023-05-07
URL redirect bug


orthzar:
This issue is submitted as per this [SN post](https://stacker.news/items/174926).

### Problem

First of all, this bug is severe, because it could be used to phish user credentials -- and SN accounts involve money.

Some SN pages currently redirect signed-in users to any URL. For instance, these two links will redirect you to Google:
https://stacker.news/signup?callbackUrl=https%3A%2F%2Fwww.google.com%2F
https://stacker.news/login?callbackUrl=https%3A%2F%2Fwww.google.com%2F

It would be trivial to send SN users emails, which contain links that seem legit, but which actually redirect them to a fake SN login page. The credentials harvested through such a page would be used to steal user funds.  The only thing the attacker would need is SN user email addresses, which have been accidentally leaked once before.

The bug exists in ```pages/signup.js``` and ```pages/login.js```. Both have a function ```getServerSideProps```, which seems to not check the content of ```callbackUrl```. The bug only affects users who are signed in, because the session variable has to be non-null in ```getServerSideProps``` for the redirect to work.

```callbackURL``` is also mentioned in ```pages/settings.js```, ```pages/invites/[id].js```, and some files in the components folder. I have not looked at or tested those files.

### Solutions

This bug could be solved by either eliminating the redirection functionality; or by checking the domain in callbackUrl (e.g. by only allowing redirects to stacker.news). I don't know which solution would be correct, because I don't know what purpose this redirection serves.

Sorry, I don't have any code to offer. I have no experience writing Javascript and there isn't enough time for me to learn.

ekzyis:
Thanks for posting! I actually didn't see the SN post but saw this ticket here.
**So thank you very much, I guess we would have totally missed it otherwise.**

This is definitely an issue and we should warn users before redirecting them to external sites like many other sites do.

For example, clicking on this link shows a prompt before redirecting: https://www.google.com/url?q=https://www.bufferzonesecurity.com :

![2023-05-07-132722_658x198_scrot](https://user-images.githubusercontent.com/27162016/236674748-061cb030-abc2-4250-a11d-b733c21d480b.png)

Therefore, an attacker could just create a very similar site that looks like SN to grab credentials from users who don't check the URL again. So I agree, this is indeed severe. 

Fortunately, the fix should be easy by just validating the callback URL. Will create a PR in a few minutes.

ekzyis:
Taking a more closer look, I realized that grabbing of credentials wasn't possible.

[...]

PR #265, opened 2023-05-07
Return 500 if callback URL is external

ekzyis:
This should fix the open redirect mentioned in #264.

**However, I wasn't able to test this since I can't login/signup on my local machine at the moment.**

I just tested that `decodeURIComponent` and the regex work as I expected manually:

```
// open_redirect_test.js
const regex = /^https?:\/\/stacker.news\/?/

let callbackUrl = "https%3A%2F%2Fwww.google.com%2F"
let decoded = decodeURIComponent(callbackUrl)
let isExternal = !regex.test(decoded) 
console.log(isExternal) // should print true

callbackUrl = "https%3A%2F%2Fstacker.news%2F"
decoded = decodeURIComponent(callbackUrl)
isExternal = !regex.test(decoded)
console.log(isExternal) // should print false
```

```
$ node open_redirect_test.js
true
false
```

These changes should return 500 Internal Server Error if there exists a session already for a user instead of redirecting them to an external site.

However, I think for users with no session yet, the callback url should also be validated. Else, they will be redirected to a external site _after_ login which is probably also not in our interest.

### DIFF
--- a/pages/login.js
+++ b/pages/login.js
@@ -6,6 +6,14 @@ import Login from '../components/login'
 export async function getServerSideProps ({ req, res, query: { callbackUrl, error = null } }) {
   const session = await getSession({ req })
 
+  const regex = /^https?:\/\/stacker.news\/?/
+  const external = !regex.test(decodeURIComponent(callbackUrl))
+  if (external) {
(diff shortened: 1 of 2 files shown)


huumn:
Thanks! I'll check this out and deploy it soon

PR #276, opened 2023-05-15
Update node to v18.16.0, prisma to v3.13

ekzyis:
Updating prisma to v3.13 fixes my issue with missing openssl 3.0 support where I get this error:
> Error: Unknown binaryTarget debian-openssl-3.0.x

See here: https://github.com/prisma/prisma/releases/tag/3.13.0.

See here for breaking changes: https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-3

I think we are not impacted by the introduction of referential actions and we don't use JSON columns.

Regarding the named constraints, I picked [option 2](https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-3/named-constraints) which updates all constraints to use the new default naming convention. This drops some indices and recreates them. I've seen that the most recent post has id 178998 and I think `Item` is the biggest table, so I think this won't take too long. However, I am not sure why some foreign key constraints are dropped and recreated under the same name while some constraints are indeed just renamed. :shrug: 

Regarding `$queryRaw(string)`, I replaced all occurrences with `$queryRawUnsafe(string)` since the former is no longer supported. Replacing with usage of template literals wouldn't work since we've been using string interpolation to build parts of the queries which can't be built using bindings with template literals. I haven't tested this but I am pretty sure.

Updating node to v18.16.0 because [v14 is EOL since May 2023](https://nodejs.dev/en/about/releases/).


ekzyis:
Ok this needs some further work, posting doesn't work atm:

2023-05-16-001624_741x557_scrot

huumn:
I haven't updated to prisma 3 in attempt to avoid all these breaking changes ... and because I'd like to move off of their generated client anyway.

If we *need* to upgrade to v3, it will need to be *thoroughly* tested and might be better left to me as I'm familiar with all the code likely to break.

Couldn't you just install openssl 2?

ekzyis:
For some reason, there is no openssl 2 for arch linux (I am running manjaro): https://archive.archlinux.org/packages/o/openssl/

Downgrading to the next available version which is openssl-1.1.1 seems to lead into dependency hell where other system dependencies would need to be downgraded as well. Since manjaro is a rolling release, I would rather not go down this path :sweat_smile: 

```
$ sudo pacman -U https://archive.archlinux.org/packages/o/openssl/openssl-1.1.1.q-1-x86_64.pkg.tar.zst        
loading packages...
warning: downgrading package openssl (3.0.8-1 => 1.1.1.q-1)
resolving dependencies...
looking for conflicting packages...
error: failed to prepare transaction (could not satisfy dependencies)
:: installing openssl (1.1.1.q-1) breaks dependency 'libcrypto.so=3-64' required by libarchive
:: installing openssl (1.1.1.q-1) breaks dependency 'libcrypto.so=3-64' required by libgit2
:: installing openssl (1.1.1.q-1) breaks dependency 'libssl.so=3-64' required by libgit2
:: installing openssl (1.1.1.q-1) breaks dependency 'libcrypto.so=3-64' required by sudo
:: installing openssl (1.1.1.q-1) breaks dependency 'libssl.so=3-64' required by sudo
:: installing openssl (1.1.1.q-1) breaks dependency 'libcrypto.so=3-64' required by systemd
:: installing openssl (1.1.1.q-1) breaks dependency 'libssl.so=3-64' required by systemd
:: installing openssl (1.1.1.q-1) breaks dependency 'libcrypto.so=3-64' required by tcpdump
:: installing openssl (1.1.1.q-1) breaks dependency 'libcrypto.so=3-64' required by unbound
:: installing openssl (1.1.1.q-1) breaks dependency 'libssl.so=3-64' required by unbound
```

But I can understand that the changes in this PR should not be taken lightly.

We could skip this upgrade for now since there is no immediate need. I only did the changes to run SN locally to connect to Polar. But I don't need to work with the Lightning code right away. I can keep running SN using docker in the meantime.

What about the node upgrade? Should I create a separate PR only for that?

huumn:
> What about the node upgrade? Should I create a separate PR only for that?

Yes I think that's the way to go.

> We could skip this upgrade for now since there is no immediate need. I only did the changes to run SN locally to connect to Polar. 

You *should* be able connect to Polar from the SN docker "cluster." It's *probably* just figuring out any docker networking config changes that are necessary to bridge the two.

huumn:
It looks like docker by default creates a bridge network between all containers: https://www.tutorialworks.com/container-networking/#default-bridge-network-easiest-option

I'm not sure what you've tried though.

huumn:
> If we need to upgrade to v3, it will need to be thoroughly tested and might be better left to me as I'm familiar with all the code likely to break.

Scratch that. If you want this sooner rather than later, you'll probably be more motivated to get it done than me. You'll just need to Q/A it thoroughly.

huumn:
We've been upgraded to 18 since 16e882 and prisma 5 shortly before then.

PR #278, opened 2023-05-16
Add subName to seed

ekzyis:
Not sure if this was overlooked?
Users can no longer create posts with no sub set. So I think it makes sense to reflect this in the seed script?


huumn:
I don't use the seed anymore when developing. This makes sense though.

PR #279, opened 2023-05-17
Add docs for local LND setup

ekzyis:


docs/local-lnd.md
```diff
@@ -0,0 +1,177 @@
...
+            "Options": {},
+            "Config": [
+                {
+                    "Subnet": "192.168.192.0/20",
+                    "Gateway": "192.168.192.1"
+                }
+            ]
+        },
+        "Internal": false,
+        "Attachable": false,
```
ekzyis, line 88: `attachable` is not required to be true. That was a false positive from me

Issue #164, opened 2022-06-13
linking multiple lightning logins


huumn:
Users are asking. Need to migrate the db from has-a to has-many and provide UI for it.

ekzyis:
Another post related to this: https://stacker.news/items/179236

Regarding security:
I think this should be handled the same way as password changes are handled in other services since functionality like this can lead to account takeovers. So this would mean that even logged-in users have to confirm that they still have access to the old wallet to be able to add a new wallet.

If old lightning logins can also be deleted (which I think would be good UX), this would be even more of a concern.

rleed:
Is this perhaps a general case of needing to be able to merge SN profiles? I noticed by logging into SN with different methods (GH vs. email e.g.) I get different profiles, and sometimes I would like to merge them.

huumn:
That's a separate issue. This is for having multiple lightning logins on a single account.

huumn:
Implementation requirements:

- lightning pubkey needs to be broken out into its own table and `User` has-many of them
- each lightning login need to be given an optional label

PR #282, opened 2023-05-17
Show user pubkey in settings

### LINKED ISSUE #281: Show pubkey of users logged in with Lightning
A user asked how they can check which wallet they used for login: https://stacker.news/items/179376

We could show users their pubkey in their profiles so they can compare with the pubkey in their wallet.

I know that at least Phoenix shows your node id (pubkey) in the channel list.

ekzyis:
Closes #281 

Looks like this:

2023-05-17-154052_1920x1080_scrot

(shows full pubkey when hovering)

I think it could look better but don't know how exactly :thinking:

### DIFF
--- a/api/typeDefs/user.js
+++ b/api/typeDefs/user.js
@@ -33,7 +33,7 @@ export default gql`
   }
 
   type AuthMethods {
-    lightning: Boolean!
+    lightning: String!
     slashtags: Boolean!
     github: Boolean!
     twitter: Boolean!
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -52,7 +52,7 @@ async function authMethods (user, args, { models, me }) {
   const oauth = accounts.map(a => a.providerId)
 
   return {
-    lightning: !!user.pubkey,
+    lightning: user.pubkey ?? '',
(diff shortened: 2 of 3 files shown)


api/typeDefs/user.js
```diff
@@ -33,7 +33,7 @@ export default gql`
   }
 
   type AuthMethods {
-    lightning: Boolean!
+    lightning: String!
     slashtags: Boolean!
```
ekzyis, line 37: Maybe do the same with slashtags?

ekzyis:
> This pubkey is specific to SN given the way ln-auth works. It is not their node pubkey.

Oh, right, it creates domain-specific `linkingKey`s... I forgot

ekzyis:
> which I was going to do when I allowed multiple ln auths on a single account

Makes sense. I'll link #164 and close this PR then. Sorry for the mistake, I realize I am getting to eager to change stuff without properly thinking first hehe. Maybe related to "too much excitement" haha

huumn:
All good. I strongly prefer you eagerly changing stuff than not. Especially if you don't mind it occasionally not getting through when it isn't right.

PR #284, opened 2023-05-18
Proof of Concept of WebSocket API

ekzyis:
I have implemented a basic example how websockets could be integrated into SN in a not too intrusive way for real-time data.

By just creating more signals and subscriptions, we could use a single websocket connection for various things like "new comments", "new tips" etc.

The following stuff is missing however:

- [ ] HTTPS (`wss://` instead of `ws://`)
- [ ] Not hardcoding `localhost:3000`
- [ ] Notification for earnings
- [ ] Reconnects
- [ ] Consideration of load balancer
- [ ] General testing to see how stable the implementation is

Example video to show difference between current 30 seconds polling for notifications and websockets:

https://github.com/stackernews/stacker.news/assets/27162016/9a2e7d70-9a30-472c-a277-5530d4f5c619


api/resolvers/item.js
```diff
@@ -927,13 +931,15 @@ export default {
       const [item] = await models.$queryRaw(`
             ${SELECT}
             FROM "Item"
-            WHERE id = $1 AND "userId" = $2`, Number(id), me.id)
-      if (item) {
+            WHERE id = $1`, Number(id))
+      if (item === me.id) {
```
huumn, line 935: Typo?

ekzyis: Oh, yes, good catch

huumn:
Also, as we've discussed, I would like to remind us (I need this reminder too) that we're going to want to be very very particular with when we decide to use websockets for stuff. It's tempting to use them for everything, but we should use polling unless we have good reason not to ... good reason most of the time is going to look a lot like this: the thing we're interested in updates very infrequently but its very expensive to check if it's updated.

ekzyis:
> Overall it's really slick! Nice work!

Thanks :)

> (tho tbh all of the state management has me worried there are going to be tricky non-deterministic bugs ... but I'm certain we can figure it out if we want)

Yes, I'm also worried and optimistic. But I have to admit, it took me like 2 hours to fix this simple bug, lol:
```diff
- socket.removeListener(onError)
+ socket.removeListener('error', onError)
```
I was very confused by the webpack errors which were pointing to a line which didn't even exist and I thought I was going crazy haha

There are most likely still bugs in there. At least when using hot reloading in dev, some event emitters are still leaking iirc.

> There are notifications for more than just earning and direct replies. Mentions for example. Invites. Referrals. Replies of replies. Most parts of the code will need to be aware of this code. How can we prevent that? If we can figure that out, then the maintenance cost of this will be 100x lower. It'll feel like it isn't even there.

Mentions are already included no? Invites and referrals I missed, yes. Regarding replies of replies, I added a TODO (needs recursive search back to root). I used the `hasNewNotes` resolver as a reference but now I see it's also mentioned there. 

What I can see there (and what has been implemented in this PR already):
- [x] replies
- [x] tips
- [x] mentions
- [x] job updates
- [ ] earnings
- [ ] deposits
- [ ] invites/referrals
- [ ] cowboy hat

Anything else? Is that resolver a full reference?

> my goto for side effects like this is db triggers

Yes, that was my first idea even though I heard bad stories about them ("hidden behaviour" makes debugging hard etc.). But I have taken a look if Prisma supports `NOTIFY/LISTEN` (which was the thing I would have done inside a trigger) but that is not the case: https://github.com/prisma/prisma/issues/905

So this was out of scope for a PoC.

> because we know where all of them are and the db will record them and list them out for us if asked

What do you mean with "record them and list them out for us"? You would have done something else than `NOTIFY/LISTEN`? Afaik, `NOTIFY` does not keep state of previous `NOTIFY`s so we can't "list them out"?

For reference, that's how I use `NOTIFY/LISTEN` for https://minesats.gg: https://gitlab.com/ekzyis/minesats/-/blob/develop/src/pages/api/_db.ts#L11-22 
The worker in `bot/main.go` (discord bot written in golang) calls `NOTIFY`.

> we'll want to mutate the client side cache in response to these 'events', which in some cases like zaps is straightforward, but things like adding new items will need additional state management (so that there isn't layout shift when the cache is updated).

Makes sense!

> This requires thinking a bit more architecturally. If we don't want new infrastructure (we don't), then we'll need to use postgres for pubsub, ie notify/listen. I do this in booger and its what I was planning on doing for SN, so perhaps calls to notify are mostly what happen in the triggers.

Yes, this article also mentions this ("pub/sub broker"): https://hackernoon.com/scaling-websockets-9a31497af051

> we're going to want to be very very particular with when we decide to use websockets for stuff. It's tempting to use them for everything, but we should use polling unless we have good great reason not to

Yes, I agree. I was also thinking about still using polling in parallel just as a fallback ... but then we would gain nothing performance wise (in fact make it worse).

> great reason most of the time is going to look a lot like this: the thing we're interested in updates very infrequently but its very expensive to check if it's updated (edit: and we've exhausted attempts to make it less expensive to check).

I would add that UX should also be considered: How valuable would it be to have less latency?
For posts about live events, I think it's a big difference if the replies for a specific item are polled every 5, 10, 30 seconds (whatever the cost is to poll, probably not much using indices) or if they are indeed live.

An example post would be this from today: https://stacker.news/items/180253

What do you think about the use case here for notifications? You think that's a good example when websockets can be used?

ekzyis:
A bad example of using websockets "just for the sake of it" would be invoices, I guess.

I think it doesn't take much resources to poll for a specific invoice (we do it every second already) + the user doesn't really care if we notice that the invoice was paid a few seconds earlier + invoices and withdrawals are not part of the "main UX" of users on SN (I assume users don't deposit and withdraw that often).

huumn:
> Yes, that was my first idea even though I heard bad stories about them ("hidden behaviour" makes debugging hard etc.). 
You can find mixed takes on anything because any bit of code is worse than none. The question we need to ask is if an automatic (aka hidden), atomic, and ephemeral behavior that exists in the DB is preferred to it being put directly in the code in quite a few different places? Should the side effect reside near the action (a request to store) or the effect (store)? It isn't super black and white to me but I generally lean toward triggers for side effects to keep the api layer action/query focused. You should choose for yourself.

> But I have taken a look if Prisma supports NOTIFY/LISTEN (which was the thing I would have done inside a trigger) but that is not the case.
We can connect establish our own connection to the database without prisma. It's garbage anyway.

> What do you mean with "record them and list them out for us"?
You can ask postgres to tell you about its triggers. It basically documents the side effects in a way that we can query and it tells us how it relates to any particular table.

> I would add that UX should also be considered: How valuable would it be to have less latency?
For posts about live events, I think it's a big difference if the replies for a specific item are polled every 5, 10, 30 seconds (whatever the cost is to poll, probably not much using indices) or if they are indeed live.

True!

The size of the difference is linear. The difference between 100ms and 1 seconds isn't much difference practically speaking but 100ms and 10 is a lot for real time chat. Do I really need to know the founder of that pre product unicorn said he was a saint sent here to save us all as fast as possible? I can probably wait 10 seconds. It does depend on the use case.

huumn:
> Anything else? Is that resolver a full reference?

Yes, it should be!

components/ws.js
```diff
@@ -0,0 +1,71 @@
...
+  const [ws, setWs] = useState(null)
+
+  useEffect(() => {
+    if (ws) return
+
+    // the endpoint needs to be hit at least once using HTTP
+    // to initialize the websocket server
+    fetch('http://localhost:3000/api/ws')
+
+    const client = new WebSocket('ws://localhost:3000/api/ws')
```
ekzyis, line 17: Race condition here regarding the first fetch from the first user after a deploy.

There is probably a better way to initialize the websocket server instead of every user first fetching the endpoint.

huumn:
Another use for this kind of push system is push notifications in a PWA. I think this would be a big upgrade in terms of people being willing to start and engage in conversations ... and its one of the most frequently and afaict "real" feature requests we get.

It's not exactly related to the websocket stuff, but it could make use of all the "this person should be notified" hooks being built for websockets.

kevkevinpal:
concept ACK for me, websockets helps the snappyness of the user experience and would be a greatly appreciated upgrade

also love the music choice on the video :)

PR #287, opened 2023-05-20
Fix no dupes check after link unshortening

### LINKED ISSUE #244: Dupe check isn't occurring after shortened link gets decoded
There was a feature added in Issue #179 so that a shortened t.co link gets decoded.  However, there appears to not then be the normal dupe check occurring for that link.

Here's a specific example:
Shortened link: https://t.co/QYsP8pIF78
Decoded link: https://blog.btcpayserver.org/btcpay-server-1-8-0/

If I paste the Decoded link, the Dupe check will properly warn that this is a dupe.  If I instead paste the shortened, t.co link, the decoding occurs but no "Dupe" warning occurs.

ekzyis:
Closes #244

https://github.com/stackernews/stacker.news/assets/27162016/66c39691-ce7a-4380-881a-c7174833e45b

### DIFF
--- a/components/link-form.js
+++ b/components/link-form.js
@@ -80,6 +80,14 @@ export function LinkForm ({ item, sub, editThreshold, children }) {
     }
   }, [data?.pageTitleAndUnshorted?.title])
 
+  useEffect(() => {
+    if (data?.pageTitleAndUnshorted?.unshorted) {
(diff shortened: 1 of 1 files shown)


huumn:
I really appreciate these videos!

PR #292, opened 2023-05-31
Use PWA with display: standalone

### LINKED ISSUE #137: Turn Stacker News into a Progressive Web App
Currently on my machine it takes ~2s to get response to the initial GET / request, which slows down the experience. 
If the App would follow the rules around PWAs and register ServiceWorker, define manifest and static resources, then the initial 2s could be completely eliminated. PWAs are using interesting set of native browser APIs and there are multiple interesting things that can be done. 

Here are basic tutorials: 
- https://web.dev/learn/pwa/
- https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Offline_Service_workers

The are various benefits, e.g.
- saving latency and number of resources that need to be sent over wire (essentially the ServiceWorker serves as a proxy between the web app and internet - you can do various caching and background-refetching schemes)
- better control around icons on the phone
- the app can fairly easily be wrapped into an **Android app and put on Play Store**: https://developers.google.com/codelabs/pwa-in-play#0
- It's easier/possible to register phone notifications even if Stacker News isn't in focus

ekzyis:
What this PR does:

- Use `display: standalone` for PWA
- includes back button for navigation in standalone mode
- includes support for notifications using the Notification API [0]
- empty custom service worker for further enhancements:
  - better caching to boost performance
  - Push API support for real-time notifications even when client is inactive
  - ...
- usage of `url_handlers` to register `https://stacker.news` such that links to SN can be opened using the PWA [1]
- basic offline page

Closes #137 

[0] https://developer.mozilla.org/en-US/docs/Web/API/notification
[1] https://developer.chrome.com/articles/pwa-url-handler/


ekzyis:
Notification API showcase:

https://github.com/stackernews/stacker.news/assets/27162016/96304653-8abb-4c28-ab8f-a542a0b5f30c

ekzyis:
I was also thinking about using `position: fixed` for the header in standalone mode so the user can use the back navigation without having to scroll all the way up. What do you think @huumn? Worth the effort + additional complexity?

huumn:
> I was also thinking about using `position: fixed` for the header in standalone mode so the user can use the back navigation without having to scroll all the way up. What do you think @huumn? Worth the effort + additional complexity?

I think we can live without for now. It'll be easier to know what we need vs what we think we need when we're all using it for a bit.

huumn:
Did you figure out email login on iOS?

ekzyis:
> Did you figure out email login on iOS?

Not sure, but I used `url_handlers` which in theory, should prompt the user if he wants to open the PWA when clicking on the login link in the mail:

> PWA as URL Handlers allows apps like music.example.com to register themselves as URL handlers for URLs that match patterns like https://music.example.com, https://*.music.example.com, or https://🎵.example.com, so that links from outside of the PWA, for example, from an instant messenger application or an email client, open in the installed PWA rather than in a browser tab.

-- https://developer.chrome.com/articles/pwa-url-handler/

If I understood the problem correctly, that should be the solution, no?

But I haven't tested it. Was not trivial to test that stuff since it requires https and installing PWAs using free `ngrok` also wasn't trivial because of the disclaimer:

> The scheme is always assumed to be https://, but it needs to be explicitly mentioned.

-- https://developer.chrome.com/articles/pwa-url-handler/#the-url_handlers-web-app-manifest-member

ekzyis:
> we can just ship and wait for someone to complain before figuring out a hack.

Yeah, I was thinking the same

huumn:
Do you now if this fetches notifications in the background?

components/header.js
```diff
@@ -52,7 +55,20 @@ export default function Header ({ sub }) {
...
+    fetchPolicy: 'cache-and-network',
+    // Trigger onComplete after every poll
+    // See https://github.com/apollographql/apollo-client/issues/5531#issuecomment-568235629
+    notifyOnNetworkStatusChange: true,
+    onCompleted: (data) => {
+      const notified = JSON.parse(localStorage.getItem('notified')) || false
+      if (!notified && data.hasNewNotes) {
+        notification.show('you have new notifications')
+      }
+      localStorage.setItem('notified', data.hasNewNotes)
```
huumn, line 69: When is this cleared?

huumn: Never mind. It's cleared on the next poll. Duh

ekzyis: yeah, all i want with this is to see the "rising edge" when we move from "no new notes" to "new notes".
not sure if this is the best way

huumn:
Overall it's solid. For future reference we use https://remixicon.com/ for most "generic" icon needs (e.g. the back arrow).

ekzyis:
This has been deployed and I tested it. I noticed following things:

- Back button works decently on mobile. It disappears when not required. But it sometimes breaks the layout:

signal-2023-06-01-224219_002

- SN URL registering worked as expected. I now get a prompt to open any link which starts with https://stacker.news using the PWA. This also fixes the email login PWA issue:

signal-2023-06-01-225027
signal-2023-06-01-225251

- Offline mode doesn't work:

signal-2023-06-01-231157

huumn:
I forgot to test it on mobile ... I'll do that, fix it, then maybe we can announce before end of day.

ekzyis:
btw, I think notifications using the Notification API doesn't work in a PWA. It needs to use the Push API.

huumn:
Yeah there's some unexpected caching behavior happening in the browser too on notifications. eg the notification dot doesn't go away immediately when I navigate to notifications. BUT overall the site is way more snappy (even in the browser) likely due to more cache hints from the pwa ... assuming that's a thing.

I'm not exactly sure what's going on though with the notification yet. Still trying to get the layout stuff to not regress.

ekzyis:
Oh ok, yeah, I noticed the same with notifications.

I'll look into the caching thing

PR #301, opened 2023-06-03
Enable push notifications in settings

### LINKED ISSUE #300: Make browser notifications double opt-in
https://stacker.news/items/187444

ekzyis:
Close #300 

https://github.com/stackernews/stacker.news/assets/27162016/cebcab78-4f06-42a7-b9a7-8b5da4522b7c

As can be seen at 1:50, there is a bug where the checkbox is still checked even though the value is `false`. I think it has to do with the checkbox being disabled and thus no longer respecting changes to its `value` (? - just a wild guess)

/cc @sime


sime:
Potentially out of scope for this PR but a tasteful CTA could exist on the notifications page.

E.g. "want to get a push notification"/"enable  push notifications"?

As for the bug, I can understand that check boxes and reloading pages have some annoying quirks.

But the PR is cool though!

huumn:
I agree with @sime that this would see more usage if it were on the /notifications page but this is fine too if we only want a minority of stackers to use it.

huumn:
Also, I'll assume this a draft until the bug is fixed.

ekzyis:
I decided to put it into /settings because I read [this](https://web.dev/push-notifications-permissions-ux/#settings-panel) and I wasn't sure what to put on /notifications without "disrupting the page". I am still a noob when it comes to UI/UX, lol. This was also noticeable since I actually put the prompt to ask for permission on the front page for every logged-in user to see which is apparently the worst thing one can do, lol:

> The worst thing you can do is to show the permission dialog to users as soon as they land on your site.

 @huumn fortunately decided to move it to /notifications.

But I think a CTA (I also had to look up what this means, haha) on /notifications would still be a good idea. It's probably also a good practice in UI/UX for me to come up with something.

> Also, I'll assume this a draft until the bug is fixed.

To be honest, I didn't consider the bug to be blocking since the real state is shown after a page reload. So I thought it's more important to get this out fast to prevent more annoyed users than to wait until the bug is fixed. I thought that would be the right trade-off here.

But I'll take another fresh look at the bug now. Maybe I just completely missed something and I can fix it now within 5 minutes. :)

huumn:
> To be honest, I didn't consider the bug to be blocking since the real state is shown after a page reload. So I thought it's more important to get this out fast to prevent more annoyed users than to wait until the bug is fixed. I thought that would be the right trade-off here.

Oh I didn't realize this was intended as a temporary fix. You think it's good to ship?

>  I am still a noob when it comes to UI/UX, lol. 

80% of good UX is to forget you're the developer and think about what you'd want as a user who knows nothing about how something works. It's all kind of role playing.

ekzyis:
> Oh I didn't realize this was intended as a temporary fix. You think it's good to ship?

Fix is pushed now. I also checked some other stuff like that I didn't break `formik` by explicitly setting `value`:

```diff
         <BootstrapForm.Check.Input
-          {...field} {...props} type='checkbox' onChange={(e) => {
+          {...field} {...props} value={value} disabled={disabled} type='checkbox' onChange={(e) => {
```

 Afaik, I didn't but I removed the `value` prop anyway since it had no effect.
I didn't realize that `formik` was actually setting the values and setting `value` like this had no effect.

Additionally, I checked that `pushNotify` is not submitted since it's only a client-side setting (for now) and it might break something if sent to the server. However, submitting the settings still worked so I was confused because I was pretty sure that the database query should break if an unexpected property is added but it didn't.

I then did some further investigation and realized that GraphQL already removes any variable which is not defined in the mutation. So I didn't add specific code to exclude the property like this:

```diff
diff --git a/pages/settings.js b/pages/settings.js
index 8a6ad7b..d7491e8 100644
--- a/pages/settings.js
+++ b/pages/settings.js
@@ -74,7 +74,7 @@ export default function Settings ({ data: { settings } }) {
             nostrRelays: settings?.nostrRelays?.length ? settings?.nostrRelays : ['']
           }}
           schema={settingsSchema}
-          onSubmit={async ({ tipDefault, nostrPubkey, nostrRelays, ...values }) => {
+          onSubmit={async ({ tipDefault, nostrPubkey, nostrRelays, pushNotify, ...values }) => {
             if (nostrPubkey.length === 0) {
               nostrPubkey = null
             } else {
```

This behavior is now implicitly relied on however. Not sure if that's the best way to go.

ekzyis:
Btw, you can test this on https://sn.ekzyis.com/ if you want.

I've added a server config for `sn.ekzyis.com` to the nginx running on a server of mine so I can have a dev environment more close to prod (HTTPS, public domain, nginx as reverse proxy)

For now, it just points back to my local machine using a SSH remote port forwarding but I intend to host it directly on the server soon.

ekzyis:
Btw, you can test this on https://sn.ekzyis.com/ if you want.

I've added a server config for sn.ekzyis.com to the nginx running on a server of mine so I can have a dev environment more close to prod (HTTPS, public domain, nginx as reverse proxy)

For now, it just points back to my local machine using SSH remote port forwarding but I intend to host it directly on the server soon.

huumn:
Another UX problem with this is that everything else in the settings pane is persisted on the server. If I enable push notifications on my phone and go to desktop, this will be unchecked in my settings. It's a pretty inconsistent experience.

I think an ideal UX might just be a dismissible Alert on the notifications page at the top like we do with lightning addresses: https://github.com/stackernews/stacker.news/blob/369bd82a84562700c52c295be7701afc0a45f3cf/pages/wallet.js#L99

Another alternative approach is something like the darkmode and lightning bolt animation toggles which similarly operate on local storage.

ekzyis:
> Another UX problem with this is that everything else in the settings pane is persisted on the server. If I enable push notifications on my phone and go to desktop, this will be unchecked in my settings. It's a pretty inconsistent experience.

Yes, it's inconsistent but isn't this related to the actual permission request being "inconsistent" since we need to ask per browser for notifications? Also, afaik, phone notifications don't work atm. We will need to use the Push API inside the service worker for that which is different from the Notification API used here.

When we use the Push API, I agree that this setting should be persisted on the server (it has to anyway if I understood the Push API architecture correctly). Not sure how to solve the "ask per browser" thing, though. Just showing "push notifications enabled" for a browser for which push notifications have in fact _not_ been enabled is also inconsistent.

ekzyis:
> I think an ideal UX might just be a dismissible Alert on the notifications page at the top like we do with lightning addresses: https://github.com/stackernews/stacker.news/blob/369bd82a84562700c52c295be7701afc0a45f3cf/pages/wallet.js#LL99C1-L107C18

Added this in 962bbff. Looks like this:

2023-06-05-002938_932x191_scrot

components/notifications.js
```diff
@@ -251,6 +251,39 @@ function Reply ({ n }) {
...
+          Enable push notifications?
+          <button
+            className='btn mx-1'
+            onClick={() => {
+              pushNotify.requestPermission()
+              close()
+            }}
+          >Yes
+          </button>
+          <button className='btn mx-1' onClick={close}>No</button>
```
ekzyis, line 280: Added `mx-1` because of https://web.dev/accessible-tap-targets/

Else there is no cursor change when going between "Yes" and "No". 

(Btw, I noticed the same with bounty posts in the past. There was no intermediate cursor change between "reply" and "pay bounty".)

huumn:
Is this good to go?

ekzyis:
Mhh, I think I this is still a draft since the permission is not stored in the database yet :thinking: 

I think this is the only thing that's missing if you are fine with the alert UX in notifications/

Issue #290, opened 2023-05-25
UX enhancement request: Ctrl-click on yellow buttons to open in new tab


cointastical:
My usage pattern with SN is mainly on web (on desktop) and oftentimes has me scrolling down the Recent/Newest. With almost every action I prefer to Ctrl-Click to open a new tab for the results of that click.  This lets me not lose my position in the recent/newest page as I scroll down to see posts in reverse chronological order.

That works great for links.  But for (most of) SN's yellow buttons, instead of Ctrl-Click, I need to use my mouse to right-click and select Open in New Tab.

Could the javascript behavior on those be (easily) modified so that Ctrl-Click on them behaves the same as if it were a Link, so that I wouldn't need to right-click and select Open in New Tab?

huumn:
Do you mean buttons like the `post` button? 

Also, you shouldn't lose your place if you use the back/forward buttons. If you do, that's a bug.

cointastical:
Yes, the Post button is why I wrote this.

I would like to be able to click the button but have the response in a new tab.

> Also, you shouldn't lose your place if you use the back/forward buttons. If you do, that's a bug.

Let's say I've navigated down a couple pages on /recent, and I'm at #100.  Then I scroll back up to the top to /post
Then I post.  

How do I get back to the page that showed me that I was down to around #100?   If I click /recent I am back to the most recent 21 entries.  So I click More, ... look at what is on the next page, ... determine I must have been further back, so I click More again, repeat N times until new stuff appears.

What I would like to do is ctlr-click on Post, then in my new tab, Post my link, close that tab, and then back in my tab with /recent scroll down to the bottom , ... I'm back to where I've left off.   Today I can right-click and at least that opens a new tab, but it doesn't have /post, it has a new tab of whatever page I was on.   So I'ld actually like both actions ... right-click, Open Link in New Tab as well as Ctrl-Click, done on the Post button, to launch /post in a new tab.

After looking at the other buttons, I can't really see where I would want this Ctrl-Enter and/or right-click behavior.  I don't think it would hurt, to be consistent maybe, but /post is actually the only button on which I would use this enhancement.

Or, maybe there's some other UI change that would provide the same benefit, ... ability to /post and then easily resume scrolling /recent (for example) without having to re-navigate through the site.  (and please don't suggest hitting Back button about 3 times)

PR #302, opened 2023-06-03
Allow CTRL+click to open /post in new tab

ekzyis:
This at least allows CTRL+click on the post button to open it in a new tab.

`window.open` seems to have good support between browsers so I think it's safe to use:
https://developer.mozilla.org/en-US/docs/Web/API/Window/open#browser_compatibility

Related to #290

https://github.com/stackernews/stacker.news/assets/27162016/2e589c9d-e6b5-4a25-a764-6cd11ae1da9e

### DIFF
--- a/components/header.js
+++ b/components/header.js
@@ -263,7 +263,15 @@ export default function Header ({ sub }) {
       ? (
         <Nav.Link eventKey='post' className={`${className}`}>
           <Link href={prefix + '/post'} passHref>
-            <button className='btn btn-md btn-primary px-3 py-1'>post</button>
+            <button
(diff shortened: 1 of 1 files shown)


huumn:
I don't think this is the right way to approach this. Why does ctrl-click work for other nav links but not the post button? It's because the post button is a button and not a link. We can just make it a link styled like a button.

```jsx
  const PostItem = ({ className }) => {
    return me
      ? (
        <Link href={prefix + '/post'} passHref>
          <a className={`${className} text-reset btn btn-md btn-primary px-3 py-1 `}>post</a>
        </Link>)
      : null
  }
```

huumn:
I went ahead and just pushed my snippet

PR #305, opened 2023-06-06
Add Markdown formatting hotkeys: CTRL+K, CTRL+I, CTRL+B

ekzyis:
As mentioned in our call, I would love to use CTRL+K to insert Markdown link formatting [like in GH](https://docs.github.com/en/get-started/using-github/keyboard-shortcuts#source-code-editing).

In my opinion, this is ready to ship but something to notice is that [`document.execCommand()` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) but still works as mentioned in [this SO answer](https://stackoverflow.com/a/62266439) and is still supported by every browser according to the [section about browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#browser_compatibility).

Also, it is mentioned that when it is unsupported, it will just return `false`: 

> **Return value**
A boolean value that is false if the command is unsupported or disabled.

-- https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#return_value

When it's not supported, only undo for this action will not be usable. It will continue to work afterwards.

I had to use it to fix CTRL+Z since the undo stack is broken after using `helpers.setValue()`.

I found using `document.execCommand()` as the solution [here](https://stackoverflow.com/a/27028258).

I also added hotkeys for bold and italic formatting since it was easy to include in this change.

Showcase:

https://github.com/stackernews/stacker.news/assets/27162016/093c13c9-ff9f-41b4-924c-12663e0abf6d

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -168,7 +168,7 @@ function InputInner ({
         )}
         <BootstrapForm.Control
           onKeyDown={(e) => {
-            if (e.keyCode === 13 && (e.metaKey || e.ctrlKey)) {
+            if (e.key === 'Enter' && (e.metaKey || e.ctrlKey)) {
               formik?.submitForm()
             }
             if (onKeyDown) onKeyDown(e)
--- a/components/form.js
+++ b/components/form.js
@@ -121,24 +121,29 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, setH
   )
 }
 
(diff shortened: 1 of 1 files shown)


ekzyis:
> I had to use it to fix CTRL+Z since the undo stack is broken after using helpers.setValue().

Btw, this may also be relevant in other parts of the code base.

PR #317, opened 2023-06-13
Fix cursor reset after innerRef initialized

### LINKED ISSUE #316: Cursor jumps to start on first keypress after entering edit mode
When entering the edit mode for a comment and pressing any character, the cursor jumps back to the start first.

This is most likely related to https://github.com/stackernews/stacker.news/pull/305

https://github.com/stackernews/stacker.news/assets/27162016/f1bcf829-f2af-43a3-9afa-90f0834e2774

ekzyis:
The error was caused because `innerRef.current` is not initialized on first render. On second render (after the user typed something), it is and thus `useEffect` runs with the default values (start: 0, end: 0); resetting the cursor.

So I fixed it by removing the dependency on `innerRef.current`.

Closes #316

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -93,7 +93,7 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, setH
       const input = innerRef.current
       input.setSelectionRange(start, end)
     }
-  }, [innerRef?.current, selectionRange.start, selectionRange.end])
+  }, [innerRef, selectionRange.start, selectionRange.end])
 
   return (
     <FormGroup label={label} className={groupClassName}>


components/form.js
```diff
@@ -93,7 +93,7 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, setH
       const input = innerRef.current
       input.setSelectionRange(start, end)
     }
-  }, [innerRef?.current, selectionRange.start, selectionRange.end])
+  }, [selectionRange.start, selectionRange.end])
```
huumn, line 96: Just fyi in modern versions of react all dependencies in useEffect must be specified.

huumn: I think the move might be to not depend on innerRef.current and instead on just innerRef which shouldn't change on rerender

ekzyis: > I think the move might be to not depend on innerRef.current and instead on just innerRef which shouldn't change on rerender

I agree. I tested it and it works: https://github.com/stackernews/stacker.news/pull/317/commits/b9286c59c084de7e8f136147b64cdc9706dc39b0

PR #324, opened 2023-06-19
Service worker rework, Web Target Share API & Web Push API

ekzyis:
I replaced `next-pwa` with `workbox-webpack-plugin` for the reasons mentioned in 900b115dd6dd4585cd31efd938c16ab71b113a99.

Also, I added support for the [Web Target Share API](https://developer.chrome.com/articles/web-share-target/) and replaced the previous notification implementation with proper push notifications using the [Web Push Protocol](https://web.dev/notifications/).

Since I was confused for a while what the relationship between the Notification API and the Push API was, here is a primer about "push notifications":
> Push messages and notifications are two separate but complementary technologies.
> Push is the technology for sending messages from your server to users even when they're not actively using your website.
> Notifications is the technology for displaying the pushed information on the user's device.
> It's possible to use notifications without push messaging.
> [...] when we say push notifications we mean the combination of pushing a message followed by displaying it as a notification.
> When we say push messages we are referring to push technology on its own.
> And when we say notifications we're referring to notification technology on its own.

-- https://web.dev/push-notifications-overview/

To generate the VAPID keys, you can run this command: `npx web-push generate-vapid-keys`

TODO:

- [x] Actually use the push subscriptions to send useful notifications to the user based on events
  - [x] send notifications to _all_ stackers in the parent thread instead of only to the parent
- [x] Check if we are safe against XSRF since this is explicitly mentioned as a warning when implementing PushManager subscriptions:

> Warning: When implementing PushManager subscriptions, it is vitally important that you protect against CSRF/XSRF issues in your app. See the following articles for more information:
> - [Cross-Site Request Forgery (CSRF) Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html)
> - [Preventing CSRF and XSRF Attacks](https://blog.codinghorror.com/preventing-csrf-and-xsrf-attacks/)

-- https://developer.mozilla.org/en-US/docs/Web/API/Push_API

- [x] add UI/UX to disable and re-enable push notifications
- [x] check for subscription changes and resubscribe like described [here](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/pushsubscriptionchange_event)

--- feedback from @huumn 

- [x] move checkbox to the right
- [x] show checkbox in both cases (user pressed "Yes" or "No" in alert)
- [x] use `LTREE` operators instead of recursive query

---

Final steps:

- [x] test new code to respect user notification settings 
- [x] test production build using `next build`

https://github.com/stackernews/stacker.news/assets/27162016/3fbf22fd-1b20-4557-aab8-276d3005d2f6

signal-2023-06-19-072122_256x512


ekzyis:
Added push notification for following events:

- replies
- mentions
- zaps


https://github.com/stackernews/stacker.news/assets/27162016/cc6e474a-3658-4ac2-8761-8ccc6ae17286

ekzyis:
Replies and mentions are now merged into one notification:


https://github.com/stackernews/stacker.news/assets/27162016/79aadca8-d867-486c-898d-70caf38bbb4b

ekzyis:
Oh, I forgot that I have to send a notification to every stacker of a thread, not only to the parent. Added TODO in issue description

ekzyis:
Regarding performance of the recursive query to send notifications to every (distinct) user in the thread:

I searched for the item with the most comments as a proxy metric for long threads:

```
stackernews=# select id, title, ncomments FROM "Item" ORDER BY ncomments DESC LIMIT 1;
   id   |                                            title                                             | ncomments
--------+----------------------------------------------------------------------------------------------+-----------
 165552 | I'm losing my faith                                                                          |       172
(10 rows)
```

In this post, I found [this comment](https://stacker.news/items/165842) which I then used to create a query plan:

```
stackernews=# EXPLAIN ANALYZE WITH RECURSIVE base AS (
  SELECT "Item".id, "Item"."parentId", "Item"."userId"
  FROM "Item"
  WHERE id = 165842
  UNION ALL
  SELECT "Item".id, "Item"."parentId", "Item"."userId"
  FROM base p
  JOIN "Item" ON "Item".id = p."parentId")
SELECT DISTINCT "userId" FROM base
WHERE "userId" <> 9506;
                                                                    QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------
 HashAggregate  (cost=858.73..859.73 rows=100 width=4) (actual time=0.130..0.134 rows=3 loops=1)
   Group Key: base."userId"
   Batches: 1  Memory Usage: 24kB
   CTE base
     ->  Recursive Union  (cost=0.42..856.21 rows=101 width=12) (actual time=0.030..0.109 rows=12 loops=1)
           ->  Index Scan using "Item_pkey" on "Item"  (cost=0.42..8.44 rows=1 width=12) (actual time=0.028..0.030 rows=1 loops=1)
                 Index Cond: (id = 165842)
           ->  Nested Loop  (cost=0.42..84.58 rows=10 width=12) (actual time=0.005..0.005 rows=1 loops=12)
                 ->  WorkTable Scan on base p  (cost=0.00..0.20 rows=10 width=4) (actual time=0.000..0.000 rows=1 loops=12)
                 ->  Index Scan using "Item_pkey" on "Item" "Item_1"  (cost=0.42..8.44 rows=1 width=12) (actual time=0.004..0.004 rows=1 loops=12)
                       Index Cond: (id = p."parentId")
   ->  CTE Scan on base  (cost=0.00..2.27 rows=100 width=4) (actual time=0.049..0.119 rows=6 loops=1)
         Filter: ("userId" <> 9506)
         Rows Removed by Filter: 6
 Planning Time: 0.423 ms
 Execution Time: 0.193 ms
(16 rows)
```

LGTM since mostly index scans (no sequential scans). Also, the cost estimate of 859.73 for the root node doesn't seem to be too high.

ekzyis:
Regarding:

> Check if we are safe against XSRF since this is explicitly mentioned as a warning when implementing PushManager subscriptions:

We use [`next-auth`](https://next-auth.js.org/getting-started/client#getcsrftoken) which (unsurprisingly) includes CSRF tokens so I _think_ we are safe. But not entirely sure since maybe this only secures the endpoints for signing in and out?

So I think it would be useful to dedicate some time to search for vulns in the near future.

huumn:
You don't need to perform a recursive query to find ancestors of a comment. We already have to do this for notifications and there are examples of it. 

This is why we store the `LTREE` path. It makes doing these types of queries trivial, e.g. `WHERE p.path @> "Item".path` gives me all `p` that are ancestors of `"Item"`.

You can find `LTREE` operators here: https://www.postgresql.org/docs/current/ltree.html

ekzyis:
Added a checkbox on the /notifications page to allow stackers to unsubscribe:

https://github.com/stackernews/stacker.news/assets/27162016/6aea8737-aef0-452b-b8c2-d3028fdf2871

huumn:
I left you a comment about catching promises and being mindful of notification logic throwing that might interfere with CRUD that should succeed regardless of the notification logic.

Given that Promise.all rejects on the first rejection, it might not be what you want to use if that's not what you intend: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#promise.all_fail-fast_behavior

ekzyis:
> I left you a comment about catching promises and being mindful of notification logic throwing that might interfere with CRUD that should succeed regardless of the notification logic.

Mhh, I don't see any other comment :thinking: 

But I agree. I already tried to prevent any interfering with CRUD but I have noticed the code can be written more defensively. I wrapped the whole body of the functions that are called in resolvers with try/catch now.

> Given that Promise.all rejects on the first rejection, it might not be what you want to use if that's not what you intend: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#promise.all_fail-fast_behavior

Also, I used `Promise.allSettled` instead of `Promise.all` since I just to run promises concurrently but didn't care about the result. So `Promise.allSettled` seems to be a better fit. Do you think I should also add a catch to `Promise.allSettled` even though the promises it wraps should never reject?

huumn:
> Do you think I should also add a catch to Promise.allSettled even though the promises it wraps should never reject?

If they'll never reject, they probably wouldn't/shouldn't be a promise right?

Even if they shouldn't theoretically reject today, they could be changed to reject tomorrow and whoever changes it tomorrow might not know we assumed that would never happen.

I don't know what best exception handling practices are, but I'd guess one is to make sure they are always caught either by the current context or the calling context.

ekzyis:
> If they'll never reject, they probably wouldn't/shouldn't be a promise right?

It wraps promises because the code is async to use `await`. They never reject because the code is wrapped in try/catch so any rejection is already handled within the code. Returning no error from a catch handler means the promise resolved. So not sure what you mean?


> I don't know what best exception handling practices are, but I'd guess one is to make sure they are always caught either by the current context or the calling context.

It's always caught in the current context now using try/catch. See the code in _api/webPush/index.js_. The question was if it should additionally be caught in the calling context to make "super sure"

sw/index.js
```diff
@@ -1,2 +1,92 @@
...
+      // convert keys from ArrayBuffer to string
+      subscription = JSON.parse(JSON.stringify(subscription))
+      const variables = {
+        endpoint: subscription.endpoint,
+        p256dh: subscription.keys.p256dh,
+        auth: subscription.keys.auth,
+        oldEndpoint: event.oldSubscription.endpoint
+      }
+      const body = JSON.stringify({ query, variables })
+      return fetch('/api/graphql', {
```
ekzyis, line 81: Regarding this:

> Although examples demonstrating how to share subscription related information with the application server tend to use [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/fetch), this is not necessarily the best choice for real-world use, since it will not work if the app is offline, for example.
>
> Consider using another method to synchronize subscription information between your service worker and the app server, or make sure your code using fetch() is robust enough to handle cases where attempts to exchange data fail.

-- https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/pushsubscriptionchange_event#usage_notes

I have read somewhere that the [Background Sync API](https://developer.chrome.com/blog/background-sync/) is recommended to use. However, I would keep it as it is now since the worst case that happens is that a user no longer receives push notifications at some point but they can easily resubscribe by checking the button in /notifications again. Also, I would like to not make the code any more complicated in the first iteration. We can make this event listener more stable in next iterations. Then I should know more about how to use Background Sync API and what the gotchas are.

huumn:
afaict this still sends users metion/reply/stacking notifications even if they've opted out of them in settings.

Also, is there anything special I need to do to test this locally?

ekzyis:
> afaict this still sends users metion/reply/stacking notifications even if they've opted out of them in settings.

no because if they opted out, their subscription is deleted in our database

> Also, is there anything special I need to do to test this locally?

you need HTTPS afaik. Service workers only work with HTTPS

huumn:
> no because if they opted out (unchecked the checkbox in /notifications), their subscription is deleted in our database. also, they unsubscribed from the push service so the push service would no longer forward push messages to this client using that subscription.

I meant that in /settings we have "notify me when..." settings that these push notifications don't respect

ekzyis:
> I meant that in /settings we have "notify me when..." settings that these push notifications don't respect

Oh, I see and you are right! Good catch. Will push fix

PR #336, opened 2023-06-28
Allow zapping, posting and commenting without funds or an account

### LINKED ISSUE #266: Allow zapping without funds or an account
This is something I've been wanting to do for awhile. 

It'd be nice to allow lurkers to zap *and* allow stackers to zap arbitrary amounts even if their wallet doesn't have enough sats ... by presenting an invoice for them to pay in a modal.

ekzyis:
Closes #266 

---

~~Current state (444ebf7):~~ _(outdated)_

https://github.com/stackernews/stacker.news/assets/27162016/d410bcb2-f384-46bd-b144-7888768961fe

TODO:

- [x] ~~Fix missing apollo cache update after an invoice was paid~~ _(outdated)_
- [x] test zapping when logged in still works as before
- [x] ~~stop polling invoice after confirmation since it will be deleted shortly after~~ ~~_(using `skip` seemed to throw a bigger error and the current "invoice not found" error is only logged to console so I think stopping the query is not needed :thinking:)_~~ _(mhh, the user may see "\<div>error\</div>" rendered for a very short amount of time before the modal is closed and the mutation is executed... this could confuse users so should be prevented)_
- [x] "allow stackers to zap arbitrary amounts even if their wallet doesn't have enough sats ... by presenting an invoice for them to pay in a modal."
- [x] ~~replace invoice deletion with a different method to make sure invoices are not double-spent~~
- [x] exclude anon user from trust algorithm
- [x] error handling: let user repeat mutation or contact admins if invoice was paid but error happened during mutation
- [x] `itemRepetition` costs are not properly updated

---

@huumn, this is ready for review now (82d94b24ec2f1b56905f58d2d469b68fbb4c8aa0)

here are some showcase videos: https://files.ekzyis.com/public/sn/266-zaps-without-account/

[001.mp4](https://files.ekzyis.com/public/sn/266-zaps-without-account/001.mp4) shows the changes for anon users

[002.mp4](https://files.ekzyis.com/public/sn/266-zaps-without-account/002.mp4) shows the error handling

[003.mp4](https://files.ekzyis.com/public/sn/266-zaps-without-account/003.mp4) shows that stackers can now pay invoices if they don't have enough funds

---

TODO:
- [x] rebase on 566e3f4 and test again


components/item-act.js
```diff
@@ -35,50 +37,110 @@ const addCustomTip = (amount) => {
...
-            sats: Number(amount)
+  const updateCache = (itemSats) => {
+    client.cache.modify({
+      id: `Item:${item.id}`,
+      fields: {
+        sats (existingSats = 0) {
+          return existingSats + itemSats
+        },
+        meAnonSats (anonSats = 0) {
+          return anonSats + itemSats
```
ekzyis, line 58: for some reason, this function is not called thus `meAnonSats` is not updated :thinking:

api/resolvers/wallet.js
```diff
@@ -189,29 +188,30 @@ export default {
   },
 
   Mutation: {
-    createInvoice: async (parent, { amount }, { me, models, lnd }) => {
-      if (!me) {
+    createInvoice: async (parent, { itemId, amount }, { me, models, lnd }) => {
+      if (!me && !itemId) {
         throw new AuthenticationError('you must be logged in')
```
ekzyis, line 193: message could be updated now:

```diff
- you must be logged in
+ you must be logged in or pay for an item
```

api/resolvers/wallet.js
```diff
@@ -189,29 +188,30 @@ export default {
...
       await ssValidate(amountSchema, { amount })
 
-      const user = await models.user.findUnique({ where: { id: me.id } })
+      let user
+      if (me) user = await models.user.findUnique({ where: { id: me.id } })
 
       // set expires at to 3 hours into future
       const expiresAt = new Date(new Date().setHours(new Date().getHours() + 3))
-      const description = `Funding @${user.name} on stacker.news`
+      const description = user ? `Funding @${user.name} on stacker.news` : `Funding item ${itemId} on stacker.news`
```
ekzyis, line 203: Allow lurkers to hide invoice descriptions ?

huumn:
An alternative to rewriting a lot of the code/data models to be user-less is creating an `@anon` user (or similar) and all anon users perform all their actions through that user.

The minimal code rewrite flow for anon zapping might be: pay invoice to `@anon`'s account, have `@anon` zap the item the amount of the invoice. (Doing this atomically might be tricky though.)

Implementing anon posting/commenting will likely be a lot easier if user-less actions are done this way afaict

ekzyis:
> An alternative to rewriting a lot of the code/data models to be user-less is creating an @anon user (or similar) and all anon users perform all their actions through that 

Done :+1: 

> Doing this atomically might be tricky though.

I did this by adding a column `itemId` to the invoice table and updating the `confirm_invoice` function. If an invoice has an `itemId` set, `confirm_invoice` will now trigger a zap by calling `item_act` with that item id and the invoice user id.

I also had to update the `serialize` function though since I wasn't able to create an invoice and set the item id in one query. I noticed that CTEs all run with the same snapshot, so the following didn't work:

```
WITH created_invoice AS (
  SELECT id FROM create_invoice(...)
)
UPDATE "Invoice"
SET "itemId" = $1
FROM created_invoice
WHERE "Invoice".id = created_invoice.id
```

> The sub-statements in WITH are executed concurrently with each other and with the main query. Therefore, when using data-modifying statements in WITH, the order in which the specified updates actually happen is unpredictable. All the statements are executed with the same snapshot (see [Chapter 13](https://www.postgresql.org/docs/current/mvcc.html)), so they cannot “see” one another's effects on the target tables.

-- https://www.postgresql.org/docs/current/queries-with.html

New state with proxy anon user instead of deep invoice integration (87ec14a):


https://github.com/stackernews/stacker.news/assets/27162016/7f231aad-23d0-4c29-be05-38057d8a57fd

api/resolvers/item.js
```diff
@@ -960,6 +961,40 @@ export default {
...
+        throw new AuthenticationError('you must not be logged in')
+      }
+
+      await ssValidate(amountSchema, { amount: sats })
+
+      const user = await models.user.findUnique({ where: { name: ANON_USER_NAME } })
+
+      // set expires at to 3 hours into future
+      const expiresAt = new Date(new Date().setHours(new Date().getHours() + 3))
+      const description = `Zapping item @${id} on stacker.news`
```
ekzyis, line 976: we should probably allow anons to hide invoice descriptions

lib/constants.js
```diff
@@ -22,3 +22,5 @@ export const DONT_LIKE_THIS_COST = 1
 // to be loaded from the server
 export const SUBS = ['bitcoin', 'nostr', 'tech', 'meta', 'jobs']
 export const SUBS_NO_JOBS = SUBS.filter(s => s !== 'jobs')
+
+export const ANON_USER_NAME = 'anon'
```
ekzyis, line 26: we should replace this and all corresponding logic with the anon user id after the user was created in production

huumn:
It looks pretty good on first glance. We have quite a few options for making it generic that we might want to consider before merging (so that we can support a wide variety of anonymous actions).

#### Options for storing on backend and waiting for payment
1. polymophic has-one relationship in "Invoice"
    - create anon versions of all tables that anons might act on by [inheriting from the non-anon table](https://www.postgresql.org/docs/current/tutorial-inheritance.html)
        - e.g. "Item" -> "PeningItem", "ItemAct" -> "PendingItemAct"
    - then add "Invoice"."pendingActionId" and "Invoice"."pendingActionType"
    - when the invoice is confirmed insert into non-anon tables, e.g. insert into "Item" select * from "PendingItem" where "Item".id = "Invoice"."pendingActionId"
2. Store the "actions" as JSON in the "Invoice"
    - we'd then create the actions when the invoice is confirmed
3. Add a "invoiceHash" field to "Item" and "ItemAct" and don't display the "Item" or "ItemAct" until confirmed

#### Options for only sending to backend when payment confirmed
There's only really one way to do this (which might indicate it's the right way):
- poll backend for payment confirmation
- when confirmed, send mutation with the invoice hash
- backend confirms invoice hash is paid, indicates somehow that the invoice hash has been spent, and stores the mutation

api/resolvers/item.js
```diff
@@ -960,6 +961,40 @@ export default {
...
     },
+    anonAct: async (parent, { id, sats }, { me, models, lnd }) => {
+      // this should never be called when logged in
+      if (me) {
+        throw new AuthenticationError('you must not be logged in')
+      }
+
+      await ssValidate(amountSchema, { amount: sats })
+
+      const user = await models.user.findUnique({ where: { id: ANON_USER_ID } })
```
ekzyis, line 972: fetching the user is obsolete since b179e70

huumn:
Another "dimension" of generic that we might want to consider is allowing logged in users to keep a zero/low balance and pay/zap only when they perform actions. 

Most of the generic options above are consistent with that though I think.

api/resolvers/item.js
huumn: Yes their description should be empty by default. We should also probably remove their cowboy hat and remove them from 'top' ... all of which we can do by setting the appropriate fields on the `anon` user record.

ekzyis:
> **Options for only sending to backend when payment confirmed**
> 
> There's only really one way to do this (which might indicate it's the right way):
> 
> - poll backend for payment confirmation
> - when confirmed, send mutation with the invoice hash
> - backend confirms invoice hash is paid, indicates somehow that the invoice hash has been spent, and stores the mutation
> 
> IMO the server-side versions seem very complicated if we're going to allow anons to really use SN. The client-side version isn't atomic but we can make cache the invoice hashes so that they can be retried on the rare failure cases

That does indeed sound good. I'll update code to work like this

huumn:
The only gotcha I can see with this approach is post/comment fee escalation race conditions, e.g.

I pay when there isn't another recent anon post, but before I submit my mutation another anon submits theirs, and then my mutation fails because I paid `1 * 10^0 sats`.

But as we discussed elsewhere, perhaps we just make anon posts/comments 1000 sats or something and disable fee escalation.

ekzyis:
a9ffee3 now uses attached invoice ids as a proof of payment when sending a mutation.

To prevent double spends, the invoice is deleted inside the same transaction.

ekzyis:
Added comment and posting as anons (links, discussions and polls) now and disabled fee escalation for the anon user.
I have set anon comment fees to 100 and post fees to 1000.

TODO:
- [ ] stop polling invoice after confirmation since it will be deleted shortly after
- [ ] "allow stackers to zap arbitrary amounts even if their wallet doesn't have enough sats ... by presenting an invoice for them to pay in a modal." (from #266)

ekzyis:
I realized that double spends are still possible since you can use existing payment hashes (which were used for depositing) to pay for stuff since the backend only checks if there was a paid invoice with that payment hash to allow actions.

So we need to use a different method than deletion to check if an invoice was already spent

huumn:
*note: I haven't looked at the code*

If a user is logged in, we could prevent them from using payment hashes. Instead, their paid invoice *always* goes to their balance, and they *always* spend from their balance.

ekzyis:
I thought I am done now but I think there is one thing missing:
@huumn, iirc, I think we discussed that the anon user should never gain any trust from a) trusted people zapping it or b) it zapping posts other people also like (is that a good summary how trust is gained? i'm still a noob regarding this) since I think the anon user should not be able to affect ranking, right?

huumn:
You'd just skip adding them to values in worker/trust.js:160

ekzyis:
Added some error handling. On first error, the user can retry the mutation. If there is still an error, contacts are shown so they can ask for support or refund:

_first error:_
2023-07-22-215351_1920x1080_scrot

_second error:_
2023-07-22-215400_1920x1080_scrot

components/bounty-form.js
```diff
@@ -49,6 +51,33 @@ export function BountyForm ({
...
+
+      if (item) {
+        await router.push(`/items/${item.id}`)
+      } else {
+        const prefix = sub?.name ? `/~${sub.name}` : ''
+        await router.push(prefix + '/recent')
+      }
+    }, [upsertBounty, router])
+
+  const anonUpsertBounty = useAnonymous(submitUpsertBounty, { requireSession: true })
```
ekzyis, line 79: Maybe `lib/anonymous.js` and all related nomenclature should be renamed the `useAnonymous` hook can now also be used to only handle invoices for stackers and specifically exclude anon users using `{ requireSession: true }` ... so it's more about "pay as you go" than enabling anon users now, I think.

ekzyis, line 79: Maybe `lib/anonymous.js` and all related nomenclature should be renamed since the `useAnonymous` hook can now also be used to only handle invoices for stackers and specifically exclude anon users ... so it's more about "pay as you go" than enabling anon users now, I think.

lib/anonymous.js
```diff
@@ -0,0 +1,214 @@
...
+        >
+          telegram
+        </a>
+        <span className='mx-2 text-muted'> \ </span>
+        <a
+          href='https://simplex.chat/contact#/?v=1-2&smp=smp%3A%2F%2F6iIcWT_dF2zN_w5xzZEY7HI2Prbh3ldP07YTyDexPjE%3D%40smp10.simplex.im%2FebLYaEFGjsD3uK4fpE326c5QI1RZSxau%23%2F%3Fv%3D1-2%26dh%3DMCowBQYDK2VuAyEAV086Oj5yCsavWzIbRMCVuF6jq793Tt__rWvCec__viI%253D%26srv%3Drb2pbttocvnbrngnwziclp2f4ckjq65kebafws6g4hy22cdaiv5dwjqd.onion&data=%7B%22type%22%3A%22group%22%2C%22groupLinkId%22%3A%22cZwSGoQhyOUulzp7rwCdWQ%3D%3D%22%7D' className='nav-link p-0 d-inline-flex'
+          target='_blank' rel='noreferrer'
+        >
+          simplex
+        </a>
```
ekzyis, line 53: The links for sphinx and simplex are group chats. Should we replace them with DM links like for TG?

lib/anonymous.js
```diff
@@ -0,0 +1,214 @@
...
+  }
+  return (
+    <>
+      <QrInvoice invoice={data.invoice} {...props} />
+      {errorCount > 0
+        ? (
+          <>
+            <InvoiceStatus variant='failed' status={errorStatus} />
+            {errorCount === 1
+              ? <div className='d-flex flex-row mt-3 justify-content-center'><Button variant='info' onClick={repeat}>Retry</Button></div>
```
ekzyis, line 86: Not sure if this retry button is very useful. As it is, it is very naive and just literally retries the action, hoping it would succeed now.

We could try to parse the error and check if there is something that can be done. For example, if the error is insufficient fees, we could update the invoice amount :thinking:

lib/anonymous.js
huumn: Let's not make more work for ourselves yet. This is good enough.

lib/anonymous.js
huumn: I think the retry button is better than nothing and its sufficient to ship it

components/bounty-form.js
huumn: Yep, seems like a good idea to rename it now if that's the way it's being used.

huumn:
In general this looks good. It's really excellent how much functionality was added with so few backend changes.  

A few off-the-bat high-level things:

- files that are mostly components should not reside in `lib`, e.g. lib/anonymous ... to date, `lib` is mostly for code needed by both the server/client but doesn't use react
- anything named `anonymous` that is used beyond anonymous actions, should be renamed
- because the invoice hash and preimage (sigh) are public, we probably shouldn't use them as a token for payments
    - while it's unlikely this public information is leveraged in a meaningful way, we can avoid it by creating a secret token for every invoice and having the client use that token for payment instead of the hash

I'll continue testing/reviewing in the meantime.

pages/invoices/[id].js
Comment (huumn) [inline pages/invoices/[id].js:1]: This isn't used

pages/invoices/[id].js
```diff
@@ -6,7 +6,7 @@ import { useRouter } from 'next/router'
 import { INVOICE } from '../../fragments/wallet'
 import { SSR } from '../../lib/constants'
 
-export default function FullInvoice () {
+export default function FullInvoice ({ id }) {
```
Comment (huumn) [inline pages/invoices/[id].js:9]: This change isn't used

components/job-form.js
```diff
@@ -50,6 +51,40 @@ export default function JobForm ({ item, sub }) {
     }`
   )
 
+  const submitUpsertJob = useCallback(
+    // we ignore the invoice since only stackers can post bounties
```
huumn, line 55: Comment applies to bounties

components/discussion-form.js
```diff
@@ -57,20 +78,8 @@ export function DiscussionForm ({
...
-        }
-
-        if (item) {
-          await router.push(`/items/${item.id}`)
-        } else {
-          const prefix = sub?.name ? `/~${sub.name}` : ''
-          await router.push(prefix + '/recent')
-        }
+      onSubmit={handleSubmit || (async ({ boost, cost, ...values }) => {
+        await anonUpsertDiscussion(cost, boost, values)
```
huumn, line 82: The naming of these functions is confusing. It makes it seem like everything is anonymous.

lib/anonymous.js
```diff
@@ -0,0 +1,214 @@
...
+    mutation createInvoice($amount: Int!) {
+      createInvoice(amount: $amount) {
+        id
+        hash
+      }
+    }`)
+  const showModal = useShowModal()
+  const [fnArgs, setFnArgs] = useState()
+
+  // fix for bug where `showModal` runs the code for two modals and thus executes `onConfirmation` twice
```
huumn, line 118: Are you sure this isn't a react strictMode thing indicating you're maybe making a mistake somewhere?

huumn:
Instead of adding a token to every "Invoice" we could solve this statelessly with using some kind of cryptography on the invoice hash. We could sign it and somehow encode the signature into the token or encrypt it. 

I think this is the way to go as it would make this change a lot easier.

components/discussion-form.js
ekzyis: I renamed `useAnonymous` to `useInvoiceable` now. The `anon` suffix was replaced with `invoiceable`.

I first wanted to use `useInvoice` but that sounded too much like it would return an invoice like `useMe` returns the current user.

ekzyis:
> Instead of adding a token to every "Invoice" we could solve this statelessly by using some kind of cryptography on the invoice hash. We could sign it and somehow encode the signature into the token or do some kind of [authenticated encryption](https://en.wikipedia.org/wiki/Authenticated_encryption).
>
> I think this is the way to go as it would make this change a lot easier.

I used a HMAC now. This meets following requirements:

| Requirement | Solution |
| --- | --- |
| Only the user who paid the invoice can use the payment hash to pay for actions on SN | Paying via payment hash also requires knowledge of the HMAC and HMAC is only sent once during invoice creation to the browser using HTTPS (out of band like payment hashes) |
| We don't need to keep any state to verify hashes | Create HMACs with a single secret key. No state required |
| HMAC can't be reused | HMACs are never reused because hashes are never reused in LN |

This is not really authenticated encryption (since there is no encryption involved) but I think this meets our requirements.

lib/anonymous.js
ekzyis: It was. Solution was to use `useEffect`

huumn:
This is really cool overall. I just push some changes to make the feature more obvious. A few things we'll need to fix before deploying:

1. anon zappers can't zap anon poster/commenters which is unexpected
2. I should be allowed to retry on failure as many times as I want (while also being given the option to contact)
3. I'm not sure if this is addressed, but there are rate limits around invoice creation that will probably need to be relaxed if we are multiplexing many people onto a single account ... they should probably have different rules all together (default invoice expiration is 3hrs for instance which doesn't make sense here)
4. we should provide a back button so someone can change their mind after viewing the invoice

This is what I've found so far. I'll continue trying to get full coverage and committing things as I go. If you end up working on any of this lmk

PR #349, opened 2023-07-07
Add nym in reply notification

ekzyis:
signal-2023-07-07-045735


ekzyis:
We can also include images.
I was thinking about parsing the body for image links and then including the first image as a proper image in the notification (I think the spec doesn't allow multiple images). The image should then be visible in your push notification.

I was thinking about this since sometimes, I see markdown image formatting in my notification which isn't useful. This markdown image formatting should then probably be removed in the body if it's included as an image.

PR #353, opened 2023-07-12
Show longest cowboy streak in profile

ekzyis:
no streak:

2023-07-13-003948_937x272_scrot

new streak:

2023-07-13-012025_924x266_scrot

long streak:

2023-07-13-012102_946x312_scrot

TODO: 
- [x] fix image edit icon offset
- [ ] ~~check if code in `computeStreaks` works as expected~~ _(outdated)_
- [x] create one-time query to calculate longest streaks for all users


huumn:
This doesn't need to be stored/denormalized in the users table. We should only be denormalizing when an expensive query (or series of queries) is on a critical path. This is only ever shown on the stacker's profile.

ekzyis:
> This doesn't need to be stored/denormalized in the users table. We should only be denormalizing when an expensive query (or series of queries) is on a critical path. This is only ever shown on the stacker's profile page.

Done in 908d201

PR #380, opened 2023-08-07
Singular/Plural "Sats" labels

SatsAllDay:
This PR updates all text in the app that use "sats" as a unit to handle singular and plural variations e.g.

* `0 sats`
* `1 sat`
* `2, 3, 4 sats...`

The implementation is done by adding one new common formatting function that is then used throughout the app. The common function also performs numerical abbreviation (e.g. `1m` instead of `1,000,000`) using an existing utility function.

One consideration: there are probably places where we want to show an exact number of sats instead of an abbreviated number, so perhaps the new common utility function can be updated to accept an optional boolean parameter that allows you to disable the number abbreviation.


SatsAllDay:
> One consideration: there are probably places where we want to show an exact number of sats instead of an abbreviated number, so perhaps the new common utility function can be updated to accept an optional boolean parameter that allows you to disable the number abbreviation.

@huumn Any thoughts on this? I think I will implement that as part of the PR, and opt out of number abbreviation in places like invoices, etc.

huumn:
Yes! I think a sat formatter tool makes a lot of sense

SatsAllDay:
> One consideration: there are probably places where we want to show an exact number of sats instead of an abbreviated number, so perhaps the new common utility function can be updated to accept an optional boolean parameter that allows you to disable the number abbreviation.

I went ahead and updated the PR to do this. By default, the sat formatting function will abbreviate numbers, but passing a `false` second param will opt out of it. I mostly kept the number abbreviation logic in place as it was, with maybe one or two exceptions. I aimed to ensure that for places where knowing exact counts is of interest, they remained non-abbreviated.

SatsAllDay:
@huumn Absolutely no rush, I just wanted to check what the process was to mark a PR as ready for review by a maintainer? I don't seem to have the ability to request a review. TIA!

huumn:
I'll be looking to clear out the PR queue before the end of the week. So this should get in the next few days.

PR #387, opened 2023-08-09
Debounce API requests on edit nym by 500ms

### LINKED ISSUE #257: Inconsistent behavior while checking if nym is already taken
https://user-images.githubusercontent.com/27162016/230557133-a3a57655-cd42-4684-a4e0-5a422ba2c45b.mp4

SatsAllDay:
Support an optional debounce prop on Input component

If provided, the debounce is applied to the validation of the containing form, imperatively invoking form validation after debounce is finalized

This also required introducing the `validateOnChange` prop on `Form` which gets passed to `Formik`, and defaults to true, just as it does in `Formik`.

Imperatively invoking form validation seemed to be the only way to debounce the validation call through formik.

Closes #257

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -219,7 +219,7 @@ function FormGroup ({ className, label, children }) {
 
 function InputInner ({
   prepend, append, hint, showValid, onChange, overrideValue,
-  innerRef, noForm, clear, onKeyDown, ...props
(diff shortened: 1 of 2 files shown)


huumn:
That was fast! Nice work!

SatsAllDay:
Thank you!🙏

Issue #390, opened 2023-08-10
Light/Dark Splash Screen on iOS PWA


SatsAllDay:
Currently when using the PWA on iOS, the splash screen is the black SN text with yellow background. For those that prefer a dark theme, this can be pretty jarring.

I would like to propose that we create a "dark" alternative splash screen - basically flip the colors of the existing one (yellow text, black background) for those iOS users who have dark theme enabled on their device. 

From a technical perspective, I believe this will require the following:

1. Create all new splash images to accommodate the various Apple device sizes with the colors swapped (black background, yellow text)
2. Update the _document.js page to add these new splash image link elements, with the same media queries _except_ add an additional _and_ condition to only enable them if dark mode is preferred. This should give us the dynamism desired. 

I don't know for sure, but I don't think we have the same options for Android since it's computed from the manifest file and doesn't expose any "dynamic" hooks like we have with the media queries on iOS?

Anyway, I'm filing an issue for this to start the discussion and reach agreement, then I'll open a PR with the changes. I've briefly started messing with this locally but I don't have it in a state to share yet.

huumn:
I agree that this would be preferred. I hadn't thought of adding other queries to the links. 

> I don't know for sure, but I don't think we have the same options for Android since it's computed from the manifest file and doesn't expose any "dynamic" hooks like we have with the media queries on iOS?

Android doesn't give you as much control afaik. They put the icon over your background color. Because I wanted a yellow icon, I used a yellow background and did the same on iOS.

SatsAllDay:
Regarding android support, I read [this article](https://web.dev/user-preference-media-features-headers/) and it gave me the following idea:

Basically, this example is how we could enable it: https://web.dev/user-preference-media-features-headers/#how-it-works

1. Update our server to solicit the `Sec-CH-Prefers-Color-Scheme` client hint via request header from the client.
2. Instead of serving the manifest as a static resource, we could serve it dynamically, meaning we could compute the manifest response payload as a function of the request. Use the client hint, if provided, to alter the manifest's color values to be light/dark accordingly. Default to light if not provided to maintain the current experience. 

One issue I see here is, I don't know how to keep the yellow-background icon for the home screen, but use a dark-background icon for the splash screen, since those are coupled...

huumn:
Dynamically serving it is a really interesting solve. Nice find!

PR #393, opened 2023-08-10
iOS splash screen for dark-theme

SatsAllDay:
Add images for each supported iOS device for a dark-theme splash screen

Conditionally use light vs dark images based on the `(prefers-color-scheme: dark)` media query

Note: Once the PWA is installed, it appears that iOS caches the splash screen image for an unknown period of time, so if you change your device's preferred theme after installation, the splash screen doesn't update to reflect it. The PWA must be re-installed for this change to take effect.

Notable resources in making this PR possible:
1. https://progressier.com/pwa-icons-and-ios-splash-screen-generator
2. https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
3. https://developer.apple.com/design/human-interface-guidelines/layout#Specifications
4. https://web.dev/learn/pwa/enhancements/#splash-screens

Addresses (part of) #390. I'll do some more research on the android aspect, but that can come in a follow-up PR, IMO.


huumn:
Looks good! I'll ship this today. The manifest pr will require a little more review

huumn:
Never mind shipping tomorrow morning so I don't have to put out fires tonight lol.

PR #395, opened 2023-08-11
Fix link to profile matching current URL, not profile

### LINKED ISSUE #84: Link for Profile matches current URL, not Profile
Though clicking on the link for my nym causes a drop down as expected, if I right-click on that link and Open in a new tab, the URL for that nym will match the current page, and not my profile.

To reproduce issue:
- Log into Stacker.News
- Navigate to something like: https://stacker.news/recent
- Hover over the link for the nym

Expected result:
The URL showing for that num to be the Profile, same as if I had clicked on the link and gotten the drop-down
i.e., Stacker.News/[nym]

Observed result:
The URL matching the current page
i.e., https://stacker.news/recent, if that was the current page

rleed:
It seems bootstrap never really contemplated the idea of supplying links to top-level dropdown buttons, which are not really links anyway and only function to toggle open the dropdown menu. So to remove the bug, I changed the top-level element from a link to a button to get rid of the nonsensical href="#". This way, anyone who right-clicks on it will not find a copy-link option, and that would be their hint to open the drop down and right-click on the actual profile link instead to copy the correct link.

Closes #84


huumn:
rleed I'll try to take a look at these tomorrow. Apologies for not acknowledging them sooner

ekzyis:
LGTM, checked following things:

- Dropdown toggle no longer behaves like a link: context menu no longer includes "Open link in ..."; CTRL+click no longer opens a new tab etc.
- Same on mobile: long press no longer opens link menu
- Thought style change may influence other Dropdowns but all of them use `a` instead of `button` so no changes there

This might not be the fix that was requested in #84 but is still a viable solution imo. Just opening the same page was definitely bad UX. Now it's clear that you need to use the profile entry as a link to your profile as @rleed mentioned.

There are some minor conflicts though. Can you rebase @rleed ?

rleed:
I've rebased...thanks

ekzyis (approved):

Issue #185, opened 2022-08-14
Trim leading and trailing whitespace for the 'forward sats to' option when doing a Link post 


cointastical:
I copied and pasted a nym and didn't realize there was a trailing space.   A message `user does not exist` showed.  In testing, I see the same occurs if there is a leading space. 

And, ... while on the subject.  If I paste the nym with the "@" (@nym) that will also result in "user does not exist".  This one i'm not sure about since the @ is the field label.   I think it would be OK to strip out any @ prefix if exists and then use that value as if that is what the user had entered.

ekzyis:
Maybe whitespace shouldn't even be allowed in the input? Can usernames contain whitespace?

cointastical:
> I don't like intransparent solutions like trimming whitespace in the backend. 

That's not a universal sentiment on trimming input field data.

There's some like this:

> Your backend SHOULD ALWAYS assume the data coming from the frontend is GARBAGE and do its own clean-up, parsing and validation.

https://www.reddit.com/r/golang/comments/oa8uqd/comment/h3k4nl5

Other examples are plentiful, and can vary (e.g., never trim passwords, which I wouldn't agree with but whatever).

Copy and paste results by those copying from html will oftentimes end up with input field data containing leading or trailing whitespace.  That we aren't frustrated by "error" messages when this happens is that these are commonly trimmed regularly, on the back-end oftentimes, when the form is submitted.

As far as invalid characters within the data (e.g., username:  Firstname Lastname) I would expect a response message (e.g., "username not found" or "contains invalid character").  But a leading / trailing whitespace on a username is certainly extraneous and unintentional, and thus reasonable that only the trimmed value is used.

ekzyis:
> Your backend SHOULD ALWAYS assume the data coming from the frontend is GARBAGE and do its own clean-up, parsing and validation.
https://www.reddit.com/r/golang/comments/oa8uqd/comment/h3k4nl5

Ah, good point, I forgot about that. You are right. It makes sense to _also_ trim in the backend etc. due to this

> Copy and paste results by those copying from html will oftentimes end up with input field data containing leading or trailing whitespace. That we aren't frustrated by "error" messages when this happens is because these are commonly trimmed regularly, on the back-end oftentimes, when the form is submitted.

[...]

PR #397, opened 2023-08-11
Forward sats to user: text trimming and menu click-out UX improvements

rleed:
Approach: on input, check for leading/trailing spaces and the leading @ symbol on the username (e.g. when pasting a username with extra characters) and strip them. This has the effect of also restricting the user from typing such characters too.

While working on this, I discovered the drop-down suggestions menu was too sticky; it covered the post button and virtually obliged the user to pick a name in the list (even if not necessary because it was pasted in, e.g.) so I added logic to hide the menu when the user clicks out or tabs away.

(issue #185)

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -218,7 +218,7 @@ function FormGroup ({ className, label, children }) {
 }
 
 function InputInner ({
-  prepend, append, hint, showValid, onInput, onChange, onBlur, overrideValue,
+  prepend, append, hint, showValid, onChange, onBlur, overrideValue,
   innerRef, noForm, clear, onKeyDown, debounce, ...props
 }) {
   const [field, meta, helpers] = noForm ? [{}, {}, {}] : useField(props)
--- a/components/form.js
+++ b/components/form.js
@@ -321,6 +321,14 @@ function InputInner ({
 }
 
 export function InputUserSuggest ({ label, groupClassName, ...props }) {
+  const formik = useFormikContext()
+  const blurFunc = (e) => {
+    formik.handleBlur(e)
+    if (!document.getElementsByClassName(groupClassName)[0].contains(e.relatedTarget)) {
+      setSuggestions(INITIAL_SUGGESTIONS)
+    }
+  }
+
   const [getSuggestions] = useLazyQuery(USER_SEARCH, {
     onCompleted: data => {
       setSuggestions({ array: data.searchUsers, index: 0 })
@@ -370,11 +378,7 @@ export function InputUserSuggest ({ label, groupClassName, ...props }) {
               break
           }
         }}
-        onBlur={(e) => {
-          if (!document.getElementsByClassName(groupClassName)[0].contains(e.relatedTarget)) {
-            setSuggestions(INITIAL_SUGGESTIONS)
-          }
-        }}
+        onBlur={blurFunc}
       />
       <Dropdown show={suggestions.array.length > 0}>
         <Dropdown.Menu className={styles.suggestionsMenu}>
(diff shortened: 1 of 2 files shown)


components/form.js
```diff
@@ -375,6 +386,11 @@ export function InputUserSuggest ({ label, groupClassName, ...props }) {
                 setOValue(v.name)
                 setSuggestions(INITIAL_SUGGESTIONS)
               }}
+              onBlur={(e) => {
+                if (!document.getElementsByClassName(groupClassName)[0].contains(e.relatedTarget)) {
+                  setSuggestions(INITIAL_SUGGESTIONS)
+                }
+              }}
```
ekzyis, line 393: I can see that you need the `if` check in the input `onBlur` listener to be able to still select items. The `onClick` listener then seems to close the dropdown.

But do you need _this_ `onBlur` listener here? It seemed to work without this.

Also, can't this be solved in a more "react-y way"; without using `document.getElementsByClassName` For example with `useRef`? I didn't try yet though.

ekzyis (commented):
It's hard to test this properly since I think `<InputUserSuggest>` in general is buggy. See #409. For example, already on 41463d7, one can't reselect users. Maybe there is more broken stuff.

But I noticed that with these changes, the icon if the input is valid no longer appears:

https://github.com/stackernews/stacker.news/assets/27162016/196f01b3-5707-4cc7-a4c3-d6f5e076db86

@rleed, can you confirm on your end?

I encountered strange things with this input with and without your changes while testing so I am not sure if your changes actually cause this.

I was also wondering why we can't use the input property [`pattern`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) for this instead of the new custom prop `autoRestrict`. When I set it, it doesn't seem to have any effect even though I can see it set in the raw HTML. It is at least mentioned in the validation chapter of Formik: https://formik.org/docs/tutorial#validation

rleed:
Re: testing, yes, the validation seems buggy or at least quirky and I also noticed that reselecting isn't recognized. This PR was intended as an incremental improvement...we could take a closer look at those other oddities in another PR.

I noticed the validation didn't happen till attempting to submit the form, and I seem to remember seeing something in the code to deliberately disable validation initially... I'll check that, because I didn't think I changed that behavior.

Using pattern would only give a validation signal but would not do anything to strip the redundant characters when pasting a username. This was considered as a possible behavior in the issue discussion, but IMO it doesn't make for a smooth UX, which is why I ended up restricting the input instead.

components/form.js
rleed: The simplest way to think about it is that if the focus leaves the whole group (inclusive of the input box and the suggested items) then and only then do we want to hide the menu. The blur can be triggered by clicking or by forward/backward tabbing. So the blur handler hides the menu IF the new focused element is anything outside the group. And for that, all elements in the group that can get focus have to have the blur function.

I wanted to use useRef (commonly used for focus-handling use cases) but unfortunately useRef is no longer supported on function components it seems.

rleed:
> But I noticed that with these changes, the icon if the input is valid no longer appears:

Good catch, thanks. (My blur handlers were overriding Formik's behavior.)

rleed:
So if all are ok with my answers to @ekzyis, I think this one is ready

huumn:
Thanks so much! Please keep in the mind the following feedback as you continue to submit prs.

I ended up removing the blur handler because while it's nice to have, it was a bit of hack and hard to follow. re: useRef. If you want to forward a ref from a parent to a child, you wrap the child in [`forwardRef`](https://react.dev/reference/react/forwardRef).

I also updated the change to reuse the code and patterns that were already present instead of writing new ones.

PR #401, opened 2023-08-14
Allow refunds using payment tokens

ekzyis:
https://github.com/stackernews/stacker.news/assets/27162016/9de44ebf-2d30-4095-b1ef-2f3f811b81af

TODO:
- [ ] set max tx fee in backend which we will pay. I think we should pay refund tx fees but there should be an upper bound.
- [ ] refactor code (for example, the hash XOR op could be put into a function)
- [ ] add link to /refunds in error modal
- [ ] test XOR reversal on failed withdrawals
- [ ] make sure using XOR to "block" payment tokens (while still being able to reverse blocking) is indeed a secure way to prevent stealing from other anons
- [ ] is showing balance in top-right corner good UX?
- [ ] fix possible privacy leak for anon withdrawals

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -1,6 +1,5 @@
 import { createInvoice, decodePaymentRequest, payViaPaymentRequest } from 'ln-service'
 import { GraphQLError } from 'graphql'
-import crypto from 'crypto'
 import serialize from './serial'
 import { decodeCursor, LIMIT, nextCursorEncoded } from '../../lib/cursor'
 import lnpr from 'bolt11'
@@ -10,6 +9,7 @@ import { msatsToSats, msatsToSatsDecimal } from '../../lib/format'
 import { amountSchema, lnAddrSchema, ssValidate, withdrawlSchema } from '../../lib/validate'
 import { ANON_BALANCE_LIMIT_MSATS, ANON_INV_PENDING_LIMIT, ANON_USER_ID, BALANCE_LIMIT_MSATS, INV_PENDING_LIMIT } from '../../lib/constants'
 import { datePivot } from '../../lib/time'
+import { createInvoiceHmac, xor } from '../../lib/crypto'
 
 export async function getInvoice (parent, { id }, { me, models }) {
   const inv = await models.invoice.findUnique({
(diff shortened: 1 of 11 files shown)


api/resolvers/wallet.js
```diff
@@ -64,7 +55,7 @@ export default {
         }
       })
 
-      if (wdrwl.user.id !== me.id) {
+      if (me ? wdrwl.user.id !== me.id : wdrwl.user.id !== ANON_USER_ID) {
```
ekzyis, line 58: Possible privacy leak: stackers could enumerate withdrawals and find withdrawals for anons including bolt11 invoice.

api/resolvers/wallet.js
```diff
@@ -328,9 +319,32 @@ export default {
...
+      }
+      // xor last half of invoice hash with HMAC of first half.
+      // this makes the invoice not usable while withdrawal is in process.
+      // if withdrawal failed, we can reverse this operation.
+      const uHash = xor(
+        Buffer.from(oHash, 'hex'),
+        Buffer.concat([
+          Buffer.alloc(16), Buffer.from(createInvoiceHmac(oHash.slice(0, 32)), 'hex').slice(0, 16)
+        ])
+      ).toString('hex')
```
ekzyis, line 343: `new_hash = hash[:16] || (hash[16:] ^ HMAC_k(hash[:16])[:16])`

This is kind of a hack. But is it a good hack? :thinking: 

This meets the following requirements afaict:
- [ ] anons can't use payment tokens again while withdrawal is in process
- [ ] payment tokens can be used again if withdrawal failed since XOR is reverseable using the hash[:32] which wasn't changed
- [ ] anons can't predict the new hash (= hash[:32] || hash[32:] ^ HMAC[32:]) since the HMAC uses a secret key. Knowing the HMAC of the full hash does not give them any knowledge of the HMAC of the first half (need to double-check this)

api/resolvers/wallet.js
```diff
@@ -328,9 +319,32 @@ export default {
...
+      }
+      // xor last half of invoice hash with HMAC of first half.
+      // this makes the invoice not usable while withdrawal is in process.
+      // if withdrawal failed, we can reverse this operation.
+      const xHash = xor(
+        Buffer.from(oHash, 'hex'),
+        Buffer.concat([
+          Buffer.alloc(16), Buffer.from(createInvoiceHmac(oHash.slice(0, 32)), 'hex').slice(0, 16)
+        ])
+      ).toString('hex')
```
ekzyis, line 343: `new_hash = hash[:16] || (hash[16:] ^ HMAC_k(hash[:16])[:16])`

This is kind of a hack. But is it a good hack? :thinking: 

This meets the following requirements afaict:
- anons can't use payment tokens again while withdrawal is in process
- payment tokens can be used again if withdrawal failed since XOR is reverseable using `hash[:16]`
- anons can't predict the new hash since the HMAC uses a secret key. Knowing the HMAC of the full hash does not give them any knowledge of the HMAC of the first half (need to double-check this)

PR #404, opened 2023-08-14
Configure imgproxy timeouts + add click to load on imgproxy errors

### LINKED ISSUE #402: imgproxy blocks too many images
**Description**

Too many images are blocked by our imgproxy. Also, stackers can't see in the preview if it will work or not.

**Steps to Reproduce**

Example image links:
- https://i.postimg.cc/cLvhR9RZ/image.gif
- https://i.postimg.cc/PJ4JFBqy/bullshit.gif

**Expected behavior**

Images do render; stackers can see in the preview if it will render or not; if images don't render, at least show original URL (without automatic rendering). We could also implement "click to load" for original URLs like some nostr clients.

**Additional context**

https://stacker.news/items/224823

ekzyis:
https://github.com/stackernews/stacker.news/assets/27162016/14b4f919-95c5-47e3-b0bf-43584add0d92

Close #402


components/text.js
```diff
@@ -226,4 +256,11 @@ export function ZoomableImage ({ src, topLevel, ...props }) {
       {...props}
     />
   )
+
+  return (
+    (!me || !me.clickToLoadImg || isImgProxy)
```
ekzyis, line 261: For UX reasons, anons don't have to click to load images and because our privacy settings are false by default. So creating an account does not suddenly require click to load for external images (since it's off by default).

ekzyis, line 261: For UX reasons, anons don't have to click to load images and because our privacy settings are false by default. So creating an account does not suddenly require click to load for external images.

huumn:
We're still mutating the input markdown in this by the looks of it. 

I thought we agreed we shouldn't. Was it too hard to stop doing that with signed urls?

ekzyis:
> We're still mutating the input markdown in this by the looks of it.
>
> I thought we agreed we shouldn't. Was it too hard to stop doing that with signed urls?

Ah yes, we still mutate the input markdown.

But I realized the original URL is still available so it was faster to fix the issue by continuing with the replacement.

Also, the replacement is reverse-able at any time. I think the biggest argument against replacement was that the original URLs are no longer accessible which is not the case. We only sign the URLs, we don't encrypt them or anything else. The original URL is still available in base64Url encoding inside the imgproxy URL. So I am not sure if not using replacement is worth it :thinking:

The only other argument against replacement is to make the original URL accessible for stackers in any case (not only on errors). But I don't know when a stacker would want the original URL.

Issue #167, opened 2022-06-18
Improvement request: UI for Withdraw to more clearly show distinct choices  


cointastical:
Currently the Withdraw page has the LNUrl withdraw and Lightning Address methods below the Withdraw button. It's easy for someone who doesn't know those exist to overlook those.  I myself had not known Lightning Address withdraw was possible until very recently (and I withdraw frequently).  This undoubtedly happens to others, as evidenced by this post, today:

*[Add withdrawals from stacker.news via Lightning Addresses*
https://stacker.news/items/37290

So this improvement request is to re-orient the options so that it is more clear that there are three wtihdrawal methods supported:

For example, using radio-button select options:

Select withdrawal method:

* Invoice
* Lightning address
* QR Code

(and Invoice can remain default)

Then below that are the fields and button (or whatever is needed) specific to that withdrawal method.

versieboer:
I see it now.  The box asking for an Invoice appears first.  As soon as I saw that, I stopped reading there, knowing that a Lightning Address was not an invoice.

What would help would be a radio button at the top, pre-selected to withdraw via Invoice.  Other radio button options would be "QR code" and "Lightning Address".  When selected, the current "Invoice" field should change accordingly.

huumn:
I guess what we are solving for here is the visibility of withdrawal options. Radio buttons could do it but it's kind of a sucky UI.

ekzyis:
What about simple input with "Enter invoice or LN address ..." as placeholder + QR code next to it?
It should be easy to detect if the user entered an invoice (starts with "lnbc1..." [0]) or LN address

[0] https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#human-readable-part

PR #414, opened 2023-08-16
Make withdrawal options more visible

rleed:
The discussion of issue #167 offered some ideas for a good way to fix the easily overlooked withdrawal options, such as radio buttons. In this PR, I implemented that idea in the form of tabs, which I think is intuitive and visually organized, but let me know your opinions. I'm not convinced it fits the look of the site.

I also moved the history link up under the balance because that seems like a more logical place for it in terms of the flow of thought as a person reviews their balance, perhaps wants to understand it by clicking the history link, or just moves down to choose a withdrawal method and/or enter the data.

(EDIT: I see my linter broke...I will push a fix for the standard errors.)

### DIFF
--- a/pages/wallet.js
+++ b/pages/wallet.js
@@ -43,7 +43,7 @@ function YouHaveSats () {
 
 function WalletHistory () {
   return (
-    <div className='pb-5' style={{fontWeight: 500}}>
+    <div className='pb-5' style={{ fontWeight: 500 }}>
       <Link href='/satistics?inc=invoice,withdrawal' className='nav-link p-0'>
         wallet history
       </Link>
@@ -131,7 +131,7 @@ export function FundForm () {
   )
 }
 
-export function WithdrawalMethods() {
+export function WithdrawalMethods () {
   const router = useRouter()
 
   return (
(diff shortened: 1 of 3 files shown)


ekzyis:
Can you share screenshots so we can give faster and more precise feedback since we don't need to checkout and run the code first?

Maybe we should mention this somewhere that UI changes should include screenshots :thinking:

rleed:
image(1)
image

huumn:
This looks good. 

Did react-bootstrap not natively have great tabbing? This seems to use quite a bit of custom css

rleed:
I feel that the amount of CSS is inordinate too, but it looked pretty tacky without it. The .content style needed to be brought into each page, flex parameters, grid parameters, margins, borderlines, etc... but honestly, in the end, I am not really happy with it though.

Today I noticed the underline tabs on the profile page, and I think it might be better/simpler to do something like that. What do you think about canceling this PR and letting me try a similar look as the profile page? I personally think that would be more satisfying (and hopefully with less CSS).

Another consideration you can't see in the picture is that the vertically centered layout makes the tabs jump vertically sometimes. The more I think about it, I do think more work needs to go into this.

huumn:
It's up to you. If you leave it as is, I'll probably end up massaging it a bit before deploying.

pages/wallet.js
```diff
@@ -127,7 +127,46 @@ export function FundForm () {
...
+            ? 'qrcode'
+            : router.query.type === 'lnaddr-withdraw'
+              ? 'lnaddr'
+              : 'invoice'
+        }
+        id='withdrawal-tabs'
+        justify
+        mountOnEnter
+        onSelect={(k, e) => {
+          router.replace(
```
huumn, line 152: Nit: This might be better as `router.push(..., ..., {shallow: true})` so that back/forward navigation can be used.

rleed:
I've improved this PR by top-aligning the tabs so they don't jump based on the vertical size of the form.

I've also created a second branch `issue167a` with a variation that I like better:

image(5)
^-- branch `issue167a` --^

This version uses the underline "tab" style of the profile page (and reduces style sheet complexity considerably).

It's your call, and feel free to tweak either one, but I wanted to provide both options.

huumn:
Closing because I prefer your other pr

Issue #340, opened 2023-07-03
Add Onboarding Banner On First Visit


kerooke:
**Is your feature request related to a problem? Please describe.**
New visitors and new stackers have no added context when they first visit Stacker News. Everyone gets the same experience regardless of their knowledge or familiarity of Stacker News. It's likely that Stacker News is losing some people who want to experience the site but need some help getting started.

**Describe the solution you'd like**
A blue alert bar (just like the Stacker News Live alert bar) should be added to Stacker News whenever someone either visits Stacker News for the first time or logs into their account for the first time. This is the MVP of an on-site onboarding experience and should help with retention of people who need a helping hand to get started.

There are early signs of marginal improvements in retention among new stackers after introducing email onboarding, and this feature could level up the onboarding experience one step further without adding much clutter to the site.

For new stackers, the copy should read: "Welcome to Stacker News! To get started, check out our [FAQs](https://stacker.news/faq), [content guidelines](https://stacker.news/guide), or go ahead and make [your first post](https://stacker.news/post)."

For new visitors, the copy should read: "Welcome to Stacker News! To get started, check out our [FAQs](https://stacker.news/faq), [content guidelines](https://stacker.news/guide), or go ahead and [sign up](https://stacker.news/signup)."

This message should have an X allowing people to remove the banner, and should only appear on their first visit or first logged-in visit to Stacker News.

**Describe alternatives you've considered**
Alternatives like step-by-step onboarding add a lot of clutter to a website, and can be frustrating when users are forced to go through them. Another alternative I've considered is a dedicated landing page with helpful links that all new stackers are re-directed to, but this doesn't help new visitors who haven't created an account. Additionally, this is a larger development effort, and could be implemented if the feature mentioned above proves to improve retention.

brihuang99:
I like the idea! I can work on a quick proof of concept to share

huumn:
That'd be amazing!

brihuang99:
So I had two ideas that came to mind: 

#### 1. A banner below the header that sits right on top of the content
![Banner 1](https://github.com/stackernews/stacker.news/assets/122472324/b89ff20e-d8ce-401d-b2af-0712f59aea50)


#### 2. A thinner banner above the header
![Banner 2](https://github.com/stackernews/stacker.news/assets/122472324/86cf9950-b682-4383-958d-36f94ad5dc86)

[...]

PR #418, opened 2023-08-17
Add new visitor welcome banner

rleed:
image(2)

This appears on first visit until dismissed or until the date changes. It uses localStorage to know if it is a first-time visit or not.

(This is part 1 of issue #340. The other part is to show a similar banner on first login.)


huumn:
Great start!

We can detect if someone is logged in by inspecting `me` and change the copy. For someone logged in, we'll also want to make it never display again on any device once it's dismissed, so it'll need to run a graphql mutation.

The localStorage thing is fine for lurkers (aka logged out peoples).

rleed:
> it'll need to run a graphql mutation

Yeah, I was thinking so too. If I'm not mistaken, it also requires a schema change in order to record that the banner was dismissed. I looked into this a bit and got as far as starting up the app with the new schema, but I saw in the logs that it needed interaction to do the migration (problematic for running in a docker container). I'm learning a lot, but at this point I'm not sure how to get past that point so I might have to bail on this one until I get smarter on prisma.

huumn:
All good. You got the party started!

ekzyis:
> but I saw in the logs that it needed interaction to do the migration (problematic for running in a docker container). I'm learning a lot, but at this point I'm not sure how to get past that point so I might have to bail on this one until I get smarter on prisma.

Don't hesitate to ask questions here or TG or SimpleX or nostr ... or however you can reach us, we don't really have a dedicated channel for contributors (yet?), haha

You probably need to run `npx prisma generate` or `npx prisma migrate dev`

You can use `docker exec` if you want to run this inside the docker container: `docker exec -it app <cmd>`

rleed:
> Don't hesitate to ask questions here or TG or SimpleX or nostr ... or however you can reach us, we don't really have a dedicated channel for contributors (yet?), haha

Thank you for the tips! I have SimpleX and nostr and will seek you out there or here. Meanwhile I have been utilizing some online resources to get up to speed, but sometimes a context-specific tip is extremely helpful, so thank you for that!

huumn:
> we don't really have a dedicated channel for contributors (yet?), haha

I mean github discussions are probably fine. Maybe we'll have a ~sub for them one day

rleed:
I believe it's fully working now.

Suggestion: to minimize banner spam during deployment, the whole return block from components/banners.js:53 to the end can be replaced with `return (<></>)` to no-op the banner for one release cycle. Then include the actual banner code in the next release, after most existing stackers have re-visited the site and gotten their localStorage updated.

huumn:
I'll give this and your other PR a review hopefully by the end of weekend. I have a hard deadline on some other stuff.

Either way, I'll keep you in the loop.

components/banners.js
```diff
@@ -0,0 +1,79 @@
...
+  const me = useMe()
+  const [state, setState] = useState(null)
+  const handleClose = async () => {
+    const obj = { date: state?.date || new Date().toJSON(), hidden: true }
+    const str = JSON.stringify(obj)
+    window.localStorage.setItem('newVisitorInfo', str)
+    setState(obj)
+    const { error } = await hideWelcomeBanner({ variables: { } })
+    if (error) {
+      throw new Error({ message: error.toString() })
```
ekzyis, line 22: This throws an error if you are not logged in.

This mutation shouldn't run if you are not logged in (to prevent unnecessary errors) and if you are logged in, it should use our new [toast](https://github.com/stackernews/stacker.news/pull/431) mechanism or something else to tell the user an error happened.

ekzyis (commented):
> This appears on first visit until dismissed or until the date changes. It uses localStorage to know if it is a first-time visit or not.

Why "until the date changes"? Don't we want to just show it once and then never again if the alert was dismissed? Afaict, this will show the banner again on the next day.

Imo, we only need to set a boolean in local storage and then check if this boolean is set or not to hide the alert or not.

If we have a session, we use `me.hideWelcomeBanner` instead.

Or am I missing something here?

> Suggestion: to minimize banner spam during deployment, the whole return block from components/banners.js:53 to the end can be replaced with return (<></>) to no-op the banner for one release cycle. Then include the actual banner code in the next release, after most existing stackers have re-visited the site and gotten their localStorage updated.

Regarding this, I think we could just manually update every existing user to set `hideWelcomeBanner` to `true` in the database. Then it won't show up for them ever. But that ignores the same date logic which I don't understand currently.

rleed:
> > This appears on first visit until dismissed or until the date changes. It uses localStorage to know if it is a first-time visit or not.
> 
> Why "until the date changes"? Don't we want to just show it once and then never again if the alert was dismissed? Afaict, this will show the banner again on the next day.

The banner never shows beyond 24 hours of first visit or first login, even if they don't dismiss it. Sorry if I explained it poorly... (the 24 hours is part of how the code decides that it is really a first visit or first login, to be safe.)

> Imo, we only need to set a boolean in local storage and then check if this boolean is set or not to hide the alert or not.
> 
> If we have a session, we use `me.hideWelcomeBanner` instead.
> 
> Or am I missing something here?

100% correct. I just strengthened it with the 24-hour check, so the banner will automatically stop appearing after 24 hours as part of ensuring that it only appears on "first visit" or "first login" per the issue description. (And I hate overzealous popups 😝)

> > Suggestion: to minimize banner spam during deployment, the whole return block from components/banners.js:53 to the end can be replaced with return (<></>) to no-op the banner for one release cycle. Then include the actual banner code in the next release, after most existing stackers have re-visited the site and gotten their localStorage updated.
> 
> Regarding this, I think we could just manually update every existing user to set `hideWelcomeBanner` to `true` in the database. Then it won't show up for them ever.

By the suggestion I was hoping to avoid them seeing it BEFORE they log in, on first visit.

components/banners.js
```diff
@@ -0,0 +1,79 @@
...
+          <Alert.Link href='/guide'>content guidelines</Alert.Link>, or go ahead and{' '}
+          {
+            me
+              ? (
+                <Alert.Link href='/post'>make a post</Alert.Link>
+                )
+              : (
+                <Alert.Link href='/signup'>sign up</Alert.Link>
+                )
+          }.
```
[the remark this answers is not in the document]
rleed: I guess it just depends on whether we want to encourage signing up or not... I don't have a strong opinion on this.

components/banners.js
rleed: Perfect! I'll try to get that in today.

components/banners.js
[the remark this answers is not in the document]
rleed: One thing to keep in mind is that we show the banner in two specific cases, so it would seem strange if the text were not different to match each case. How about changing the lurker welcome message similar to the following:

> To get started, check out our FAQs or content guidelines, or go ahead and signup (or start off by making an anonymous post).

This would highlight the difference between posting as anon vs. signing up and posting under a username, so they can have a basis to decide which action they prefer to take.

components/banners.js
```diff
@@ -0,0 +1,86 @@
...
+    let obj = JSON.parse(str)
+    if (!str) {
+      obj = { date: new Date().toJSON(), hidden: false }
+      str = JSON.stringify(obj)
+      window.localStorage.setItem('newVisitorInfo', str)
+    }
+    if (me && (me.hideWelcomeBanner || !sameDay(new Date(me.createdAt), today))) {
+      obj.hidden = true
+      str = JSON.stringify(obj)
+      window.localStorage.setItem('newVisitorInfo', str)
```
ekzyis, line 55: These two lines are in both `if` clauses so you can just remove them in both and put them outside

ekzyis (commented):
> The banner never shows beyond 24 hours of first visit or first login, even if they don't dismiss it. Sorry if I explained it poorly... (the 24 hours is part of how the code decides that it is really a first visit or first login, to be safe.)

No need to be sorry. I could have just looked at the code, lol

However, I am not convinced we need this 24 hours check logic. I have taken a look at the code now and I think it's not worth it. Please correct me if I am wrong but I made the following observations:

- the code would be a lot simpler without this
- since this code only exists to hide a banner after 24 hours automatically it only replaces a click by the user to dismiss the banner manually
- if a new user visits the website, but then immediately leaves without reading the banner and then visits again after 24 hours, they won't see the banner again

So since this logic also possibly introduces undesired behavior (third bullet point above) while also making it more complicated than we probably need, I hope you can understand why I am not convinced we need this.

But I am open to change my mind :)

rleed:
> However, I am not convinced we need this 24 hours check logic. I have taken a look at the code now and I think it's not worth it...[snip]...I hope you can understand why I am not convinced we need this.

Fair enough...you gave good reasoning and I appreciate you weighing the pros and cons. Sometimes I am not sure where to draw the line in the tradeoff between UX vs simplicity of code, but conversation helps me get the feel for it. I appreciate that.

> But I am open to change my mind :)

> > By the suggestion I was hoping to avoid them seeing it BEFORE they log in.
> 
> Ah, I see. But I think most stackers don't log in every time they visit SN and then logout again. So most stackers probably will see the other banner first, not the anon banner. But I have no hard data on this.

Ok, no worries then from my side.

components/banners.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+    if (me?.hideWelcomeBanner) {
+      obj.hidden = true
+    }
+    str = JSON.stringify(obj)
+    window.localStorage.setItem('newVisitorInfo', str)
+    setState(obj)
+  }, [me?.hideWelcomeBanner])
+
+  if ((me && !me.hideWelcomeBanner) || (!me && state && !state.hidden)) {
```
ekzyis, line 54: You can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining):

```suggestion
  if (!me?.hideWelcomeBanner || !state?.hidden) {
```

ekzyis (commented):
This look much better now! But I think you don't need to store a date anymore, do you?

In that case, you can also rename the state from `state` to simply `hidden`.

I would also just combine `localStorage.setItem` and `JSON.stringify` / `JSON.parse` in a single line as we do in other parts of the code base.

Then you also don't need to use blank variables names like `obj` and `str`.

But you don't even need `JSON.parse` and friends if just check if there is anything in `newVisitorInfo` (which I would also rename to `hideWelcomeBanner` now)

I think the whole `useEffect` could be this:

```
useEffect(() => {
  setHidden(me?.hideWelcomeBanner || window.localStorage.getItem('hideWelcomeBanner'))
}, [me?.hideWelcomeBanner])
```

similar to in `pages/wallet.js`:

```
useEffect(() => {
  setShowAlert(!window.localStorage.getItem('hideLnAddrAlert'))
}, [])
```

components/banners.js
ekzyis: BTW, don't be confused if I propose code changes to code that I already suggested to change in other ways which would make this change unnecessary, lol

Just want to let you know how this line of code could be better written _if_ there is no reason to not include it at all :)

rleed: My very explicit logic was actually intentional for clarity, and there is a subtle difference between `me && !me.x` vs. `!me?.x` that makes the shorthand not work here. To be clear, in this case, `!me?.hideWelcomeBanner` evaluates to true when not logged in, which causes the banner to appear regardless of the other flag. But with all the other reductions, at least the right-hand side of the condition has now been simplified.

ekzyis: interesting, you are right. Nice catch!

components/banners.js
```diff
@@ -0,0 +1,70 @@
...
+    window.localStorage.setItem('hideWelcomeBanner', true)
+    setHidden(true)
+    if (me) {
+      let error
+      try {
+        ({ error } = await hideWelcomeBanner())
+      } catch (e) {
+        error = e
+      }
+      if (error) toaster.danger(error.toString())
```
ekzyis, line 23: This can be simplified like this:

```js
try {
  await hideWelcomeBanner()
} catch (err) {
  console.error(err)
  toaster.danger('mutation failed')
}
```

See `components/bookmark.js:28-34`

pages/~/index.js
```diff
@@ -17,6 +18,7 @@ export default function Sub ({ ssrData }) {
   return (
     <Layout sub={variables.sub}>
       <Snl />
+      <NewVisitorBanner />
```
ekzyis, line 21: Since you named the mutation `hideWelcomeBanner`, I would name this `WelcomeBanner` instead of `NewVisitorBanner`.

Consistency is key in code and life :)

components/banners.js
```diff
@@ -0,0 +1,70 @@
...
+          }
+        }
+      })
+    }
+  })
+  useEffect(() => {
+    setHidden(me?.hideWelcomeBanner || window.localStorage.getItem('hideWelcomeBanner'))
+  }, [me?.hideWelcomeBanner])
+
+  if ((me && !me.hideWelcomeBanner) || !hidden) {
```
ekzyis, line 42: I may have already asked this but can't find the conversation:

Is `(me && !me.hideWelcomeBanner)` needed anymore?

Since in `setHidden(me?.hideWelcomeBanner || window.localStorage.getItem('hideWelcomeBanner'))`, there already is the whole logic if we should hide or not, no?

If we're logged in, we check `hideWelcomeBanner`. If it's true, we will hide the banner. If it's false, we fallback to local storage which is fine.

If we're not logged in `me?.hideWelcomeBanner` is false and we again fallback to local storage which is fine.

ekzyis (changes requested):
These are my last nitpicks. After that, you only need to fight @huumn, the final boss (lol)

Just kidding, I think this PR is great now and will probably be merged as-is :)

components/banners.js
rleed: It _was_ needed, but with a slight modification to the `setHidden(...)` line it could be eliminated. I also changed the `if` to an early return like in the bookmark banner.

ekzyis (approved):

Issue #217, opened 2022-12-01
Feature Request: Export CSV with full transaction history


kerooke:
As Stacker News sees more and more Bitcoin transaction volume on the platform, it will become important to add in features that make tax reporting easy.

CSV file exports are the preferred format for accountants to keep track of transaction history today, and Stacker News could add a button on a user's wallet page that offers a full download of their transaction history with corresponding fiat values.

For example, a table with the following column names would cover basically everything an accountant would need to know about someone's Stacker News usage:

- Transaction type (boost, tip, reward, funding, withdrawing, etc...)
- Date (and time if possible)
- Sats value spent (when applicable)
- Sats value received (when applicable)
- Fiat value spent (when applicable)
- Fiat value received (when applicable)
- Invoice (for funding/withdrawals)

The fiat values above could be computed by pulling in the preferred fiat currency of Stacker News users as per their profile settings.

huumn:
Re: fiat values

These should be the values that the time of the transaction right? So we'd need some record of spot price history.

kristapsk:
I see this as very useful feature too! (was going to open a feature request issue myself, but then found this)

Fiat value aren't required if there is date/time. You can after export get this yourself, by querying some website, that provides historical BTC/fiat prices.

huumn:
Minimizing the requirement is a good idea - the price part is the hardest requirement really. It's otherwise pretty trivial.

kristapsk:
Yes, I think it's ok to, for at least start, just add CSV export for what is already shown when you look at wallet history.

PR #420, opened 2023-08-17
add CSV download button

rleed:
Added the green button to download a CSV file (issue #217):
image(3)

The downloaded file looks something like this:
image(4)

The filter options are taken into effect for the CSV file.


huumn:
Nice! As far as I can tell though, this will only allow you to export the visible page. People probably want to export everything.

The csv will probably need to be made on the server like our rss feed, `lib/rss.js`

rleed:
Ah ok, makes sense.

rleed:
I think this is ready for (re)review now.

huumn:
Looking good!

The `walletHistory` resolver by default returns a single page (21 items) so you'll need to page over it to return the full walletHistory.

huumn:
Btw I'm going to start doing pr related work on the weekend. So I'll go through all prs this weekend and merge anything that's ready.

huumn:
I put this in draft until we get paging in it.

huumn:
This is going to require more work before deploying. I've enhanced it but it's still not there for prod yet. I'm going to put it in draft.

At the very least it needs:
- [ ] rate limits

Ideally it'd also have:
- [ ] cancelation
- [ ] caching

rleed:
Good point about rate limit. I expected there'd be ongoing improvements to this functionality beyond the scope of this PR/issue, but I see that rate limiting is important to include before deploying on the live dataset.

Do we already have guidelines/requirements about rate limits? In the simplest case I could introduce a paging delay so that time becomes the limiting factor for large datasets. This should at least provide basic protection (leaving room for future improvement in UX), but I'm not sure if this is enough because those who care most about this feature are the ones who are likely to have large amounts of data.

A progress indicator would also be important I think.

pages/api/satistics.js
```diff
@@ -0,0 +1,33 @@
...
+  let facts = []; let cursor = null
+  res.setHeader('Content-Type', 'text/csv')
+  res.setHeader('Content-Disposition', 'attachment; filename=satistics.csv')
+  console.log(req.query.inc)
+  res.write('time,type,sats\n')
+  do {
+    // query for items
+    ({ data: { walletHistory: { facts, cursor } } } = await apollo.query({
+      query: WALLET_HISTORY,
+      variables: { cursor, limit: 1000, inc: req.query.inc }
```
rleed, line 15: Nice...I wanted the limit variable too

huumn:
> Do we already have guidelines/requirements about rate limits? In the simplest case I could introduce a paging delay so that time becomes the limiting factor for large datasets. This should at least provide basic protection (leaving room for future improvement in UX), but I'm not sure if this is enough because those who care most about this feature are the ones who are likely to have large amounts of data.

If an operation is very expensive, then we don't do it or do it infrequently. This operation is very expensive. For my account, a full wallet history CSV is 2.2MB. 

A paging delay solves a certain kind of DoS vector, but it doesn't solve the one where someone requests tens of thousands of csv downloads simultaneously.

rleed:
Requests can be queued for files to be generated inexpensively in the background and a notification sent when ready. Realistically I think what people need is to do a monthly download of a month-long time range or a yearly download of a year-long time range, or a download of everything since the last download, or a download of everything once. All of those scenarios equate to the same: a single download of all data once per user. More than that, charge sats...it's the ideal rate-limiter!

huumn:
> Requests can be queued for files to be generated inexpensively in the background and a notification sent when ready.

This is a great idea!

If we cache these files, new downloads can also check the cache for existing ones and append any new data to it.

huumn:
> Requests can be queued for files to be generated inexpensively in the background and a notification sent when ready.

This is a great idea!

If we cache these files, new downloads can also check the cache for existing ones and append any new data to it.

rleed:
I seem to have hosed my system somehow by switching builds. Now even after a fresh clone and docker-compose up --build I get the following error repeating in the log:

2023-08-30 16:09:22 app       | prisma:error 
2023-08-30 16:09:22 app       | Invalid `prisma.user.update()` invocation:
2023-08-30 16:09:22 app       | 
2023-08-30 16:09:22 app       | 
2023-08-30 16:09:22 app       | An operation failed because it depends on one or more records that were required but not found. Record to update not found.
2023-08-30 16:09:22 app       | PrismaClientKnownRequestError: 
2023-08-30 16:09:22 app       | Invalid `prisma.user.update()` invocation:
2023-08-30 16:09:22 app       | 
2023-08-30 16:09:22 app       | 
2023-08-30 16:09:22 app       | An operation failed because it depends on one or more records that were required but not found. Record to update not found.
2023-08-30 16:09:22 app       |     at Hr.handleRequestError (/app/node_modules/@prisma/client/runtime/library.js:122:6999)
2023-08-30 16:09:22 app       |     at Hr.handleAndLogRequestError (/app/node_modules/@prisma/client/runtime/library.js:122:6388)
2023-08-30 16:09:22 app       |     at Hr.request (/app/node_modules/@prisma/client/runtime/library.js:122:6108)
2023-08-30 16:09:22 app       |     at async l (/app/node_modules/@prisma/client/runtime/library.js:126:10298) {
2023-08-30 16:09:22 app       |   code: 'P2025',
2023-08-30 16:09:22 app       |   clientVersion: '5.1.1',
2023-08-30 16:09:22 app       |   meta: { cause: 'Record to update not found.' }
2023-08-30 16:09:22 app       | }

I've tried docker-compose rm -f without success. Any tips on what else I can do to reset the whole thing and get a fresh start?

huumn:
`docker-compose down --rmi all -v` might work

rleed:
That was annoying... it turned out to be caused by the browser, that's why no amount of cleaning helped, not even the command you suggested. Clearing all browser data finally fixed it.

ekzyis:
> That was annoying... it turned out to be caused by the browser, that's why no amount of cleaning helped, not even the command you suggested. Clearing all browser data finally fixed it.

Sorry for not chiming in earlier. I think you mentioned a similar problem somewhere else; not sure.

Clearing the browser data fixed it because you probably still had a cookie for the old database data. So it tried to update an user with an valid cookie which doesn't exist.

rleed:
> Clearing the browser data fixed it because you probably still had a cookie for the old database data. So it tried to update an user with a valid cookie which doesn't exist.

Thanks for the insight!

> But regarding the branch / PR status: Did the branch got messed up? I see duplicate commits (same commit messages) with changes that don't seem to be related to this PR. Maybe something got messed up during a merge?

Yes, sorry.

> I would recommend a clean rebase on master, no merge.

Agreed! I will take care of it.

I am hoping to push some updated code for this PR today as well, to hook up to the worker process!

SatsAllDay:
I'm very eager to see what this PR turns into. I think the idea of doing this kind of data export with async jobs in the background is very interesting. Keep it up!

rleed:
Partial update...still in work.
These diagrams document the interactions...

Untitled

rleed:
Partial update...still in work.
These diagrams document how the interactions work...

Untitled

ekzyis:
Going to take a look at this within the next 24 hours. Mentioning this here to keep myself accountable, lol

rleed:
> Going to take a look at this within the next 24 hours. Mentioning this here to keep myself accountable, lol

Great! Appreciate experienced feedback at this point. I think its working well, but I feel like some parts of the code still have room for improvement/simplification where I might have overlooked some better ways of doing things.

- Files are currently created in the root folder and shared between worker and app. Ok or can we improve?
- Is there a way to share GQL queries between app (ES6) and worker (non-module)?
- I added a user id variable to the query for when it is called by the worker. Is there a better way? (to provide "me"?)
- Do you like the UI (button color changes, anim, etc)? Maybe it's too much complexity again for UX?

Surely you will find other things too, but those are areas I am questioning.

api/resolvers/wallet.js
```diff
@@ -73,9 +73,9 @@ export default {
     connectAddress: async (parent, args, { lnd }) => {
       return process.env.LND_CONNECT_ADDRESS
     },
-    walletHistory: async (parent, { cursor, inc }, { me, models, lnd }) => {
+    walletHistory: async (parent, { cursor, inc, limit = LIMIT, id }, { me, models, lnd }) => {
       const decodedCursor = decodeCursor(cursor)
-      if (!me) {
+      if (!me && !id) {
```
ekzyis, line 78: Doesn't this allow anyone to fetch the wallet history of any user?

For example:

```
$ curl -XPOST -H "Content-type: application/json" --data @wallet_history.json https://sn.ekzyis.com/api/graphql
{"data":{"walletHistory":{"facts":[]}}}
```

_wallet\_history.json_:

```json
{
  "query": "query walletHistory($id: Int) { walletHistory(id: $id) { facts { id createdAt factId bolt11 sats satsFee status type description item { id } } } }",
  "variables": {
    "id": 1
  }
}
```

However, I created a new user and funded my account. I then used the id of that new user and expected to see their wallet history but the history in the response was empty :thinking: 

I am too tired at the moment to debug why it's empty; had a long day... will take a look at this PR again after I slept!

pages/api/satistics.js
```diff
@@ -0,0 +1,38 @@
+import fs from 'fs'
+import getSSRApolloClient from '../../api/ssrApollo'
+import { ME } from '../../fragments/users'
+import { CsvRequest, CsvRequestStatus } from '../../api/constants'
+import { gql } from '@apollo/client'
+
+export default async function handler (req, res) {
```
ekzyis, line 7: Calling this endpoint without a session results in a 500 error:

command:

```
$ curl https://sn.ekzyis.com/satistics.csv
```

server logs:
```
app  | - error pages/api/satistics.js (10:24) @ id
app  | - error Error [TypeError]: Cannot read properties of null (reading 'id')
app  |     at handler (webpack-internal:///(api)/./pages/api/satistics.js:29:9)
app  |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
app  |   digest: undefined
app  | }
app  |    8 |   const apollo = await getSSRApolloClient({ req, res })
app  |    9 |   console.log('satistics handler')
app  | > 10 |   const { data: { me: { id, csvRequest, csvRequestStatus } } } = await apollo.query({ query: ME })
app  |      |                        ^
app  |   11 |   console.log(id)
app  |   12 |   const fname = `satistics_${id}.csv`
app  |   13 |   res.setHeader('Content-Type', 'text/csv')
app  | satistics handler
app  | 14912
app  | - wait compiling...
app  | - event compiled successfully in 63 ms (425 modules)
app  | LND GRPC connection successful
app  | - error pages/api/satistics.js (9:24) @ id
app  | - error Error [TypeError]: Cannot read properties of null (reading 'id')
app  |     at handler (webpack-internal:///(api)/./pages/api/satistics.js:28:9)
app  |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
app  |   digest: undefined
app  | }
app  |    7 | export default async function handler (req, res) {
app  |    8 |   const apollo = await getSSRApolloClient({ req, res })
app  | >  9 |   const { data: { me: { id, csvRequest, csvRequestStatus } } } = await apollo.query({ query: ME })
app  |      |                        ^
app  |   10 |   const fname = `satistics_${id}.csv`
app  |   11 |   res.setHeader('Content-Type', 'text/csv')
app  |   12 |   res.setHeader('Content-Disposition', 'attachment; filename=satistics.csv')
```

api/constants.js
```diff
@@ -0,0 +1,11 @@
+export const CsvRequest = {
+  NO_REQUEST: 'NO_REQUEST',
+  FULL_REPORT: 'FULL_REPORT'
+}
+
+export const CsvRequestStatus = {
+  NO_REQUEST: 'NO_REQUEST',
+  FULL_REPORT: 'FULL_REPORT',
```
ekzyis, line 8: Why is `FULL_REPORT` used here? Wouldn't be `DONE` or `COMPLETED` be more intuitive?

pages/satistics.js
```diff
@@ -166,6 +171,64 @@ export default function Satistics ({ ssrData }) {
...
 
+  const me = useMe()
+  const getCsvBtnClass = () => {
+    return me.csvRequest === CsvRequest.NO_REQUEST
+      ? 'btn-grey-darkmode'
+      : me.csvRequestStatus === CsvRequestStatus.FULL_REPORT
+        ? 'btn-success'
+        : 'btn-danger'
+  }
+  const getIconClass = () => {
```
ekzyis, line 182: Why are this and `getCsvBtnClass` functions?

ekzyis (commented):
Okay, I have taken a quick look over and tested generating + downloading CSVs. That worked like a charm and I like the "generating CSV" animation :)

Left some comments and will do another pass after I slept since it got late for me.

pages/api/satistics.js
rleed: Thanks..fixed to gracefully fall into the 400 error case.

pages/satistics.js
rleed: At some point refreshes weren't happening when I needed them, so I put functions there (but I am not even sure anymore whether making them functions did anything for my original problem). Anyway, it is working without them, so that's good.

api/constants.js
rleed: Another case of me thinking more complicated than necessary... to distinguish / match the completion status to potentially different kinds of requests in the future. Nothing for this PR...I've now changed it to DONE (and I shortened GENERATING_REPORT to a simpler IN_PROGRESS).

api/constants.js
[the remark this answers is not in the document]
rleed: Ahhh wonderful. I was looking for a way to share code between app and worker... now I understand how to do it via CommonJS. I've now done so with the enum constants.

I would still like a solution for sharing the GQL template literals...I don't think putting those in the lib/constants.js file would be the right organization. Do you have any suggestions?

api/resolvers/wallet.js
rleed: I'm learning all this as I go, but I believe you might be correct that this exposes the wallet history of all users in a way that it wasn't before. (You need the "inc" variable to get non-empty results.) This is partly why I was questioning whether there's a better way to handle this "me" issue here.

In essence, the problem is that to fulfill its jobs, the worker is querying the GQL API just like anybody else. It needs its own privileged access somehow.

ekzyis: > You need the "inc" variable to get non-empty results.

Ah, will check this out when I have more time

> In essence, the problem is that to fulfill its jobs, the worker is querying the GQL API just like anybody else. It needs its own privileged access somehow.

The worker has direct database access. Whatever you want to do, you don't need to go through the GraphQL API.

api/constants.js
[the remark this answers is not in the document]
ekzyis: > I would still like a solution for sharing the GQL template literals...I don't think putting those in the lib/constants.js file would be the right organization. Do you have any suggestions?

My answer [here](https://github.com/stackernews/stacker.news/pull/420#discussion_r1318986964) should answer your question: I don't think you need to share GQL template literals with the worker

ekzyis:
> Files are currently created in the root folder and shared between worker and app. Ok or can we improve?

Good question. I can think of following options:

- store files in `/tmp`. This way, we would also not clutter the local project folder with CSV files and the files get automatically deleted during a reboot. (This may be irrelevant for production. I think everything gets deleted during a new deployment) 
- use some in-memory solution like [`memfs`](https://www.npmjs.com/package/memfs) but we want to minimize the amount of dependencies we have
- store CSV files in database as binary blobs

I think the database option would be the best option since then the CSV files can be persistent across deployments.

Another reason is that I think it _could_ be nice to give users the option to have a history of CSV exports. Afaict, the current implementation only allows that a single CSV file per user exists on the server, right?

I think it's quite common to allow multiple exports to exist at the same time on the server. For example, here a screenshot from Kraken:

2023-09-07-210429_1920x1080_scrot

However, this also might not be needed for us or at least for the first iteration on this feature.

> - Is there a way to share GQL queries between app (ES6) and worker (non-module)?
> - I added a user id variable to the query for when it is called by the worker. Is there a better way? (to provide "me"?)

See [here](https://github.com/stackernews/stacker.news/pull/420#discussion_r1318986964)

> Do you like the UI (button color changes, anim, etc)? Maybe it's too much complexity again for UX?

Nah, I think it's fine for the first iteration. @huumn may have more to say about it though

api/resolvers/wallet.js
rleed: Alright. I've pushed another commit that switches from GraphQL to Postgres queries. This solves the problem with the id parameter (and removes it). So I'll mark this conversation/thread as resolved.

rleed:
I'm not liking the copy-paste duplication of the Postgres queries and related functions any more than the GraphQL queries. I will do some cleanup and try to share these via CommonJS.

api/resolvers/user.js
```diff
@@ -773,6 +767,12 @@ export default {
       })
 
       return !!subscription
+    },
+    csvRequest: async (user) => {
+      return user.csvRequest
+    },
+    csvRequestStatus: async (user) => {
+      return user.csvRequestStatus
```
ekzyis, line 775: I think resolvers like these are obsolete. GraphQL automatically resolves fields of the parent object with same names.

huumn:
This is solid. There are some changes I'll want to make before shipping to prod, which are nontrivial like storing the CSV on S3 and giving authenticated access to the link, but in general @rleed you can consider this PR done.

I'm going to move to draft until I get around to those changes (which might take me a couple weeks).

rleed:
Awesome. Thank you!

huumn:
Remaining todos (for me):

- [ ] use s3 and authenticated links in prod

staab:
Is this still planned? I'd like to use this for this tax year if possible.

huumn:
> Is this still planned? I'd like to use this for this tax year if possible.

It's still planned. I'll try to get it done before tax time. Worst case I can do something manually for people that need it.

staab:
In case you're not able, or other people want to pull the data, here's the client-side script that I'm using:

```

async function expand() {
  let misses = 0

  while (misses < 5) {
    const buttons = document.querySelectorAll('.btn.btn-primary.btn-md')
    const button = Array.from(buttons).find(e => e.textContent === 'more')

    if (!button) {
      misses++
    } else {
      misses = 0
      button.click()
    }

    await new Promise(resolve => setTimeout(resolve, 1000))
  }
}

function gather() {
  const types = Array.from(document.querySelectorAll('.satistics_type__0Rabd'))
  const details = Array.from(document.querySelectorAll('.satistics_detail__EYLCt'))
  const amounts = Array.from(document.querySelectorAll('.satistics_sats__RNvnG'))
  const data = []

  for (let i = 1; i < types.length; i++) {
    data.push({
      type: types[i].textContent,
      amount: amounts[i].textContent,
      date: details[i].lastChild.textContent,
    })
  }

  return data
}

async function main() {
  await expand()

  console.log(JSON.stringify(gather()))
}

main()
```

Just go to the statistics page, run this, and the data will get logged to the console.

Issue #269, opened 2023-05-09
Enhancement request: Improve feedback when attempting to withdraw without being logged in


sgornick:
If I am logged out but have navigated to the /wallet page and attempt to Withdraw, the message shown will be:

    Error: Cannot read property 'id' of null

This gives the user the impression the withdrawal issue is with the Stacker.News website, when it is actually user error.  

This occurred, in my instance, on my Android where I switched to my LN wallet (WoS) to get the invoice for my withdrawal to pay.  When I switched back to SN, my browser session did not persist, so I was essentially no longer logged in..  This can occur with incognito mode / Private browsing when switching back and forth between the browser and other apps.

This request is to have SN respond with a better error message when a user attempts to withdraw when not logged in.  It may even be good to display a warning message when providing a response for GET on /wallet and the user is not logged in / not authenticated.

PR #421, opened 2023-08-18
make wallet page prompt for login if needed

rleed:
This improves UX to address issue #269


huumn:
I'd prefer for this to be handled more generically the backend in `getGetServerSideProps` ... it'll require deeper backend changes but then we can reuse it in other parts of the code.

PR #423, opened 2023-08-18
Make withdrawal options more visible (issue167a)

rleed:
For convenience... this PR could replace PR 414.

image(5)


huumn:
Thank you again! This pr was noticeably better (less code!) than the prior withdrawal pr.

In the future, please try to keep your prs scoped to the problem they aim to solve. If I disagree with you on those extra changes, I don't want to have to ask you to remove them.

Otherwise, the pr did what it was supposed to very well!

PR #424, opened 2023-08-19
Add new user setting to hide bookmarks from other users (and anon)

SatsAllDay:
A new feature - give stackers the ability to hide their bookmarks from other stackers (and non-logged-in users).

Includes:
1. A new boolean flag on the Users table
2. A new checkbox in the user settings page
3. Some special logic in the item API resolver to ensure that a stackers' bookmarks are meant to be shared, and returns empty results if not.

Let me know if I overlooked anything, this ended up requiring fewer changes than expected...

EDIT: I originally posted about this here: https://stacker.news/items/206443

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -319,7 +319,7 @@ export default {
         // the calling user is either not logged in, or not the user upon which the query is made,
         // so we need to check authz
         user = await models.user.findUnique({ where: { name } })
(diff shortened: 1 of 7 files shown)


huumn:
Afaict this looks right. Looking forward to reviewing this tomorrow morning along with your other pr.

huumn:
This is good to go. We'll merge this after the forward user one.

Issue #377, opened 2023-08-03
Improving Referral Rewards


ghost:
**Is your feature request related to a problem? Please describe.**
Referrals on Stacker News has seen limited adoption to date, and I believe part of the reason is the barriers to participating in the referral program. Today, stackers must either find the share button on a post or comment or manually add their referral tag to a post prior to sharing it. 

In addition, stackers do not receive notifications when a referral earns them sats. The only way to see that information is to open ones referral page in the settings pane. I often open up my referrals page and am surprised when i see how many sats my referrals have earned for me.

**Describe the solution you'd like**
To improve access to the referral program, Stacker News should append all links with the referral tag of a stacker when they are signed in. This means that moving forward any links shared by a logged in stacker will give them a chance of earning referral rewards.

Furthermore, Stacker News should also create a new notification type that specifically calls out the rewards each stacker earns from their referrals. For example, the copy could read "You stacked X sats today for referring @insertUsername".

The net effect of these two changes is that referrals are more visible, stackers have a higher chance of earning sats, and they are able to better attribute those sats to the links they shared once they begin earning.

byoungdale:
It looks like the first improvement of adding referral tag on copy already exists [here](https://github.com/stackernews/stacker.news/blob/679f766f079d0f1d1ff3a5ffa7fe05a26e5ffb2b/components/share.js#L8). But, the notifications when earning sats for referrals is not. I will take further a look at that.

huumn:
I'd love to see the notifications!

#### re: referral tags

I think @kerooke means appending the referral fragment to urls in the url bar. If it were to be done, it'd probably require one of two hacks:

1. middleware to redirect incoming requests from `/some/url` to `/some/url/r/k00b`
2. nextjs router hacking to manipulate the url clientside somehow avoiding navigation

In either case, with our current fragment style `/r/usernym`, its very nontrivial and will affect (and potentially introduce bugs) on every route across the site given the way nextjs works.

If this is something we want to do, we'll probably want to modify referral codes to use GET params instead, e.g. `/some/url?r=k00b`.

byoungdale:
I think I see how I can get that working without dramatic changes. Having problems with the docker compose setup right now for some unknown reason. Once I sort that out, I'll get moving.

huumn:
If it's a problem we can fix or document, please let us know

byoungdale:
Is this what we are thinking for the automatic referral param for logged in users:

[...]

PR #425, opened 2023-08-19
referral query param updates

byoungdale:
This is for the query param changes for referrals mentioned in #377 .


huumn:
> The issue is this didn't update the url at all.

`router.replace` takes two paths, one as the first argument and one as the second. The first argument, the one you're adding referral to, can be thought of as known to the application but *hidden from the user*. If you'll notice, I'm adding `nodata=true` but it also doesn't show in the url bar.

The second argument is what you want to modify. Here's a link to the nextjs docs for router.push (same args with same meanings as router.replace but it has more thorough param descriptions ... you'll still want to use router.replace): https://nextjs.org/docs/pages/api-reference/functions/use-router#routerpush

byoungdale:
Gotcha. Thank you. Should of looked a little closer at the code and docs. Got it working now with router.replace in `_app.js`. The thing is now the `?r=username` query param will be added to every single page. Not only the item pages like the current referral fragment works. Is that what we want? Seems like it's not a bad idea. Users might just want to copy any url they are on and send it to someone...

pages/_app.js
```diff
@@ -54,7 +54,7 @@ function MyApp ({ Component, pageProps: { ...props } }) {
     router.replace({
       pathname: router.pathname,
       query: { ...router.query, nodata: true }
-    }, router.asPath, { ...router.options, shallow: true }).catch((e) => {
+    }, props.me ? `?r=${props.me.name}` : router.asPath, { ...router.options, shallow: true }).catch((e) => {
```
huumn, line 57: Hmmm I'm not sure this is right. Doesn't this make the url bar always `stacker.news?r=<nym>` regardless of what page we're on? 

I think we want to "add" the query parameter to `router.asPath`

byoungdale: It adds the param to the url like this as examples:
- root: `stacker.news/~?r=username`
- top: `stacker.news/~/top/posts/day?r=username`
- recent: `stacker.news/~/recent?r=username`
- item: `stacker.news/items/4?r=username`

huumn: I see. I would like to keep our url pattern otherwise the same while introducing this change.

If you'd like I can try to take this the last yard on my own. You've done the lionshare of the effort already.

> It adds the param to the url like this as examples:
> - root: stacker.news/~?r=username
> - top: stacker.news/~/top/posts/day?r=username
> - recent: stacker.news/~/recent?r=username
> - item: stacker.news/items/4?r=username

This is what ideally we'd get instead:
- root: stacker.news?r=username
- top: stacker.news/top/posts/day?r=username
- recent: stacker.news/recent?r=username
- item: stacker.news/items/4?r=username

byoungdale: Ah, I see. Without the tilde added. I don't actually know where that's coming from. I see it's apart of the pages:

```sh
└── ~
    ├── index.js
    ├── post.js
    ├── recent
    │   ├── [type].js
    │   └── index.js
    ├── rss.js
    ├── search.js
    └── top
        ├── [type]
        │   └── [when].js
        ├── cowboys.js
        └── stackers
            └── [when].js
```

I don't know why my changes are forcing it though.

Either way, go ahead and take it to the finish line. I don't think I'll be able to find time to work on it this coming week. But, now I'm that much more familiar with the codebase for next time.

huumn: Cool, will do!

huumn:
Thanks for this!

You did a really precise job of making these changes and I probably would've procrastinated doing this forever otherwise.

Please drop a lightning address or something when you get a change so I can tip you when I deploy this.

byoungdale:
Thanks! You can send any tip to my stacker news lightning address: brandonsbytes@stacker.news

huumn:
Remaining todos:

- [ ] provide migration path for old referral links
- [ ] make paths as expected (discussed above)
- [ ] meybe fix lightning login not attributing referrals

huumn:
We've decided not to do this now that everything is an implicit referral link

PR #426, opened 2023-08-21
Fix youtube player x-overflow

ekzyis:
Closes #408 

https://files.ekzyis.com/public/sn/youtube-player-x-overflow.mp4

PR #427, opened 2023-08-21
Fix click to load in preview

### LINKED ISSUE #416: Don't use "click to load" in preview
2023-08-16-064748_607x268_scrot

2023-08-16-064758_605x218_scrot

ekzyis:
Closes #416

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -161,7 +161,7 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, setH
           : (
             <div className='form-group'>
               <div className={`${styles.text} form-control`}>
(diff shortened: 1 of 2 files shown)


ekzyis:
See https://github.com/stackernews/stacker.news/pull/426#issuecomment-1687327636

ekzyis:
Yup, lol, just pushed another branch and now this branch got deleted since it only exists in Github.

It's basically this problem: https://github.blog/changelog/2022-06-15-block-potentially-destructive-git-pushes/

Guess I have to only push to Gitea now ... or I should use pull mirrors, not push mirrors.

PR #431, opened 2023-08-21
Toast

### LINKED ISSUE #428: Global Toast mechanism for better feedback
There are several actions in SN that don't offer any visual feedback to indicate success/failure e.g.
1. Subscribing to a post
2. Bookmarking an item
3. Probably others

Additionally, in places that do offer feedback, it's usually statically placed on the page. If you're scrolled down too far, like editing user settings or making a post, you don't immediately see the feedback.

I'd like to propose a global toast system that displays notifications in a fixed position on the screen so that you can see them regardless of your scrolling position on the page.

The toast messages should probably be auto-dismissed, at least for success messages. Maybe not for error, but they should still be dismissable by the user.

As far as styling goes, they can be styled just like how messages are displayed today.

N/A

**Additional context**
react-bootstrap has [toasts](https://react-bootstrap.netlify.app/docs/components/toasts/) that we should be able to use, though I haven't explored them yet.

SatsAllDay:
A very quick and dirty prototype of what a global toast system could look like. See screenshot for example on the subscribe action.

I've put this into a draft PR to gather feedback, it's nowhere near ready to merge.

Thoughts?

image

Closes #428

### DIFF
--- a/components/delete.js
+++ b/components/delete.js
@@ -5,6 +5,7 @@ import Alert from 'react-bootstrap/Alert'
 import Button from 'react-bootstrap/Button'
 import Dropdown from 'react-bootstrap/Dropdown'
 import { useShowModal } from './modal'
+import { useToast } from './toast'
 
 export default function Delete ({ itemId, children, onDelete }) {
   const showModal = useShowModal()
@@ -61,6 +62,7 @@ export default function Delete ({ itemId, children, onDelete }) {
 
 function DeleteConfirm ({ onConfirm }) {
   const [error, setError] = useState()
+  const dispatchToast = useToast()
 
   return (
     <>
@@ -71,6 +73,7 @@ function DeleteConfirm ({ onConfirm }) {
           variant='danger' onClick={async () => {
             try {
               await onConfirm()
+              dispatchToast({ body: 'Delete successful!', variant: 'success', autohide: true, delay: 5000 })
             } catch (e) {
               setError(e.message || e)
             }
--- a/components/delete.js
+++ b/components/delete.js
@@ -7,7 +7,7 @@ import Dropdown from 'react-bootstrap/Dropdown'
 import { useShowModal } from './modal'
 import { useToast } from './toast'
 
-export default function Delete ({ itemId, children, onDelete }) {
+export default function Delete ({ itemId, children, onDelete, type = 'post' }) {
   const showModal = useShowModal()
 
   const [deleteItem] = useMutation(
@@ -41,6 +41,7 @@ export default function Delete ({ itemId, children, onDelete }) {
         showModal(onClose => {
           return (
             <DeleteConfirm
+              type={type}
               onConfirm={async () => {
                 const { error } = await deleteItem({ variables: { id: itemId } })
                 if (error) {
@@ -60,7 +61,7 @@ export default function Delete ({ itemId, children, onDelete }) {
   )
 }
 
-function DeleteConfirm ({ onConfirm }) {
+function DeleteConfirm ({ onConfirm, type }) {
   const [error, setError] = useState()
   const toaster = useToast()
 
@@ -73,7 +74,7 @@ function DeleteConfirm ({ onConfirm }) {
           variant='danger' onClick={async () => {
             try {
               await onConfirm()
-              toaster.success('deleted item')
+              toaster.success(`deleted ${type.toLowerCase()}`)
             } catch (e) {
               setError(e.message || e)
             }
--- a/components/delete.js
+++ b/components/delete.js
@@ -62,7 +62,7 @@ export default function Delete ({ itemId, children, onDelete }) {
 
 function DeleteConfirm ({ onConfirm }) {
   const [error, setError] = useState()
-  const dispatchToast = useToast()
+  const toaster = useToast()
 
   return (
     <>
@@ -73,7 +73,7 @@ function DeleteConfirm ({ onConfirm }) {
           variant='danger' onClick={async () => {
             try {
               await onConfirm()
-              dispatchToast({ body: 'Delete successful!', variant: 'success', autohide: true, delay: 5000 })
+              toaster.success('deleted item')
             } catch (e) {
               setError(e.message || e)
             }
--- a/components/toast.js
+++ b/components/toast.js
@@ -1,0 +1,43 @@
+import React, { createContext, useCallback, useContext, useState } from 'react'
+import Toast from 'react-bootstrap/Toast'
+import ToastBody from 'react-bootstrap/ToastBody'
+import ToastContainer from 'react-bootstrap/ToastContainer'
+import ToastHeader from 'react-bootstrap/ToastHeader'
+import { randInRange } from '../lib/rand'
+
+const ToastContext = createContext(() => {})
+
+export const ToastProvider = ({ children }) => {
+  const [toasts, setToasts] = useState([])
+  const dispatchToast = useCallback((toastConfig) => {
+    const id = Array(10).fill(randInRange(0, 9)).join('')
+    toastConfig = {
+      ...toastConfig,
+      id
+    }
+    if (toastConfig.autohide && toastConfig.delay > 0) {
+      setTimeout(() => {
+        setToasts(toasts.filter(toast => toast.id !== id))
+      }, toastConfig.delay)
+    }
+    setToasts([...toasts, toastConfig])
+  }, [toasts])
+  const removeToast = useCallback(id => {
+    setToasts(toasts.filter(toast => toast.id !== id))
+  }, [toasts])
+  return (
+    <ToastContext.Provider value={dispatchToast}>
+      <ToastContainer position='top-center'>
+        {toasts.map(toast => (
+          <Toast key={toast.id} bg={toast.variant} show autohide={false} onClose={() => removeToast(toast.id)}>
+            <ToastHeader closeButton>{toast.variant}</ToastHeader>
+            <ToastBody>{toast.body}</ToastBody>
+          </Toast>
+        ))}
+      </ToastContainer>
+      {children}
+    </ToastContext.Provider>
+  )
+}
+
+export const useToast = () => useContext(ToastContext)
--- a/components/toast.js
+++ b/components/toast.js
@@ -1,16 +1,18 @@
-import React, { createContext, useCallback, useContext, useState } from 'react'
+import React, { createContext, useCallback, useContext, useMemo, useState } from 'react'
+import Button from 'react-bootstrap/Button'
 import Toast from 'react-bootstrap/Toast'
 import ToastBody from 'react-bootstrap/ToastBody'
 import ToastContainer from 'react-bootstrap/ToastContainer'
-import ToastHeader from 'react-bootstrap/ToastHeader'
-import { randInRange } from '../lib/rand'
+import CloseIcon from '../svgs/close-line.svg'
 
 const ToastContext = createContext(() => {})
+
+let toastId = 0
 
 export const ToastProvider = ({ children }) => {
   const [toasts, setToasts] = useState([])
   const dispatchToast = useCallback((toastConfig) => {
-    const id = Array(10).fill(randInRange(0, 9)).join('')
+    const id = toastId++
     toastConfig = {
       ...toastConfig,
       id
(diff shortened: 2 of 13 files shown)


huumn:
Sexy. 

Subjectives (nits): 
- I'd prefer it be positioned to the bottom-right of the primary container
- The close icon should be right aligned

SatsAllDay:
Btw I'm not trying to flood you with PRs to review, I'm just having fun working on the app. Don't feel pressure to review quickly :)

components/toast.js
```diff
@@ -0,0 +1,51 @@
...
+      danger: 'Danger',
+      info: 'Info'
+    }[toastConfig.variant] ?? toastConfig.variants
+  }, [])
+  return (
+    <ToastContext.Provider value={dispatchToast}>
+      <ToastContainer position='bottom-end' containerPosition='fixed'>
+        {toasts.map(toast => (
+          <Toast key={toast.id} bg={toast.variant} show autohide={false} onClose={() => removeToast(toast.id)}>
+            <ToastHeader closeButton closeLabel='close'><span className='flex-grow-1'>{getHeaderText(toast)}</span></ToastHeader>
```
ekzyis, line 41: I am wondering if the header is required? Looks a bit unnecessary. The color already tells you if it's a success or error. Is it possible to put the close button into the body?

I like the styling of the [evergreen toaster](https://evergreen.segment.com/components/toaster) a lot more:


https://github.com/stackernews/stacker.news/assets/27162016/3365c607-6404-454a-9012-b54d6ebfd3b7

components/toast.js
```diff
@@ -0,0 +1,51 @@
...
+import ToastContainer from 'react-bootstrap/ToastContainer'
+import ToastHeader from 'react-bootstrap/ToastHeader'
+import { randInRange } from '../lib/rand'
+
+const ToastContext = createContext(() => {})
+
+export const ToastProvider = ({ children }) => {
+  const [toasts, setToasts] = useState([])
+  const dispatchToast = useCallback((toastConfig) => {
+    const id = Array(10).fill(randInRange(0, 9)).join('')
```
ekzyis, line 13: You just need an id that is random enough for no collisions, right?

`randInRange` returns float and the value in `fill` is just executed once so the resulting id looks kind of weird:

```
$ node
Welcome to Node.js v20.5.0.
> const randInRange = (min, max) => Math.random() * (max - min) + min
undefined
> randInRange(0, 10)
1.13396597332184
> randInRange(0, 10)
1.2429912934983878
> randInRange(0, 10)
3.6246351810399124
> Array(10).fill(randInRange(0, 9)).join('')
'0.482431026012052340.482431026012052340.482431026012052340.482431026012052340.482431026012052340.482431026012052340.482431026012052340.482431026012052340.482431026012052340.48243102601205234'
```

Can't you just use `toasts.length` as the id? I assume the id just has to be unique among all _existing_ toasts

(You then also need to add `toasts` as a dependeny of `useCallback`)

ekzyis (commented):
Looks good, just noticed three small things:

1. Header looks unnecessary. As mentioned, I like the [evergreen toasters](https://evergreen.segment.com/components/toaster) more because they look more "clean". But I guess styling is always to some degree personal preference
2. `id` generation looks weird
3. Maybe some options like `autohide: true` and `delay: 5000` can be encapsulated in `useToast`? For example, you could return a `toaster` object which has a `success` and `danger` function property and then you can just call `toaster.success('Item flagged successfully!')` etc. This is also inspired by the [evergreen toaster API](https://evergreen.segment.com/components/toaster#types_of_toasts), haha
4. Have you checked how this looks on mobile? The toasts shouldn't be too big.

edit: Okay, it became four things lol

edit 2: regarding

> Also, I could use some feedback on the wording of the toast header and body messages.

As mentioned, I would remove the header if possible. For the body messages, I wouldn't use any punctuation like `!` and try to keep them as short and consistent as possible. I think using `successfully` is just "noise" since the color already signals success + the user knows what he wanted to do. I would maybe even put everything in lower case.

Some examples:
`Bookmarked successfully removed!` -> `bookmark removed`
`Bookmarked successfully added!` -> `bookmark added`
`Failed to remove bookmark` -> `failed to remove bookmark`
`Failed to bookmark` -> `failed to bookmark`
`Delete successful!` -> `${itemType} deleted` where `itemType` depends on context like "item" or "account" and is thus passed into the component
`Item flagged successfully!` -> `item flagged`
`Failed to flag this item` -> `failed to flag item`
...

But not sure if @huumn agrees, i think he'll have the last word haha

components/toast.js
SatsAllDay: Yes, I do just need it to be unique to avoid collisions. I did the random code in lieu of access to a `uuid` generator function, I didn't want to add a dependency for it.

I don't think using `toasts.length` is safe enough. Consider the following:

1. dispatch a sticky toast -> id: 0. `toasts` looks like: `[{ id: 0}]`
2. dispatch an auto-dismissed toast: id: 1. `toasts` looks like: `[{ id: 0}, {id: 1}]`
3. user dismisses toast 0 (makes toasts' length 1). `toasts` looks like: `[{id: 1}]`
4. dispatch an auto-dismissed toast: id: 1. `toasts` looks like: `[{id: 1}, {id: 1}]` -> duplicate id

Admittedly, I was thinking `randInRange` would give an integer, not a float.

I think I'll just refactor to use a globally incrementing integer, starting at 0. If a user has `Number.MAX_SAFE_INTEGER` toasts in one page session, they can afford to refresh the page ;)

components/toast.js
SatsAllDay: I agree, it does feel like overkill to have the header when it just says 'success'. Let me explore some options here. I do like the evergreen styling more as well.

SatsAllDay:
Thank you for the feedback!

> 1. Header looks unnecessary. As mentioned, I like the [evergreen toasters](https://evergreen.segment.com/components/toaster) more because they look more "clean". But I guess styling is always to some degree personal preference

Replied inline.

> 2. `id` generation looks weird

Replied inline.

> 3. Maybe some options like `autohide: true` and `delay: 5000` can be encapsulated in `useToast`? For example, you could return a `toaster` object which has a `success` and `danger` function property and then you can just call `toaster.success('Item flagged successfully!')` etc. This is also inspired by the [evergreen toaster API](https://evergreen.segment.com/components/toaster#types_of_toasts), haha

I like this idea. I was thinking of applying some default values in the implementation of `dispatchToast` but I like this utility function approach better. I'll update the PR accordingly. 

> 4. Have you checked how this looks on mobile? The toasts shouldn't be too big.

I did, IMO it looks ok. I'll attach some screenshots. Would look even better w/o a header, if we can do that.


> > Also, I could use some feedback on the wording of the toast header and body messages.
> 
> As mentioned, I would remove the header if possible. For the body messages, I wouldn't use any punctuation like `!` and try to keep it them as short and consistent as possible. I think using `successfully` is just "noise" since the color already signals success + the user knows what he wanted to do. I would maybe even put everything in lower case.
> 
> Some examples: `Bookmarked successfully removed!` -> `bookmark removed` `Bookmarked successfully added!` -> `bookmark added` `Failed to remove bookmark` -> `failed to remove bookmark` `Failed to bookmark` -> `failed to bookmark` `Delete successful!` -> `${itemType} deleted` where `itemType` depends on context like "item" or "account" and is thus passed into the component `Item flagged successfully!` -> `item flagged` `Failed to flag this item` -> `failed to flag item` ...
> 
> But not sure if @huumn agrees, i think he'll have the last word haha

I like these suggestions. short and sweet :)

components/toast.js
ekzyis: I don't like the X button though. Wouldn't it be better if you could just click the toast anywhere to close it? Are there cases where I would click the toast which are not about closing it?

Just some ideas, you don't have to do everything I am saying haha

We can always iterate on features again later

components/toast.js
ekzyis: > I don't think using toasts.length is safe enough. Consider the following:

Ah, makes sense. I thought for some reason the IDs would always correspond to the index.

> I think I'll just refactor to use a globally incrementing integer, starting at 0. If a user has Number.MAX_SAFE_INTEGER toasts in one page session, they can afford to refresh the page ;)

Yes, globally incrementing integer makes sense.

components/toast.js
SatsAllDay: Generally speaking, I have seen toast-like messages that are "actionable", meaning clicking on them takes you somewhere within the app. This is where an explicit close button/icon helps users know that they can dismiss the toast w/o it launching them into some other flow.

I don't think we have such a use case right now, but something to consider, I suppose.

SatsAllDay:
Mobile screenshot (using chrome dev tools, not on an actual mobile device)

image

components/delete.js
```diff
@@ -71,6 +73,7 @@ function DeleteConfirm ({ onConfirm }) {
           variant='danger' onClick={async () => {
             try {
               await onConfirm()
+              toaster.success('deleted item')
```
SatsAllDay, line 76: @ekzyis AFAICT this is only used in the dropdown for an item, so I didn't make the message generic to support multiple item types.

ekzyis: I can't check right now but I think this also shows up when you remove an authentication method. (accidentally used "account" instead of "auth method" in my previous comment)

components/toast.js
huumn: 10 out of 10 programmers think close buttons aren't necessary (most of our modals didn't have them to begin with) and 10 out of 10 UX designers think you should have them.

I think the UX designers are right. People hate overlays and seeing/knowing that they can close them, comforts them.

components/delete.js
SatsAllDay: It has its own implementation. I updated the PR to add toast to the auth method operations, too.

components/toast.js
ekzyis: To admit, I didn't meant to remove the X button but just to make the hitbox of the X button the whole toast

SatsAllDay:
Some updated screenshots. I adjusted the width of the toast for screens <400px wide. I also borrowed the "X" close button from the modals⚡️

Another consideration: we might want to clear all toasts upon page navigation. It appears we can do that with [`router events`](https://nextjs.org/docs/pages/api-reference/functions/use-router#routerevents). Thoughts on that?

Screen Shot 2023-08-24 at 3 31 00 PM
image

huumn:
> Another consideration: we might want to clear all toasts upon page navigation. It appears we can do that with [router events](https://nextjs.org/docs/pages/api-reference/functions/use-router#routerevents). Thoughts on that?

Yes, that sounds like the right thing to do until we begin to apply them to other types of events (like non-custodial payment failures that we optimistically had shown as succeeding).

components/toast.js
```diff
@@ -0,0 +1,85 @@
...
+  const router = useRouter()
+  const [toasts, setToasts] = useState([])
+  const dispatchToast = useCallback((toastConfig) => {
+    const id = toastId++
+    toastConfig = {
+      ...toastConfig,
+      id
+    }
+    if (toastConfig.autohide && toastConfig.delay > 0) {
+      setTimeout(() => {
```
huumn, line 23: Not a big thing but why aren't we using the delay prop?

https://react-bootstrap.netlify.app/docs/components/toasts/#autohide

huumn: I'll make the change myself. I'm just curious if I'm missing something

SatsAllDay: I didn't want to leave the toast instance in the component tree after auto hide, so I chose to hide it by removing it from the array of toasts instead of using the delay prop.

Perhaps we could use the delay prop and also remove it from the array state? That might provide a transition out effect that would be nice. 

Idk if that made any sense...

huumn: I think it makes sense. 

It looks like we can use the delay prop and `onClose` to remote it from the toasts array. I'll push that in a second.

huumn:
Very solid work! We'll deploy master tomorrow probably

huumn:
Do you have another lightning address/lnurl-pay I can send sats to?

SatsAllDay:
You can send to my SN address WeAreAllSatoshi@stacker.news

SatsAllDay:
Unless you mean one other than that one?

huumn:
Oh yeah that's fine

huumn:
We have max balances on SN though so if you're going to keep tearing through issues, we'll need somewhere to send you way more sats than you can have on SN.

SatsAllDay:
Heh fair enough. Next time I'll provide somewhere else to receive sats

Issue #405, opened 2023-08-14
Combine "fund wallet" + "pay invoice"


ekzyis:
**Is your feature request related to a problem? Please describe.**

If the user wants to do something on the site but does not have enough balance, we ask if they want to fund their wallet or pay via invoice. This is an unnecessary step and prevents wallets like Alby to immediately prompt to pay an invoice.

**Describe the solution you'd like**

We could just immediately show an invoice. Wallets like Alby then prompt immediately to pay the invoice.
To still allow account funding, we could ask while the invoice is shown if they also want to fund their account.
We then create another invoice and any excess

**Additional context**

https://stacker.news/items/223799

PR #432, opened 2023-08-22
Use HODL invoices

### LINKED ISSUE #415: Use HODL invoices for atomic anon payments

The current implementation of anon payments is not 100% reliable since it's not atomic.

This is the current UI flow:
1. user commits to action
2. invoice with payment hash + hmac is created, stored in localStorage and shown to user
3. user pays invoice which funds balance of anon account
4. user agent uses payment hash + hmac to retry action
5. on success, we're done.
on error, the user should be able to request a refund using the hash + hmac stored in localStorage
(TBD in #401)

I see following problems with this:

1. We rely on the user session to not die between step 3 and 4. Sometimes, when I switch from my browser to a different application and then switch back (which is required for paying the invoices on mobile), the browser loads the site again which closes the modal and thus the action is not retried
=> bad UX and user has to request refund (and pay tx fees twice) / retry action in a different way
2. We rely on `localStorage` to save the payment hash + hmac in case an error happens after payment. However, in incognito mode, [browsers clear localStorage after each session](https://stackoverflow.com/questions/4386358/when-does-localstorage-get-cleaned-by-the-browser) or `localStorage` might not even work with some privacy settings enabled as mentioned in [this article](https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/)
3. We need to allow withdrawals from the anon balance (which multiple anons share) which comes with security risk
4. To prevent abuse, stackers would need to pay tx fees again for refunds.
5. These payments may be the first interaction some may have with SN. So any error in this interaction could lead to a very bad first impression since it involves money.

We could use [HODL invoices](https://lightningwiki.net/index.php/HODL_Invoice) to make the payments atomic.

Advantages:
- no refunds required anymore hence no additional security risk related to withdrawals and no double tx fees for (potentially new) stackers
- no hmacs required anymore
- vastly more reliable which leads to better UX and a foundation we can build additional features on top

ekzyis:
Not working yet, see FIXME in api/resolvers/item.js:72

>  // FIXME:
>  //   secret is null for HODL invoices since you have to specify your own hash with LND:
>  //   https://api.lightning.community/api/lnd/invoices/add-hold-invoice#messages
>  //   Therefore, LND does not know the preimage and we have to save it ourselves ...

Closes #415 

TODO:
- [x] find way to get preimage to settle hodl invoice after payment and successful action: _added new column `preimage` to invoice table_
- [x] make sure hodl invoices are canceled to not risk force closures.
According to this [comment](https://github.com/lightningnetwork/lnd/pull/2022#issuecomment-1175574865), HODL invoices are automatically canceled after they expire but I would like to see some official docs on this.
-- _invoices are now canceled if they expire in `checkInvoice` of worker_
- [x] delete code which is no longer needed like contacts: just add retry or cancel/refund button. refund will cancel the HODL invoice

---

Works now:

https://files.ekzyis.com/public/sn/hodl_invoices.mp4

---

New TODOs:
- [x] rebase on master
- [x] Fix invoices not used if session exists


huumn:
Zero out you anon user's account balance and test this.

Because the invoice isn't settled before the action, anon's account has an insufficient balance. 

This will probably require adding held funds to anon's account balance and reversing on cancel/expire.

ekzyis:
> Because the invoice isn't settled before the action, anon's account has an insufficient balance.

Good catch! Fixed that now. I picked

> adding held funds in another statement (within the same tx) before the action

since it required the least code and rollback is automatic which seems like the most secure solution

https://files.ekzyis.com/public/sn/hodl_invoices_2.mp4

I'll keep this as a draft now though since I want to sleep over this and do another pass over the code. Maybe I missed something.

api/resolvers/wallet.js
```diff
@@ -308,7 +328,8 @@ export default {
 
   Invoice: {
     satsReceived: i => msatsToSats(i.msatsReceived),
-    satsRequested: i => msatsToSats(i.msatsRequested)
+    satsRequested: i => msatsToSats(i.msatsRequested),
+    expired: i => new Date(i.expiresAt) <= new Date()
```
ekzyis, line 332: required to re-render component when invoice expired

huumn: What?

ekzyis: Without this, it will never show expired in the frontend since the component only renders again if the query actually returned something else.

huumn:
As we were philosophizing about code style on monday ...

One blanket observation about this PR that I think is worth noting is that we are introducing several more places where we are interacting with LND, ie when I imagine changing something about our interaction with lnd in the future I now have several more places to look/change.

This "interaction creep" is natural and it's not obvious to me how we'd reduce the "interaction points" in this case (as I'm not the author), but it's something we always want to keep in mind; ie how easy will it be for me to change this code in the future without introducing bugs.

api/resolvers/wallet.js
huumn: Ah, I see. I guess you could `setTimeout` on the frontend instead. Kind of weird to control rendering from the backend.

Come to think of it, this something we could add to our countdown component because it'd be nice to do this when someone is editing something and the edit window expires.

ekzyis: I used `onComplete` from the `<Countdown>` component now

ekzyis:
> As we were philosophizing about code style on monday ...

We did? :see_no_evil: :smile: 

> One blanket observation about this PR that I think is worth noting is that we are introducing several more places where we are interacting with LND, ie when I imagine changing something about our interaction with lnd in the future I now have several more places to look/change.

You are right that it's used in more places now but afaict, the only _new_ place is in api/resolvers/item.js to settle the hodl invoices, no?

Other than that, I only updated _existing_ LND code in api/resolvers/wallet.js (to conditionally create HODL invoices instead of normal invoices which automatically settle) and in the worker.

> This "interaction creep" is natural and it's not obvious to me how we'd reduce the "interaction points" in this case (as I'm not the author), but it's something we always want to keep in mind; ie how easy will it be for me to change this code in the future without introducing bugs.

I agree. I was thinking about if it's worth to also move the settlement code to the worker by confirming the invoice during item creation and adding a pgboss job. Then the worker would pick up the job and settle the invoice. Then all LND code for finalizing hodl invoices (settlement or cancellation) would be in the worker.

But I don't think it's worth it since it's more changes for not much benefit imo

huumn:
> I agree. I was thinking about if it's worth to also move the settlement code to the worker by confirming the invoice during item creation and adding a pgboss job. Then the worker would pick up the job and settle the invoice. Then all LND code for finalizing hodl invoices (settlement or cancellation) would be in the worker.
> 
> But I don't think it's worth it since it's more changes for only moving a single line to a different file imo

I think you're thinking about this right. It's not important that we do it *just this minute* but it needs to be on The List™️.

huumn:
Also sorry for my recent push, but bright side: it should allow you to delete a satisfying amount of duplicate code.

ekzyis:
> I'll keep this as a draft now though since I want to sleep over this and do another pass over the code. Maybe I missed something.

Yep, I did. The following current logic prevents paying per invoice for logged-in stackers:

_api/resolvers/item.js:1123 @ 41600f8_:
```
let invoice
if (me) {
  item.userId = Number(me.id)
} else {
  if (!invoiceHash) {
    throw new GraphQLError('you must be logged in or pay', { extensions: { code: 'FORBIDDEN' } })
  }
  invoice = await checkInvoice(models, invoiceHash, invoiceHmac, item.parentId ? ANON_COMMENT_FEE : ANON_POST_FEE)
  item.userId = invoice.user.id
  spamInterval = ANON_ITEM_SPAM_INTERVAL
}
```

api/resolvers/item.js
```diff
@@ -1127,12 +1142,19 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
...
-    models,
+  const trx = [
     models.$queryRawUnsafe(`${SELECT} FROM create_item($1::JSONB, $2::JSONB, $3::JSONB, '${spamInterval}'::INTERVAL) AS "Item"`,
-      JSON.stringify(item), JSON.stringify(fwdUsers), JSON.stringify(options)),
-    ...trx)
-  item = Array.isArray(result) ? result[0] : result
+      JSON.stringify(item), JSON.stringify(fwdUsers), JSON.stringify(options))
+  ]
+  if (invoice) {
+    trx.unshift(models.$queryRaw`UPDATE users SET msats = msats + ${invoice.msatsReceived}`)
```
ekzyis, line 1150: ooof, lol, I just added the sats from the invoice to **_every_** user ... :see_no_evil:

huumn: lol wouldn't it be *amazing* if postgres knew which user it was though lol

ekzyis:
I was thinking about adding #405 in this PR but I think it's better to do this in a separate PR now.

It should be fairly easy though. With the current logic, any additional paid sats will already go to the user balance.
So just frontend changes required

huumn:
If it interests you, please also consider refactoring "invoicable" stuff into a Form error handler.

huumn:
Love seeing all these code deletions! So satisfying!

components/invoice.js
```diff
@@ -238,12 +231,13 @@ export const useInvoiceable = (fn, options = defaultOptions) => {
...
       }
     }
-    setFnArgs(args)
-    await createInvoice({ variables: { amount } })
+    setFormValues(formValues)
+    setSubmitArgs(submitArgs)
+    await createInvoice({ variables: { amount: cost } })
     // tell onSubmit handler that we want to keep local storage
     // even though the submit handler was "successful"
     return { keepLocalStorage: true }
```
huumn, line 239: do we still need this now that Form is invoicable-aware?

(This was one of the things that tipped us off that original pattern might be improved.)

ekzyis: Mhh, yes, that was one of the things that tipped of us off to improve the original pattern. But I realized that moving the error handling code into the form code wouldn't be enough. For example, at this line here, it's not about errors but communicating back on success when the form should clear local storage or not.

So I am not sure how to fit that logic into the form code. It would have to be able to distinguish between different submit handlers successes which it can't since we just pass an opaque `onSubmit` to it and then additionally (and conditionally) wrap it with `useInvoiceable`. If this wrapped function is called successfully, form can not know by itself if it should clear local storage since the action was submitted successfully or if only an invoice was created and the invoice modal opened afaict

huumn: Cool sounds good. I still haven't pulled to take a close look. I'll let you know if any simplifications jump out.

huumn:
Oh dang I just realized updating by rebasing re-authors them. I'll stop doing that.

huumn:
First of all, this is AMAZING.

One thing I've noticed so far: local storage isn't cleared on success ... better than losing it on failure though.

ekzyis:
> Oh dang I just realized updating by rebasing re-authors commits. I'll stop doing that.

Oh, I thought rebases only update "committed by" not "authored by"

> First of all, this is AMAZING.

:blush: 

> One thing I've noticed so far: local storage isn't cleared on success ... better than losing it on failure though.

Oh, right, I'll try to think of something

Issue #365, opened 2023-07-22
Search field no longer exists at bottom, except for /search


cointastical:
Not sure if that was intended behavior, or if that was a regression.   But it's gone now, ... as far as I can tell, site-wide, except for /search.

If that change was intended behavior, then perhaps modify /search to put it prominently up in the body, like how Google search UI (and most every other search page) has it, at least for the first /search page, maybe only when there's no ?q= yet, or ... if that's now the main method for search, then fix the serach fields to be up above the search results all the time, if /search is the only page that has those fields now.

Others requesting a fixed search bar up in the body (when in /search page): 

*Stacker Search box proposal - search box similar to google*
https://stacker.news/items/148535

ekzyis:
makes sense to have the search field at the top, thanks for the issue!

brihuang99:
Sharing @cryptofolyo's mockup here:

![image](https://github.com/stackernews/stacker.news/assets/122472324/6f3c2daa-19f1-401c-8dfc-ddc5bd375cc4)

huumn:
While I agree that search being at the top is ideal, it's not as trivial as drawing it at the top. 

That mockup does not account for mobile or other context generally.

brihuang99:
> While I agree that search being at the top is ideal, it's not as trivial as drawing it at the top.
> 
> That mockup does not account for mobile or other context generally.

No problem, I understand. I'll see about providing some designs that give more context without changing the current UI drastically.

brihuang99:
Here are some of my ideas: 

## Desktop

#### Empty State
![Search 1](https://github.com/stackernews/stacker.news/assets/122472324/60bc6643-7c55-45f3-839d-73c1c45d4846)

#### Search State
![Search 2](https://github.com/stackernews/stacker.news/assets/122472324/aca79f7f-d739-497d-9f5b-46e3242f89fc)


## Mobile

#### Empty State
![Mobile Search 1](https://github.com/stackernews/stacker.news/assets/122472324/ed239c6c-c593-4977-8361-915cb3950756)

#### Search State
![Mobile Search 2](https://github.com/stackernews/stacker.news/assets/122472324/de125546-a9f0-40f1-8cfa-ea6ce5bd58f3)

#### Scrolled State??
![Mobile Search 3](https://github.com/stackernews/stacker.news/assets/122472324/41355f18-1466-42a7-bfcc-395d267aadc8)

[...]

PR #433, opened 2023-08-22
move search bar to top of page

rleed:
The approach is straightforward: put the search controls at the top of the results list, similarly to the illustrations in the mockups of issue #365. If the user scrolls, the search bar assumes a floating position at the top of the screen and the list of items scrolls underneath it (as also suggested in the mockups).

Representative screenshots non-scrolled and scrolled:

Untitled

Notes on the CSS changes:
- the searchSection and searchPadding classes are sort of complementary (only one is display: block at a time)
- they both have an associated class "float" that toggles on and off depending on scroll position
- a media query was needed to fill in missing container padding on xs screens (not sure why it was missing). I would have preferred an @include directive not to hard-code the bootstrap xs width in the @media query, but that is only possible in scss files, not modules, and I didn't think it would be good to risk global scss changes just for this unique case.
- the form left/right were set to 0 because the containers already provide the correct (responsive) shape.
- the form flex-flow was set to nowrap to fix unsightly overlaps of the search inputs with other parts of the page on very narrow window sizes.

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -146,7 +146,7 @@ const orderByClause = async (by, me, models, type) => {
       return 'ORDER BY "Item".ncomments DESC'
     case 'sats':
       return 'ORDER BY "Item".msats DESC'
-    case 'votes':
+    case 'zaprank':
       return await topOrderByWeightedSats(me, models)
     default:
       return `ORDER BY ${type === 'bookmarks' ? '"bookmarkCreatedAt"' : '"Item".created_at'} DESC`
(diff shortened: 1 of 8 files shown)


huumn:
I'd like the selects to be moved below the dropdown. Do you want to take a stab at that or should I?

huumn:
Yes below the search box please!

rleed:
Untitled

Besides moving the inputs, I also moved the `my-3` class from one div to the other to keep the spacing equivalent.

huumn:
Looks great!

ekzyis:
This branch also got messed up somehow. Maybe a rebase will help?

huumn:
I have rebased thee

rleed:
Sorry about the merge mess... I think I've got the hang of it now after fixing the other one.

PR #436, opened 2023-08-23
Fix expireSecs ignored for anons

ekzyis:
I think it's weird that `expireSecs` is totally ignored for anons.

At least it should be possible to set a lower expiry (even if not used at the moment)

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -215,7 +215,7 @@ export default {
       let balanceLimit = BALANCE_LIMIT_MSATS
       let id = me?.id
       if (!me) {
-        expirePivot = { minutes: 3 }
+        expirePivot = { seconds: Math.min(expireSecs, 180) }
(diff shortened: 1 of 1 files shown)


huumn:
This change allows someone to DoS anon invoice creation.

huumn:
We should probably comment about this though because I feel you

ekzyis:
> This change allows someone to DoS anon invoice creation.

Does it? I used `Math.min(expireSecs, 180)` so it's always maximally as long as before the change. You can only make it shorter with this change

huumn:
Fuck I'm sorry hadn't had coffee yet. This is legit.

ekzyis:
Haha no worries, I can see this happening to me, too

huumn:
Um this got deleted

ekzyis:
> Um this got deleted

Damn, I probably accidentally pushed to github instead of to my gitea instance. Need to fix my setup lol

PR #440, opened 2023-08-24
Add meta sub to rss popover in footer

SatsAllDay:
Add meta sub to rss popover in footer and move nostr sub up to top row for 3 and 3 instead of 2 and 4

I mirrored the order from the sub picker select component (home, bitcoin, nostr, tech, meta, jobs)

image


huumn:
I noticed that too and keep forgetting to do anything about it!

SatsAllDay:
I should have taken my own advice, opened an issue, and marked it "Good first issue" lol

PR #443, opened 2023-08-26
Subscribe to a user

### LINKED ISSUE #442: Subscribe to a user
# Subscribe to a user

As a stacker, I'd like to be notified when another stacker has activity (posts, comments, etc.)

## UX Proposal

Add an overflow menu next to stacker's nym in their profile page with a "subscribe" dropdown item. Toast feedback just like when subscribing to a post

## Technical Proposal

new DB model: `UserSubscription`
```
followerId    Int
followeeId    Int
createdAt     DateTime @default(now()) @map("created_at")
follower      User     @relation(fields: [followerId], references: [id], onDelete: Cascade)
followee      User     @relation(fields: [followeeId], references: [id], onDelete: Cascade)

@@id([followerId, followeeId])
@@index([createdAt], map: "UserSubscription.created_at_index")
```

new gql mutation: `subscribeUser(id: ID)` which toggles the user subscription for the specified user

update the `user` typeDef to include a boolean `meSubscription` field, indicating whether the calling user is subscribed to the target user
update the `user` api resolver to implement the `meSubscription` field using the above model

update the notifications api resolver for the notifications query to also search for items (posts, comments) that are from any followed users.

Question: Introduce a new notification type, or re-use `reply`?

I'm sure there's more details I've overlooked, but this is what I expect it to take to make this happen.

Thoughts? Concerns? Questions?

SatsAllDay:
Closes #442 

See #442 for the details of how this is implemented. It encapsulates the technical proposal.

TODOs:
- [x] Send Push Notification for posts, comments being created by a followed user
- [x] Consolidate some of the re-used UI components
- [ ] see if we can de-dupe notifications like being mentioned in an item by a user you follow so it doesn't show up twice in the notifications page
- [x] don't show user action overflow menu for unauthenticated users

### DIFF
--- a/api/resolvers/notifications.js
+++ b/api/resolvers/notifications.js
@@ -91,6 +91,16 @@ export default {
               AND "Item"."userId" <> $1 AND "Item".created_at <= $2
             ${await filterClause(me, models)}
             ORDER BY "sortTime" DESC
+            LIMIT ${LIMIT}+$3)
+          UNION DISTINCT
+          (SELECT DISTINCT "Item".id::TEXT, "Item".created_at AS "sortTime", NULL::BIGINT as "earnedSats",
+            'FollowActivity' AS type
+            FROM "Item"
+            JOIN "UserSubscription" ON "Item"."userId" = "UserSubscription"."followeeId"
+            WHERE "UserSubscription"."followerId" = $1
+              AND "Item".created_at <= $2
+            ${await filterClause(me, models)}
+            ORDER BY "sortTime" DESC
             LIMIT ${LIMIT}+$3)`
       )
 
@@ -272,6 +282,9 @@ export default {
     item: async (n, args, { models, me }) => getItem(n, { id: n.id }, { models, me })
   },
   Reply: {
+    item: async (n, args, { models, me }) => getItem(n, { id: n.id }, { models, me })
+  },
+  FollowActivity: {
     item: async (n, args, { models, me }) => getItem(n, { id: n.id }, { models, me })
   },
   JobChanged: {
--- a/api/resolvers/notifications.js
+++ b/api/resolvers/notifications.js
@@ -99,6 +99,8 @@ export default {
             JOIN "UserSubscription" ON "Item"."userId" = "UserSubscription"."followeeId"
             WHERE "UserSubscription"."followerId" = $1
               AND "Item".created_at <= $2
+              -- Only show items that have been created since subscribing to the user
+              AND "Item".created_at >= "UserSubscription".created_at
             ${await filterClause(me, models)}
             ORDER BY "sortTime" DESC
             LIMIT ${LIMIT}+$3)`
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1137,6 +1137,29 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
 
   await createMentions(item, models)
 
+  const notifyUserSubscribers = async () => {
+    try {
+      const userSubs = await models.userSubscription.findMany({
+        where: {
+          followeeId: Number(item.userId)
+        },
+        include: {
+          followee: true
+        }
+      })
+      const isPost = !!item.title
+      await Promise.allSettled(userSubs.map(({ followerId, followee }) => sendUserNotification(followerId, {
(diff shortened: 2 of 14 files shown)


SatsAllDay:
> see if we can de-dupe notifications like being mentioned in an item by a user you follow so it doesn't show up twice in the notifications page

This might be a larger effort, if we even want to do it.

For example, if someone replies to you and also tags you, you'd get a notification for the reply, and a notification for the mention. So it goes further than the changes being added in this PR in support of User Subscriptions. Perhaps it warrants a separate effort to reduce/consolidate notifications across the board.

Same thing would also apply for push notifications - not sending a push for a reply _and_ user activity (and other types once implemented) when it all comes from one new comment, for example.

I'm open to input on this, but I will hold off on making any changes related to it since it's pretty open ended as far as how we want to implement it.

api/resolvers/notifications.js
```diff
@@ -91,6 +91,18 @@ export default {
               AND "Item"."userId" <> $1 AND "Item".created_at <= $2
             ${await filterClause(me, models)}
             ORDER BY "sortTime" DESC
+            LIMIT ${LIMIT}+$3)
+          UNION DISTINCT
```
SatsAllDay, line 95: This `union distinct` probably won't do any elimination since the `type` is `FollowActivity`, not `Reply`.

huumn: You're right, it won't.

We'll need to `group by` the item id to make them distinct.

huumn:
> For example, if someone replies to you and also tags you, you'd get a notification for the reply, and a notification for the mention. So it goes further than the changes being added in this PR in support of User Subscriptions. Perhaps it warrants a separate effort to reduce/consolidate notifications across the board.

Yep, this has been a stick in my boot. It's mostly a matter of forming a sub-union over these potentially duplicate notifications and grouping by id ... this sub-union is then unioned with the other notifications.

SatsAllDay:
> > For example, if someone replies to you and also tags you, you'd get a notification for the reply, and a notification for the mention. So it goes further than the changes being added in this PR in support of User Subscriptions. Perhaps it warrants a separate effort to reduce/consolidate notifications across the board.
> 
> Yep, this has been a stick in my boot. It's mostly a matter of forming a sub-union over these potentially duplicate notifications and grouping by id ... this sub-union is then unioned with the other notifications.

Yea, that approach makes sense. I think it would be best to tackle that in a separate PR. What do you think?

api/resolvers/notifications.js
SatsAllDay: Let me know if you want me to address this in this PR, or if it should be lumped in with the general de-dupe effort mentioned at the root-level of this PR.

huumn:
> > > For example, if someone replies to you and also tags you, you'd get a notification for the reply, and a notification for the mention. So it goes further than the changes being added in this PR in support of User Subscriptions. Perhaps it warrants a separate effort to reduce/consolidate notifications across the board.
> > 
> > 
> > Yep, this has been a stick in my boot. It's mostly a matter of forming a sub-union over these potentially duplicate notifications and grouping by id ... this sub-union is then unioned with the other notifications.
> 
> Yea, that approach makes sense. I think it would be best to tackle that in a separate PR. What do you think?

You're right, that's the move.

SatsAllDay:
I think this is ready for a review. No rush, just wanted to make it clear that I _think_ I am done tweaking it. I am working on a write-up for the notification de-dupe effort but that'll land in a separate issue.

huumn:
Solid!

SatsAllDay:
Here's a LN invoice for whenever this gets deployed

lnbc1pjwmmdxpp5damqrqcl5nusmx60n62cfvtxsve7usc3h69jx7kcwgs7mz8jptksdqqcqpjsp53hj7283ctnyzx54p56kzghsw7dewernxxq65xrvf0pvmme9vqewq9q7sqqqqqqqqqqqqqqqqqqqsqqqqqysgqmqz9gxqyjw5qrzjqwryaup9lh50kkranzgcdnn2fgvx390wgj5jd07rwr3vxeje0glcllcn67flrvwvn5qqqqlgqqqqqeqqjq8vjrmkcu56y7vw5mkle5u6jj86tccaqpq29a4ect667cfkzmkhhjlfu395zz767xm25qme97kpm6w25tke75dfml4cj975hs2s0hl8qqxjtyja

PR #450, opened 2023-08-28
WIP: Support server-side debugging when running local in docker

SatsAllDay:
Update ports exposed, and run the next server with inspect flag

Currently does not work due to bug in next.js
See related PR:
https://github.com/vercel/next.js/pull/53683

Leaving this in draft mode until next.js fixes their issue.


huumn:
ek and I were talking about it this morning! 🙌

SatsAllDay:
Woo hoo! Upstream support has merged https://github.com/vercel/next.js/pull/65006

Now we wait for it to be GA'd and then we can bump nextjs dependency versions and then we can debug!

huumn:
Looks like this still hasn't been cherry picked into v14.2. Well dang.

SatsAllDay:
Big sad

huumn:
On v14.2.17, it's still not picked in.

ekzyis:
According to https://github.com/vercel/next.js/commit/ab86fcf5e6c77d0885ab5597d0bcf75fb7ded503, this commit entered a stable release for the first time in v15.0.0. Doesn't look like this will ever be included in a Next v14 release.

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "audit:blast-radius": "node scripts/audit-blast-radius.js",
-    "dev": "NODE_OPTIONS='--trace-warnings --enable-source-maps' next dev --webpack",
+    "dev": "NODE_OPTIONS='--trace-warnings --enable-source-maps --inspect=0.0.0.0:9229' next dev --webpack",
```
Soxasora, line 7: I'm not sure about enabling `inspect` all the time, the local dev server is already slow as-is. Maybe something like `dev:inspect`?

This also means that we'd have to find a way to run sndev with `npm run dev:inspect` (Dockerfile). Maybe with something like `./sndev start --inspect`

Soxasora (commented):
A really welcomed addition, especially now that we're on Next.js 16 and it *should* support `--inspect`.

If you're still interested in this PR, I gave my opinion on the way `inspect` should be enabled down below. Otherwise we'll pick it ourselves ^^

Thanks!

PR #455, opened 2023-08-29
Replace FundError with InvoiceModal

### LINKED ISSUE #405: Combine "fund wallet" + "pay invoice"

ekzyis:
* invoices are no longer deleted to prevent double-spends but marked as confirmed.
      therefore, during checkInvoice, we also check if the invoice is already confirmed.
* instead of showing FundError (with "fund wallet" and "pay invoice" as options), we now always immediately show an invoice
* since flagging, paying bounties and poll voting used FundError but only allowed spending from balance, they now also support paying per invoice

Based upon #432, so #432 should be merged first

Closes #405


ekzyis:
I'll rebase this now since #432 was merged

huumn:
I'm not sure the info message actually scratches the issue's itch. It's just kind of telling people they can overpay an invoice which most wallets don't support and most people aren't aware of (not to mention there are limitations on how much you can overpay an invoice). So I suspect this message is more confusing than helpful.

After seeing this invoicable stuff live for a bit, I don't think we *need* the ability to fund an account from these modals. It might not be worth the UX bloat to even do it how I imagined it and people can always put sats in their wallet if they want to.

I'm going to remove the info message for now.

PR #457, opened 2023-08-30
De-dupe item-driven notifications

### LINKED ISSUE #453: De-dupe notifications
# De-dupe notifications

There are many different kinds of notifications support by SN. Sometimes, one action by a user can trigger several notifications to appear for another user. It would be nice if these were de-duped/aggregated so as to reduce noise and visual clutter.

### Creating a top-level post

* If a user is mentioned in the content of the post, this will trigger a notification for the mentioned user (if enabled)

* If a user is subscribed to the user who posts, this will trigger a notification for the subscriber.

### Replying to a top-level post

* If a user is mentioned in the content of the comment/reply, this will trigger a notification for the mentioned user (if enabled)

* A notification is triggered for a user to which the comment is replying, e.g. the OP of the post, or a parent comment in the reply tree (this is conditional based on whether descendent notifications are enabled)

* If a user is subscribed to the thread upon which the reply is made, this will trigger a notification for the subscriber.

* If a user is subscribed to the user who posts, this will trigger a notification for the subscriber.

## Summary

Based on the above, we can see that the duplicate notifications come when users create items, and when other users are subscribed to the posting user, subscribed to the containing thread, mentioned in the item, or are being replied to in one way or another.

Ideally, we would only dispatch 1 notification per item, as applicable.

How do we represent this notification? I propose a ranking of precedence, in descending order:

1. mentions

2. replies

3. thread subscriptions

4. user subscriptions

In other words, if you are mentioned in an item from a user that you subscribe to, the notification is labeled as a mention. If you are replied to by a user that you subscribe to, the notification is labeled as a reply, etc.

## Technical details

From a technical perspective, as @huumn mentioned [here](https://github.com/stackernews/stacker.news/pull/443#issuecomment-1694722055), we should be able to reduce duplication by performing unions of the sources of dupes, grouped by item id, then unioning with the rest of the notifications.

I'm not sure how exactly we will implement the above order of precedence, but that can probably also be baked in to the SQL queries

### Push notifications

I think once we get the "in-app" notifications sorted, a similar kind of logic should be applied to our push notifications, as well. In other words, let's not spam users with push notifications for one item being created. Perhaps that can come separately, since push notifications aren't widely integrated yet, and getting this work done first would set clear expectations for the push implementation.

### Zapping an item

This isn't actually a source of duplicate notifications, but does have some lacking behavior. It can be addressed separately, IMO.

* When an item is zapped, the creator of the item gets notified (if enabled)

* Note: If the zap is for a post that has forwarding rules configured, the forward recipients _DO NOT_ currently get notified. IMO this is something we should fix

SatsAllDay:
Update how we query for item-driven notifications to de-dupe them for:

1. mentions
2. replies
3. thread subscriptions
4. user subscriptions

users should recieve only 1 notification for any given item created by another user, following the above priority order

this is accomplished by querying for replies to current user, replies on subscribed threads, items from subscribed users, and mentions (if enabled for the current user), sorting those results by item id and priority order within item id based on notification type, then selecting the highest priority notification from each item id

these results are then union all'ed with all other notification types, unchanged.

Closes #453

### DIFF
--- a/api/resolvers/notifications.js
+++ b/api/resolvers/notifications.js
@@ -71,32 +71,40 @@ export default {
 
       const queries = []
 
-      queries.push(
-        `(SELECT DISTINCT "Item".id::TEXT, "Item".created_at AS "sortTime", NULL::BIGINT as "earnedSats",
-            'Reply' AS type
-            FROM "Item"
-            JOIN "Item" p ON ${meFull.noteAllDescendants ? '"Item".path <@ p.path' : '"Item"."parentId" = p.id'}
-            WHERE p."userId" = $1 AND "Item"."userId" <> $1 AND "Item".created_at <= $2
-            ${await filterClause(me, models)}
-            ORDER BY "sortTime" DESC
-            LIMIT ${LIMIT}+$3)
-          UNION DISTINCT
-          (SELECT DISTINCT "Item".id::TEXT, "Item".created_at AS "sortTime", NULL::BIGINT as "earnedSats",
-            'Reply' AS type
-            FROM "ThreadSubscription"
-            JOIN "Item" p ON "ThreadSubscription"."itemId" = p.id
-            JOIN "Item" ON ${meFull.noteAllDescendants ? '"Item".path <@ p.path' : '"Item"."parentId" = p.id'}
-            WHERE
-              "ThreadSubscription"."userId" = $1
-              AND "Item"."userId" <> $1 AND "Item".created_at <= $2
-            ${await filterClause(me, models)}
-            ORDER BY "sortTime" DESC
-            LIMIT ${LIMIT}+$3)`
+      const itemDrivenQueries = []
 
+      // Replies
+      itemDrivenQueries.push(
+        `SELECT DISTINCT "Item".id::TEXT, "Item".created_at AS "sortTime", NULL::BIGINT as "earnedSats",
+        'Reply' AS type
+        FROM "Item"
+        JOIN "Item" p ON ${meFull.noteAllDescendants ? '"Item".path <@ p.path' : '"Item"."parentId" = p.id'}
+        WHERE p."userId" = $1 AND "Item"."userId" <> $1 AND "Item".created_at <= $2
+        ${await filterClause(me, models)}
+        ORDER BY "sortTime" DESC
+        LIMIT ${LIMIT}+$3`
       )
 
-      queries.push(
-        `(SELECT DISTINCT "Item".id::TEXT, "Item".created_at AS "sortTime", NULL::BIGINT as "earnedSats",
+      // Thread subscriptions
+      itemDrivenQueries.push(
+        `SELECT DISTINCT "Item".id::TEXT, "Item".created_at AS "sortTime", NULL::BIGINT as "earnedSats",
+        'Reply' AS type
+        FROM "ThreadSubscription"
+        JOIN "Item" p ON "ThreadSubscription"."itemId" = p.id
+        JOIN "Item" ON ${meFull.noteAllDescendants ? '"Item".path <@ p.path' : '"Item"."parentId" = p.id'}
+        WHERE
+          "ThreadSubscription"."userId" = $1
+          AND "Item"."userId" <> $1 AND "Item".created_at <= $2
+          -- Only show items that have been created since subscribing to the thread
+          AND "Item".created_at >= "ThreadSubscription".created_at
+        ${await filterClause(me, models)}
+        ORDER BY "sortTime" DESC
+        LIMIT ${LIMIT}+$3`
+      )
+
+      // User subscriptions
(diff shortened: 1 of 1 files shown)


SatsAllDay:
Side note, this should also remove duplicate notifications if you're mentioned more than once in an item :)

huumn:
> Side note, this should also remove duplicate notifications if you're mentioned more than once in an item :)

That shouldn't have been an issue before unless we regressed somewhere along the way.

SatsAllDay:
> > Side note, this should also remove duplicate notifications if you're mentioned more than once in an item :)
> 
> That shouldn't have been an issue before unless we regressed somewhere along the way.

You are right. I misread some of my notifications in the past (on one of the SN release posts). I thought it had 2 mention notes but it indeed only has 1.

api/resolvers/notifications.js
```diff
@@ -71,32 +71,40 @@ export default {
...
+        `SELECT DISTINCT "Item".id::TEXT, "Item".created_at AS "sortTime", NULL::BIGINT as "earnedSats",
+        'Reply' AS type
+        FROM "ThreadSubscription"
+        JOIN "Item" p ON "ThreadSubscription"."itemId" = p.id
+        JOIN "Item" ON ${meFull.noteAllDescendants ? '"Item".path <@ p.path' : '"Item"."parentId" = p.id'}
+        WHERE
+          "ThreadSubscription"."userId" = $1
+          AND "Item"."userId" <> $1 AND "Item".created_at <= $2
+          -- Only show items that have been created since subscribing to the thread
+          AND "Item".created_at >= "ThreadSubscription".created_at
```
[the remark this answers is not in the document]
huumn: I think this is fine. More of what I'd expect intuitively

api/resolvers/notifications.js
```diff
@@ -106,25 +114,40 @@ export default {
...
+      // mentions
       if (meFull.noteMentions) {
-        queries.push(
-          `(SELECT "Item".id::TEXT, "Mention".created_at AS "sortTime", NULL as "earnedSats",
+        itemDrivenQueries.push(
+          `SELECT "Item".id::TEXT, "Mention".created_at AS "sortTime", NULL as "earnedSats",
             'Mention' AS type
             FROM "Mention"
             JOIN "Item" ON "Mention"."itemId" = "Item".id
-            LEFT JOIN "Item" p ON "Item"."parentId" = p.id
```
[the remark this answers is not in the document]
huumn: Nice, I would've missed that

api/resolvers/notifications.js
```diff
@@ -106,25 +114,40 @@ export default {
...
         )
       }
+      // Inner union to de-dupe item-driven notifications
+      queries.push(
+        // Only record per item ID
+          `(SELECT DISTINCT ON (id) *
+          FROM (
+            SELECT *
+            FROM (
+              ${itemDrivenQueries.map(q => `(${q})`).join(' UNION ALL ')}
```
huumn, line 142: I removed this `UNION DISTINCT` because the distinct on (id) already covers distinctness

api/resolvers/notifications.js
```diff
@@ -106,25 +114,40 @@ export default {
...
+            ) as inner_union
+            ORDER BY id ASC, CASE
```
huumn, line 144: Really nice hack!

SatsAllDay: Thank you! I was struggling with accomplishing this via `GROUP BY`, fortunately I came up with this alternative.

SatsAllDay:
Here's another LN invoice

lightning:lnbc1pjwlu8wpp5nva3yeggnf5kuqfd9azv6lyd2wmk5wmcsw93gn9ex0usc3kedt9qdqqcqpjsp54aueuuvrpkqqvnpv37cusn73708tls7ddctj42pekq0a5ztys7sq9q7sqqqqqqqqqqqqqqqqqqqsqqqqqysgqmqz9gxqyjw5qrzjqwryaup9lh50kkranzgcdnn2fgvx390wgj5jd07rwr3vxeje0glcllcn67flrvwvn5qqqqlgqqqqqeqqjqc27w5u7x20vyn85aypugpkfjxpmpppjrj8kxed2zzat9230swpkyxc4vgjltxajlp5yx5m4rfh7hytpjc2sccdd0hg9gav827n7hr2gqdlu09q

Should I just put these in the PR description when I open it? I think they're valid for a week upon creation.

huumn:
If you can make them 2 weeks that'd put me less under the gun, but we can give it a shot.

If you're using an lnurl-pay compatible wallet an lnurl-pay code would be even better! (At least until we have non-custodial overflow on SN.)

SatsAllDay:
Let me explore and see what I can come up with. I don't want to apply unnecessary pressure :)

Issue #411, opened 2023-08-15
Push Notifications setting lost again


SatsAllDay:
**Description**
It appears I have again lost the setting where I enabled push notifications. I think this was an issue in the past that was fixed at one point. 

**Steps to Reproduce**
Open PWA and find Push Notifications disabled on the notifications page when it was enabled previously. 

**Expected behavior**
Push notifications remains enabled if enabled. 

**Screenshots**
N/A

**Logs**
N/A

**Environment:**
iOS PWA

**Additional context**
N/A 

cc @ekzyis

SatsAllDay:
No, I am not sure it was fixed. I thought it was reported as fixed, but perhaps that was just optimistic :)

Let me know if there's any additional information I can provide to help debug further.

ekzyis:
> No, I am not sure it was fixed. I thought it was reported as fixed, but perhaps that was just optimistic :)

Ah, I see, haha, yes, we were hoping [in this release](https://stacker.news/items/221420) it would be fixed now with [these changes](https://github.com/stackernews/stacker.news/pull/370).

> Let me know if there's any additional information I can provide to help debug further.

Thanks for the offer, now that you ask, there might indeed be something: Can you confirm if anything is saved as `subscription` in your IndexedDB? Don't share the full URL / the keys of the subscription since anyone could use that to send push subscriptions to your device. If anything is stored there, it should be the current subscription and thus fine.

![2023-08-15-164527_598x281_scrot](https://github.com/stackernews/stacker.news/assets/27162016/e4d36264-0754-43e7-88a6-3c0bdce58863)

Other than that, I don't think there is any more information we can request at the moment. I also just noticed we don't log user ids during web push errors so we can't associate errors with users with specific subscriptions at the moment. Will push a fix soon.

Opt-in client-side logging would be useful at this point for debugging now. I have only worked with [Sentry](https://sentry.io/) so far and Sentry seems overkill for us at this stage. [This website](https://www.sitepoint.com/logging-errors-client-side-apps/) mentions some simpler alternatives including just rolling our own code.

@huumn, what do you think about client-side logging?

huumn:
I'm for it. Two requirements:

1. doesn't in any way violate user privacy
2. without a doubt helps us help stackers

SatsAllDay:
[...]

PR #463, opened 2023-08-31
Add setting to send diagnostics back to SN

ekzyis:
This should help to investigate #411.

This adds a new opt-in setting `allow diagnostics`:

2023-09-09-013253_296x208_scrot

2023-09-09-021358_510x404_scrot

I added a new endpoint `/api/log` where diagnostics data from devices of stackers which have this setting enabled are sent. **This endpoint is unauthenticated for privacy reasons.**

To know which events belong to the same client, a fancy random name is generated which is only known to the client. Hence if a stacker uses two different devices, it will show up with two different names. This should make sending these events **pseudonymous** since we could trace back to the individual stacker if the stacker tells us their fancy name. It would only be anonymous if this step is not reversible.

Additionally, anything sent to that endpoint is also sent into a Slack channel:

_sample log in slack:_
2023-09-09-013046_1058x773_scrot

Since the endpoint is unauthenticated, you don't even need an account to send something to it. This may be a problem but if we require authentication, this seems antithetical to 

> without a doubt doesn't violate user privacy

-- https://github.com/stackernews/stacker.news/issues/411#issuecomment-1679100707

Additionally, I think using authentication doesn't even really help against potential spam since anyone can easily create an account.


ekzyis:
Asked for feedback here: https://stacker.news/items/248388

components/logger.js
```diff
@@ -0,0 +1,107 @@
...
+  const pickRandom = (array) => array[Math.floor(Math.random() * array.length)]
+  const adj = pickRandom(fancyNames.adjectives)
+  const noun = pickRandom(fancyNames.nouns)
+  const id = Math.floor(Math.random() * fancyNames.maxSuffix)
+  return `${adj}-${noun}-${id}`
+}
+
+function detectOS () {
+  const userAgent = window.navigator.userAgent
+  const platform = window.navigator?.userAgentData?.platform || window.navigator.platform
```
huumn, line 16: Here you check for navigator but you don't the line before. I think I recall navigator not being defined in some browsers. It might be best to bail out of os detection if navigator isn't defined.

ekzyis: Copied that code [from SO](https://stackoverflow.com/a/38241481). Not sure why there is this inconsistency.

Also, `navigator` should be [supported by every browser](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform#browser_compatibility). We may had this error because of SSR?

But I'll use `window.navigator?` anyway. Better be safe than sorry :)

ekzyis:
Since we discussed hiding the setting and only enabling it for us manually ...  should we just hide the setting?

The advantage would be that we don't need to explain what it does. 

The disadvantage would be that it's less transparent that this exists and some stackers may actually want to help us.

I'll also try to fix the error boundary in this PR. So we'll get more useful information out of these changes already.

pages/api/log/index.js
```diff
@@ -0,0 +1,34 @@
...
+const slackPostMessage = ({ id, level, name, message, env, context }) => {
+  const text = `\`${new Date().toISOString()}\` | \`${id} [${level}] ${name}\` | ${message} | ${toKV(context)} | ${toKV({ os: env.os })}`
+  return fetch(slackApiUrl + '/chat.postMessage', {
+    method: 'post',
+    headers: {
+      Authorization: `Bearer ${botToken}`,
+      'Content-type': 'application/json'
+    },
+    body: JSON.stringify({ channel: channelId, text })
+  })
```
ekzyis, line 20: Mhh, we may need to buffer requests:

> 
Posting messages | 1 per second | Short bursts >1 allowed. If you attempt bursts, there is no guarantee that messages will be stored or displayed to users. If the burst exceeds available limits, users will see an error message indicating that some messages from your app are not being displayed.
-- | -- | --

-- https://api.slack.com/docs/rate-limits

ekzyis: [This article](https://medium.com/slack-developer-blog/handling-rate-limits-with-slacks-apis-f6f8a63bdbdc) from someone working at Slack mentions the Slack SDK which uses a queue. I'll just use that. Better than figuring out how large messages can be since and rolling our own queue imo.

huumn:
> Since we discussed hiding the setting and only enabling it for us manually ... should we just hide the setting?

Nah let's keep it there. I'll do a pass on the copy before/after merging.

ekzyis:
Using the [Slack SDK](https://github.com/slackapi/node-slack-sdk) now.

The only (minor, imo) issue is that messages from a client arrive out of order in slack since we don't wait until the previous slack requests are done before submitting new slack requests.

Doing this would require a mutex in the server or client afaict. We can't just use `await` somewhere since the slack requests are triggered in different API requests which are triggered at different points in the client code.

But I think this fine for the first iteration.

I also looked into the error boundary. I noticed that `getDerivedStateFromError` is triggered but not `componentDidCatch`. Also, `ErrorBoundary` has no access to `me`, `logger` etc. so it would require a more elaborate solution to send errors from there to `/api/log`.

So I think it's also not worth it to dig into that further - at least for now. We can use this as-is and see if that even works well. If too many people start using it, the messages being out of order may become more important so that should be fixed in that case first before anything else.

huumn:
The only real concern I have with this is storing the log messages in the application db when this may be kind of noisy/a lot to store.

It might be worth removing the db storage until we have somewhere less permanent to store them. idk I need to think about this a little more.

huumn:
Going to merge (and possibly comment out the db storage)

PR #466, opened 2023-08-31
Visual Character counter for post titles, poll options

SatsAllDay:
Live counter update to help authors know how many more chars they have to use in their post titles, and also poll options

applies across all post types (discussion, link, poll, bounty, job)

image

### DIFF
--- a/components/bounty-form.js
+++ b/components/bounty-form.js
@@ -9,7 +9,7 @@ import { bountySchema } from '../lib/validate'
 import { SubSelectInitial } from './sub-select-form'
 import CancelButton from './cancel-button'
 import { useCallback } from 'react'
-import { normalizeForwards, titleInputHint } from '../lib/form'
+import { normalizeForwards } from '../lib/form'
 import { MAX_TITLE_LENGTH } from '../lib/constants'
 
 export function BountyForm ({
@@ -105,7 +105,6 @@ export function BountyForm ({
         required
         autoFocus
         clear
-        hint={titleInputHint}
         maxLength={MAX_TITLE_LENGTH}
       />
       <Input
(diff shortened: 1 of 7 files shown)


huumn:
We really need this!

I think this is fine as is, but I suspect there might be room for improvement. Let me know if I'm off base:

1. If all of these length capped inputs use `InputInner`, could we avoid directly handling/passing the formik context around and just do it as part of `InputInner` when `maxLength` is passed (which afaik should have access to the value length already)? It'd be nice to "activate" this feature with a single prop I guess would be the difference.
2. It might be too big of a lift for an initial pass but I'd prefer more of a twitter-like UI for character counting. This might be better done in a follow-on PR though.

SatsAllDay:
1. Yes we probably can refactor it to be encapsulated and implemented by `InputInner`. That would certainly be nicer for future adoption. I can play around with that. I suppose in that case, if a `hint` was passed in as a prop, it would take precedence over this generated hint, if `maxLength` was also supplied.
2. Bold of you to assume I know what twitter's character counting UI looks like :P I'm sure I can figure that out though and see if I can incorporate that style, or something like it.

huumn:
> take precedence over this generated hint

It might be better to just show both hints. Idk probably not a lot of instances of this in the code right now so it might not even matter.

Again, you've done the overwhelming majority of the effort already. I'm happy to always take it the last mile, but it seems worthwhile to give feedback on what I'll probably change before shipping.

SatsAllDay:
I do appreciate the feedback, it is helpful! I'll update the PR to incorporate it.

SatsAllDay:
Actually, on second thought, I will leave the UX aspect of it to you to finish off - that's not really my favorite part of the work, it's pretty subjective, and you'll probably make some changes anyway (which is great!) I made the refactoring changes as you described though, and it's a much cleaner implementation!

huumn:
Yeah it looks GRRRREAT. It makes me feel at peace, you know? 

That's kind of my goal post especially for heavily trafficked parts of the code base.

SatsAllDay:
Clean implementations and abstractions give me chills

huumn:
😆 Goosebumps >> LOC

ekzyis (approved):

PR #464, opened 2023-08-31
Automatically subscribe forward users to threads

SatsAllDay:
Update create and update item sql functions to subscribe forward recipients to the post

When the post is updated, ensure that if a forward user has already unsubscribed, keep it that way. Also ensure that any forward users that are no longer forward users are removed from subscription

Closes #459

PR #467, opened 2023-09-01
Notifications for when you are forwarded sats from a post

### LINKED ISSUE #458: Notify forwarded zap recipients (with opt-out capability)
Forked from https://github.com/stackernews/stacker.news/issues/453#issuecomment-1697750926

When a post that has forward rules configured is zapped, we currently notify the OP of the post, but not the forward recipient(s). We should add notifications for this flow. It should also be opt-out-able via a boolean flag in user settings.

SatsAllDay:
Closes #458 
Support configurable notifications when a post for which you are forwarded gets zapped

This is controlled by a new boolean flag in user settings

image
image

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -815,12 +815,12 @@ export default {
 
           // send push notifications to forwarded recipients
           if (mappedForwards.length) {
-            await mappedForwards.map(forward => sendUserNotification(forward.user.id, {
+            await Promise.allSettled(mappedForwards.map(forward => sendUserNotification(forward.user.id, {
               title: `you were forwarded ${numWithUnits(msatsToSats(updatedItem.msats) * forward.pct / 100)}`,
               body: updatedItem.title ?? updatedItem.text,
               item: updatedItem,
               tag: `FORWARDEDTIP-${updatedItem.id}`
-            }))
+            })))
           }
         } catch (err) {
           console.error(err)
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -812,6 +812,16 @@ export default {
             item: updatedItem,
             tag: `TIP-${updatedItem.id}`
           })
+
+          // send push notifications to forwarded recipients
+          if (mappedForwards.length) {
+            await mappedForwards.map(forward => sendUserNotification(forward.user.id, {
+              title: `you were forwarded ${numWithUnits(msatsToSats(updatedItem.msats) * forward.pct / 100)}`,
+              body: updatedItem.title ?? updatedItem.text,
+              item: updatedItem,
+              tag: `FORWARDEDTIP-${updatedItem.id}`
+            }))
+          }
         } catch (err) {
           console.error(err)
         }
--- a/api/resolvers/notifications.js
+++ b/api/resolvers/notifications.js
@@ -176,6 +176,23 @@ export default {
         )
       }
 
+      if (meFull.noteForwardedSats) {
+        queries.push(
+          `(SELECT "Item".id::TEXT, MAX("ItemAct".created_at) AS "sortTime",
+            MAX("Item".msats / 1000 * "ItemForward".pct / 100) as "earnedSats", 'ForwardedVotification' AS type
+            FROM "Item"
+            JOIN "ItemAct" ON "ItemAct"."itemId" = "Item".id
+            JOIN "ItemForward" ON "ItemForward"."itemId" = "Item".id AND "ItemForward"."userId" = $1
+            WHERE "ItemAct"."userId" <> $1
+            AND "Item"."userId" <> $1
+            AND "ItemAct".created_at <= $2
+            AND "ItemAct".act IN ('TIP', 'FEE')
+            GROUP BY "Item".id
+            ORDER BY "sortTime" DESC
+            LIMIT ${LIMIT}+$3)`
+        )
+      }
+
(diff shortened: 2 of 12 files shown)


api/resolvers/item.js
```diff
@@ -812,6 +812,16 @@ export default {
             item: updatedItem,
             tag: `TIP-${updatedItem.id}`
           })
+
+          // send push notifications to forwarded recipients
+          if (mappedForwards.length) {
+            await mappedForwards.map(forward => sendUserNotification(forward.user.id, {
```
ekzyis, line 818: is this not missing a `Promise.allSettled`?

SatsAllDay: Eek yep I believe it is. I'll correct that tomorrow

ekzyis: > since we're not waiting on the successful notification sends for anything further in this function

to be pedantic, I believe the `catch` will only work with a proper `await` (so with `Promise.allSettled`). However, not 100% sure on that :)

SatsAllDay: I think if you use `await` on a non-promise, it just returns an immediately-resolved promise, resolved with the `await`ed value, so yea it wouldn't catch exceptions from the inner promises.

Also, `Promise.allSettled` won't reject, even if any of the contained promises reject, so `await`ing `Promise.allSettled` also won't trigger the `catch` block, which is what I originally meant. We'd have to iterate through the results and see if any of the statuses were `rejected` and log them accordingly.

Interesting discussion 🍻

ekzyis: > Also, Promise.allSettled won't reject, even if any of the contained promises reject, so awaiting Promise.allSettled also won't trigger the catch block, which is what I originally meant. We'd have to iterate through the results and see if any of the statuses were rejected and log them accordingly.

Haha, I will just pretend that this is what I meant with

> However, not 100% sure on that in this case :)

lol

> Interesting discussion 🍻

Yes :) I love nerding out, lol

ekzyis:
tACK f6f62ac

Did reset my database and created two new accounts; one on desktop, one on mobile. Enabled push notifications for both of them. Created a new anon post and forwarded 33% to both users.

Zapped that post with anon (anon is great for local development, lol) and both received push notifications.

However, I can see a `you were forwarded 0 sats` notification. (I only received 29 sats because of the 10% fee so that's fine)

2023-09-11-134500_922x235_scrot

SatsAllDay:
@ekzyis looking at that screenshot again... are those two notifications from the same post? I see the top one has 100 sats, and the bottom one has 0 sats, so I am wondering if somehow there's a duplicate post there? I haven't had a chance to try to recreate it locally yet, I'll follow up once I do.

ekzyis:
Ah, nice catch, didn't see the difference. Already deleted my database. I only created one post with that title but not sure when that notification popped up

SatsAllDay:
I just followed the above steps but was not able to recreate the issue. To double-check:

1. Create user 1, user 2, user 3, user 4
2. Post as user 1, forwarding to user 2 (33%) and user 3 (33%)
3. Zap post 100 sats as user 4
4. Check notifications for user 2, only 1 notification present (with 33 sats being forwarded, the correct amount)

image

Could it be due to `anon`? I need to get LND setup so I can do invoices (I think?) to post as anon locally, so I didn't try that yet.

I also wonder if maybe there was some weird stale notification that indicated 0 sats? Not sure, I'll keep experimenting...

SatsAllDay:
I _mostly_ got the LND setup going using the docs in the repo. I'm able to generate invoices successfully. However, how am I supposed to pay them? I setup a second LND node in Polar (aliased as bob), opened channels between it and the first LND node I created (aliased as alice) which acts as SN, but the invoice payment seems to just hang infinitely if I try to pay from bob to alice. When trying to pay from alice, it fails saying no self-payments allowed, which makes sense. I also added bob to the sn docker network, in case that is necessary. I realize this isn't directly related to this issue, but any idea what I might be missing?

ekzyis:
> but the invoice payment seems to just hang infinitely if I try to pay from bob to alice

the worker must be running to confirm the invoices :)

SatsAllDay:
Fantastic, thank you! I didn't realize the worker wasn't running... I thought it would have started up given it's included in the docker-compose file...

SatsAllDay:
Oh hey this looks familiar 
image

SatsAllDay:
It's coming from [here](https://github.com/stackernews/stacker.news/pull/467/files#diff-afd1c0bdb4a6b60f9cff2b32622acf69027d0b9ae70265cea8a0f21c6233ded9R189), the `'FEE'` option. Since it's including unzapped items, you then see the "forwarded 0 sats" note.

@huumn why did we originally include `'FEE'` for zapped notifications? Any concern with me removing `'FEE'` from this `IN` clause? Otherwise, I could add `AND "ItemAct"."userId" <> "Item"."userId"` to filter out self-actions, which should also eliminate fees, but we could just remove fees from the query altogether?

SatsAllDay:
I went ahead and removed `FEE` from the query. @ekzyis if you get a chance to re-test this, lmk if this works. I think this is the fix we need.

ekzyis:
tested 8aacd87 which includes #464 

I no longer get a `you were forwarded 0 sats`:

2023-09-12-173323_931x280_scrot

so LGTM :)

Oh, @huumn already merged this, lol

PR #469, opened 2023-09-03
toast border color based on variant

### LINKED ISSUE #468: Danger toast has success border
Title says it all :) 

image

SatsAllDay:
Closes #468 

Side note: I wanted to just do `.toast.bg-success { ... }`, `.toast.bg-danger { ... }` in the CSS module, but with how Next.JS loads CSS modules via className, that didn't seem to work how I'd like, so I went with this approach instead. I am open to different implementations if there's a cleaner option.

### DIFF
--- a/components/toast.js
+++ b/components/toast.js
@@ -56,7 +56,7 @@ export const ToastProvider = ({ children }) => {
         {toasts.map(toast => (
           <Toast
             key={toast.id} bg={toast.variant} show autohide={toast.autohide}
-            delay={toast.delay} className={styles.toast} onClose={() => removeToast(toast.id)}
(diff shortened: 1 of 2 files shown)


ekzyis (approved):
Actually not sure why we even need a border if it's the same color as the background but I don't think we'll ever touch this code again (or at least not for a long time) and it's a very small change, so I think it's fine.

Maybe there is a cleaner solution but it's probably not worth to think about it.

I almost started bikesheeding on this PR but then I caught myself, lol

SatsAllDay:
Idk how I've never heard of "bike shedding" before LOL. But yea, this is a really minor issue so it's probably not worth any more mental cycles

PR #470, opened 2023-09-04
Proposal: User Subscriptions: separate posts and comments

SatsAllDay:
A proposed solution to allows stackers to subscribe to other stackers' posts and comments separately. This is to address feedback that subscribing to some stackers causes the subscriber's notification feed to get blown up because some stackers are very active with their comments, and the subscriber is really only interested in the stacker's posts.

The backend implementation adds columns to the `UserSubscription` table to track posts and comments individually, along with `updatedAt` timestamps for each so we know which items from the subscribee should trigger notifications for the subscriber. It default posts and comments to `true` for any existing subscriptions, also migrates existing subscriptions `updatedAt` times sourced from `createdAt` to maintain the time-based filtering.

It also replaces the singular `subscribeUser` mutation with `subscribeUserPosts`, and `subscribeUserComments` for individual actions. The `meSubscription` user field is also replaced with `meSubscriptionPosts` and `meSubscriptionComments` accordingly.

As far as UX, I took a simple approach of having 2 dropdown/overflow menu items to replace the current 1:
image
image

I considered keeping the single "subscribe" action, and having it popup a modal for you to choose which items you wanted to subscribe to, but that felt too heavy for this relatively simple use-case. Perhaps we could explore that pattern if there are more configurable options for user subscriptions in the future.

As always, I am open to feedback! And no rush, I just happened to have some time to look at this this weekend.


SatsAllDay:
Note to self: need to honor posts vs comments settings in push notifications as part of this PR, too.

ekzyis:
> I considered keeping the single "subscribe" action, and having it popup a modal for you to choose which items you wanted to subscribe to, but that felt too heavy for this relatively simple use-case. Perhaps we could explore that pattern if there are more configurable options for user subscriptions in the future.

Yes, I agree, that would be too heavy. But I also don't like the current solution much but I have no better idea so far.

Maybe @huumn can chime in here

_note: I haven't looked at the code_

huumn:
My only nit with this is that the models don't need the booleans. It can be controlled with two timestamps postsSubscribedAt and commentsSubscribedAt and setting to null when there isn't one or the other. Unless I'm missing something.

I'm going to continue merging other stuff, but I can come back and follow through on this if you can't get it done today.

> Yes, I agree, that would be too heavy.

Yes, this is perfect as is. The other way *sounds* correct but it isn't worth it so it's wrong.

SatsAllDay:
> My only nit with this is that the models don't need the booleans. It can be controlled with two timestamps postsSubscribedAt and commentsSubscribedAt and setting to null when there isn't one or the other. Unless I'm missing something.
> 
> I'm going to continue merging other stuff, but I can come back and follow through on this if you can't get it done today.

Yep, this makes sense. I added the timestamps after the booleans, so I didn't think to replace the booleans with the timestamps. I think it should work fine with just the timestamps. I'll try to get to it today, will update the PR accordingly. If it's not done today, feel free to make the changes yourself.

SatsAllDay:
PR has been updated as discussed ^, lmk of any other feedback!

huumn:
Looks great! I probably won't get around to merging until this weekend and shipping until next week though. (I'm trying to block time and mostly unsuccessfully so far lol)

SatsAllDay:
Thank you! No worries, take your time. Time for me to find something else to work on... :)

huumn:
Sent for 470, 474, and 496

PR #474, opened 2023-09-06
Verified contributors

### LINKED ISSUE #471: Verified stacker.news contributors
# Verified stacker.news contributors

This feature proposal originated from a comment made on the most recent SNL episode, where one of my posts was mentioned and @huumn spoke about my recent contributions to the open source project, and Car pulled up my profile and there was discussion of having some kind of "contributor" indicator on my SN profile page.

The idea is to display a `verified stacker.news contributor` decoration on user profile, when applicable.

## How it works

* Users must link their associated GitHub account to their SN account as an auth method. This must be the GitHub account that has committed to the stacker.news repo.

* A new user setting boolean flag will be added, allowing the user to opt-in to showing this verified contributor status, if applicable. This will be configurable in the privacy section of user settings.

* A new field on user GraphQL API response payload -> `isVerifiedContributor: Boolean!`

* In the API resolver layer, we will query:
```bash
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/stackernews/stacker.news/contributors
```
[reference](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-contributors)

* This API gives a paginated list of contributors of the repo. Load 100 at a time (max), load all pages until no more, and keep a map of id:username in memory.

* Cache this total set of contributors for an hour so as to reduce network calls (GitHub API w/o access token has strict rate limitations per hour)

When a user is requested via GraphQL API, we'll resolve the `isVerifiedContributor` field by checking:

1. Does the user have this setting enabled? No -> return false

2. Yes? Check if the user has a GitHub account (in the `accounts` table) linked? No -> return false

3. Yes? Check if the linked GitHub account ID exists in the current set of cached contributors? No -> return false

4. Yes? Return true

* The in-memory cache should be updated every hour, lazily on demand. This will cause one user query to take longer than others, but only once an hour.

* Only request this field for full user view (profile page)

* Display some kind of read-only decorator/icon on user profile pages based on the API response (maybe include the GitHub username so we can link to their contributions in GitHub or something?)

I want to emphasize that this would be opt-in only, and we are not persisting any additional information, nor are we requesting additional information from GitHub on the user's behalf. We're simply connecting the dots, if the user wants us to, to show stackers who has contributed to the project. Privacy is important!

Feedback? Is this something we'd like to have? I don't have an issue scrapping the idea if we don't collectively like it.

SatsAllDay:
Closes #471 

Allow verified contributors to stacker.news to proudly display this fact on their SN profile page 🎉 

## Implementation details
* `isContributor` boolean flag added to users
* `hideIsContributor` boolean flag added to user settings, only exposed in the UI to contributors, allowing contributors to hide that they're contributors if they want
* a once-per-startup script added to the worker which sets `isContributor` to true based on config that's baked into the source code
* a simple UI decoration text on the user profile page (see screenshot) when the setting is enabled for a contributor

image

image

**Note:** I am very open to different UI for the decoration itself. I just went with something simple.


worker/index.js
```diff
@@ -63,6 +64,9 @@ async function work () {
   await boss.work('views', views(args))
   await boss.work('rankViews', rankViews(args))
 
+  // Run once
```
SatsAllDay, line 67: I added this to the worker because I didn't see a nice way to add it on start-up in the main `app` container. Also, this isn't really critical functionality, so I figured it made sense to put it in the separate worker process anyway. Open to feedback on this.

SatsAllDay:
cc @ekzyis thank you for the inspiration, I like how this turned out!

worker/contributors.js
```diff
@@ -0,0 +1,57 @@
...
+const contributorMap = {
+  616: {
+    github: 'huumn',
+    sn: 'k00b'
+  },
+  946: {
+    github: 'kerooke',
+    sn: 'kr'
+  },
+  6030: {
```
ekzyis, line 17: how do you know my user ID? have you received a database dump? lol

wanted to mention that contributors not necessarily know their user IDs on SN

worker/index.js
ekzyis: Was thinking more about adding a line to `.platform/hooks/predeploy/00_build.sh` but I guess is fine here, too

Probably also easier to do it here since you already have database access

components/user-header.js
```diff
@@ -205,6 +205,7 @@ function HeaderHeader ({ user }) {
             className='me-1'
           />{user.name}@stacker.news
         </Button>
+        {user.isContributor && <div>Verified stacker.news contributor</div>}
```
ekzyis: I am surprised this is styled like other `text-muted` and small text in the profile according to the screenshot since you only use a simple `<div>` here.

ekzyis (approved):
utACK 3eaf7c9e3d2671cffd8ae0d7b512248afc187a6a

That moment when you review a PR and think: Wait... that's it?

Looks very clean to me, nice work :)

Regarding the decoration, I think it's good but maybe it can be improved. Just not sure how exactly.

> cc @ekzyis thank you for the inspiration, I like how this turned out!

Thanks, no worries :)

worker/index.js
SatsAllDay: > Was thinking more about adding a line to .platform/hooks/predeploy/00_build.sh but I guess is fine here, too

So add a new `npm script`, or maybe just a script in the `scripts` folder, then e.g. `node scripts/seedContributors.js` in the build.sh? That would also be fine with me. I wasn't sure the best place, so I just stuck it here.

components/user-header.js
SatsAllDay: Take a look at the later commits, this simple `div` was just the placeholder. I am using `small` with `.text-muted` like the other text in that area.

worker/index.js
ekzyis: Nah, I think it's fine / better here because of no need to configure database access.

worker/contributors.js
ekzyis: I tried that too but I thought it's not possible to see the data because of SSR.

But I'll look more closely now, lol

components/user-header.js
ekzyis: Ah, that's what I get for reviewing immediately when I got the mail, lol

worker/contributors.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * Instructions: Once you make a contribution to the repo, add yourself to this map,
+ * where the key is your user id in stacker.news prod,
+ * and the value map contains your GitHub username, and your SN nym.
+ * Right now, the GitHub usernames and SN nyms are not used, but maybe we'll use them
+ * in the future for a fancier UI decoration.
+ */
+const contributorMap = {
```
ekzyis, line 8: Was thinking about pinging all persons mentioned in https://github.com/stackernews/stacker.news/graphs/contributors

but maybe it's better to just announce this on SN when it's released and then we can add them or they create a PR :thinking:

SatsAllDay: I like the idea of them opening a PR to add themselves. A little bit of proof of work to discover their user ID, plus it makes it opt-in from the get-go

huumn:
I mean if we're going to have a contributor map in the code, what good is it storing it in the database?

ekzyis:
> I mean if we're going to have a contributor map in the code, what good is it storing it in the database?

That is a very good point, lol

huumn:
I think ideally this might be implemented as a file in the repo `contributors.json`:

The people can add themselves pretty naturally.

```json
{
    k00b: {
        github: huumn
    },
    ekzyis: {
        github: ekzyis
    }, ....
}
```

We could also just do a text file of newline seperated nyms if we add the ability to add other social links to a profile (which has been a persistent feature request).

ekzyis:
> if we add the ability to add other social links to a profile (which has been a persistent feature request).

in that case, it would make sense to store it in the database since I don't think we want to tell people to create a PR to add a social link which is not Github? (Since we can't just write to the JSON during runtime else the deployed code gets into a dirty version control state which may get lost)

huumn:
I think we might be misunderstanding each other. I'm proposing that for contributors we have a flat text file `contributors.txt`:

```text
k00b
ekzyis
WeAreAllSatoshi
```

Then as a feature for every stacker, we allow them to add any number of social links to their profile so that we don't have to track the map on github.

-------------------------

#### yet-another-alternative

We don't store anything in the repo BUT we allow adding links to a profile. THEN we call github's api to see if the github usernym matches ... the only problem with this is there's not authentication.

huumn:
I think the short term mvp is probably just a file though.

SatsAllDay:
OK, so just to summarize:

1. Add a text file which contains a newline-separated list of SN nyms which represents the contributors list
2. Remove the `isContributor` field from the DB
3. Keep the `hideIsContributor` field in the DB, and keep the user setting mutation
4. Update the user resolver to consult the text file in (1) above to determine `user.isContributor`

Is this right?

I think I had `isContributor` in the DB as a relic of the original proposal which was to source it on-demand via GitHub API. I agree that we can remove it from the DB and reference the in-memory list instead, if we're going to capture contributors in the code anyway.

huumn:
> Update the user resolver to consult the text file in (1) above to determine user.isContributor

We probably shouldn't read and scan from disk every profile visit, so it should probably be loaded into memory (maybe as a `Set` on start).

Everything else makes sense imo

SatsAllDay:
Sorry yes that's what I meant but I didn't express it well.

huumn:
> Sorry yes that's what I meant but I didn't express it well.

Figured! Just wanted to reduce another round just in case.

SatsAllDay:
OK, I updated the PR to incorporate the feedback from yesterday's discussion. LMK what you think!

huumn:
Looks excellent. It brings me peace 🧘‍♂️😄

Will merge this weekend.

SatsAllDay:
Sounds good, thank you!

PR #477, opened 2023-09-06
Add scripts for automated social posting

rleed:
README.md says a lot. This basically just curls the GraphQL api and generates tweets/events from that.

There is just one minor feature missing: to exclude ads. Not sure how to create an ad yet for testing. (Hints, anyone?)

### DIFF
--- a/scripts/social/README.md
+++ b/scripts/social/README.md
@@ -34,4 +34,4 @@ For twitter, we use [twitter/twurl](https://github.com/twitter/twurl). Follow ex
 
 ## Notes and more info
 
-See issue #434.
+See [issue #434](https://github.com/stackernews/stacker.news/issues/434).
--- a/scripts/social/README.md
+++ b/scripts/social/README.md
@@ -1,0 +1,30 @@
+# Scripts for automated social posting
+
+Let it rip:
+
+`./run.sh &`
+
+## Overview
+
+`./run.sh` # This is the main script. It checks the time and decides when to do stuff.
+`./check-top-story.sh` # This is called regularly. If a new post hits the top rank, it blares it out.
+`./check-rewards.sh` # This checks for pending rewards and if greater than threshold, publicizes it.
+`./check-weekly-stories.sh` # This gathers a summary of how many stackers posted stories and posts it.
+`./check-stackers.sh` # This ranks the top 3 stackers and reports about it.
+`./send-all.sh <text>` # This broadcasts `<text>` to the social networks.
+`./send-nostr.sh <text>` # This publishes `<text>` as an event on Nostr.
+`./send-twitter.sh <text>` # This tweets out `<text>` on Twitter.
+
+You can run any of the scripts manually for testing purposes. (And they will call any of the scripts below them as needed.)
+
+## Configuration
+
+Install and configure your social network tools and accounts as follows:
+
+For nostr, we use @fiatjaf's `noscl`. Follow existing installation and setup docs.
+
+For twitter, we use `twurl`. Follow existing installation and setup docs.
+
+## Notes and more info
+
+See issue #434.
--- a/scripts/social/README.md
+++ b/scripts/social/README.md
@@ -7,23 +7,30 @@ Let it rip:
 ## Overview
 
 `./run.sh` # This is the main script. It checks the time and decides when to do stuff.
+
 `./check-top-story.sh` # This is called regularly. If a new post hits the top rank, it blares it out.
+
 `./check-rewards.sh` # This checks for pending rewards and if greater than threshold, publicizes it.
+
 `./check-weekly-stories.sh` # This gathers a summary of how many stackers posted stories and posts it.
+
 `./check-stackers.sh` # This ranks the top 3 stackers and reports about it.
+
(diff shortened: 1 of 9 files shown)


SatsAllDay:
> Not sure how to create an ad yet for testing. (Hints, anyone?)

Can you just insert items into the database manually, and associate them to the ad account?

ekzyis:
I haven't looked through all of the code but my first impression is that I don't think we want to maintain a lot of bash code.

Don't get me wrong, I like bash, but not 299 lines of it, lol.

Imo, it's a good scripting language for "quick and dirty" solutions for one-time jobs or for one-liners using a clever combination of CLI tools thanks to the UNIX philosophy (do one thing and do it well) where other scripting languages like Python or NodeJS may be more verbose. But I think it's not the right choice for this task here:

- It's hard to read: for example, `-z` stands for "check if empty" - I even had to look this up to make sure
- It's error prone: forget some `"` and you may run into a hard to find bug since it only happens if there is whitespace in a variable or even worse: arbitrary code execution
- I think not many devs have a lot of experience with bash
- probably other reasons that I don't remember right now

Also, why split these into different files? If you really want to go down this path (without guarantees that we will merge this), you should define functions in these different files which you can source in a single bash file. Then you don't need to call one script from the other script.

Also, I am not sure if we can install command line tools in the environment in which SN is deployed.

I think all of this code should be placed in the worker.

rleed:
> > Not sure how to create an ad yet for testing. (Hints, anyone?)
> 
> Can you just insert items into the database manually, and associate them to the ad account?

Yeah, I searched the code and figured out how to do it...just set the `"Item"."userId"` to `9` (the global ad user). Now the script also checks that to filter out ads. So the functionality is now complete.

rleed:
> I think all of this code should be placed in the worker.

Lol that was my original plan, but I thought this was supposed to be independent of the SN site. Perhaps being in the worker is independent enough...from the app portion at least? @huumn

rleed:
If we do this in the worker, what do you think about adding `twitter-api-v2` and `blastr` as the dependencies?

rleed:
> If we do this in the worker, what do you think about adding `twitter-api-v2` and `blastr` as the dependencies?

I guess for completeness it's worth mentioning some of the benefits of the bash solution:

- The twurl and noscl command-line tools are likely more stable as dependencies, since they are designed for bots, exactly for this type of use case, and they are end-user tools that seem to be widely used and relied upon.
- The main maintenance surface would be around the GraphQL API. (I don't know how much the fields it depends on might change in the future, but maybe you have a feel for that. I wouldn't necessarily expect those fields to be changing much.)
- The bash solution is completely stand-alone and isn't even required to run on the same server.
- Minor maintenance (such as changing the scheduling or wording of the tweets) would arguably be easier in bash.

Just FWIW....not intending to dig heels in if the sentiment is strong the other way.

huumn:
While this technically satisfies the issue, I'd strongly prefer these were written in JS like everything else. 

> If we do this in the worker, what do you think about adding twitter-api-v2 and blastr as the dependencies?

It need not be in the worker, but if it can be if that's easier. Ideally, we'd run this as its own microservice.

As for deps, we'd likely only need a twitter api one and some websocket lib that allows us to send events to wss://nostr.mutinywallet.com.

I'm going to move this to a draft pending conversion to javascript. @rleed if you're not interested in completing the conversion I can partially compensate you for the work here. You've logically modeled the solution.

rleed:
> I'm going to move this to a draft pending conversion to javascript. @rleed if you're not interested in completing the conversion I can partially compensate you for the work here. You've logically modeled the solution.

Interested!? More than happy to finish the job! (Honestly it would be nice to have partial compensation now, but I'd also be delighted to take it across the finish line too.)

Thanks for weighing in...

huumn:
Yeah sorry for the late reply. Been conferencing for the last week. 

I'm reviewing your other straightforward PRs now and I'll send along funds for them all at once.

kandycoder:
> Ideally, we'd run this as its own microservice.

I've added a container for it and ported the script logic already.

> As for deps, we'd likely only need a twitter api one and some websocket lib that allows us to send events to wss://nostr.mutinywallet.com.

Sounds good. I'm working on that now.

rleed:
> Ideally, we'd run this as its own microservice.

I've added a new container for it and ported the main logic already.

> As for deps, we'd likely only need a twitter api one and some websocket lib that allows us to send events to wss://nostr.mutinywallet.com.

Sounds good. I'm working on that now...

rleed:
This is now fully functional!

That was fun...
- I didn't notice till now that we already have a nostr dependency, which works beautifully.
- The twitter dependency I added has zero additional dependencies, and it works great too, so I think it was a good choice.
- The PgBoss scheduling simplified things a lot.
- Everything is nicer.

In the end, I'm soo happy we did this in JS. Simply nicer. @ekzyis, you were right from the beginning ;)

huumn:
Looks great at a glance! 

I'm going to deploy some other stuff, but I'll send you the partial sum as discussed along with the other stuff once I've deployed the stuff I've already reviewed.

PR #481, opened 2023-09-08
Hide wallet balance

### LINKED ISSUE #480: Hide wallet balance with long press

When stackers share screenshots, they expose their balance in the top right corner.

We could add a carousel for the balance like for the price. So if you click it, it gets replaced with ****

No alternatives

**Additional context**

https://stacker.news/items/247853

ekzyis:
Closes #480 

I tried to keep `<Link>` by using `event.preventDefault()` in `onLongPress` (and additionally in `onShortPress`) but it kept navigating.

So I used `router.push('/wallet')` in `onShortPress` instead.

Does this introduce undesired behavior?

Also wondering if we should mention this feature in the FAQ somewhere. But afaict, there is no "privacy" section yet.

Maybe just show a help icon next to the privacy section header and mention this?

### DIFF
--- a/api/typeDefs/user.js
+++ b/api/typeDefs/user.js
@@ -25,7 +25,7 @@ export default gql`
       noteInvites: Boolean!, noteJobIndicator: Boolean!, noteCowboyHat: Boolean!, hideInvoiceDesc: Boolean!,
       hideFromTopUsers: Boolean!, hideCowboyHat: Boolean!, clickToLoadImg: Boolean!,
       wildWestMode: Boolean!, greeterMode: Boolean!, nostrPubkey: String, nostrRelays: [String!], hideBookmarks: Boolean!,
-      noteForwardedSats: Boolean!): User
+      noteForwardedSats: Boolean!, hideWalletBalance: Boolean!): User
     setPhoto(photoId: ID!): Int!
     upsertBio(bio: String!): User!
     setWalkthrough(tipPopover: Boolean, upvotePopover: Boolean): Boolean
@@ -85,6 +85,7 @@ export default gql`
     hideCowboyHat: Boolean!
     hideBookmarks: Boolean!
     hideWelcomeBanner: Boolean!
+    hideWalletBalance: Boolean!
     clickToLoadImg: Boolean!
     wildWestMode: Boolean!
     greeterMode: Boolean!
--- a/components/header.js
+++ b/components/header.js
@@ -10,7 +10,7 @@ import Price from './price'
 import { useMe } from './me'
 import Head from 'next/head'
 import { signOut } from 'next-auth/react'
-import { useCallback, useEffect, useState } from 'react'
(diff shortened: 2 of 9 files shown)


huumn:
I'm going to move this to draft so it can be updated to reflect our discussion about the visibility of this feature: ideally, this is a stacker setting and the balance should be revealed when cursed over.

ekzyis:
Added the setting now. Also made sure that there are no layout shifts using `useLayoutEffect`.

I used 5x `*` since that should have a larger width than any rendered balance width.
For example, if balance is 1e8, it will render `100m`. If balance is 9e5, it will render `900k`. Both fit inside `*****`.

https://github.com/stackernews/stacker.news/assets/27162016/6618ad33-1530-4c02-a127-931ff1c63508

huumn:
Looks great! I think the widest it ever is is 4 characters.

ekzyis:
Wondering if we should allow using long press to hide the balance for better accessibility on mobile? As it is now, pressing on it counts as "hover" afaict. But this redirects you to `/wallet`.

Also, not sure if a mutation should be triggered in that case. So if we use no mutation, it's messy because of conflicts with the setting. And if we use a mutation, it's also messy since currently, we only change settings at `/settings`. Also, we need to send all settings for every `setSettings` call (but this could easily be changed, I think).

I guess the real question is just if stackers with this setting enabled would like to easily see their balance on mobile without having to update their setting.

I think we can ship it as-is but just wanted to raise awareness about this :)

huumn:
I think there are other events we can listen for that are mobile specific that aren't detect as a click.

Does this also hide the balance in the wallet "you have ***** sats"?

ekzyis:
> I think there are other events we can listen for that are mobile specific that aren't detected as a click.

Oh, interesting!

> Does this also hide the balance in the wallet "you have ***** sats"?

No, is that important?

> I was wrong. It's 5 (excluding an optional period).

Ah, hehe. I think it also depends on the individual character width since it's not using a monospaced font (?)

huumn:
> No, is that important?

I mean importance is subjective, but it'd be nice I think.

> Ah, hehe. I think it also depends on the individual character width since it's not using a monospaced font (?)

Yep. 5 is the right choice I think.

huumn:
Is this still ready for review?

ekzyis:
> I mean importance is subjective, but it'd be nice I think.

I'll add it :) Not worth it to think if it's worth it, lol. Just Do It™

ekzyis:
> Is this still ready for review?

Ha, we posted at the same time, I guess.

I'll rebase on master because of the conflicts and also hide the balance in `/wallet`

ekzyis:
Done:

https://github.com/stackernews/stacker.news/assets/27162016/944e9742-3695-40ea-a1a8-6fa309017ddf

I think layout shifts here are fine else I would have to use a ridiculous amount of `*`, lol

PR #484, opened 2023-09-09
Add block height to price carousel

### LINKED ISSUE #435: Block height clock
It's been a year since another user recommended this , so i hope this is not spam.
https://stacker.news/items/22352/r/llamabyte
In this bear market it would be nice have the option to open SN and see the blockheight instead

SatsAllDay:
Source the current block height from [mempool.space API](https://mempool.space/docs/api/rest#get-block-tip-height)

The price carousel now shows (in order):
1. price per BTC in fiat
2. sats per fiat unit
3. 1sat=1sat
4. current block height (I am open to feedback on how to present this number in the carousel)

image

Closes #435

### DIFF
--- a/api/resolvers/blockHeight.js
+++ b/api/resolvers/blockHeight.js
@@ -7,7 +7,7 @@ async function fetchBlockHeight () {
     .then(res => res.text())
     .then((body) => Number(body))
     .catch((err) => {
-      console.error(err)
+      console.error('blockHeight', err)
       return 0
     })
   cache.set('block', { height: blockHeight, createdAt: Date.now() })
@@ -23,7 +23,7 @@ async function getBlockHeight () {
   } else {
     fetchBlockHeight().catch(console.error)
   }
-  return null
+  return 0
 }
 
 export default {
--- a/api/resolvers/blockHeight.js
+++ b/api/resolvers/blockHeight.js
@@ -1,0 +1,35 @@
+const cache = new Map()
+const expiresIn = 30000 // in milliseconds
+
+async function fetchBlockHeight () {
+  const url = 'https://mempool.space/api/blocks/tip/height'
+  const blockHeight = await fetch(url)
+    .then(res => res.text())
+    .then((body) => Number(body))
+    .catch((err) => {
+      console.error(err)
+      return 0
+    })
+  cache.set('block', { height: blockHeight, createdAt: Date.now() })
+  return blockHeight
+}
+
+async function getBlockHeight () {
+  if (cache.has('block')) {
+    const { height, createdAt } = cache.get('block')
+    const expired = createdAt + expiresIn < Date.now()
+    if (expired) fetchBlockHeight().catch(console.error) // update cache
+    return height // serve stale block height (this on the SSR critical path)
+  } else {
+    fetchBlockHeight().catch(console.error)
+  }
+  return null
+}
+
+export default {
+  Query: {
+    blockHeight: async (parent, opts, ctx) => {
+      return await getBlockHeight()
+    }
+  }
+}
--- a/api/resolvers/blockHeight.js
+++ b/api/resolvers/blockHeight.js
@@ -1,5 +1,5 @@
 const cache = new Map()
-const expiresIn = 30000 // in milliseconds
+const expiresIn = 1000 * 60 // in milliseconds
 
 async function fetchBlockHeight () {
   const url = 'https://mempool.space/api/blocks/tip/height'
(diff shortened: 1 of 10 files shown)


SatsAllDay:
It seems like we can probably also use
```js
import lndService from 'ln-service'
import lnd from '../lnd'
// ...
lndService.getHeight({ lnd }).then(res => res.current_block_height)
```

However, I don't have LND setup in my dev env, and I found the mempool API first :) If we'd rather me use the above approach with LND, I can try to set it up to test it out locally, just lmk.

api/resolvers/blockHeight.js
```diff
@@ -0,0 +1,35 @@
...
+  cache.set('block', { height: blockHeight, createdAt: Date.now() })
+  return blockHeight
+}
+
+async function getBlockHeight () {
+  if (cache.has('block')) {
+    const { height, createdAt } = cache.get('block')
+    const expired = createdAt + expiresIn < Date.now()
+    if (expired) fetchBlockHeight().catch(console.error) // update cache
+    return height // serve stale block height (this on the SSR critical path)
```
ekzyis, line 22: This is basically the [`stale-while-revalidate` caching strategy](https://developer.chrome.com/docs/workbox/caching-strategies-overview/#stale-while-revalidate), right?

api/resolvers/blockHeight.js
```diff
@@ -0,0 +1,35 @@
+const cache = new Map()
+const expiresIn = 1000 * 60 // in milliseconds
```
ekzyis, line 2: lol, I tried to find out the rate limits for the mempool API but only found this

> Our API limits aren't disclosed, basically because if you have to ask then you will hit them. For this type of application it's best if you run your own node and query it locally.

-- https://github.com/mempool/mempool/discussions/752#discussioncomment-1248907

Now I am confused. Is 1 minute too often? lol

components/price.js
```diff
@@ -80,6 +87,14 @@ export default function Price ({ className }) {
     )
   }
 
+  if (asSats === 'blockHeight') {
+    return (
+      <div className={compClassName} onClick={handleClick} variant='link'>
+        {numWithUnits(blockHeight, { abbreviate: false, unitSingular: 'block', unitPlural: 'blocks' })}
```
ekzyis, line 93: I don't think you need `numWithUnits`. We will never reach 1 block again :)

I would have used `block 807175` [like in the block explorer](https://mempool.space/block/000000000000000000026b7f47d18638c29e9bbdb84fa4652f4332bead8c5244) but no strong opinions on that or using `numWithUnits`.

ekzyis (commented):
Looks good except that the rate limits of the mempool API are unclear.

At first, I would have said it's fine to use the API instead of the local LND node since it's not part of the code is not _that_ important and the worst that can happen is a stale cache afaict.

But this may have changed my mind.

> However, I don't have LND setup in my dev env, and I found the mempool API first :) If we'd rather me use the above approach with LND, I can try to set it up to test it out locally, just lmk.

So if you want to go down the LND path, you can check out [docs/local-lnd.md](https://github.com/stackernews/stacker.news/blob/master/docs/local-lnd.md). This should explain everything to work with a local LND node :) If not, let us know!

Having a local LND node setup would also be useful for other features in the future.

SatsAllDay:
Now that I have LND setup locally, I'll refactor this to use it instead of the mempool.space API. It should be more performant, anyway.

components/price.js
SatsAllDay: > I don't think you need `numWithUnits`. We will never reach block 1 again :)

lol yea, I was just doing it for completeness, but you're right, it's not necessary.

> I would have used `block 807175` [like in the block explorer](https://mempool.space/block/000000000000000000026b7f47d18638c29e9bbdb84fa4652f4332bead8c5244) but no strong opinions on that or using `numWithUnits`.

I like that, too. I'll refactor to `block n`

api/resolvers/blockHeight.js
SatsAllDay: That's... an interesting choice.

Regardless, I'll refactor to use LND instead.

api/resolvers/blockHeight.js
SatsAllDay: Yea, pretty much. I just copied it from `api/resolvers/price.js`, so I can't take credit :)

SatsAllDay:
Updated to use LND instead of mempool.space.

Also updated the UI per feedback.

image

api/resolvers/blockHeight.js
huumn: That was one of ekzyis' first contributions :)

ekzyis: That's right but I looked at `git blame` and you updated the code to use that better caching strategy :)

See 3c8ea0db

PR #485, opened 2023-09-10
Disallow forwarding sats to yourself

SatsAllDay:
Implemented via dependency injecting `me` into post validation schemas (link, discussion, bounty, poll), then comparing `me?.name` to the specified name for the forward entry.

I validated this on client-side, and also server-side by hitting the API directly, since you can't send to the API through the UI with this invalid state. Same error came back in both cases.

image

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -58,7 +58,7 @@ async function usernameExists (client, name) {
   return !!user
 }
 
-export function advPostSchemaMembers (client) {
+export function advPostSchemaMembers (client, me) {
   return {
     boost: intValidator
       .min(BOOST_MIN, `must be blank or at least ${BOOST_MIN}`).test({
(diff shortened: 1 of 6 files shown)


lib/validate.js
```diff
@@ -58,26 +58,33 @@ async function usernameExists (client, name) {
...
-            if (!name || !name.length) return true
-            return await usernameExists(client, name)
-          },
-          message: 'stacker does not exist'
-        }),
+        nym: string().required('must specify a stacker')
+          .test({
+            name: 'nym',
+            test: async name => {
+              if (!name || !name.length) return false
```
SatsAllDay, line 76: This line also changed, I think this was a bug from before (changed return from `true` to `false`)

huumn: I think this was there because of the compact call or something ... I remember posts being invalid without forwards or something. I'll double check before merging.

PR #486, opened 2023-09-11
Fix processing images to potatos

### LINKED ISSUE #451: "Potato" image quality via proxy
Context: https://stacker.news/items/237871

Images of text generally get messed up. We probably need to tune the quality params of the proxy.

ekzyis:
Fix #451 

Not sure why I used these processing options. I think I found them in a tutorial and just copied them without updating them later. The default processing options are to fit images (no cropping):

- Original: https://files.ekzyis.com/public/sn/imgproxy_test/imgproxy_test.png
```
$ identify imgproxy_test.png
imgproxy_test.png PNG 1920x1080 1920x1080+0+0 8-bit sRGB 241507B 0.000u 0:00.000
```
- Default processing options: https://files.ekzyis.com/public/sn/imgproxy_test/imgproxy_test_default.png
(basically no difference to original image)
```
$ identify imgproxy_test_default.png
imgproxy_test_default.png PNG 1920x1080 1920x1080+0+0 8-bit sRGB 240165B 0.000u 0:00.000
```
- Previous processing options: https://files.ekzyis.com/public/sn/imgproxy_test/imgproxy_test_rs_600_500_0_g_no.png
(resized to fit 600x500)
```
$ identify imgproxy_test_rs_600_500_0_g_no.png
imgproxy_test_rs_600_500_0_g_no.png PNG 600x338 600x338+0+0 8-bit sRGB 88785B 0.000u 0:00.000
```

Will test with more images like [this one](https://i.imgur.com/qjZvB3t.png). ~~Maybe we should limit to 1920x1080.~~ It's more about image size than image resolution imo

---

UPDATE:

- Original: https://files.ekzyis.com/public/sn/imgproxy_test/imgproxy_test_2.png
```
$ identify imgproxy_test_2.png
imgproxy_test_2.png PNG 1419x1583 1419x1583+0+0 8-bit sRGB 370184B 0.000u 0:00.000
```
- Default processing options: https://files.ekzyis.com/public/sn/imgproxy_test/imgproxy_test_2_default.png
```
$ identify imgproxy_test_2_default.png
imgproxy_test_2_default.png PNG 1419x1583 1419x1583+0+0 8-bit sRGB 367236B 0.000u 0:00.000
```
- Previous processing options: https://files.ekzyis.com/public/sn/imgproxy_test/imgproxy_test_2_rs_600_500_0_g_no.png
```
$ identify imgproxy_test_2_rs_600_500_0_g_no.png
imgproxy_test_2_rs_600_500_0_g_no.png PNG 448x500 448x500+0+0 8-bit sRGB 104215B 0.000u 0:00.000
```

Need a bigger image. I think if an image is 1MB+, we shouldn't serve this image as-is.


ekzyis:
> t's more about image size than image resolution imo

> Need a bigger image. I think if an image is 1MB+, we shouldn't serve this image as-is.

Actually, is about desktop (higher size for higher resolution is ok) vs mobile (lower resolution so lower size).

So I'll look into [`srcset`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) for this PR already

ekzyis:
Closing this since I will fix several things with our current image proxy in one big PR since they are all related.

For example, to use [`srcset`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images), it would make even more sense to not replace image URLs with URLs from our image proxy since we need to sign one URL per resolution of an image.

So I'll also do that.

Issue #492, opened 2023-09-14
non-nullable field Item.depth is null


Leon-Sander:
**Description**
Hello, I am working on a python library for stacker.news.
When executing the get items query, the Error "Cannot return null for non-nullable field Item.depth." occurs.:

```python
from main_client import StackerNewsGraphQL
client = StackerNewsGraphQL(endpoint="https://stacker.news/api/graphql")
output = client.get_items(limit=2, sort="TOP")
print(output)
```

```
def get_items(self, limit=10, cursor=None, sort="NEW", type=None, sub=None, name=None, when=None, by=None):
    variables = {
        "limit": limit,
        "cursor": cursor,
        "sort": sort,
        "type": type,
        "sub": sub,
        "name": name,
        "when": when,
        "by": by
    }
    response = self.execute(get_items_query, variables)
    return {
        "items": response["items"],
        "cursor": cursor,
        "limit": limit
    }

get_items_query = """
    query ($limit: Int, $cursor: String, $sort: String, $type: String, $sub: String, $name: String, $when: String, $by: String) {
        items(limit: $limit, cursor: $cursor, sort: $sort, type: $type, sub: $sub, name: $name, when: $when, by: $by) {
            items {
                id
                createdAt
                updatedAt
                title
                url
                text
                user {
                    id
                }
                depth
                boost
                upvotes
                wvotes
                ncomments
            }
        }
    }
"""
```

**Steps to Reproduce**
1. You can clone my repo: https://github.com/Leon-Sander/stacker.news_python_library
2. In the queries.py file under get_items_query add depth.
3. run test.py

**Expected behavior**
There should be someting in the field.

**Logs**
ERROR:logger:Error executing query: {'message': 'Cannot return null for non-nullable field Item.depth.', 'locations': [{'line': 22, 'column': 7}], 'path': ['items', 'items', 0, 'depth'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR'}}

huumn:
Nice catch. That's a vestigal part of the API that we don't use but it should be made to work

PR #493, opened 2023-09-14
made Item.depth nullable

Leon-Sander:
Quickfix for the following issue by changing the depth field to Int from Int! in the item.js typeDef
https://github.com/stackernews/stacker.news/issues/492

### DIFF
--- a/api/typeDefs/item.js
+++ b/api/typeDefs/item.js
@@ -77,7 +77,7 @@ export default gql`
     root: Item
     user: User!
     userId: Int!
-    depth: Int!
+    depth: Int
     mine: Boolean!
     boost: Int!
     bounty: Int


ekzyis:
To find the history for related code, I ran the following command:

```
$ git log -S "depth: async" --stat
commit 95d6d789fafdaf332bd629f5c6570a8aba62b740
Author: keyan <keyan.kousha+huumn@gmail.com>
Date:   Thu Apr 15 14:41:02 2021 -0500

    pre recursive comments

 api/resolvers/item.js | 40 ++++++++++++++++++----------------------
 1 file changed, 18 insertions(+), 22 deletions(-)

commit 9acde2df1cba14ba0c1802f58fa07a92d1a9c017
Author: keyan <keyan.kousha+huumn@gmail.com>
Date:   Mon Apr 12 13:05:09 2021 -0500

    a bunch of increments

 api/resolvers/item.js | 61 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 61 insertions(+)
```

The resolver for this field was added in 9acde2df1cba14ba0c1802f58fa07a92d1a9c017 and used to be this:

```
Item: {
  user: async (item, args, { models }) =>
    await models.user.findUnique({ where: { id: item.userId } }),
  depth: async (item, args, { models }) => {
    if (item.path) {
      return item.path.split('.').length - 1
    }

    // as the result of a mutation, path is not populated
    const [{ path }] = await models.$queryRaw`
      SELECT ltree2text("path") AS "path"
      FROM "Item"
      WHERE id = ${item.id}`

    return path.split('.').length - 1
  }
}
```

but was then removed in 95d6d789fafdaf332bd629f5c6570a8aba62b740.

So I think we can just add this resolver again to still support it per @huumn's opinion:

> Nice catch. That's a vestigal part of the API that we don't use but it should be made to work

-- https://github.com/stackernews/stacker.news/issues/492#issuecomment-1719518099

I also think it's useful to have this field resolved.

Issue #410, opened 2023-08-15
Custom Date Selector (250,000 Sat Bounty)


kerooke:
**Is your feature request related to a problem? Please describe.**
The tools for filtering content by date on Stacker News are limited, and do not allow people to fully customize their date parameters for top stackers, posts, on user profiles, or on the search page.

**Describe the solution you'd like**
In addition to the existing date filters that allow one to filter by day, week, month, year, or forever, there should be tools that allows anyone to insert custom date parameters to find content on any given date/dates.

People should be able to search for content on a given day, week, month, or year with the ability to scroll between previous periods at the click of a button. This is similar to how [Plausible](https://plausible.io/stacker.news?period=month&date=2023-07-15) allows anyone to segment their site traffic by various custom date options.

This would enable more robust leaderboards and improve access to historical content on SN. 

**Additional context**
For the completion of this feature, Stacker News is offering a 250,000 sat bounty.

huumn:
Hints: 
- backend queries will need to change (hard part):
    -  `/api/resolvers/search.js` (in `search` resolver) ... this is an OpenSearch query 
    -  `/api/resolvers/item.js` (`items` resolver) ... this is a sql query
- this'll require a date range picker on the frontend (tedious part)
- graphql queries will need to be modified to take date ranges (relaxing part)

SuperAtic:
Have been playing around a bit, but I'll be definitely not able to build/code the functionality. Not sure I understand it properly... @kerooke you'd need something like this?
👇

`1.` Select Custom
![Select Custom](https://snipboard.io/SBHch1.jpg)

`2.` Select Custom Date-Range
![Select Custom Date-Range](https://snipboard.io/0zwM9s.jpg)

`3.` Visualize** selected Date-Range as current option on the `select` input
![Visualize selected Date-Range as current option on the `select` input](https://snipboard.io/h9uiH1.jpg)

Hope it helps, and happy to make some changes to the mockup if needed

brihuang99:
> Have been playing around a bit, but I'll be definitely not able to build/code the functionality. Not sure I understand it properly...  @kerooke you'd need something like this?
> 👇
> 
> `1.` Select Custom
> ![Select Custom](https://snipboard.io/SBHch1.jpg)
> 
> `2.` Select Custom Date-Range
> ![Select Custom Date-Range](https://snipboard.io/0zwM9s.jpg)
> 
> `3.` Visualize selected Date-Range as current option on the `select` input
[...]

PR #494, opened 2023-09-14
Add a date selector to the search function

rleed:
I'm opening this in draft state (pertaining to issue #410) because I've gotten stuck and I want to ask if someone has insight on what I'm missing.

I've added the date picker to the UI in this PR, and it works fine up to and including updating the `values` object and the query parameters in the location bar. However, when clicking the search button to actually perform the search, the `values` revert back to what seem to be the initial values, and I do not understand why or where those are coming from.

Does anyone have insight on why that is happening?


huumn:
It's because the search button trigger `onSubmit` which calls `search` with the `Form` values. You are storing your date range in the search component's state and the values sent for `onSubmit` come from the `Form`'s state. 

Make sense?

rleed:
Yeah, that makes sense. I was thinking `formik.values` took care of it, but now I see I need to use `helpers.setValue` as well.

huumn:
The code for this looks simple and good but the UI is too bad on mobile to ship. 

Going to drop in draft until you or I can fix that.

rleed:
Updated to wrap the date selector on mobile.

image

PR #496, opened 2023-09-15
Format bounty sats with separators

SatsAllDay:
Introduce `format` option on `numWithUnits` fn, consumed by bounty listing. Format option is disabled by default, and only applicable when `abbreviate` is `false`.

Uses [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat), so it should honor each user's browser locale settings automatically.

Inspired by [this](https://stacker.news/items/255020/r/WeAreAllSatoshi) thread.

Can be used elsewhere in the codebase, but this is the place of inspiration, so I only adopted it here.

image

### DIFF
--- a/components/item-full.js
+++ b/components/item-full.js
@@ -150,7 +150,7 @@ function TopLevelItem ({ item, noReply, ...props }) {
                 </div>)
               : (
                 <div className='px-3 py-1 d-inline-block bg-grey-darkmode rounded text-light'>
-                  {numWithUnits(item.bounty, { abbreviate: false })} bounty
+                  {numWithUnits(item.bounty, { abbreviate: false, format: true })} bounty
                 </div>)}
           </div>}
       </div>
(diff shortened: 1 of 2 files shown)


ekzyis:
> Can be used elsewhere in the codebase, but this is the place of inspiration, so I only adopted it here.

I think it makes sense to also use this for the `stacked` amount in user profiles and in `/wallet`.

SatsAllDay:
> > Can be used elsewhere in the codebase, but this is the place of inspiration, so I only adopted it here.
> 
> I think it makes sense to also use this for the `stacked` amount in user profiles and in `/wallet`.

Makes sense. I can do that as part of this PR.

huumn:
Going to merge. I can make the changes to the other parts

SatsAllDay:
Ok sounds good, sorry I have been busy today and didn't get to it.

huumn:
Dood no worries!

ekzyis:
Maybe we should also have done this for invoices?

2023-09-20-173055_1920x1080_scrot

huumn:
Probably

Issue #274, opened 2023-05-15
Enhancement request: Add to Search a link to a guide


cointastical:
[Search filters were added in October 2022](https://stacker.news/items/83923), and I still don't really know how to use them. 

For instance, today I want to find a comment that I had made by searching for some set of words, and this comment would have been before Jan 2023. 

Another [post with "refinements"](https://stacker.news/items/85615) gives more info, in that i would add nym:cointastical to see my own comments.  But this is not intuitive.

A comprehensive guide for search and search filtering would be useful   (just as how for composing comments replies, a link to a guide for markdown is just a click away).

ekzyis:
Multiple users (me included) noticed some lacking of UX with the search.

Thanks for the issue, the search definitely needs a rework

huumn:
Current hidden filters that definitely need to be documented are: 

- `~subname`, search within a specific sub
- `nym:ausersnym`, search for items from a certain user
- `url:aurl`, search for certain urls ... currently this is only domain names iirc

These should probably be added to the FAQ @kerooke and we can link to it from the empty search page.

PR #497, opened 2023-09-16
Add search hints to initial search page

rleed:
This is to address issue #274.

Untitled

(The ~sub syntax didn't seem to work for me...note to self to check into it.)

### DIFF
--- a/pages/~/search.js
+++ b/pages/~/search.js
@@ -3,6 +3,7 @@ import { getGetServerSideProps } from '../../api/ssrApollo'
 import { useRouter } from 'next/router'
 import { SUB_SEARCH } from '../../fragments/subs'
 import Items from '../../components/items'
+import styles from './search.module.css'
 
 export const getServerSideProps = getGetServerSideProps({
   query: SUB_SEARCH,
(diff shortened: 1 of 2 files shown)


rleed:
> (The ~sub syntax didn't seem to work for me...note to self to check into it.)

Following up on this, there seems to be a disconnect in `resolvers/search.js`. I added the changes to this PR that I think are needed, but I could not find a way to test the fix. Is there a way to test the search function (OpenSearch) on a local dev setup? (...I need to be able to test it for another PR too.)

huumn:
> (The ~sub syntax didn't seem to work for me...note to self to check into it.)

I think I broke support for this at some point and you search within a sub by going to `/~sub/search`

>  Is there a way to test the search function (OpenSearch) on a local dev setup?

If you're using docker we'd have to add another container to the cluster. This is something I'd like to add so happy to sponsor you doing that if you want.

rleed:
> If you're using docker we'd have to add another container to the cluster. This is something I'd like to add so happy to sponsor you doing that if you want.

Great, yeah, I'm using docker (on Windoze). I'm happy to take that on!

rleed:
Revision to the hint:

Untitled

huumn:
Looks great! Nice work!

Issue #230, opened 2023-01-21
Support lightning address/lnurl-pay "descriptions"


huumn:

cointastical:
This was again requested on an SN post:

*What about sending sats directly to a user in SN ?*
https://stacker.news/items/219912

huumn:
> Or is this to surface descriptions when your SN wallet is funded, or both?

I think this is the one people ask for most. We have a minimal lightning address implementation and there's a lot more they can do. This is kind of the next thing we should add, but it be cool to fully feature the spec.

SatsAllDay:
> > Or is this to surface descriptions when your SN wallet is funded, or both?
> 
> I think this is the one people ask for most. We have a minimal lightning address implementation and there's a lot more they can do. This is kind of the next thing we should add, but it be cool to fully feature the spec.

Cool. I will take a look at this issue, I'm interested in learning more about the lightning address spec and SN's implementation.

SatsAllDay:
Doing some research...

[LUD-6](https://github.com/lnurl/luds/blob/luds/06.md) - where `payRequest` is specified, does not define a way for the payer to specify the description for the subsequently-generated invoice, so I think that would be left up to SN in our situation, which is what we're doing now with the hardcoded `Funding @ user on stacker.news` description.

[LUD-12](https://github.com/lnurl/luds/blob/luds/12.md) - where `comments` is specified, adds a way for the payee to allow comments on a `payRequest`. This is something that SN does not currently support, since we don't specify `commentsAllowed` in our response to `.well-known/lnurlp/[username]`. So, this is one feature we could implement.

However, the spec is not clear about _what to do_ with the comment. I think it's open-ended on purpose.

I think technically we could use the comment, if provided, as the description of the invoice that we generate in response to `GET /.well-known/lnurlp/[username]/pay?amount=<>&comment=<>`. I did find some [discussion](https://github.com/Dolu89/ligess/issues/10) that suggests this is a bad idea, so one alternative could be that we just alter our invoice table to store `comment` as a new field, and display it in the UI, but don't use it as the description of the invoice itself.

Thoughts?

SatsAllDay:
I created #498 just to give us something to reference in further discussions. We don't have to keep any of it, but I thought it would be helpful to see what this might look like.

huumn:
GIT MERGITH

PR #498, opened 2023-09-16
Implementing LUD-12 comments on payRequest in LNURLP Lightning Address flow (sending and receiving)

SatsAllDay:
Related to #230 

This PR enhances SN on the receiving end to support comments when someone sends to a user@stacker.news lightning address.

It also enhances the withdraw to lightning address flow _from_ SN to allow comments to be included, if the lightning address provider supports them.

Sample wallet history UI:
image

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -486,9 +486,11 @@ export function Form ({
   initial, schema, onSubmit, children, initialError, validateImmediately, storageKeyPrefix, validateOnChange = true, invoiceable, ...props
 }) {
   const toaster = useToast()
+  const initialErrorToasted = useRef(false)
   useEffect(() => {
-    if (initialError) {
+    if (initialError && !initialErrorToasted) {
       toaster.danger(initialError.message || initialError.toString?.())
+      initialErrorToasted.current = true
     }
   }, [])
 
--- a/pages/api/lnurlp/[username]/index.js
+++ b/pages/api/lnurlp/[username]/index.js
@@ -1,6 +1,7 @@
 import { getPublicKey } from 'nostr'
 import models from '../../../../api/models'
 import { lnurlPayMetadataString } from '../../../../lib/lnurl'
+import { LNURLP_COMMENT_MAX_LENGTH } from '../../../../lib/constants'
 
 export default async ({ query: { username } }, res) => {
   const user = await models.user.findUnique({ where: { name: username } })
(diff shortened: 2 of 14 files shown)


huumn:
Looks like the right approach to me. Nice work!

SatsAllDay:
Cool, thanks for taking a look! I'd like to implement the "wallet" side of LUD-12 on SN, too, so withdrawing to a lightning address allows you to specify a comment, if the recipient supports it. Full circle! :)

components/form.js
```diff
@@ -486,9 +486,11 @@ export function Form ({
   initial, schema, onSubmit, children, initialError, validateImmediately, storageKeyPrefix, validateOnChange = true, invoiceable, ...props
 }) {
   const toaster = useToast()
+  const initialErrorToasted = useRef(false)
   useEffect(() => {
-    if (initialError) {
+    if (initialError && !initialErrorToasted) {
       toaster.danger(initialError.message || initialError.toString?.())
+      initialErrorToasted.current = true
     }
```
SatsAllDay, line 494: These changes shouldn't have been necessary, but I think something about how the wallet withdraw component is implemented is causing the form to be re-rendered several times upon error, which was causing error toasts to show up several times.

It might be because we switch from `Form` to `WithdrawSkeleton` components while the request is processing.

Similarly, when an error occurs on withdrawal, the form inputs are re-initialized, so I suspect we could remove _this_ code if we fix that re-render issue.

SatsAllDay, line 494: These changes shouldn't have been necessary, but I think something about how the wallet withdraw component is implemented is causing the form to be re-rendered several times upon error, which was causing error toasts to show up several times.

It might be because we switch from `Form` to `WithdrawSkeleton` components while the request is processing.

Similarly, when an error occurs on withdrawal, the form inputs are re-initialized, so I suspect we could remove _this_ code if we fix that re-render issue.

pages/api/lnurlp/[username]/index.js
```diff
@@ -13,8 +14,9 @@ export default async ({ query: { username } }, res) => {
     minSendable: 1000, // Min amount LN SERVICE is willing to receive, can not be less than 1 or more than `maxSendable`
     maxSendable: 1000000000,
     metadata: lnurlPayMetadataString(username), // Metadata json which must be presented as raw string here, this is required to pass signature verification at a later step
+    commentAllowed: LNURLP_COMMENT_MAX_LENGTH, // LUD-12 Comments for payRequests https://github.com/lnurl/luds/blob/luds/12.md
     tag: 'payRequest', // Type of LNURL
-    nostrPubkey: getPublicKey(process.env.NOSTR_PRIVATE_KEY),
-    allowsNostr: true
+    nostrPubkey: process.env.NOSTR_PRIVATE_KEY ? getPublicKey(process.env.NOSTR_PRIVATE_KEY) : undefined,
+    allowsNostr: !!process.env.NOSTR_PRIVATE_KEY
```
Comment (SatsAllDay) [inline pages/api/lnurlp/[username]/index.js:20]: I made these changes only because I didn't have `NOSTR_PRIVATE_KEY` setup locally. Shouldn't cause an issue in prod, but without these changes, I couldn't do this work. Up to you if you want to keep them, or force devs to configure a valid nostr private key.

huumn:
Great work! I'll probably deploy this weekend's merges monday

huumn:
All good. It took me awhile to get in the habit of deleting the functions when signatures change. It was only after subtle bugs started appearing where I was using old versions of functions unknowingly

PR #508, opened 2023-09-21
FIX #499 - nostr: URIs doesn't work in bio

### LINKED ISSUE #499: nostr: URIs doesn't work in bio
Was editing my bio - https://stacker.news/kristapsk.

If I change my Nostr description to "nostr:npub1vxpwwydpmmc4usmxf2zkr8shdjq888g532ydyls3nadxxd9wyx5ssjq5nd" (like a proper nostr URI), it becomes hidden and not visible / clickable.

image

image

tom335:
This PR removes `remarkDirective` plugin from `text.js` component.

According to issue #499 on Github, text patterns starting with a colon `:` were removed from rendered Markdown.

This is partly expected, since the plugin works for expressions conforming to [generic directives proposal](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444), for example `:cite[smith04]`. However, if the pattern is incomplete, it fails and text doesn't appear.

If this plugin cannot be removed temporarily, its behaviour could be fixed on a plugin fork, as suggestion.

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -5,7 +5,7 @@ import { LightAsync as SyntaxHighlighter } from 'react-syntax-highlighter'
 import atomDark from 'react-syntax-highlighter/dist/cjs/styles/prism/atom-dark'
 import mention from '../lib/remark-mention'
 import sub from '../lib/remark-sub'
-import remarkDirective from 'remark-directive'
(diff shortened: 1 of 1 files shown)


huumn:
I've removed this in the latest commit. Should go live soon.

Issue #505, opened 2023-09-20
Add OpenSearch to docker setup for development


rleed:
Searching SN posts in the development environment isn't supported out-of-box. We need to add an OpenSearch instance in its own container to provide the back-end search functionality.

rleed:
Do we have existing index and mapping definitions that can be used to initialize the OpenSearch instance?

huumn:
Mapping should happen automatically. The index called `item` will need to be created in some kind of initialization step, but that's it.

PR #509, opened 2023-09-21
Add OpenSearch to docker setup for development

rleed:
For issue #505... "In theory" this seems done, but I'm still not understanding why I don't get search results. I'm opening this PR in draft for now (EDIT: at least that was my intention). I really want to understand how OpenSearch ingests our data; I couldn't find in the code anywhere where we explicitly hand off newly posted content to OpenSearch (e.g. in the mutation resolver or such).

This PR follows a pretty straightforward setup per OpenSearch's instructions, with the variation of removing HTTPS since it is for development environments only. The optional dashboards are included by specifying the `aux` profile as in `docker-compose --profile aux up --build` and `docker-compose --profile aux down`.

On initial runs, some index and mapping errors will be seen in the logs, which can be resolved by trial-and-error with the following commands or similar:

```
curl -X PUT -H "Content-Type: application/json" http://localhost:9200/item --data "{}"

curl -X PUT -H "Content-Type: application/json" http://localhost:9200/item/_mapping --data "{\"properties\":{\"wvotes\":{\"type\":\"integer\"}}}"

curl -X PUT -H "Content-Type: application/json" http://localhost:9200/item/_mapping --data "{\"properties\":{\"sats\":{\"type\":\"integer\"}}}"
```

After that, everything appears to run smoothly, but searches on the SN site still don't return any results...even if new posts are made. (Nor do searches in the OpenSearch dashboard, as far as I can tell.) Something is still missing... but what?


api/search/index.js
```diff
@@ -1,7 +1,7 @@
 const os = require('@opensearch-project/opensearch')
 
 const options = process.env.NODE_ENV === 'development'
-  ? { node: 'http://localhost:9200' }
+  ? { node: 'http://opensearch:9200' }
```
huumn, line 4: Can you figure out a way to not need to hard code this?

This will break local dev for anyone not running in docker

huumn:
Cool! I'll try to run it and give more detailed feedback tomorrow

api/search/index.js
rleed: I've now moved it to .env, which should cover all bases assuming non-docker dev setups keep it blank (or set it to localhost) in their .env.

huumn: Woot

huumn:
I managed to get it to work but pretty much had to start from scratch. 

The main issue was that the worker was trying to reach `app` via `localhost` - among other things.

rleed:
> The main issue was that the worker was trying to reach `app` via `localhost` - among other things.

Interesting...it's slowly starting to make sense to me. Looking forward to working with it tomorrow.

I like how you put the script and disabled security directly in `docker-compose.yml`... I didn't know it could be done that way 😳

PR #511, opened 2023-09-21
Toast success and danger for donate action

SatsAllDay:
Danger case was already handled by the common `Form` component, but success wasn't toasted. I like having feedback on actions like this one.

Now, the danger toast message is also specific to the donate action, and not just the underlying JS error that occurred (which is logged, instead).

sn-donate-success-toast

sn-donate-danger-toast


SatsAllDay:
I wonder if we should also use the lightning visual effect on successful donation. That feels more appropriate? Or both? Idk

huumn:
Yeet! The lightning bolt would be cool for a little more visual feedback but I'm going to merge this and that can be done in another pr.

PR #500, opened 2023-09-18
Convert worker to ESM

ekzyis:
To use ESM for the worker, I created a package.json file in worker/ with `{ type: "module" }` as its sole content.

I then rewrote every import to use ESM syntax.

I also tried to set `{ type: "module" }` in the root package.json file to also use ESM in next.config.js.

However, this resulted in a weird problem: default imports were now getting imported as objects in this shape: `{ default: <defaultImport> }`.

Afaik, this should only be the case if you use "import * as foo from 'bar'" syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#default_import

This is fixed by not using `{ type: "module" }` for some reason. However, then, next.config.js also doesn't support ESM import syntax anymore.

The documentation says that if you want to use ESM, you can use next.config.mjs: https://nextjs.org/docs/pages/api-reference/next-config-js

But I didn't want to use MJS extension since I don't have any experience with it. For example, not sure if it's good style to mix JS with MJS etc. So I kept the CJS import syntax there.

PR #513, opened 2023-09-21
Images v2

### LINKED ISSUE #451: "Potato" image quality via proxy

ekzyis:
Depends on #500 

Fixes #451

TODO:
- [x] add placeholder image while images are not processed by worker yet (condition `imgproxyUrls IS NULL`)
- [x] fix fullscreen stays open on navigation
- [ ] fix alignment of `open original` / `open in new tab` in fullscreen mode
- [x] ~~image upload _(maybe in next PR?)_~~ _in next PR_
- [x] add script to process existing items
- [x] evaluate [lazy loading of images](https://www.freecodecamp.org/news/how-to-lazy-load-images-in-react/) and add if considered useful _-> evaluation: don't use it_

---

- [x] add video to demonstrate all changes related to images
- [x] improve accuracy of script by adding known positives (image hosting services like https://i.postimg.cc/) and known negatives (links to tweets etc.). currently, there might be false negatives for image detection because of timeouts (rate limiting?)
- [x] [make sure GIFs are not cut off if possible](https://stacker.news/items/263772) - use [this GIF](https://i.postimg.cc/1zGtmXVy/ao-Kdnr2-460svav1.gif) for testing
- [x] fixme: privacy setting is not respected since we load the original image anyway to check if it's an image. only afterwards, we check if the original image should be displayed

---

videos:

if "click to load external images" is enabled, there will be a placeholder for unprocessed images. on click, it loads the original image: https://files.ekzyis.com/public/sn/images_v2_processing_click_to_load.mp4

if "click to load external images" is not enabled, it will automatically load the original url if images are not processed yet (`imgproxyUrls` is falsy): https://files.ekzyis.com/public/sn/images_v2_processing_automatically_load_original.mp4

image processing is done in worker + responsive images: https://files.ekzyis.com/public/sn/images_v2_processing_in_worker.mp4
_(note: it will always load the best resolution if it was already cached)_

on imgproxy errors, it will show "click to load image" (if privacy setting enabled) or automatically load the original image (if privacy setting not enabled): https://files.ekzyis.com/public/sn/images_v2_imgproxy_error.mp4

---

TODO:

- [x] replace regexp usage with hardcoded list of known image hosting services and known not-image hosts, based on our existing data
- [x] ~~sign URLs in the database path for these known image hosting services to improve UX (no processing delay)~~ _decided not to since that would only be partial processing which breaks frontend assumptions about when items have been processed by the worker_


ekzyis:
I looked at https://web.dev/browser-level-image-lazy-loading and I think that's not something we need at the moment.

- For example, the thresholds can be quite large:

> On fast connections (4G), we reduced Chrome's distance-from-viewport thresholds from 3000px to 1250px and on slower connections (3G or lower), changed the threshold from 4000px to 2500px. This change achieves two things:

But I tested on https://stacker.news/items/178769 (with "click to load external images" turned off) which has a height of 10,000 pixels and all images were loaded nonetheless in Brave.

- [we shouldn't use lazy loading for images already in the first visible viewport](https://web.dev/browser-level-image-lazy-loading/#avoid-lazy-loading-images-that-are-in-the-first-visible-viewport) since [it might make the LCP metric worse](https://web.dev/lcp-lazy-loading/). This makes applying it more complicated and might actually decrease UX instead of improving it.

- [it's more about saving bandwidth than improving performance](https://web.dev/browser-level-image-lazy-loading/#why-browser-level-lazy-loading). While it would be nice to save bandwidth, there might be better options to do this without unknown risks.

ekzyis:
TODO:
- add video to demonstrate all changes related to images
- improve accuracy of script by adding known positives (image hosting services like https://i.postimg.cc/) and known negatives (links to tweets etc.). currently, there might be false negatives for image detection because of timeouts (rate limiting?)

ekzyis:
Only thing left to do is to fix the CSS (if we don't want to ship as is):

Screenshots from desktop and mobile which show the misalignment of the link to the original (bottom right corner):

2023-09-25-225514_1920x1080_scrot

sn ekzyis com_items_187221(iPhone SE)

I wanted to have the link aligned with the image. So on desktop it's too far to the left and on mobile, it's too far below the image.

worker/imgproxy.js
```diff
@@ -0,0 +1,180 @@
...
+  /^https:\/\/youtu\.be/,
+  /^https:\/\/(www\.)?youtube\.com/,
+  /^mailto:/,
+  /^https:\/\/stacker\.news\/items/,
+  /^https:\/\/news\.ycombinator\.com\/(item|user)\?id=/,
+  /^https:\/\/\w+\.substack.com/,
+  /^http:\/\/\w+\.onion/,
+  /^http:\/\/nitter\.priv\.loki/,
+  /^http:\/\/\w+\.b32\.i2p/,
+]
```
ekzyis, line 39: Might make sense to replace this with a [less vulnerable logic](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS)

ekzyis: Going to reprocess all images and collect metrics about URLs: how often they were used, how often errors were encountered, how often the HEAD requests returned true or false, how often the GET request returned true or false.

With these metrics, I will create a better list such that we don't need to make any request for most image services our stackers use. We can also skip requests for hosts where we are sure that they are not images. Currently, we try every URL to make sure we don't get false negatives.

ekzyis: Collected metrics. See https://github.com/stackernews/stacker.news/pull/513#issuecomment-1735788877

ekzyis: > If we immediately sign URLs for common hosting services, this shouldn't be too much of a problem.

I decided not to do this since that would introduce "partial processing" which breaks frontend assumptions about when items have been processed.

prisma/migrations/20230920192620_item_imgproxy_urls/migration.sql
```diff
@@ -0,0 +1,175 @@
...
+    IF med_votes >= 0 OR item."userId" = 27 THEN
+        med_votes := 0;
+    ELSE
+        med_votes := ABS(med_votes);
+    END IF;
+
+    -- there's no great way to set default column values when using json_populate_record
+    -- so we need to only select fields with non-null values that way when func input
+    -- does not include a value, the default value is used instead of null
+    SELECT string_agg('"' || key || '"', ',') INTO select_clause
```
huumn, line 56: This needs to use quote_ident

prisma/migrations/20230920192620_item_imgproxy_urls/migration.sql
```diff
@@ -0,0 +1,175 @@
...
+        UPDATE "Item" SET boost = boost + item.boost WHERE id = item.id;
+        PERFORM item_act(item.id, item."userId", 'BOOST', item.boost);
+    END IF;
+
+    IF item.status IS NOT NULL THEN
+        UPDATE "Item" SET "statusUpdatedAt" = now_utc()
+        WHERE id = item.id AND status <> item.status;
+    END IF;
+
+    SELECT string_agg('"' || key || '"', ',') INTO select_clause
```
huumn, line 130: same quote_ident

prisma/migrations/20230920192620_item_imgproxy_urls/migration.sql
ekzyis: Oh, yes, copied outdated migration

worker/imgproxy.js
```diff
@@ -0,0 +1,194 @@
...
+    const res = await fetchWithTimeout(url, { timeout: 10000 })
+    const buf = await res.blob()
+    isImage = buf.type.startsWith('image/')
+  } catch (err) {
+    console.log(url, err)
+  }
+
+  cache.set(url, isImage)
+  return isImage
+}
```
huumn, line 184: How frequently is this needed? Are there images that take 10s to return a HEAD request?

worker/imgproxy.js
```diff
@@ -0,0 +1,194 @@
...
+    }
+    if (!(await isImageURL(url, { forceFetch }))) {
+      console.log("[imgproxy] id:", id, "-- not image url:", url)
+      continue
+    }
+    imgproxyUrls[url] = {}
+    for (const res of resolutions) {
+      const [w, h] = res.split('x')
+      const processingOptions = `/rs:fill:${w}:${h}`
+      imgproxyUrls[url][`${w}w`] = createImgproxyUrl(url, processingOptions)
```
huumn, line 118: Does this also store the image width and height? 

Currently because we don't know the width and height, we can't reserve space for loading images so when they do load, they go from 0 height to whatever height they are, causing significant layout shift. It'd be nice to remove that.

worker/imgproxy.js
ekzyis: Some hosts don't support HEAD at all or respond with wrong headers.

For example, zapread.com (which is in the top 5 hosts used by stackers) responds with 404:

```
curl -I https://www.zapread.com/i/9QP
HTTP/2 404
date: Sat, 30 Sep 2023 20:45:06 GMT
content-type: text/html; charset=utf-8
access-control-expose-headers: Request-Context
cache-control: private
set-cookie: ARRAffinity=24410e45ca5e57f20082acb1c6e848971a37deecc97f93db2394281f8d4f09c7;Path=/;HttpOnly;Secure;Domain=www.zapread.com
set-cookie: ARRAffinitySameSite=24410e45ca5e57f20082acb1c6e848971a37deecc97f93db2394281f8d4f09c7;Path=/;HttpOnly;SameSite=None;Secure;Domain=www.zapread.com
strict-transport-security: max-age=31536000; includeSubDomains
x-aspnet-version: 4.0.30319
request-context: appId=cid-v1:1f8ce37e-229f-450c-a0fd-1303f2fc9b8c
x-powered-by: ASP.NET
x-content-type-options: nosniff
referrer-policy: no-referrer
cf-cache-status: DYNAMIC
report-to: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=sARzDZKmKzHoC8OHea0UssbRwy4EL%2FFI0Fl8vZB21iqoPgFawu7lSPePyri24k9AQeUX6lvGjZ1UDBob%2BVtjEj6FUj5Pbvv4G0Xw20i0xPwpLBm0tcap%2Br9j0x1PCm7%2BL1w%3D"}],"group":"cf-nel","max_age":604800}
nel: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}
server: cloudflare
cf-ray: 80ef4ec20918421d-EWR
alt-svc: h3=":443"; ma=86400
```

But if you go to https://www.zapread.com/i/9QP, there is an image.

So for reliable image detection, a GET request is definitely needed imo.

Also, you're looking at the GET request here. The timeout for HEAD requests is 1s.

worker/imgproxy.js
ekzyis: > Does this also store the original image width and height?

Not the original, this `w` here is only the upper bound for the width of the processed image.

But with the `w`, we also know the upper bound for height. So we could at least minimize layout shift by using heuristics?

If we want to know the exact width and height, we need to know the original width and height. From that, we should be able to tell which width and height the processed image in the various resolutions will have.

But for this, we either need to roll our own code or pay for imgproxy since the image info feature is a Pro feature:

> imgproxy can fetch and return a source image info without downloading the whole image.

-- https://docs.imgproxy.net/getting_the_image_info

PR #522, opened 2023-09-26
Crosspost discussion items to nostr

AustinKelsay:
Crosspostr

This PR implements crossposting of discussion items from stacker.news to nostr as nip-23 long-form parameterized replaceable events.

### High level summary:
- crosspost discussion items to nostr
- requires NIP-07 extension for signing
- we use your NIP-05 relays if set
- otherwise we default to these relays:
    - wss://nostrue.com/
    - wss://relay.damus.io/
    - wss://relay.nostr.band/
    - wss://relay.snort.social/
    - wss://nostr21.com/

### User facing changes:
- User can check the nostr crossposting checkbox on their settings to enable automatic attempted crossposting on all of their discussion items
- User can check the nostr crossposting checkbox under advanced for a one-off crossposting of a discussion item
  - If the user already has nostr crossposting checked on their settings they can uncheck it under advanced to NOT crosspost a particular discussion item.
- User gets an info modal for each of these checkboxes explaining the feature
- While the user is crossposting they get a success toast for each relay that there event was successfully published to
- Any failed relays get grouped up in a danger toast and shown to the user giving them the option to recursively retry until all failed relays succeed or skip the failed relays and finish the post
- If the user edits a post within the edit window they will re-crosspost it if crossposting is enabled. Since these events are parameterized and replaceable, when they are edited and republished the relays will discard the old version of the event and only keep the latest version.
- The nostr nip-05 section that was hidden under an accordion is now simply shown as the 'nostr' section and it is not hidden by default

### Internal code changes:
- Added migration for nostrCrossposting boolean
- Added crosspostDiscussion function in lib/nostr
- Added generic publishNostrEvent function in lib/nostr
- Added functionality to toast component to optionally allow triggering removal of the toast from the component that invokes it
- Added functionality to toast component to trigger an optional callback when the toast is closed
- Added crosspost handler function in discussion-form
- Added relayError function in discussion-form


SatsAllDay:
Really nice PR description, I need to get better about that!

lib/nostr.js
```diff
@@ -25,3 +32,93 @@ export function nostrZapDetails (zap) {
...
+    const userPubkey = await window.nostr.getPublicKey();
+    const timestamp = Math.floor(Date.now() / 1000);
+
+    const event = {
+      created_at: timestamp,
+      kind: 30023,
+      content: item.text,
+      tags: [
+        ['d', `https://stacker.news/items/${id}`],
+        ['a', `30023:${userPubkey}:https://stacker.news/items/${id}`, 'wss://nostr.mutinywallet.com'],
```
huumn, line 93: This `a` tag should probably list out the relays from the function parameter, right?

PR #526, opened 2023-09-27
Quote reply support on text-based posts and comments

### LINKED ISSUE #516: Quote reply option
If I want to reply to multiple parts of someone's item (post or comment), I need to copy-paste their text into the reply input, annotate it with quote markdown (`>`), then go from there.

I'd like a "quote reply" action that can be taken which copies the quoted item into the reply, automatically annotated with quotation markdown syntax. This action could live in the item overflow menu so as to avoid visual clutter.

Alternative is as I described above: manually copy-paste and annotate by hand.

**Additional context**
https://stacker.news/items/264216/r/WeAreAllSatoshi

See GitHub's implementation: 
image

SatsAllDay:
Closes #516 

Add support for a "Quote Reply" action on text-based posts and comments, available in the item info overflow menu.

Demo in 2 parts:
https://github.com/stackernews/stacker.news/assets/128755788/fd7ac334-492a-45b5-b7b1-e7a1545f4e28
https://github.com/stackernews/stacker.news/assets/128755788/8e226e09-b44f-4106-bb2e-1c27dd67a1ab


components/reply.js
```diff
@@ -33,10 +34,32 @@ function FreebieDialog () {
...
+      updatedValue = `${replyInput.current.value}\n${quote(item.text)}`
+    }
+    if (updatedValue) {
+      replyInput.current.value = updatedValue
+      formInnerRef.current.setValues({ text: updatedValue })
+      window.localStorage.setItem(`reply-${parentId}-text`, updatedValue)
+    }
+  }, [reply, item])
+  if (innerRef) {
+    innerRef.current = { quoteReply }
```
SatsAllDay, line 61: Apparently I just re-invented https://react.dev/reference/react/useImperativeHandle lol, I'll refactor to use that

SatsAllDay:
...and here's quote reply selection. Note it's not going to give you full markdown quoted since it's based on the rendered text, not the raw markdown. I think that should suffice. You get full markdown quoted when you quote the entire item instead of a selection

cc @ekzyis 

https://github.com/stackernews/stacker.news/assets/128755788/d8c107c8-6584-46a8-9b85-8335e48c7b75

components/action-dropdown.js
```diff
@@ -8,7 +8,7 @@ export default function ActionDropdown ({ children }) {
   }
   return (
     <Dropdown className={`pointer ${styles.dropdown}`} as='span'>
-      <Dropdown.Toggle variant='success' as='a'>
+      <Dropdown.Toggle variant='success' as='a' onMouseDown={e => e.preventDefault()}>
```
SatsAllDay, line 11: This is to prevent clearing a text selection when opening the action dropdown menu, thus enabling quoting selections.

components/comment.js
```diff
@@ -197,7 +200,7 @@ export default function Comment ({
               />
               )
             : (
-              <div className={styles.text}>
+              <div className={styles.text} ref={contentContainerRef}>
```
SatsAllDay, line 203: Mark the DOM tree root where the item's text can be found, so we can check to see if a selection is made in the item when doing a quote reply, not just any selection in the entire page.

components/item-info.js
```diff
@@ -140,6 +142,8 @@ export default function ItemInfo ({
           !item.mine && !item.deletedAt && <DontLikeThisDropdownItem id={item.id} />}
         {item.mine && !item.position && !item.deletedAt &&
           <DeleteDropdownItem itemId={item.id} type={item.title ? 'post' : 'comment'} />}
+        {(item.parentId || item.text) && onQuoteReply &&
```
SatsAllDay, line 145: If it has a parent, or if it has text, and if we have a handler for the quote reply action, render the menu item.

components/reply.js
```diff
@@ -33,10 +34,35 @@ function FreebieDialog () {
...
 
-export default function Reply ({ item, onSuccess, replyOpen, children, placeholder }) {
+export default forwardRef(function Reply ({ item, onSuccess, replyOpen, children, placeholder, contentContainerRef }, ref) {
   const [reply, setReply] = useState(replyOpen)
   const me = useMe()
   const parentId = item.id
+  const replyInput = useRef(null)
+  const formInnerRef = useRef()
+  useImperativeHandle(ref, () => ({
+    quoteReply: () => {
```
SatsAllDay, line 44: An imperative quoteReply method to be triggered from above in the react tree.

components/reply.js
```diff
@@ -33,10 +34,35 @@ function FreebieDialog () {
...
+      if (!reply) {
+        setReply(true)
+      }
+      const selection = window.getSelection()
+      const selectedText = selection.isCollapsed ? undefined : selection.toString()
+      const isSelectedTextInTarget = contentContainerRef?.current?.contains(selection.anchorNode)
+      const textToQuote = isSelectedTextInTarget ? selectedText : item.text
```
SatsAllDay, line 51: Check to see what we should quote, the entire item, or a selection of text.

components/reply.js
```diff
@@ -96,7 +122,6 @@ export default function Reply ({ item, onSuccess, replyOpen, children, placehold
     setReply(replyOpen || false)
   }, [upsertComment, setReply, parentId])
 
-  const replyInput = useRef(null)
```
SatsAllDay, line 99: This was just moved up above for earlier use.

components/reply.js
```diff
@@ -115,38 +140,38 @@ export default function Reply ({ item, onSuccess, replyOpen, children, placehold
...
-            {reply &&
-              <div className='mt-1'>
-                <FeeButton
-                  baseFee={1} parentId={parentId} text='reply'
-                  ChildButton={SubmitButton} variant='secondary' alwaysShow
-                />
-              </div>}
-          </Form>
-        </div>}
+      <div className={styles.reply} style={{ display: reply ? 'block' : 'none' }}>
```
SatsAllDay, line 143: Instead of conditionally rendering this div, I am toggling `display` CSS because I need the `Form` to exist even when the reply isn't shown, for the ref to be used.

huumn:
omg this looks amazing! (haven't looked at the code yet)

maybe we should insert a newline after the quote or is that unergonomic for some reason?

SatsAllDay:
> maybe we should insert a newline after the quote or is that unergonomic for some reason?

Done

ekzyis (commented):
To summarize and it helps me to make sure I understand what's going on:

You create two refs in `<Comment>`. One to reference the reply (`replyRef`) and one to reference the container holding the content we are replying to (`contentContainerRef`).

The `contentContainerRef` is simply passed to a `<div>` inside `<Comment>`. Nothing more to follow there.

Both refs get passed to `<Reply>`.

In `<Reply>`, `replyRef` gets transformed using `useImperativeHandle` to only expose the `quoteReply` function. That way you can pass `replyRef?.current?.quoteReply` as `quoteReply` to `<ItemInfo>` in `<Comment>`. `<ItemInfo>` then renders conditionally the menu item for `quote item`.

`contentContainerRef` is used in `quoteReply` to detect if we selected something in the text we're replying to.

In `<Reply>` two more refs are created: `replyInput` and `formInnerRef`.

`replyInput` is passed to `<MarkdownInput>` which then references the `<input>` element (after going through `<InputInner>`, too). This one then allows us to reference the existing reply in `quoteReply` so we don't overwrite it.

`formInnerRef` is passed to `<Form>` and then to `<Formik>`. That one is required to actually update the text in `quoteReply` since it's controlled by `formik`.

LGTM! :smile: 

But while playing around, I think I found a way to make `quote item` disappear as a menu item ... then I started recording to reproduce it but I wasn't able to, lol

Will take another look later, but now I have gotten a good overview of what you did :)

SatsAllDay:
You got it! As I mentioned above, the imperative stuff is kind of a mess. I wish this code was simpler but it just isn't. I welcome any simplification suggestions.

> But while playing around, I think I found a way to make quote item disappear as a menu item ... then I started recording to reproduce it but I wasn't able to, lol

I have noticed it sometimes takes a moment for the "quote reply" action to show up when loading a page locally. I figured it was still loading the `item` into memory via GraphQL API call so that the conditional rendering logic could be satisfied, but I also was not able to reliably reproduce it. I figured it was a side-effect of my slow dev machine :P

ekzyis:
> I wish this code was simpler but it just isn't. I welcome any simplification suggestions.

I always wondered if we should make use of more local [context](https://react.dev/reference/react/useContext). Afaik, we currently only use it on the global level to provide global context. But what if we use it to provide context for individual components, like to make `Reply` aware of what it's replying to etc.?

SatsAllDay:
I actually didn't realize you could use the same context provider in a nested fashion. That could be useful in this context, no pun intended lol

SatsAllDay:
Since the reply component already has the item to which it's replying as a prop, local context may only help with the "is the selection part of the item" aspect, which is cool, but we'll still need a ref for the DOM node, so I'm not sure it'll save us much in this case. Just my quick thinking, could be missing something...

huumn:
> Afaik, we currently only use it on the global level to provide global context.

fwiw we also use context to provide item roots to child comments

huumn:
All this imperative stuff scares the crap out of me, but it's well written and super useful so we have to let it fly

I changed it to use `onPointerDown` to work on mobile. I also made it so that selecting text in the parent then hitting `reply` also quote replies.

Excellent work! 🚀🚀🚀

PR #531, opened 2023-09-29
LUD-18 Wallet implementation

SatsAllDay:
Query the lightning address provider client-side to learn of capabilities

Conditionally render LUD-12 and LUD-18 fields based on what the remote server says is supported

Allow identifier, name, and email to be sent from the SN side during the withdrawal flow. Auth seems too complicated for our use case, and idk about pubkey?


huumn:
> Allow identifier, name, and email to be sent from the SN side during the withdrawal flow.

I haven't looked at the code yet, but I'd prefer this be optional.

> Auth seems too complicated for our use case, and idk about pubkey?

Hmmm I don't even know what this refers to. I'll take a look tomorrow.

SatsAllDay:
Every piece of payer data is opt in by default. Lightning address providers can specify whether certain pieces of payer data are required, but it's still ultimately up to the sender if they want to satisfy those requirements by supplying payer data or not. I have yet to encounter a lightning address provider that requires any payer data.

Regarding auth, it is whether you want to sign the random challenge provided by the lightning address provider with your private key. Very much like LNAuth. Since stacker news does not manage private keys for users, I don't think it's worth trying to implement Wallet implementation for auth payer data.

SatsAllDay:
I'll record a demo of this a little later on and attach it to this PR for reference. I think that would help to review, once it's ready.

SatsAllDay:
Two recordings to show functionality. I partially mocked the API response in the shorter recording to simulate the `identifier` use case.

Alby supports name, email. WoS doesn't support any LUD18 fields. You can also see the dynamic LUD12 comment length when I switch between different lightning address providers.

The `could not decode invoice` errors are just because I'm trying to process mainnet invoices using a regtest LND node locally.

https://github.com/stackernews/stacker.news/assets/128755788/f9366db1-be56-454d-bc8b-a70bc5f06048

https://github.com/stackernews/stacker.news/assets/128755788/16558ab7-6cb6-4eae-bb43-760efc4d1be2

huumn:
This looks great!

SatsAllDay:
I think this is ready for review now. Of course, now that I've marked it as such, I'll probably realize something else that I want to do with it 😅

hsjoberg:
FYI, I've added `identifier` support to the chat.blixtwallet.com playground now. In case you guys wanna test it there too.

SatsAllDay:
darth coin is going to be so happy

PR #532, opened 2023-10-01
Mention auto-complete

### LINKED ISSUE #529: mention autocompletes
It'd be nice in a post or comment when `@`ing someone to have some help getting the spelling right.

SatsAllDay:
Closes #529 

Support mention auto-complete in markdown input fields. Closely follows the suggestions in #529, which are largely inspired by GitHub's experience.

* When `@` is typed, show a list of top stackers for the day to pick from (5)
* Use the text immediately following `@` to search for similarly named users to pick from
* Use up/down arrow to navigate between search options
* Use Tab/Enter to pick the current selection
* Use escape to close the dropdown
* Selected user is inserted into the textarea input, and your cursor position is set to the end of the mention
* Suggestions do not show up if there's non-whitespace characters immediately preceding `@` e.g. `x@`
* Suggestions are shown as you're deleting an existing mention, all the way back to `@`. Suggestions are based on the remaining mention text.
* Suggestions appear below the caret position in the textarea, so it's easy to see but doesn't block your current text

### DIFF
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -151,7 +151,7 @@ export default {
         users
       }
     },
-    topUsers: async (parent, { cursor, when, by }, { models, me }) => {
+    topUsers: async (parent, { cursor, when, by, limit = LIMIT }, { models, me }) => {
       const decodedCursor = decodeCursor(cursor)
       let users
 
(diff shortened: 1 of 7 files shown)


SatsAllDay:
Demo cc @ekzyis I believe this captures all of your input

Ignore the delay on searching users, that's just my slow machine :P 

https://github.com/stackernews/stacker.news/assets/128755788/8883ccda-f542-4f26-9391-8da0cdd666bb

SatsAllDay:
Thank you! This is going to be a big release!

huumn:
YUGE 😄

PR #534, opened 2023-10-01
Add category tag to RSS feed

### LINKED ISSUE #528: Provide category tag per item containing the sub in the 'home' RSS feed
I'm using the RSS feeds to get the latest posts. The user can select a specific sub, or it defaults to the home RSS feed. To determine the active sub, you can parse the URL. However, in the 'home' RSS feed all subs are visible in random order.

To determine the sub per item in the 'home' RSS feed, I would like to see the `<category>` tag being used. This tag will contain the name of the sub per item. Optionally, a 'domain' attribute can be provided with the full URL to the sub as the attribute's contents. This tag is part of the [RSS 2.0 specification](https://www.rssboard.org/rss-specification#ltcategorygtSubelementOfLtitemgt).

**Additional context**
In `lib/rss.js` the `generateRssFeed()` function gets a list of items and separately the sub. Here the category tag could also be used for all RSS feeds but the 'home' feed to provide the sub for all items at once. The `generateRssItem()` function gets only the item object. Does this item object contain the sub as well (as in `item.sub`)? Then it can be added conditionally if it is available. The link tag off the channel part of the RSS feed should be a function so it can be reused in `generateRssFeed()` and in the attribute 'domain' of the category tag in `generateRssItem()`.

ekzyis:
Close #528

### DIFF
--- a/lib/rss.js
+++ b/lib/rss.js
@@ -23,6 +23,7 @@ const generateRssItem = (item) => {
   if (item.isJob) {
     title = item.title + ' \\ ' + item.company + ' \\ ' + `${item.location || ''}${item.location && item.remote ? ' or ' : ''}${item.remote ? 'Remote' : ''}`
   }
+  const category = item.subName ? `<category domain="${SITE_URL}/~${item.subName}">${item.subName}</category>` : '<category></category>'
   return `
     <item>
       <guid>${guid}</guid>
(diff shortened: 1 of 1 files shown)


lib/rss.js
```diff
@@ -23,6 +23,7 @@ const generateRssItem = (item) => {
   if (item.isJob) {
     title = item.title + ' \\ ' + item.company + ' \\ ' + `${item.location || ''}${item.location && item.remote ? ' or ' : ''}${item.remote ? 'Remote' : ''}`
   }
+  const category = item.subName ? `<category domain="${SITE_URL}/~${item.subName}">${item.subName}</category>` : '<category></category>'
```
ekzyis, line 26: Even though this should never be relevant since only comments have no sub, I think it's better to always include tags and not conditionally include them.

PR #538, opened 2023-10-02
Fix image detection in client not decoding original URL

ekzyis:
The worker is decoding proxy URLs to use the original URL since we want to process the original images, not already processed images. It then uses the original URL as the key to save the signed proxy URLs.

The client however does not do this. So for proxy URLs, this check always fails since the worker never uses proxy URLs as the key.

This commit fixes that during populating the image cache and when querying it.

### DIFF
--- a/components/image.js
+++ b/components/image.js
@@ -38,7 +38,10 @@ export function useImgUrlCache (text, { imgproxyUrls, tab }) {
     const urls = extractUrls(text)
 
     urls.forEach((url) => {
-      if (IMG_URL_REGEXP.test(url) || !!imgproxyUrls?.[url]) {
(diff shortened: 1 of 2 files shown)


huumn:
This conflicts with my changes. I'm not sure if they're still needed. Can you check?

ekzyis:
> This conflicts with my changes. I'm not sure if they're still needed. Can you check?

No longer needed. You decode the original URL to get `srcSet` [here now](https://github.com/stackernews/stacker.news/commit/62b53e1b3e5e57bcdce553bca7b4fb52075931a3#diff-7d53f6879a63a7d12daa48865f00f34663615bfd0e29512151b59e4918b17efdR103)

PR #541, opened 2023-10-03
Image refactor refactor

ekzyis:
- Removed comments about `srcSet` which no longer made sense
- Update comments in `<ImageOriginal>`
- Simplify condition when image is loaded in `<ImageOriginal>`
- Rename column from `clickToLoadImg` to `imgproxyOnly` since we no longer use clicking to bypass setting. We simply show the link for the user to click. (I also never liked the name of the setting)
- Add info about setting in /settings
- Fix misleading URL on imgproxy errors: I read [your comment here](https://github.com/stackernews/stacker.news/commit/62b53e1b3e5e57bcdce553bca7b4fb52075931a3#r129010564) and will make sure it supports markdown links, too

### DIFF
--- a/api/typeDefs/user.js
+++ b/api/typeDefs/user.js
@@ -23,7 +23,7 @@ export default gql`
     setSettings(tipDefault: Int!, turboTipping: Boolean!, fiatCurrency: String!, noteItemSats: Boolean!,
       noteEarning: Boolean!, noteAllDescendants: Boolean!, noteMentions: Boolean!, noteDeposits: Boolean!,
       noteInvites: Boolean!, noteJobIndicator: Boolean!, noteCowboyHat: Boolean!, hideInvoiceDesc: Boolean!,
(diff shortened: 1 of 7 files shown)


huumn:
😂 the title

ekzyis:
Mhh, or wait. I have a better solution:

If `[text](url)` is used and text is not empty, I think we shouldn't render this as an image at all.

This would also fix the markdown in my bio. I use `[click here](https://ekzyis.com/simplex.jpeg)` there but it displays the image which was precisely what I tried to avoid.

huumn:
> Also https://github.com/stackernews/stacker.news/commit/19d2f5f393c8e51394149b3d81cb4ba88bad6bfb throws an errors on [](https://postimg.cc/34sFW6GD) as the input (but apparently not consistently 🤔

Is this good to merge or does this need to be fixed?

ekzyis:
Fixed in https://github.com/stackernews/stacker.news/pull/541/commits/9b204aee4089271df6f6bd93de5f894913ede773 using optional chaining: `children?.[0]`

So this is good to merge now

PR #543, opened 2023-10-04
Add useful dev commands doc for others to reference

SatsAllDay:
I figured I'd share this file I'd been carrying around locally. I could also add some of these to `package.json` as scripts, if that would be better. Let me know if you think this is valuable to share!


huumn:
Definitely valuable to share. Going to merge.

PR #544, opened 2023-10-05
Fix quote selection reply in iOS Safari

SatsAllDay:
Approach borrowed from https://stackoverflow.com/a/72537632

I don't know how the licensing works for using content from stack overflow, but here's the applicable license: https://creativecommons.org/licenses/by-sa/4.0/ ¯\\\_(ツ)_/¯ 

Basically this makes a copy of the selection when the `touchend` event occurs, so we can use it for processing later (when the imperative `quoteReply` method is invoked). It also stores a reference to the containing node, so we can do the DOM tree lookup to confirm the selection is from the parent item.

This code listens to that event for each instance of the reply component, removing the event listener on unmount

I also threw in a minor tweak to the dev commands file that I noticed should be made.

### DIFF
--- a/components/reply.js
+++ b/components/reply.js
@@ -40,15 +40,42 @@ export default forwardRef(function Reply ({ item, onSuccess, replyOpen, children
   const parentId = item.id
   const replyInput = useRef(null)
   const formInnerRef = useRef()
+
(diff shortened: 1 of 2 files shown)


huumn:
Works! Still can't figure out how to prevent the highlight from laying on top of the dropdown though.

PR #545, opened 2023-10-05
LUD-18 fixes

SatsAllDay:
* Debounce the `onAddrChange` event handler when sending to a LN Address, so we more accurately display the input form for LUD-12 and LUD-18 options

* Remove explicit URI encoding of Payer Data when sending to a LN Addr, since we're getting encoding for free via URLSearchParams

* Append `@stacker.news` to identifier values sent in payer data

* Don't do extra decoding when receiving LUD-18 data

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -290,7 +290,7 @@ export default {
       if (payer) {
         payer = {
           ...payer,
-          identifier: payer.identifier ? me.name : undefined
+          identifier: payer.identifier ? `${me.name}@stacker.news` : undefined
         }
         payer = Object.fromEntries(
           Object.entries(payer).filter(([, value]) => !!value)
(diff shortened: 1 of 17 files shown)


hsjoberg (approved):
Code looks good to me 💯

Darth-Coin:
Welcome to LNURL Mafia...
image

SatsAllDay:
That does sound plausible. I can refactor it to use a different pattern, like we've done elsewhere. Or if you're messing with the code locally, feel free. I won't get to it til tomorrow.

huumn:
I made a custom hook for it. Basically we need to clear the timeout on unmount or else we'll leak memory.

It's unlikely the memory would amount to much in this case or that it would even happen (it'd require someone filling out the form rapidly and hitting submit in < 500ms).

huumn:
It's still a little wonky though wrt to deps and rerendering so going to probably need to revisit it tomorrow.

SatsAllDay:
> I made a custom hook for it. Basically we need to clear the timeout on unmount or else we'll leak memory.

Awesome! Makes sense to me. 

> It's unlikely the memory would amount to much in this case or that it would even happen (it'd require someone filling out the form rapidly and hitting submit in < 500ms).

I agree, probably not a big deal in this case, but it's good to set the correct pattern to model in the future 👍

> It's still a little wonky though wrt to deps and rerendering so going to probably need to revisit it tomorrow.

Sounds good. Let me know if I can help.

PR #548, opened 2023-10-06
Copy lightning address on click

SatsAllDay:
When you click on the button that has the lightning address on a user's profile, in addition to opening a modal with the LNURLP QR Code, it now also copies the lightning address to your clipboard. I've found that I want to pay via LN Addr this way all the time but it's not convenient.

https://github.com/stackernews/stacker.news/assets/128755788/cb4a126f-ddae-4cc8-881e-18937e839041

### DIFF
--- a/components/user-header.js
+++ b/components/user-header.js
@@ -22,6 +22,8 @@ import SubscribeUserDropdownItem from './subscribeUser'
 import ActionDropdown from './action-dropdown'
 import CodeIcon from '../svgs/terminal-box-fill.svg'
 import MuteDropdownItem from './mute'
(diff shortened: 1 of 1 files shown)


SatsAllDay:
I'm not sure in which conditions we'd prefer `navigator.share` over just `copy`, so I can refactor to use `navigator.share` if it's available in the browser. Open to feedback here

huumn:
Looks good. If this is for paying from SN to SN, it might make sense to autocomplete in the lighning address form as well.

SatsAllDay:
> it might make sense to autocomplete in the lighning address form as well.

Good idea!

PR #550, opened 2023-10-08
SN wallet transaction details

### LINKED ISSUE #503: SN wallet transactions details
**Description**
Based on my experiment post on SN https://stacker.news/items/259459/ I notice that paying a beer with sats from SN account, it doesn't have enough details to be like a real payment.

**Solution proposal**
Until SN account will be converted into a lndhub account (see #490 ) I would like to see in SN "wallet" section more details for payments done from there (not withdrawals to my own other wallets).

In that thread "Pay it forward" I tried to use as much is possible the sats received from other kind stackers, to pay directly the beers for people joining to this initiative.
I also wanted to give them some kind of proof, posting the payment hash, just in case somebody will come up with complaints that the payment wasn't done or I am some kind of liar.
Some other wallet apps provide this information and is good.

**What I would like to see in SN wallet?**
- payment hash of the tx
- description taken from the payment details. This is good for own tracking later. Or at least a field to add my own notes.
- pre-image (if is not too much)

SatsAllDay:
Closes #503 

This PR introduces a new `Bolt11Info` component to show info extracted from a BOLT11 after decoding. This component is used for Invoices and Withdrawals. It shows:
1. payment hash
2. payment description (if exists)
3. preimage (for invoices only, not withdrawals)

The preimage is fetched from LND via lookup by hash.

Fields are displayed in `CopyInput`, so they're readonly and easily copyable, if that's useful. 

This PR also updates the `CopyInput` component to use our toast hook for success/failure instead of showing the Thumb icon, just to standardize the UX. We don't have to keep this, but I thought the consistency would be good.

Are there other bits of info we should include? I am not really sure what's worth exposing.

## Invoice screenshot
image

## Withdrawal screenshot
image

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -76,13 +76,6 @@ export default {
         throw new GraphQLError('not ur withdrawal', { extensions: { code: 'FORBIDDEN' } })
       }
 
-      try {
-        const lnInv = await getInvoiceFromLnd({ id: wdrwl.hash, lnd })
-        wdrwl.preimage = lnInv.secret
-      } catch (err) {
-        console.error('error fetching invoice from LND', err)
-      }
-
       return wdrwl
     },
     connectAddress: async (parent, args, { lnd }) => {
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -1,4 +1,4 @@
-import { createHodlInvoice, createInvoice, decodePaymentRequest, payViaPaymentRequest, cancelHodlInvoice } from 'ln-service'
+import { createHodlInvoice, createInvoice, decodePaymentRequest, payViaPaymentRequest, cancelHodlInvoice, getInvoice as getInvoiceFromLnd } from 'ln-service'
 import { GraphQLError } from 'graphql'
 import crypto from 'crypto'
 import serialize from './serial'
@@ -40,6 +40,13 @@ export async function getInvoice (parent, { id }, { me, models, lnd }) {
   } catch (err) {
   }
 
+  try {
+    const lnInv = await getInvoiceFromLnd({ id: inv.hash, lnd })
+    inv.preimage = lnInv.secret
+  } catch (err) {
+    console.error('error fetching invoice from LND', err)
+  }
+
   return inv
 }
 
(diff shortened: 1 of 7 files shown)


huumn:
Looks sweet!

SatsAllDay:
cc @Darth-Coin does this satisfy your original request in #503 ? Any other information you'd want surfaced?

huumn:
Just double checking, but this doesn't reveal preimages before invoices are paid right (even via API this is a problem)?

SatsAllDay:
> Just double checking, but this doesn't reveal preimages before invoices are paid right (even via API this is a problem)?

Out code doesn't explicitly check paid status of the invoice to determine whether to include the preimage, so unless the LND API does, then I assume it would be included before the invoice is paid. I don't recall if I explicitly tested that locally, I know I verified it on paid invoices. I'll try it and report back.

I am still pretty green when it comes to preimage stuff, but I thought it was ok to expose it to the creator of the invoice? Is that only true after the invoice has been paid?

SatsAllDay:
Tried it: yes, it does reveal the preimage on unpaid invoices. Should I make a change so that it does not do that?

SatsAllDay:
From my (brief-ish) research, it seems like it shouldn't be a problem to reveal the preimage to the payee prior to the invoice being paid, since they're intended to get the funds anyway. But in theory they could leak it to someone else accidentally who could then steal the funds.

Please correct me if I'm missing anything here once the invoice is paid.

If we want to hide it until the invoice is paid, I'll just assert `inv.msatsReceived === inv.msatsRequested` before including the preimage.

huumn:
> Tried it: yes, it does reveal the preimage on unpaid invoices. Should I make a change so that it does not do that?

Yeah we definitely need to. More below.

> I am still pretty green when it comes to preimage stuff, but I thought it was ok to expose it to the creator of the invoice? Is that only true after the invoice has been paid?

Preimages unlock access to funds in a channel onchain, so if we reveal the preimage without someone paying for it, they could theoretically rob *us* (although not trivially).

> If we want to hide it until the invoice is paid, I'll just assert inv.msatsReceived === inv.msatsRequested before including the preimage.

We'd want to check if the payment is confirmed. I believe we set `confirmedAt` in the invoice record. `inv.msatsReceived === inv.msatsRequested` can actually be not equal in some cases (according to the spec `inv.msatsRequested <= inv.msatsReceived <= 2*inv.msatsRequested`).

huumn:
I put a bounty on SN for understanding this because I don't know how funds get stolen actually: https://stacker.news/items/282282

PR #552, opened 2023-10-10
Add a setting for Default Withdrawal max fee

### LINKED ISSUE #157: Feature request: Add a setting for Default Withdrawal max fee
For withdraw from an account on SN,  the 10 sats (default) is a higher amount than is needed for most any transaction amount that would be typically withdrawn from SN.

This request is to add to /settings a: `Default Withdrawal max fee`.

That value then would be the value that is used to prefill the `max fee` field in  /withdraw.

So if having Default Withdrawal max fee set to 1 sats, most withdrawal attempts will likely succeed.  If, for whatever reason, that withdraw fail, the user can then try again but raise the max fee a little higher on the next attempt.

The initial value for a new account on SN (for the `Default Withdrawal max fee`) could remain 10 sats, ... just that this setting gives users the option to customize that value so that they no longer need to edit the `max fee` when performing each withdrawal.

rleed:
This should resolve #157.

### DIFF
--- a/pages/settings.js
+++ b/pages/settings.js
@@ -63,6 +63,7 @@ export default function Settings ({ ssrData }) {
             tipDefault: settings?.tipDefault || 21,
             turboTipping: settings?.turboTipping,
             fiatCurrency: settings?.fiatCurrency || 'USD',
+            withdrawMaxFeeDefault: settings?.withdrawMaxFeeDefault || 21,
             noteItemSats: settings?.noteItemSats,
             noteEarning: settings?.noteEarning,
             noteAllDescendants: settings?.noteAllDescendants,
@@ -87,7 +88,7 @@ export default function Settings ({ ssrData }) {
             hideIsContributor: settings?.hideIsContributor
           }}
           schema={settingsSchema}
-          onSubmit={async ({ tipDefault, nostrPubkey, nostrRelays, ...values }) => {
+          onSubmit={async ({ tipDefault, withdrawMaxFeeDefault, nostrPubkey, nostrRelays, ...values }) => {
             if (nostrPubkey.length === 0) {
               nostrPubkey = null
             } else {
@@ -103,6 +104,7 @@ export default function Settings ({ ssrData }) {
               await setSettings({
                 variables: {
                   tipDefault: Number(tipDefault),
+                  withdrawMaxFeeDefault: Number(withdrawMaxFeeDefault),
                   nostrPubkey,
                   nostrRelays: nostrRelaysFiltered,
                   ...values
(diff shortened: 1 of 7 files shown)


prisma/schema.prisma
```diff
@@ -13,88 +13,89 @@ model Snl {
...
+  noteForwardedSats     Boolean              @default(true)
+  lastCheckedJobs       DateTime?
+  noteJobIndicator      Boolean              @default(true)
+  photoId               Int?
+  upvoteTrust           Float                @default(0)
+  hideInvoiceDesc       Boolean              @default(false)
+  wildWestMode          Boolean              @default(false)
+  greeterMode           Boolean              @default(false)
+  fiatCurrency          String               @default("USD")
+  withdrawMaxFeeDefault Int                  @default(10)
```
rleed, line 51: This was the only line added (expanding the whole column)

SatsAllDay: If there's a default here at the DB level, do we need to do all the `|| 21` in the code? Can't we assume there's always a value in `me.withdrawMaxFeeDefault`?

huumn: Yes, what @SatsAllDay is saying is true if `me` is always queried with `withdrawMaxFeeDefault`

huumn:
This looks good but I have local migrations (that I don't want to lose) and which will interfere with me testing it. I'll keep you updated.

ekzyis:
> This looks good but I have local migrations (that I don't want to lose) and which will interfere with me testing it. I'll keep you updated.

I'll try to check today

pages/settings.js
```diff
@@ -163,6 +165,13 @@ export default function Settings ({ ssrData }) {
             items={SUPPORTED_CURRENCIES}
             required
           />
+          <Input
+            label='default max fee for withdrawals'
+            name='withdrawMaxFeeDefault'
+            groupClassName='mb-0'
```
ekzyis, line 171: There is no visual separation to the next item:

2023-10-15-200820_723x101_scrot

This suggestion fixes that:

```suggestion
```

Including this fix, the settings page looks like this:

2023-10-15-201132_1920x1080_scrot

I am wondering if we should create a new section for zap default, fiat currency and default max fee for withdrawals? Starts to look a bit crowded

prisma/schema.prisma
ekzyis: Also agree with @SatsAllDay but just want to mention that additionally to the default, the column is not nullable. If it would be nullable this would be a slightly different story.

So I also think the `|| 21` are not necessary

ekzyis (changes requested):
Found only the following which should be addressed:

- missing margin in /settings.
- unnecessary `|| 21` 

Other than that, looks good :+1: 

What I looked out for:

- Checked out /settings, changed `withdrawMaxFeeDefault` and saved settings and then made sure this value gets used in the withdrawal forms.

- `withdrawMaxFeeDefault` does not allow negative numbers or 0. only allows positive integers

Not much else to test afaict

prisma/schema.prisma
rleed: Removed, thanks all!

PR #553, opened 2023-10-10
Add auto-complete support when withdrawing to ln addr

SatsAllDay:
adds auto-complete support for other stacker.news users when withdrawing to a lightning address

implemented via adding an optional `transformUser` prop to the `UserSuggest` and `InputUserSuggest` components, which allows you to transform fetched user results before displaying in the suggestion dropdown

this is used to transform a user nym to nym@stacker.news, the corresponding lightning address

by default, `transformUser` is an identity fn aka no transformation

this change also clears suggestions when the surrounding input field is blurred, which is a better UX IMO

https://github.com/stackernews/stacker.news/assets/128755788/656970be-d7df-4d59-99c1-58382c302f50

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -394,15 +394,28 @@ function InputInner ({
   )
 }
 
-export function UserSuggest ({ query, onSelect, dropdownStyle, children, transformUser = user => user }) {
+export function UserSuggest ({
+  query, onSelect, dropdownStyle, children,
+  transformUser = user => user, selectWithTab = true, filterUsers = () => true
+}) {
   const [getUsers] = useLazyQuery(TOP_USERS, {
     onCompleted: data => {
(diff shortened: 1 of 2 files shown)


huumn:
This is great! 

Nit: Should it maybe also clear suggestions when recognizing it's not `@stacker.news` or maybe just on `@` because they'd only be proceeding if it weren't an SN user? I don't know how hard that is

SatsAllDay:
> Nit: Should it maybe also clear suggestions when recognizing it's not `@stacker.news` or maybe just on `@` because they'd only be proceeding if it weren't an SN user? I don't know how hard that is


That's good feedback, thank you! I'll explore implementing it 

I also think maybe not having `Tab` select the value for this specific input component, since tab is often used to switch to the next field in the form. Can keep `enter` though.

huumn:
Really elegant changes!

PR #558, opened 2023-10-12
Add optional text input for link posts

### LINKED ISSUE #504: Add optional text input for link posts

See https://stacker.news/items/259936

Add optional markdown text input field for link posts. This will then be the first comment. The text input field may be put into `options` so it's hidden first to keep the UI/UX clean.

\-

**Additional context**

See discussion in https://stacker.news/items/258906

rleed:
To resolve #504.

image


prisma/migrations/20231011205259_free_link_comment/migration.sql
```diff
@@ -0,0 +1,107 @@
...
+    item "Item";
+    med_votes FLOAT;
+    select_clause TEXT;
+    poll_options JSONB;
+BEGIN
+    PERFORM ASSERT_SERIALIZED();
+
+    -- access fields with appropriate types
+    item := jsonb_populate_record(NULL::"Item", jitem);
+    poll_options = (options->'poll_options');
```
rleed, line 23: to allow general-purpose options to be passed

prisma/migrations/20231011205259_free_link_comment/migration.sql
```diff
@@ -0,0 +1,107 @@
...
+
+    SELECT msats INTO user_msats FROM users WHERE id = item."userId";
+
+    IF item."maxBid" IS NOT NULL THEN
+        cost_msats := 1000000;
+    ELSIF (options->'noCost') IS NOT NULL THEN
+        cost_msats := 0;
```
rleed, line 30: added this

prisma/migrations/20231011205259_free_link_comment/migration.sql
```diff
@@ -0,0 +1,107 @@
...
+    ELSE
+        cost_msats := 1000 * POWER(10, item_spam(item."parentId", item."userId", spam_within));
+    END IF;
+    -- it's only a freebie if it's a 1 sat cost, they have < 1 sat, and boost = 0 ...OR the item has no cost
+    freebie := ((cost_msats <= 1000) AND (user_msats < 1000) AND (item.boost = 0)) OR (cost_msats = 0);
```
rleed, line 35: added OR cost_msats = 0

components/adv-post-form.js
```diff
@@ -17,14 +17,34 @@ export function AdvPostInitial ({ forward, boost }) {
...
+                  <Info>
+                    <span className='fw-bold'>Use this initial comment to do e.g. any of the following:</span>
+                    <ul>
+                      <li>Provide context</li>
+                      <li>Summarize the link</li>
+                      <li>Note the reason why you are posting the link</li>
+                      <li>Respect readers time by helping them decide whether to click the link or not</li>
+                      <li>Spur great conversation!</li>
+                    </ul>
+                  </Info>
```
ekzyis, line 41: I would change the info text to this:

> You can use this initial comment to ...
> - provide context
> - summarize the link
> - spur great conversation!
> - or for anything else that comes to your mind

Reasons:

- I don't like long text where shorter text can convey the same in a more efficient manner  - I assume users prefer "efficient text", too
- "Note the reason why you are posting the link" feels too similar to "provide context".
- "Respect readers time ..." is imo setting too much of a tone - I don't know how to explain it better, lol

components/adv-post-form.js
```diff
@@ -17,14 +17,34 @@ export function AdvPostInitial ({ forward, boost }) {
...
 
   return (
     <AccordianItem
       header={<div style={{ fontWeight: 'bold', fontSize: '92%' }}>options</div>}
       body={
         <>
+          {!edit && isLink &&
+            <MarkdownInput
+              label={
+                <div className='d-flex align-items-center'>comment
```
ekzyis, line 31: If you go down the `item.text` path, maybe this should be renamed to `text`? Then it's consistent with the other form fields since their names match exactly the respective database column.

But maybe "comment" is still a better user description than "text".

ekzyis (commented):
After looking at these changes, I realized that it should be a lot easier to just also set `text` and not only `url` of the item instead of creating a comment as a separate item. This has the following advantages:

- can reuse existing `create_item` postgres function, so no db migration with support for `noCosts` needed
- fee escalation can work as before: no need to mention something like `free, but compounds any future spam penalty` which not every user might understand
- it just works™ without any backend changes :tada:  (did not test if it really requires zero backend changes but i have good faith belief in that)
- consistent edit UX: you can edit everything you were able to create in the same form

Another thing is that it doesn't "[smell good](https://en.wikipedia.org/wiki/Code_smell)" to add `noCosts` as an exception to `create_item` (which was meant to be a generic function) for a single use case.

It also might be better UX if the provided summary actually shows up at the top for every user instead of as a separate comment which might get lost if there are many comments? Using `item.text` kind of works like a pinned comment in that case (except it's just regular UX like for discussion posts)

Thanks for putting in the work! It wasn't clear to me before that creating a separate comment might lead to all of this.

ekzyis:
Thinking more about this, I wondered if it would break some assumptions in our code if link posts now suddenly also can have `text` set to something - like code where we want to distinguish between link posts and discussion posts.

However, searching for `item.text` didn't look like there is anything that would break in that case.

huumn:
I agree that this would ideally not require special treatment and simply occupy the `text` column on the post.

rleed:
Great idea to simplify with the text field... I will pursue that!

components/adv-post-form.js
rleed: I went ahead and changed it to "text" since it looks different to the user than a comment. (If there's any doubt, the info button is there to clarify any doubt.)

PR #563, opened 2023-10-13
Fix duplicate push notifications on item edit if mentioned

### LINKED ISSUE #557: Duplicate push notifications when you're mentioned in an item that's edited
**Description**
If you are mentioned in an item (post/comment), every time that item is edited, if you're still mentioned in the content, you get another push notification (if you have them enabled).

**Steps to Reproduce**
1. Enable push notifications
2. Enable notifications for mentions
3. Be mentioned by another user in a post or comment
4. Have the other user edit the post or comment while keeping your mention in place.

**Expected behavior**
I'd like to only receive a push notification one time in this sequence, not once for every edit made where my mention is still present.

**Screenshots**
N/A

**Logs**
N/A

**Environment:**
N/A

**Additional context**
N/A

ekzyis:
- refactored service worker code by putting listeners into own file
- service worker now keeps track for which items it has already seen a MENTION push notification (fix #557)

https://github.com/stackernews/stacker.news/assets/27162016/0a92d6d3-eb38-4bba-95a5-e7b4b41854f0


SatsAllDay:
Cool... I was going to fix this in the item resolver by doing a diff on the old vs new mentions, and conditionally sending pushes based on that. I wouldn't have thought of this approach. Nice!

ekzyis:
> Cool... I was going to fix this in the item resolver by doing a diff on the old vs new mentions, and conditionally sending pushes based on that. I wouldn't have thought of this approach. Nice!

Interesting, how would you have done the diff? Store 2-tuples (userId, itemId) which tracks which users received a MENTION notification for an item already to not send another one? I wanted to avoid to add state to the server so I moved that state to the client instead. Basically keeping the server as dumb as possible.

However, there is one potentially major and one minor issue with my approach:

Minor issue: Unbound state array. Since I only add to `itemMentions` and never remove anything, this only grows in size with no upper limit. It will get reset between deployments but maybe I should add some logic to only keep max 100 items or something.

Potentially major issue: Silent push notifications may result in a default push notification:

> With push events, there is an additional requirement that you must display a notification before the promise you passed in has settled.

-- https://web.dev/articles/push-notifications-handling-messages

(_also some other sources I can't find anymore where I read more about this behavior_)

I haven't witnessed this on Brave on Linux Desktop but I will check on Chrome Desktop and my Android phone (Chrome + Brave). Can someone check macOS and iOS?

If any device shows default push notifications (and we can't avoid it), we need to drop this approach and need to do something server side.

SatsAllDay:
> Interesting, how would you have done the diff? Store 2-tuples (userId, itemId) which tracks which users received a MENTION notification for an item already to not send another one? I wanted to avoid to add state to the server so I moved that state to the client instead. Basically keeping the server as dumb as possible.

My thinking was, within `createMentions` for a given item:
1. Extract the mentions in the text (no change)
2. Fetch any existing mentions in the DB for this item
3. Create any new mentions in the DB based on the current text value
4. Only send push notifications to mentions in the current text that didn't exist in the DB already (in step 2)

Since we don't delete mentions in the DB if the mention is removed from the text, this approach _should_ work. While it does make the server-side impl more complicated, it does avoid trying to send notifications that will ultimately just be hidden/ignored in the client. 


> Minor issue: Unbound state array. Since I only add to itemMentions and never remove anything, this only grows in size with no upper limit. It will get reset between deployments but maybe I should add some logic to only keep max 100 items or something.

You could purge per item 10 minutes (the edit window) after the first entry is stored to help keep it trimmed.

> I haven't witnessed this on Brave on Linux Desktop but I will check on Chrome Desktop and my Android phone (Chrome + Brave). Can someone check macOS and iOS?

I am not sure I understand the repro steps to try it, but am willing to try if the steps can be outlined for me.

ekzyis:
> 2. Fetch any existing mentions in the DB for this item

Ah, I forgot that we can use the DB, lol

> I am not sure I understand the repro steps to try it, but am willing to try if the steps can be outlined for me.

You need to create two accounts and enable push notifications on the first one. (For multiple accounts, I use Login with Lightning and https://legend.lnbits.com/)

Then you mention the first account from the second account. If you now edit the notification (while keeping the mention inside), the first account shouldn't receive another notification per edit.

Makes sense?

SatsAllDay:
> You need to create two accounts and enable push notifications on the first one. (For multiple accounts, I use Login with Lightning and https://legend.lnbits.com/)
> 
> Then you mention the first account from the second account. If you now edit the item (while keeping the mention inside), the first account shouldn't receive another notification per edit.
> 
> Makes sense?

Yes, that makes sense. I guess I was confused about the "default push notification" described above. But I guess I can pull this branch, do the above steps, and should be able to see it, if it's going to happen.

SatsAllDay:
I tried it on Chrome on macOS and didn't see any issues with the notifications - behaved as we hoped.

I can't seem to enable push notifications on iOS when running on `localhost`, so we might need to try it on a hosted environment like sn.ekzyis.com.

huumn:
> Edit: speaking of which, a pre-prod aka staging deployment of SN might be nice to have. Maybe something to consider in the future.

Most definitely. I've been fantasizing about giving all regular contributors their own staging envs with up to date scubbed prod data.

Early next year I'm hoping to go back to transitioning us to a declarative infrastructure microservice system which should make this pretty simple.

ekzyis:
> Edit: speaking of which, a pre-prod aka staging deployment of SN might be nice to have. Maybe something to consider in the future.

I might be able to just copy my nginx config for the reverse proxy of https://sn.ekzyis.com and change the upstream IP addresses to host a staging environment for contributors as a temporary solution. This is the config that I am using:

```nginx
upstream sn {
  server 127.0.0.1:3003;
}

upstream sn-imgproxy {
  server 127.0.0.1:3004;
}

map $http_upgrade $connection_upgrade {
  default upgrade;
  ''      close;
}

map $uri $x_content_length {
    default     $http_content_length;
    /api/lnauth '';
}

server {
    server_name sn.ekzyis.com;
    listen      80;
    listen      [::]:80;

    return 301 https://sn.ekzyis.com$request_uri;
}

server {
    server_name         sn.ekzyis.com;
    listen              443;
    listen              [::]:443;

    ssl                 on;
    ssl_certificate     /etc/letsencrypt/live/sn.ekzyis.com/fullchain.pem;
    ssl_certificate_key /etc/letsencrypt/live/sn.ekzyis.com/privkey.pem;
    client_max_body_size 10M;

    location /img/ {
      proxy_set_header Host $http_host;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header X-Forwarded-Proto $scheme;
      proxy_set_header X-Forwarded-Host $host;
      proxy_set_header X-Forwarded-Prefix /;
      proxy_pass http://sn-imgproxy/;
    }

    location /_next/webpack-hmr {
      proxy_set_header Host $http_host;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header X-Forwarded-Proto $scheme;
      proxy_set_header X-Forwarded-Host $host;
      proxy_set_header X-Forwarded-Prefix /;
      proxy_set_header X-Real-IP $remote_addr;
      proxy_set_header Upgrade $http_upgrade;
      proxy_set_header Connection $connection_upgrade;
      proxy_pass http://sn$request_uri;
    }

    include letsencrypt.conf;
}
```

I then use SSH remote port forwarding so there is indeed something listening on 127.0.0.1 from the server's point of view.

To not have to give people SSH access to my server (or do some complex configuration), I could replace that part by setting up a WireGuard VPN and then using the internal IP addresses from each peer.

People would then just need to setup WireGuard on their machines and make sure their local app is also listening on the WireGuard interface (not 127.0.0.1).

I would be very happy to provide something like this. The work on my side would be literally just a few minutes, I think. But not sure how @huumn feels about this?

huumn:
Works for me!

ekzyis:
> Works for me!

You mean that I offer this WireGuard solution or that the changes of this PR work as intended on iOS? :smile:

PR #564, opened 2023-10-14
hasNewNotes for your own comments in subscribed threads

SatsAllDay:
Don't include your own comments when determining if there are new thread subscription replies for user hasNewNotes. 

Basically, if you subscribed to a thread and then made a comment on it, your notification bell would suggest you have new notifications to read, but when you go to the notifications page, there weren't any. Fun!


huumn:
Oh thank goodness! I was wondering why that was happening

ekzyis (approved):

PR #565, opened 2023-10-15
Dynamic protocol, host for share URLs so they work on localhost and in prod

SatsAllDay:
Basically the title... I wanted to use shared URLs in localhost and they were pointing to prod.

### DIFF
--- a/components/share.js
+++ b/components/share.js
@@ -3,13 +3,22 @@ import ShareIcon from '../svgs/share-fill.svg'
 import copy from 'clipboard-copy'
 import { useMe } from './me'
 import { useToast } from './toast'
+import { SSR } from '../lib/constants'
+
+const getShareUrl = (item, me) => {
+  const path = `/items/${item?.id}${me ? `/r/${me.name}` : ''}`
+  if (!SSR) {
+    return `${window.location.protocol}//${window.location.host}${path}`
+  }
+  return `https://stacker.news${path}`
+}
 
 export default function Share ({ item }) {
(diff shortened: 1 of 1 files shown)


ekzyis:
Why not use `process.env.PUBLIC_URL`?

SatsAllDay:
> Why not use `process.env.PUBLIC_URL`?

Because that thought never occurred to me lol

Is it set to `https://stacker.news` in prod? If so, I can make this change.

SatsAllDay:
Oh, looks like `PUBLIC_URL` only works server-side, we need to prefix it with `NEXT_PUBLIC` to get it to work client-side, so I think that's a blocker since I don't have the ability to configure new env vars for prod?

ekzyis:
> Is it set to https://stacker.news in prod? If so, I can make this change.

Yes, but please confirm @huumn 

> Oh, looks like PUBLIC_URL only works server-side, we need to prefix it with NEXT_PUBLIC_

Ah yes, I think you can do that. I don't think that's a problem for anything.

> so I think that's a blocker since I don't have the ability to configure new env vars for prod?

I don't think that's a blocker. You don't need to configure new env vars. @huumn can set up env vars during deployment as he did for changes of mine in the past

huumn:
> Yes, but please confirm @huumn

Confirmed

> You don't need to configure new env vars for prod. @huumn can set up env vars during deployment as he did for changes of mine in the past

I can but I also think using window.location is fine.

PR #567, opened 2023-10-16
Custom date selector for more pages

### LINKED ISSUE #410: Custom Date Selector (250,000 Sat Bounty)

rleed:
To close #410.

I have virtually finished this, but I have two pesky problems I'm struggling to solve on the charts.... see code comments... please if anyone can offer any assistance, I'm stuck on those strange behaviors.


api/resolvers/growth.js
```diff
@@ -17,6 +17,7 @@ export function interval (when) {
 
 export function timeUnit (when) {
   switch (when) {
+    case 'custom':
     case 'week':
     case 'month':
       return 'day'
```
rleed, line 23: So right here it should set the unit to `day`, not `hour`. But the charts always show x-axis labels like `9pm` or such, as if hours were selected. I do not understand exactly where the problem is.

api/resolvers/growth.js
```diff
@@ -44,50 +57,63 @@ export function withClause (when) {
...
-    return and ? '' : 'TRUE'
+export function viewIntervalClause (when, from, to, view, and) {
+  function daysBetween (one, another) { // horrible hack for custom case
+    return Math.round(Math.abs((+new Date(one)) - (+new Date(another))) / 8.64e7)
+  }
+  switch (when) {
+    case 'forever':
+      return and ? '' : 'TRUE'
+    case 'custom':
+      return `"${view}".day >= date_trunc('day', timezone('America/Chicago', now() - interval '${daysBetween(new Date(), from) + ' days'}')) AND "${view}".day <= date_trunc('day', timezone('America/Chicago', now() - interval '${daysBetween(new Date(), to) + ' days'}')) ${and ? 'AND' : ''} `
```
rleed, line 80: And this is stranger than strange. As written, it "works" but obviously its the hack of all hacks. I would simply like to put something such as `>= $1` (or `>= ${from}`), but when I do so, the results are as if the $1 parameter is always "now()" even when it isn't.

api/resolvers/growth.js
huumn: I'd guess that's something happening clientside in `components/charts.js > dateFormatter`

api/resolvers/growth.js
huumn: This seems a little hairy. I won't be able to diagnose without pulling which I won't be able to do for another couple days.

rleed:
So I believe this is now fully functional for the user.

ekzyis:
Unfortunately no time for a full review but this is some feedback that I gather just by interacting with the component:

---

Imo, if you can edit the text, it should also be possible to update the date by editing the text. So either remove the cursor blink and make it impossible to change the date via text input, or support changing the date via text input.

_me trying to edit date manually without using picker. First attempt is just blur, second attempt is trying to use Enter:_

https://github.com/stackernews/stacker.news/assets/27162016/11dff43c-367d-4d2d-9d8a-d11e8a39f221

The field should never be in an unclear state. If I just select the start date and then leave the picker, there is no end date. It should probably default to the start date:


https://github.com/stackernews/stacker.news/assets/27162016/f3e28dc4-6c36-4834-af8a-683c3ab62c96

Not sure how date picker usually work so this might be a nitpick but I expected that I can select a range. It doesn't mean that I have to select the start date first. So I if pick a date and then pick an earlier date, the range should just pick the earlier date as the start date:


https://github.com/stackernews/stacker.news/assets/27162016/6513180e-17ae-4efa-a169-7c9aec5d7144

Arrows not vertically aligned with month:

2023-10-17-225744_265x83_scrot

rleed:
Thanks for the feedback. I'll take a look to see if these aspects can be improved in a practical way.

rleed:
> Imo, if you can edit the text, it should also be possible to update the date by editing the text. So either remove the cursor blink and make it impossible to change the date via text input, or support changing the date via text input.

Yeah, I'm not sure why that didn't work out of the box. Now I added logic to handle editing the text field.

> The field should never be in an unclear state. If I just select the start date and then leave the picker, there is no end date. It should probably default to the start date:

I agree 100%, but unfortunately the component doesn't provide a blur or close event for the picker itself as far as I have been able to determine. (Clicking out of the text box works now.) I did fix the logic so the search isn't attempted until (or unless) the range is complete.

> Not sure how date picker usually work so this might be a nitpick but I expected that I can select a range. It doesn't mean that I have to select the start date first. So I if pick a date and then pick an earlier date, the range should just pick the earlier date as the start date:

I had the same expectation at first, but this behavior is apparently by design. I think we can live with it.

> Arrows not vertically aligned with month:

Good catch...our CSS broke some assumptions of the date picker. Now it is fixed.

> Selecting a single date doesn't seem to work:

This is a very thorny issue. It boils down to the fact that the date picker is picking "moments" in time. We can't drop the time component because we do two-way conversion of the dates (i.e. the date picker drives the URL query, and the URL query drives the date picker), and changing the values would break that two-way conversion or the date picker itself. Similarly, offsetting the start point by -24 hours for example would be problematic too. We might be able to do something in the underlying queries, but I don't think that is good practice. Furthermore, if we disregard the time component there would always still be the problem that the user selects dates according to his time zone, but the server searches according to its time zone, which would make more (and weirder) problems. I think just to leave it as-is is the best option. We could consider making the time selectable too, but personally I think that would be more inconvenient than it would be worth most of the time.

That's what I found so far, but if anyone has a good idea for this case, I'm happy to try something else!

ekzyis (commented):
>> Imo, if you can edit the text, it should also be possible to update the date by editing the text. So either remove the cursor blink and make it impossible to change the date via text input, or support changing the date via text input.
>
> Yeah, I'm not sure why that didn't work out of the box. Now I added logic to handle editing the text field.

Editing the text field works now, but you can enter dates which you couldn't pick:

https://github.com/stackernews/stacker.news/assets/27162016/8d38d15d-cd92-4c75-917a-b23290890b5c

---

There is a bug where the search doesn't update. I believe it's because it's related to the console error because of duplicate keys since in these cases, the search does not update (`useMemo` not triggered):

https://github.com/stackernews/stacker.news/assets/27162016/193f1706-529d-41ef-b8d6-2f22f50d27e6

So maybe the search returns the same item multiple times for some reason, leading to this error?

I also just noticed that when I searched for 05/18/2023 - 05/19/2023, there was no error in the console but the search still showed a result from 05/30/2023 :thinking: (at the 16 second mark)

---

I am wondering if we should use unix timestamps in the URL to make them shorter. So like this

```
/search?q=suggestions&sort=recent&when=custom&from=1684267200&to=1684353600
```

instead of 

```
/search?q=suggestions&sort=recent&when=custom&from=2023-05-16T22%3A00%3A00.000Z&to=2023-05-17T22%3A00%3A00.000Z
```

? Unix timestamps are less readable though but the URL encoding makes these YYYY-MM-DDTHH:mm:ss date strings also hard to read. But since we can only filter down to a day, we might be able to get away with just YYYY-MM-DD in the URL?

api/resolvers/growth.js
ekzyis: I couldn't get this to work as-is. Using `forever` shows this:

2023-10-30-131208_1920x1080_scrot

Using `custom` from `01/01/22` to `10/29/23`:

2023-10-30-131232_1920x1080_scrot


Shouldn't I get results with this query if it would work? :thinking: 

I wanted to investigate this:

> when I do so, the results are as if the $1 parameter is always "now" even when it isn't.

but since you mentioned, as it is, it works but it doesn't for me, I am not sure where to start.

huumn:
`/top/posts/day` errors for me when I visit it even after fixing a bug in it.

`ERROR: operator does not exist: timestamp without time zone <= bigint
    HINT: No operator matches the given name and argument types. You might need to add explicit type casts.`

huumn:
- [ ]  top is still broken (It's tricky to modify which is not your fault)
- [ ] we shouldn't use days as the unit when the date range is > ~90 days

I'm going to move onto reviewing other things that require less work from me for the time being.

huumn:
All good. I'm almost done fixing this. I decided to just get it done because reviewing it got me stuck thinking about how do this right 😄

huumn:
Holy cow! That was way more work than I wanted it to be. I can't blame you for giving up on it.

**_In the future PLEASE keep incomplete PRs in draft._**

rleed:
> Holy cow! That was way more work than I wanted it to be. I can't blame you for giving up on it.

Now that I see just how extensive of changes were needed, I do realize I bit off more than I could chew with that PR and grossly underestimated the scope. Although I am too stubborn to have officially given up, lol, I did realize that my test environment (esp. database content) was inadequate, among the other challenges, and that was holding me back to a degree (plus it's been crazy for me locally lately). So I wholeheartedly welcome and appreciate you stepping in to finish this one (esp. right now to not keep expectant stacker(s) waiting longer). Thank you!

> **_In the future PLEASE keep incomplete PRs in draft._**

Will do!

huumn:
> Although I am too stubborn to have officially given up, lol

lol we are stubborn together. Once I tried to fix one part I couldn't stop until it was done

PR #570, opened 2023-10-17
Ephemeral items

### LINKED ISSUE #472: Self-destructing comments

Some stackers manually delete their comments after some amount of time for privacy reasons (I assume).

This is bad UX and we could maybe accommodate their needs better.

We could help them by providing a setting in their settings and/or during posting/commenting that will delete their comment automatically.

We could charge extra sats for that since I think we don't want everyone to delete all their posts and comments after some amount of time, lol. The faster it gets deleted, the more sats it requires?

- a script that we could provide which they can run to do it themselves

**Additional context**

https://stacker.news/items/245650

SatsAllDay:
Closes #472 

Introduce support for ephemeral items, i.e. items that delete themselves after a specified amount of time.

Stackers can enable this by including text in their items that follows this pattern: 

`@delete in <number> <unit>`

e.g. `@delete in 1 minute`, `@delete in 5 days`, etc.

* `<number>` must be a positive integer
* `<unit>` must be one of: `second`, `minute`, `hour`, `day`, `week`, `month`, `year`, or the plurals of each

If there are multiple "delete commands" found in the text of an item, the last command found in the text is honored.

If an item is edited, the delete commands are re-evaluated. In other words, if there was a delete command originally and it was removed in the edit, the deletion will not occur. Likewise, if there was not a delete command originally but one is added, it will take effect. Finally, if the delete command is changed in an edit, the old one will no longer take effect but the new one will.

Making an item ephemeral does not cost the stacker any extra sats (see discussion on this PR for context).

Note that since link posts don't have a text property, they cannot be ephemeral.

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -16,7 +16,7 @@ import { parse } from 'tldts'
 import uu from 'url-unshort'
 import { advSchema, amountSchema, bountySchema, commentSchema, discussionSchema, jobSchema, linkSchema, pollSchema, ssValidate } from '../../lib/validate'
 import { sendUserNotification } from '../webPush'
-import { defaultCommentSort, isJob } from '../../lib/item'
+import { defaultCommentSort, isJob, deleteItemByAuthor, getDeleteCommand, hasDeleteCommand } from '../../lib/item'
 import { notifyItemParents, notifyUserSubscribers, notifyZapped } from '../../lib/push-notifications'
 import { datePivot } from '../../lib/time'
 
@@ -654,21 +654,7 @@ export default {
         throw new GraphQLError('item does not belong to you', { extensions: { code: 'FORBIDDEN' } })
       }
 
-      const data = { deletedAt: new Date() }
-      if (old.text) {
-        data.text = '*deleted by author*'
-      }
-      if (old.title) {
-        data.title = 'deleted by author'
-      }
-      if (old.url) {
-        data.url = null
-      }
-      if (old.pollCost) {
-        data.pollCost = null
-      }
-
-      return await models.item.update({ where: { id: Number(id) }, data })
+      return await deleteItemByAuthor({ models, id, item: old })
     },
     upsertLink: async (parent, { id, hash, hmac, ...item }, { me, models, lnd }) => {
       await ssValidate(linkSchema, item, { models, me })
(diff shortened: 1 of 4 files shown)


huumn:
I'm torn on charging for ephemeral items. Happy to be to be outvoted. It seems kind of big brother-y to me. @ekzyis?

ekzyis:
> I'm torn on charging for ephemeral items. Happy to be to be outvoted. It seems kind of big brother-y to me and complicates things more than we might want. Every fee has to be communicated as a receipt. 
> 
> @ekzyis?

Make it free and let's see!

I agree, it's big brother-y. We're not trying to prevent any abuse like spam here. Charging fees for deletes makes us less privacy focused.

So I changed my mind on that, especially because you mentioned yesterday that we charge sats to prevent abuse, not to make them pay for storage costs since they're negligible :)

SatsAllDay:
I'm good with making it free. It will make the changes a lot simpler. Just for the sake of discussion, I thought it made sense to charge because deleting messages kinda lowers the quality of the experience for other stackers, so increasing the reward pool felt like a nice compensation for that. Plus, manual deletes are still free, this was just charging a fee to have it handled automatically. But if we think free is the right first approach, I'll make the necessary adjustments!

SatsAllDay:
https://media3.giphy.com/media/xT9DPr4VjeCgeiLoMo/giphy.gif

Will remove the fee portion. This PR is probably pretty close to being done, then.

api/resolvers/item.js
```diff
@@ -1066,21 +1075,39 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
...
+
+const enqueueDeletionJob = async (item, models, incomingCommand) => {
+  const deleteCommand = incomingCommand ?? getDeleteCommand(item.text)
+  if (deleteCommand) {
+    await models.$queryRawUnsafe(`
+      INSERT INTO pgboss.job (name, data, startafter)
+      VALUES ('deleteItem', jsonb_build_object('id', ${item.id}), now() + interval '${deleteCommand.number} ${deleteCommand.unit}s');`)
+  }
+}
+
```
SatsAllDay, line 1110: @huumn would you rather me try to incorporate the pgboss job sql queries into the `create_item` and `update_item` SQL functions, so they are part of the same serialized transaction, or can they be done separate from those function calls, similar to how we do `createMentions` now? That'll dictate the direction I take this PR a bit, so I wanted to get your feedback on that

huumn: Let's go with a mention-style approach for now. Those postgres functions are getting bloated.

PR #576, opened 2023-10-19
Image uploads

ekzyis:
[To not lose my code again](https://stacker.news/items/285911), here is the WIP code.

This PR adds the following button:

_button to upload images_
2023-10-19-194613_617x219_scrot

The button opens a file explorer. When you've selected an image, it shows up beneath the text input for you to insert into the text:

_unsubmitted images show up beneath text input_

2023-10-19-195255_612x264_scrot

This works by requesting a presigned AWS S3 URL on image select, immediately uploading the image on the client to S3 (similar to how avatar upload works) and then storing the URL to the image in a state array to keep track of _submitted_ and _unsubmitted_ image URLs.

An image URL is _submitted_ if it's included in an item. We can check if an image URL is included in an item by looking at the `Upload.itemId` column of the row with `Upload.id = <s3_key>`. This column is populated by the worker during the imgproxy job. If it encounters a S3 URL in an item, it parses the S3 key from the URL and updates the corresponding upload row with the id of the item that the imgproxy job is currently processing:

```javascript
if (url.startsWith(AWS_S3_URL)) {
  const s3Key = url.split('/').pop()
  await models.upload.update({ data: { itemId: Number(id) }, where: { id: Number(s3Key) } })
}
```

`Upload.itemId` therefore also allows us to show the stacker which images they haven't used yet. This is implemented with a new resolver `User.images`:

```javascript
images: async (user, { submitted }, { me, models }) => {
  if (!me) return null
  const uploads = await models.upload.findMany({
    where: {
      userId: me.id,
      itemId: submitted !== undefined ? submitted ? { not: null } : null : undefined
    }
  })
  return uploads
}
```

To not run unnecessary database queries by adding this simply to the `ME` query which polls the database every second, I added a new provider `ImagesProvider` which fetches this once on session init.

When posting something (comment, discussion, link, ...), I used the `onCompleted` callback to update the state on the client.

That's the current state. Simple showcase:

https://github.com/stackernews/stacker.news/assets/27162016/a82e0ace-f184-4c66-98a7-3fd91648101b

---

TODO:

- [x] remove unique item id constraint on table `Upload`
- [x] make sure index was not required for any query:
_I searched for `Upload` (case-sensitive) and `"Upload"` (since the table might be mentioned using `"`) and `models.update` but I haven't found anything of interest. So I think this table wasn't used much so far._

~~There can be multiple uploads (= multiple images) per item now. However, this requires a data migration since jobs currently rely on there being only exactly one upload per item via `Item.uploadId`~~

[I have decided to only drop the unique index](https://github.com/stackernews/stacker.news/pull/576#issuecomment-1771864742)

<details>
<summary>desired schema diff</summary>

```diff
$ g d
diff --git a/prisma/schema.prisma b/prisma/schema.prisma
index d012499..a94a1c3 100644
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -161,7 +161,7 @@ model Upload {
   size      Int
   width     Int?
   height    Int?
-  itemId    Int?     @unique(map: "Upload.itemId_unique")
+  itemId    Int?
   userId    Int
   item      Item?    @relation(fields: [itemId], references: [id])
   user      User     @relation("Uploads", fields: [userId], references: [id], onDelete: Cascade)
@@ -252,7 +252,6 @@ model Item {
   company            String?
   weightedVotes      Float                 @default(0)
   boost              Int                   @default(0)
-  uploadId           Int?
   pollCost           Int?
   paidImgLink        Boolean               @default(false)
   commentMsats       BigInt                @default(0)
@@ -284,7 +283,7 @@ model Item {
   PollOption         PollOption[]
   PollVote           PollVote[]
   ThreadSubscription ThreadSubscription[]
-  upload             Upload?
+  uploads            Upload[]
   User               User[]
   itemForwards       ItemForward[]
```

</details>

_by the way, afaict, `Upload.itemid` was not used before. Possibly just a relict of a previous approach which wasn't pursued further but replaced with `Item.uploadId`?_

- [x] implement fees to receive presigned URLs
- [x] ~~refund fees if image is deleted (which is only possible when it wasn't submitted / used in an item yet)~~

As the code is currently, stackers could upload infinite amount of images to S3 for free. I thought about giving stackers some free images per day such that they don't have to think if uploading an image is worth the sats all the time. However, while writing this and trying to explain my reasoning, I realized it might actually be easier to code if every upload costs sats instead of the other way around because of less edge cases, lol

So image uploading could work similar to fee escalation for posting, just with a different slope. It could even be a staircase pattern: first 5 images per day cost 1 sat, then next 5 cost 10 sat, then next 5 100 etc.

The fee should also depend on the size however. So your first 5 images _OR_ your first 5 MB cost 1 sat per image, whatever is hit first. If you reached 5 MB, you get escalated to the next level which is 10 sat (the image you currently want to upload should already be included in this calculation so you can't upload 100 MB and still pay 1 sat if you haven't uploaded 5 MB yet - a common programming mistake)

However, avatar uploading should still be free and thus an exception imo. This brings me to the next point:

- [x] use same key for avatars such that stackers can upload avatars still for free but we only store the latest one
  - [x] follow-up: invalidate cache for avatars after update
- [x] drag and drop images
- [x] camera support
- [x] ~~use trash instead of cross icon for delete~~
- [x] also delete image in S3 on delete by user
but make sure delete only works if the URLs weren't submitted already! else there will be 404s for images in items
- [x] ~~[use FileReader API to show progress bar while uploading](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/progress_event)~~
- [x] ~~make sure anons have to pay per image (1000 sats per image?)
but submitted items should only show within a session, not across all anon sessions. but this opens up bad UX if anon pays for image but then doesn't include it in an item (= unsubmitted) and then forgets the image URL. mhhh :thinking:~~ 
- [x] refactor: unify `<Upload>` in _components/upload.js_ and `<ImageUpload>` in _components/image.js_
- [x] allow anons to use image uploads
---

New TODOs because of [this conclusion](https://github.com/stackernews/stacker.news/pull/576#issuecomment-1773881193):

- [x] move from upfront images fees for presigned URLs to image fees during item creation
- [x] remove client logic around unsubmitted images since if there are no more fees for presigned URLs, we don't need to keep track of unsubmitted images on the client - they can just upload another image if they lost an image link
- [x] periodically delete unused images (= images which weren't used in an item)
- [x] show image fees in frontend
- [x] ~~refactor image queries into own wrapper like `serializeInvoiceable`: `serializeImages`?~~ _image queries are now run inside `create_item` and `update_item`_
- [x] allow uploading multiple images at once
- [x] ~~run image queries in same transaction as `create_item` ... need to access item id somehow though~~ _image queries are now run inside `create_item` and `update_item`_
- [x] make sure fees show in frontend are in sync with actual fees calculated in backend

---

Description how [new approach](https://github.com/stackernews/stacker.news/pull/576#issuecomment-1773881193) works:

- Stackers can upload multiple files using presigned URLs. Getting these presigned URLs is free.
- When the upload starts, `[Uploading <file>...]()` is pasted into the text input like in Github
- When the upload finishes, this is replaced with the image URL
- Additionally, the frontend fetches image fee info from backend to show any image fees to the user
- 10 MB per 24 hours are free; anons always pay 100 sats per image. See postgres function `image_fees_info`.
- On blur, the frontend also fetches image fee info since image URLs might have been deleted
- On item creation, `image_fees_info` is called inside `create_item` to deduct image fees and update rows in the `Upload` table to mark images as paid (images can be reused in multiple items but only need to be paid once)
- On item update, `image_fees_info` is called inside `update_item` to deduct image fees
- `image_fees_info` knows which images were paid and which not using the `Upload.paid` column
- if you drag files into the markdown text input, the border is highlighted in blue


ekzyis:
>  - [ ] remove unique item id constraint on table Upload
> 
> There can be multiple uploads (= multiple images) per item now. However, this requires a data migration since jobs currently rely on there being only exactly one upload per item via Item.uploadId

I've decided to just drop the unique index. So no other changes required.

This might look weird in the schema now that an item can have a dedicated upload id _and_ multiple uploads can point to it. The row in `Upload` with `Upload.id = Item.uploadId` might not even point to the item via `itemId`.

But this was already the case.

So I think this is ok to guarantee 100% backwards compatibility and if we don't like that, we can change that in another PR.

ekzyis:
Regarding:

> - [ ] implement fees to receive presigned URLs
> 
> As the code is currently, stackers could upload infinite amount of images to S3 for free. I thought about giving stackers some free images per day such that they don't have to think if uploading an image is worth the sats all the time. However, while writing this and trying to explain my reasoning, I realized it might actually be easier to code if every upload costs sats instead of the other way around because of less edge cases, lol
> 
> So image uploading could work similar to fee escalation for posting, just with a different slope. It could even be a staircase pattern: first 5 images per day cost 1 sat, then next 5 cost 10 sat, then next 5 100 etc.
>
> The fee should also depend on the size however. So your first 5 images OR your first 5 MB cost 1 sat per image, whatever is hit first. If you reached 5 MB, you get escalated to the next level which is 10 sat (the image you currently want to upload should already be included in this calculation so you can't upload 100 MB and still pay 1 sat if you haven't uploaded 5 MB yet - a common programming mistake)

[We pay $0.023 per GB per month (for the first 50 TB).](https://aws.amazon.com/s3/pricing/)
$0.023 are 80 sats according to https://www.kraken.com/learn/satoshi-to-usd-converter (October 20, 2023).
This rounded up are 100 sats.
So we pay roughly 100 sats per GB per month (at the current exchange rate).
Per megabyte, this is 100/1024 = ~0.1 sats / MB
Since we will store the images indefinitely, we can let stackers pay for 12 months upfront.
So this makes ~1 sat / MB.

Still quite cheap. 

I also realized that it doesn't make sense to limit amount of images. Only the size should matter.

For better numbers and to limit storage per stacker, I am therefore thinking about using the following algorithm to calculate cost per image:

```javascript
// factor for bytes to megabyte
const MB = 1024 * 1024
// factor for msats to sats
const SATS = 1000

async function uploadCosts (models, userId, photoId, size) {
  let { _sum: { size: sumSize } } = await models.upload.aggregate({
    _sum: { size: true },
    where: { userId, createdAt: { gt: datePivot(new Date(), { days: -1 }) } }
  })
  // assume the image was already uploaded in the calculation
  sumSize += size
  if (sumSize <= 5 * MB) {
    return 0 * SATS
  }
  if (sumSize <= 10 * MB) {
    return 10 * SATS
  }
  if (sumSize <= 25 * MB) {
    return 100 * SATS
  }
  if (sumSize <= 100 * MB) {
    return 1000 * SATS
  }
  // check return value and throw error on -1?
  // or maybe not use -1 since in case of a bug, we might actually pay stackers to upload images, lol
  return -1 
}
```

huumn:
> I also realized that it doesn't make sense to limit amount of images. Only the size should matter.

We're not charging to make stackers pay for the hosting costs. We are charging to prevent abuse. The costs of even large images is negligible. 

No one wants to think about the size of the image they're uploading.

huumn:
I'm not sure if you get re-notified on edits, but I edited my above comment to be like "ignore me, never mind."

ekzyis:
No, I don't get new notifications per edit and that's good because else you would have gotten a ton of notifications from me editing all my posts on Github, haha

Even though you said, I should ignore what you said, I want to respond anyway since I think these are good points you mentioned and I think it's good to have them written down even if we Already agree:

> We're not charging to make stackers pay for the hosting costs. We are charging to prevent abuse. The costs of even large images is negligible.
>
> No one wants to think about the size of the image they're uploading.

The question I'm trying to answer is when it's starts to resemble abuse. Is it 25MB per day? 100MB per day? 1GB per day? Or something completely different? 

Even ignoring that there is no protection about just spinning up new accounts.

Was also wondering about anon image uploads. That's going to be the most trickiest UX wise I think since they shouldn't see each other's unsubmitted images vs. what if they pay but then forget to submit the image. The former is more for UX reasons than privacy reasons though since the images are public anyway.

ekzyis:
> Are there good reasons to not do it that way

The only reason I'm doing it this way is because if they pay, they pay immediately, not only when they submitted an image.

If we do it like Github, stackers might accidentally delete the link. Then they wasted sats because now they don't know anymore for which link they paid.

So I show them all images they haven't submitted yet including the ability to get a refund if they delete the image. 

And I used pay immediately for presigned URL since else it seemed unnecessary complex to detect if URLs were just not submitted yet or if they are just abusing our storage without paying for it.

So I could implement the payment logic within `create_item` but it's harder to secure since images were already uploaded.

And I wanted to keep client upload with presigned URLs to not diverge too far from what we're already doing with avatars.

huumn:
I see. I'm not sure if what I was thinking makes sense in light of some of these problems, but here's how I always thought it would work:

1. you're writing a post/comment 
2. add an image (works like github)
3. when you submit, we parse the text and associate any sn uploaded images with the item
4. after 24 hours (or some other period of time), if we have any images hosted that aren't associated with a user/item, then we delete it

We could even be more or less aggressive with the deletion timing depending on other factors like how often the stacker does this, or how much trust they have.

> So I could implement the payment logic within create_item but it's harder to secure since images were already uploaded.

Can you elaborate on the difficulties of securing images this way?

To be clear, the way you've done it is really nice. It's just trading some UX for other things and I want to make sure we really need the other things.

ekzyis:
> I see. I'm not sure if what I was thinking makes sense in light of some of these problems, but here's how I always thought it would work:
> 
> 1. you're writing a post/comment
> 2. add an image (works like github)
> 3. when you submit, we parse the text and associate any sn uploaded images with the item
> 4. after 24 hours (or some other period of time), if we have any images hosted that aren't associated with a user/item, then we delete it

That's also how I imagined it to work initially and I actually would love to do it like this.

But then I thought about how I would abuse this. I would upload as much images as I can with zero _variable_ costs to me. This ignores bandwidth and other factors since most likely, an attacker already pays for it with a flat fee; so it's independent of how much they exactly use these resources. (You could potentially even argue that _not_ using these resources to attack is actually the cost because of opportunity costs, lol)

Yes, we will delete them at some point if they weren't used to post on SN, but the damage might already have been done in forms of storage costs or known and unknown unknowns.

And yes, this attack might be more theoretically since what exactly has the adversary to gain? Being able to host a lot of images for a limited amount of time? I am not sure if that is a valuable goal, but I wanted to err on the side of caution here. Basically for "Fear of the Unknown"-reasons. So if it's possible to completely shut down such attacks by introducing variable costs, why not do it?

But thanks to the discussion with you, it's clear now how much I am trading UX for security.

UX problems if we introduce immediate variable costs (pay per presigned URL):

- we now need to add refunds if we not want to be a simple hosting service but only want to allow image uploads to post them on SN. solution: show unsubmitted images to delete for a refund or still use them
- how to deal with unsubmitted images of anons? we don't want to show unsubmitted images from anons to every anon for UX reasons (not necessary for privacy reasons).
- paying something before you post is a new pattern which we have to explain
- unsubmitted images show up below every text input, so they always show up beneath text input to reply to the root item + if you click on `reply` in a comment

Also, since we want to allow refunds (imo), this actually introduces a new, potentially even more serious risk: we add code which increases the balance of users. So instead of potentially being able to storage a lot of images, in case of a bug, it would now be possible to steal funds, lol 

So I think after having explored this "pay per presigned URL" approach enough, I would also prefer the "Github way". But just want to mention that:

> With you current UX, it's a two step process of (1) uploading then (2) linking. Github's is a one step process.

Is not inherent to the current UX. I could immediately link them on upload _and_ show them as unsubmitted. It's just work I didn't put in yet and I wasn't 100% sure yet if it's worth the work of more cursor tracking code (I think this means we need to refactor some stuff so we can reuse existing cursor tracking code to paste content at the cursor easily). But funnily, my showcase actually demonstrates the UX problem quite well since I clicked on "reply" but I didn't link the image yet, haha

So I'll use this approach now: Stackers don't have to pay and we can now just paste the image link without the UX risk of stackers wasting their sats. If they accidentally delete the link, they can just do a new upload for free.

And then during `create_item`, we deduct image fees if applicable.

> Can you elaborate on the difficulties of securing images this way?

Only two reasons:
- we can't enforce variable costs this way
- I need to update existing code rather than write new code. I prefer to write code in layers so existing, battle-tested code can work as it did before. And putting this logic explicit for uploaded images into `create_item` seemed unnecessarily more complex to me than just dealing with all this image upload stuff in a completely different layer. Which makes it again less secure because of more complexity.

huumn:
Your initial approach is really nice and demonstrates your ability to think through all the adversary's moves. Regardless of the exact approach we take (here or elsewhere) we *need* to forever be capable of doing this exercise. 

In our efforts to starve our enemies though, we have to be careful to not starve our allies.

> So I'll use this approach now: Stackers don't have to pay and we can now just paste the image link without the UX risk of stackers wasting their sats. If they accidentally delete the link, they can just do a new upload for free.
> 
> And then during create_item, we deduct image fees if applicable.

Cool, I do think this is the right approach. I think we can make the incentives for abuse low enough that we don't have to be too concerned about taking upfront variable fees.

If we're wrong, we can reconsider.

---------------

Whenever introducing fees come up, I think of this coin operated park bench

bGXgQQfV6juS_KueETppqhaBKBSHYdzvOMcvNqA_vIk

Adding fees should strictly increase the UX of SN. If it's not clear we're making SN better by introducing a fee, I'd rather we get some real life experience with the fee-free version (which should make it clear if the fee is needed). In the general case, fees make SN's UX worse so it's worth being very selective with them.

I should probably start writing some of these principles down.

ekzyis:
This is done now. Only testing and code cleanup is left (see PR description).

1. _upload multiple files at once; anon always pays 100 sats per image:_

https://github.com/stackernews/stacker.news/assets/27162016/8bab1c06-017c-4bad-95f6-c8f708dde889

2. _stackers have 10MB free upload, then every image starts to cost 10 sats:_

https://github.com/stackernews/stacker.news/assets/27162016/e125370b-140e-4b66-ae74-5a0c912d4668

3. _during editing, only new images (images which weren't paid yet) are considered (also, UX with updating image fees on blur is not ideal):_

https://github.com/stackernews/stacker.news/assets/27162016/813655af-3bd2-4406-8c4f-c5b8d5973c47

4. _drag and drop is supported:_

https://github.com/stackernews/stacker.news/assets/27162016/9c809afc-781c-40f9-836a-6cc8e6b3109c

5. _camera is automatically supported on mobile file explorer_

6. _unused images are regularly deleted using `pgboss.schedule`_

I think main areas which can be improved is the UX around image fees.

I currently don't show in the receipts when fees per image will start or rise. I made image fees dependent on size (so a specific size must be hit before the next level of fees per image is reached) since I think this way, most stackers won't exhaust their free quota per day compared to a free quota of images based on image amount (and not size). Also, if we limit per image amount, using fees as defense against abuse doesn't really work.

Also, I update image fees in the frontend on upload (images were added) and on text input blur (images might have been removed). Checking if images were removed on blur is not ideal. Maybe I should update fees on every text change but debounced?

api/resolvers/item.js
```diff
@@ -1120,14 +1125,19 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
...
+  const { queries: imgQueriesFn, fees: imgFees } = await imageFeesFromText(item.text, { models, me })
+  const enforceFee = (me ? undefined : (item.parentId ? ANON_COMMENT_FEE : (ANON_POST_FEE + (item.boost || 0)))) + imgFees
   item = await serializeInvoicable(
     models.$queryRawUnsafe(
       `${SELECT} FROM create_item($1::JSONB, $2::JSONB, $3::JSONB, '${spamInterval}'::INTERVAL) AS "Item"`,
       JSON.stringify(item), JSON.stringify(fwdUsers), JSON.stringify(options)),
     { models, lnd, hash, hmac, me, enforceFee }
   )
 
+  // TODO run image queries in same transaction as create_item
```
ekzyis, line 1137: Oh, forgot about this TODO. This is important since else, the user might get an error but the item actually was created. Added TODO to PR description.

api/resolvers/item.js
```diff
@@ -1120,14 +1125,19 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
...
+  const imgQueries = imgQueriesFn(item.id)
+  if (imgQueries.length > 0) await serialize(models, ...imgQueries)
```
ekzyis, line 1139: This should be refactored to something similar like `serializeInvoiceable`.

components/fee-button.js
```diff
@@ -79,46 +79,43 @@ export default function FeeButton ({ parentId, hasImgLink, baseFee, ChildButton,
...
-            </tr>
-            <tr>
-              <td>- {numWithUnits(paidSats, { abbreviate: false })}</td>
-              <td align='right' className='font-weight-light'>already paid</td>
-            </tr>
-          </>}
+        <tr>
+          <td>{numWithUnits(0, { abbreviate: false })}</td>
+          <td align='right' className='font-weight-light'>edit fee</td>
+        </tr>
```
ekzyis, line 113: Added `0 sats edit fee` so I can use `+ X sats imageFees` ... Also, I might should not have removed the usage of `paidSats` in the edit receipt.

components/image.js
```diff
@@ -132,3 +136,95 @@ export default function ZoomableImage ({ src, srcSet, ...props }) {
...
+        onChange={async (e) => {
+          const fileList = e.target.files
+          for (const file of Array.from(fileList)) {
+            if (UPLOAD_TYPES_ALLOW.indexOf(file.type) === -1) {
+              toaster.danger(`image must be ${UPLOAD_TYPES_ALLOW.map(t => t.replace('image/', '')).join(', ')}`)
+              continue
+            }
+            await s3Upload(file)
+            // TODO find out if this is needed and if so, why (copied from components/upload.js)
+            e.target.value = null
```
ekzyis, line 221: I copied this from _components/upload.js_. @huumn, you know what is the purpose of this? You added this in 29fb37b7632960871e41c4055f84b06d8911b997.

I also need to refactor this code with that code anyway.

components/job-form.js
```diff
@@ -47,6 +47,7 @@ export default function JobForm ({ item, sub }) {
         location: $location, remote: $remote, text: $text,
         url: $url, maxBid: $maxBid, status: $status, logo: $logo, hash: $hash, hmac: $hmac) {
         id
+        text
```
ekzyis, line 50: This and the other similar changes are no longer needed. Added them in e129d85a but that was for the old approach.

lib/constants.js
```diff
@@ -6,8 +6,9 @@ export const SUBS_NO_JOBS = SUBS.filter(s => s !== 'jobs')
 export const NOFOLLOW_LIMIT = 1000
 export const BOOST_MULT = 5000
 export const BOOST_MIN = BOOST_MULT * 5
-export const UPLOAD_SIZE_MAX = 2 * 1024 * 1024
+export const UPLOAD_SIZE_MAX = 10 * 1024 * 1024
```
ekzyis, line 9: Increased size so camera upload works since most phones take higher quality pictures.

Also, maybe we can drop the upload size completely since we have fees now?

ekzyis: We can set a different limit for avatars since avatars are free

api/resolvers/item.js
huumn: Prisma has interactive txs now: https://www.prisma.io/docs/concepts/components/prisma-client/transactions#interactive-transactions

We can begin doing this but be mindful that each query will require RTT to the db and we are using serializable txs ... interactive txs will make the tx slow to complete and will increase item creation/update failure probability. 

Depending on the nature of the image queries, it might be worth doing them inside `create_item` and `update_item` ... images are kind of fundamental so it seems well worth adding them.

components/image.js
huumn: I'm not sure, it might be dead code. Worth testing with/without. If I had to guess, it might be clearing the file field so they can upload another image if they don't like the one they just uploaded.

lib/constants.js
huumn: Dealer's choice.

On GrepMed, we added a bunch of limits because people would occasionally upload really odd images. The two that I remember were

1. huge image data sizes
2. normal image data size with massive image dimensions

api/resolvers/item.js
ekzyis: > Depending on the nature of the image queries, it might be worth doing them inside create_item and update_item ... images are kind of fundamental so it seems well worth adding them.

I'll do this now. I also put the calculation of image fees into a postgres function now (not pushed yet).

prisma/migrations/20231026154807_image_fees_info/migration.sql
```diff
@@ -0,0 +1,241 @@
...
+            COALESCE(COUNT(id) FILTER(WHERE paid = 'f' AND id = ANY(upload_ids)), 0)::INTEGER AS "nUnpaid"
+        FROM "Upload"
+        WHERE "Upload"."userId" = user_id
+    ) uploadinfo;
+    RETURN;
+END;
+$$;
+
+-- add image fees
+CREATE OR REPLACE FUNCTION create_item(
```
ekzyis, line 42:
```diff
diff --git a/exclude/create_item.sql b/exclude/create_item.sql
index dcd646d..149b92d 100644
--- a/exclude/create_item.sql
+++ b/exclude/create_item.sql
@@ -1,5 +1,5 @@
 CREATE OR REPLACE FUNCTION create_item(
-    jitem JSONB, forward JSONB, poll_options JSONB, spam_within INTERVAL)
+    jitem JSONB, forward JSONB, poll_options JSONB, spam_within INTERVAL, upload_ids INTEGER[])
 RETURNS "Item"
 LANGUAGE plpgsql
 AS $$
@@ -23,6 +23,13 @@ BEGIN
     ELSE
         cost_msats := 1000 * POWER(10, item_spam(item."parentId", item."userId", spam_within));
     END IF;
+
+    -- add image fees
+    IF upload_ids IS NOT NULL THEN
+        cost_msats := cost_msats + (SELECT "nUnpaid" * "imageFeeMsats" FROM image_fees_info(item."userId", upload_ids));
+        UPDATE "Upload" SET paid = 't' WHERE id = ANY(upload_ids);
+    END IF;
+
     -- it's only a freebie if it's a 1 sat cost, they have < 1 sat, and boost = 0
     freebie := (cost_msats <= 1000) AND (user_msats < 1000) AND (item.boost IS NULL OR item.boost = 0);
```

prisma/migrations/20231026154807_image_fees_info/migration.sql
```diff
@@ -0,0 +1,241 @@
...
+    -- schedule imgproxy job
+    INSERT INTO pgboss.job (name, data, retrylimit, retrybackoff, startafter)
+    VALUES ('imgproxy', jsonb_build_object('id', item.id), 21, true, now() + interval '5 seconds');
+
+    RETURN item;
+END;
+$$;
+
+-- add image fees
+CREATE OR REPLACE FUNCTION update_item(
```
ekzyis, line 154: I added these changes to `update_item`:

```diff
diff --git a/exclude/update_item.sql b/exclude/update_item.sql
index b3f3654..d066046 100644
--- a/exclude/update_item.sql
+++ b/exclude/update_item.sql
@@ -1,10 +1,11 @@
 CREATE OR REPLACE FUNCTION update_item(
-    jitem JSONB, forward JSONB, poll_options JSONB)
+    jitem JSONB, forward JSONB, poll_options JSONB, upload_ids INTEGER[])
 RETURNS "Item"
 LANGUAGE plpgsql
 AS $$
 DECLARE
     user_msats INTEGER;
+    cost_msats BIGINT;
     item "Item";
     select_clause TEXT;
 BEGIN
@@ -12,6 +13,23 @@ BEGIN

     item := jsonb_populate_record(NULL::"Item", jitem);

+    SELECT msats INTO user_msats FROM users WHERE id = item."userId";
+    cost_msats := 0;
+
+    -- add image fees
+    IF upload_ids IS NOT NULL THEN
+        cost_msats := cost_msats + (SELECT "nUnpaid" * "imageFeeMsats" FROM image_fees_info(item."userId", upload_ids));
+        UPDATE "Upload" SET paid = 't' WHERE id = ANY(upload_ids);
+    END IF;
+
+    IF cost_msats > 0 AND cost_msats > user_msats THEN
+        RAISE EXCEPTION 'SN_INSUFFICIENT_FUNDS';
+    ELSE
+        UPDATE users SET msats = msats - cost_msats WHERE id = item."userId";
+        INSERT INTO "ItemAct" (msats, "itemId", "userId", act)
+        VALUES (cost_msats, item.id, item."userId", 'FEE');
+    END IF;
+
     IF item.boost > 0 THEN
         UPDATE "Item" SET boost = boost + item.boost WHERE id = item.id;
         PERFORM item_act(item.id, item."userId", 'BOOST', item.boost);
```

prisma/migrations/20231026154807_image_fees_info/migration.sql
ekzyis, line 1: I added these changes to `create_item`:

```diff
diff --git a/exclude/create_item.sql b/exclude/create_item.sql
index dcd646d..149b92d 100644
--- a/exclude/create_item.sql
+++ b/exclude/create_item.sql
@@ -1,5 +1,5 @@
 CREATE OR REPLACE FUNCTION create_item(
-    jitem JSONB, forward JSONB, poll_options JSONB, spam_within INTERVAL)
+    jitem JSONB, forward JSONB, poll_options JSONB, spam_within INTERVAL, upload_ids INTEGER[])
 RETURNS "Item"
 LANGUAGE plpgsql
 AS $$
@@ -23,6 +23,13 @@ BEGIN
     ELSE
         cost_msats := 1000 * POWER(10, item_spam(item."parentId", item."userId", spam_within));
     END IF;
+
+    -- add image fees
+    IF upload_ids IS NOT NULL THEN
+        cost_msats := cost_msats + (SELECT "nUnpaid" * "imageFeeMsats" FROM image_fees_info(item."userId", upload_ids));
+        UPDATE "Upload" SET paid = 't' WHERE id = ANY(upload_ids);
+    END IF;
+
     -- it's only a freebie if it's a 1 sat cost, they have < 1 sat, and boost = 0
     freebie := (cost_msats <= 1000) AND (user_msats < 1000) AND (item.boost IS NULL OR item.boost = 0);
```

and these changes to `update_item`:

```diff
diff --git a/exclude/update_item.sql b/exclude/update_item.sql
index b3f3654..d066046 100644
--- a/exclude/update_item.sql
+++ b/exclude/update_item.sql
@@ -1,10 +1,11 @@
 CREATE OR REPLACE FUNCTION update_item(
-    jitem JSONB, forward JSONB, poll_options JSONB)
+    jitem JSONB, forward JSONB, poll_options JSONB, upload_ids INTEGER[])
 RETURNS "Item"
 LANGUAGE plpgsql
 AS $$
 DECLARE
     user_msats INTEGER;
+    cost_msats BIGINT;
     item "Item";
     select_clause TEXT;
 BEGIN
@@ -12,6 +13,23 @@ BEGIN

     item := jsonb_populate_record(NULL::"Item", jitem);

+    SELECT msats INTO user_msats FROM users WHERE id = item."userId";
+    cost_msats := 0;
+
+    -- add image fees
+    IF upload_ids IS NOT NULL THEN
+        cost_msats := cost_msats + (SELECT "nUnpaid" * "imageFeeMsats" FROM image_fees_info(item."userId", upload_ids));
+        UPDATE "Upload" SET paid = 't' WHERE id = ANY(upload_ids);
+    END IF;
+
+    IF cost_msats > 0 AND cost_msats > user_msats THEN
+        RAISE EXCEPTION 'SN_INSUFFICIENT_FUNDS';
+    ELSE
+        UPDATE users SET msats = msats - cost_msats WHERE id = item."userId";
+        INSERT INTO "ItemAct" (msats, "itemId", "userId", act)
+        VALUES (cost_msats, item.id, item."userId", 'FEE');
+    END IF;
+
     IF item.boost > 0 THEN
         UPDATE "Item" SET boost = boost + item.boost WHERE id = item.id;
         PERFORM item_act(item.id, item."userId", 'BOOST', item.boost);
```

lib/constants.js
ekzyis: Using this now:

```
export const UPLOAD_SIZE_MAX = 25 * 1024 * 1024
export const UPLOAD_SIZE_MAX_AVATAR = 5 * 1024 * 1024
```

components/image.js
ekzyis: [You were right, it's to clear the file field](https://bobbyhadz.com/blog/react-reset-file-input#reset-a-file-input-in-react)

components/fee-button.js
ekzyis: Example receipts:

_reply:_

2023-10-27-031624_509x218_scrot

_edit:_

2023-10-27-031644_506x188_scrot



I realized `paidSats` wasn't used before since `addImgLink` was not used.

So I think it's not a breaking UX change. I think it's fine to mention that editing itself does not cost anything but fees were added because of images.

What might be confusing though is why image suddenly cost something since they don't cost anything until 10 MB were uploaded within the last 24 hours. There is no mention of this in the receipt.

I played a bit around with a [ProgressBar](https://react-bootstrap.netlify.app/docs/components/progress/), but that also didn't feel right. Maybe mentioning it in the FAQ is enough? Or a info box in the receipt? But then we have a info box inside a info box, lol.

ekzyis:
> Checking if images were removed on blur is not ideal. Maybe I should update fees on every text change but debounced?

Using `debounce` with 1 second debouncing. The only UX problem is now that the user might hit reply before the fees are updated and then more fees might be deducted in the backend without the user knowing in advance.

We could disable the reply button until the image fee info was updated; similar to how the dupe check works for links (`post` is disabled until dupes are found or 3 sec timeout is hit).

But I think disabling the reply button after one second for every text change is going to be annoying.

Thinking about a pure client-side implementation of the fee calculation so we don't need to do a request at all. We can probably assume that the user is not typing multiple items at once so we can assume that during one "item session", the state of unpaid images does not change. Paid images always stay paid.

ekzyis:
> Thinking about a pure client-side implementation of the fee calculation so we don't need to do a request at all. We can probably assume that the user is not typing multiple items at once so we can assume that during one "item session", the state of unpaid images does not change. Paid images always stay paid.

Putting this out of draft since I believe this is not a blocking UX thing. Can be done in a follow-up PR imo.

huumn:
1. Are you still making changes to this? I can hold off on reviewing
2. Do you want me to nitpick and review or should I just make the changes myself?

api/resolvers/upload.js
```diff
@@ -1,62 +1,50 @@
...
+        width,
+        height,
+        userId: me?.id || ANON_USER_ID,
+        // avatar uploads are always free
+        paid: avatar && !!me ? undefined : false
+      }
+
+      let uploadId
+      // avatar uploads overwrite the previous avatar
+      if (avatar && photoId) uploadId = photoId
```
huumn, line 38: This is all kind of hard to follow. Control changes for avatars in 3 different places. I suspect this could be made clearer

components/form.js
```diff
@@ -209,6 +230,28 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, setH
...
+  const onDragLeave = useCallback((e) => {
+    setDragStyle(null)
+  }, [setDragStyle])
+
+  useEffect(debounce(() => {
+    const text = innerRef?.current.value
+    const s3Keys = text ? [...text.matchAll(AWS_S3_URL_REGEXP)].map(m => Number(m[1])) : []
+    updateImageFeesInfo({ variables: { s3Keys } })
+  }, 1000), [innerRef?.current?.value])
+
```
huumn, line 254: Is there a way to do this without relying on a side effect of a reference value changing? 

Might it be simpler to call this in `onChange`?

components/invoice.js
```diff
@@ -232,8 +232,9 @@ export const useInvoiceable = (onSubmit, options = defaultOptions) => {
   // this function will be called before the Form's onSubmit handler is called
   // and the form must include `cost` or `amount` as a value
   const onSubmitWrapper = useCallback(async (formValues, ...submitArgs) => {
-    let { cost, amount } = formValues
+    let { cost, imageFeesInfo, amount } = formValues
     cost ??= amount
+    if (imageFeesInfo?.totalFees) cost += imageFeesInfo.totalFees
```
huumn, line 237: Should this function used throughout the code be aware of images at all? Could the calling context do this calculation instead?

components/user-header.js
```diff
@@ -71,14 +71,22 @@ function HeaderPhoto ({ user, isMe }) {
             }
           }
         })
+      },
+      onCompleted ({ setPhoto: photoId }) {
+        // add random query param for cache busting
+        const randomParam = (Math.random() + 1).toString(16).substring(2)
+        const src = `https://${process.env.NEXT_PUBLIC_AWS_UPLOAD_BUCKET}.s3.amazonaws.com/${user.photoId}?r=${randomParam}`
+        setSrc(src)
```
huumn, line 79: Is this needed because you are reusing the `photoId`?

prisma/migrations/20231025203103_delete_unused_images_schedule/migration.sql
```diff
@@ -0,0 +1,23 @@
...
+AS $$
+DECLARE
+BEGIN
+    INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('trust', '0 2 * * *', 'America/Chicago') ON CONFLICT DO NOTHING;
+    INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('auction', '* * * * *', 'America/Chicago') ON CONFLICT DO NOTHING;
+    INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('earn', '0 0 * * *', 'America/Chicago') ON CONFLICT DO NOTHING;
+    INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('streak', '15 0 * * *','America/Chicago') ON CONFLICT DO NOTHING;
+    INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('views', '0 0 * * *', 'America/Chicago') ON CONFLICT DO NOTHING;
+    INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('rankViews', '* * * * *', 'America/Chicago') ON CONFLICT DO NOTHING;
+    INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('rankViews', '* * * * *', 'America/Chicago') ON CONFLICT DO NOTHING;
```
huumn, line 15: These don't need to be rescheduled. See the bottom of `20231013002652_pwot_views` for scheduling a new job.

prisma/migrations/20231025203103_delete_unused_images_schedule/migration.sql
```diff
@@ -0,0 +1,23 @@
...
+    INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('deleteUnusedImages', '* * * * *', 'America/Chicago') ON CONFLICT DO NOTHING;
```
huumn, line 16: Do we need to run image deletion every minute?

huumn (changes requested):
Overall it looks great! 

This is the first pass. I commented on areas that I think should be revisited. 

There are few places where encapsulation is violated for these changes and another area where side-effects are used when they probably don't need to be. I'm re-mentioning these because these are kind of meta-issues that are always worth keeping in mind. They aren't never-do-this rules, but you should only break them in exceptional circumstances and usually leave a comment as to why you're breaking the rule.

components/user-header.js
ekzyis: Yes. Without this, the avatar won't update after upload since the browser won't know that the image has been updated. The random parameter makes the browser fetch the new image. When you revisit the page, the browser seemed to reload the image even without a random parameter. But maybe I didn't test the caching behavior enough since I thought it will load the old image again with the plain URL.

prisma/migrations/20231025203103_delete_unused_images_schedule/migration.sql
ekzyis: At first this was a copy-paste error but then I realized that if we want to have a consistent 24 hour delete  threshold for images, it makes sense to run this more instead of once per day since else, the delay may become too large.

For example, if someone uploaded something 23h59min ago when the cleanup job starts, it won't get deleted. And then it will stay for another full 24 hours.

Running the cleanup job every minute might be too much but I imagined it doesn't hurt since it runs a query limited by `WHERE paid = 'f' AND created_at < date_trunc('hour', now() - interval '24 hours')` and there is an index on `created_at`. There is no index on `paid` though which may make sense.

So maybe run this every hour or every 3-12 hours?

api/resolvers/upload.js
ekzyis: Okay, will try :)

components/form.js
ekzyis: Ah yes, `debounce` isn't limited to use in `useEffect`, lol

components/invoice.js
ekzyis: > Could the calling context do this calculation instead?

Probably. I was thinking about using React Context more. So for example, we could wrap `useFormik` and expose the text input value via a wrapper hook like `useText` and the cost field via `useCost`. I was also thinking about creating a `useCursor` for cursor based operations. This might make things easier? :thinking: 

But need to check. This might be more "unrelated" changes than I initially wanted to include in this PR though.

But I can do it and then see which changes I'll put in this PR and which into a different, stacked PR :)

prisma/migrations/20231025203103_delete_unused_images_schedule/migration.sql
huumn: Every hour seems like the right cadence.

You don't need to truncate the date btw. This works too:

`WHERE paid = 'f' AND created_at < now() - interval '24 hours'`

api/resolvers/upload.js
ekzyis: Done in 58898dc

prisma/migrations/20231025203103_delete_unused_images_schedule/migration.sql
ekzyis: > You don't need to truncate the date btw.

Yes but I like clean cutoffs, lol. I don't like to depend on the randomness of the exact moment something was executed even though it's not really important here.

components/invoice.js
ekzyis: Decided to simply calculate the total fees in the calling context now in 4b4da33

prisma/migrations/20231025203103_delete_unused_images_schedule/migration.sql
ekzyis: Done in 1825cb6 together with the [other comment](https://github.com/stackernews/stacker.news/pull/576#discussion_r1378094163)

components/form.js
ekzyis: Done in 44d60f1

components/user-header.js
ekzyis: Okay. not entirely sure why this works, but it seems to work. The stale image in the cache seems to have been overwritten, even though we used a random param :thinking: 


https://github.com/stackernews/stacker.news/assets/27162016/e6b58bd9-9271-474e-a262-3c90d6ed623d

huumn (approved):

huumn:
Why'd you remove `itemId` from `Upload`?

It seems like useful metadata to keep. Imagine we have a profile tab for `media` that a stacker has uploaded. We could list all their uploads, but how do we link to the context it was shared in given we removed the relationship?

components/form.js
```diff
@@ -675,7 +751,14 @@ export function Form ({
...
       if (onSubmit) {
-        const options = await onSubmit(values, ...args)
+        // extract cost from formik fields
+        // (cost may also be set in a formik field named 'amount')
+        let cost = values?.cost || values?.amount
+        // add potential image fees which are set in a different field
+        // to differentiate between fees (in receipts for example)
+        cost += (values?.imageFeesInfo?.totalFees || 0)
+
+        const options = await onSubmit(cost, values, ...args)
```
huumn, line 761: This changes the function signature of `onSubmit` for everything that uses `Form`. I pulled and tried to update my settings and started getting errors. 

I suspect this breaks other things too.

ekzyis:
> Why'd you remove itemId from Upload?

Because I realized that for the fee calculation, I only needed to know if it was already paid instead of knowing in which items it was linked.

Previously, the column wasn't used and it was only a one-to-one-relationship. So to continue such a column, it becomes a many-to-many relationship: one item can have multiple uploads, and the same upload can be included in multiple items.

So I decided to replace the column with a simple `boolean` column instead of maintaining a many-to-many relationship which seemed unnecessary complex for what I was trying to achieve.

In your commit, [this line](https://github.com/stackernews/stacker.news/pull/576/commits/4688962abcd54fdc5850109372a7ad054cf9b2e4#diff-97f846e99583f779f9f3f3eb06f7726e4cf482f968473909dc12e51f9adaf4dbR179) replaces any previous existing upload id with the new one. So this doesn't track all links, only the latest link from an upload to an item.

huumn (changes requested):
I've updated your code to keep `itemId`s on `Upload` then realized your code has bugs. 😄

Please fix the bugs on top of my changes.

------------

To speak broadly on changes, when we change something generic we need to consider *everything* else that uses that code. `Form` is such a generic thing. Ideally updates to `Form` should be made in a way that they have no effect on existing things that use it.

When I can't avoid making fundamental changes to a piece of code meant to be generic: I search for everything that uses it, walk through it still working logically after the changes, and manually test it just in case I did something unintentional.

Spaghetti code happens when we fail to make our changes isolated and work with generic parts.

huumn:
> In your commit, [this line](https://github.com/stackernews/stacker.news/pull/576/commits/4688962abcd54fdc5850109372a7ad054cf9b2e4#diff-97f846e99583f779f9f3f3eb06f7726e4cf482f968473909dc12e51f9adaf4dbR179) replaces any previous existing upload id with the new one. So this doesn't track all links, only the latest link from an upload to an item.

This means we lose the original `itemId` right? I think that's okay but you're right it's not ideal. Ideally I guess we'd create a many-to-many relationship between `Item` and `Upload` so that we have this metadata when we want to use it later.

components/form.js
ekzyis: Okay, will revert the signature

ekzyis:
> This means we lose the original itemId right? I think that's okay but you're right it's not ideal. Ideally I guess we'd create a many-to-many relationship between Item and Upload so that we have this metadata when we want to use it later.

It also breaks trying to insert two images into the same post because of the unique index.

So I think we should replace this column since it's not complete anyway and create a table to maintain this many-to-many relationship

huumn:
> It also breaks trying to insert two images into the same post because of the unique index.

Oh, I see! Cool, obliterate my changes.

huumn:
Sorry if I seem impatient. Totally not personal.

I've just been in a bit of a mood all day and I left myself a bunch of code I promised to review 😄

I'll give your code another shot and add the many-to-many relationship. Thanks!

huumn (approved):

huumn:
my todos:

- [ ] many to many item<->upload
- [ ] give upload urls a pretty domain name by aliasing the bucket

huumn:
This is excellent.

Can you walk me through why we have static avatar upload ids again? It's causing issues for me possibly because my bucket is configured slightly differently, also see below, and I'm wondering if we can figure out a way to not need to be conscious of caching (I know you know how hard it is to debug that stuff).

Under the following circumstance, won't I see stale profile images?
1. I visit your profile
2. cache your profile image
3. you upload a new one
4. When I visit your profile again, won't I get served the stale image?
     - you cache bust it for the uploader, but we can't control everyone else's cache 

Reasons I can imagine we have static avatar ids (with possible alternatives):
1. we don't want to charge fees
    - we can exclude profile images in the image fee function 
2. we want to make sure they aren't deleted
    - we can exclude active profile images in the delete function

But I might be missing something.

huumn:
Other than that, this is good to go! I haven't tested the fee part yet, but we have some other pending PRs that involve `FeeButton` so I'll try to QA all of that at once.

ekzyis:
> This is excellent.
> 
> Can you walk me through why we have static avatar upload ids again? It's causing issues for me possibly because my bucket is configured slightly differently, also see below, and I'm wondering if we can figure out a way to not need to be conscious of caching (I know you know how hard it is to debug that stuff).
> 
> Under the following circumstance, won't I see stale profile images?
> 
> 1. I visit your profile
> 2. cache your profile image
> 3. you upload a new one
> 4. When I visit your profile again, won't I get served the stale image?
>    
>    * you cache bust it for the uploader, but we can't control everyone else's cache

Mhh, I didn't test it with two users [but only with one](https://github.com/stackernews/stacker.news/pull/576#discussion_r1379471845). I expected it wouldn't already work with one user since on page reload, the browser would fetch the URL without the random parameter and thus it may use the cached one, as you mentioned with

> you cache bust it for the uploader, but we can't control everyone else's cache

But it didn't, it used the up-to-date one even though there were `Cache-Control` headers present. But maybe it's still this:

> It's causing issues for me possibly because my bucket is configured slightly differently

And maybe testing this stuff out locally is irrelevant since we use a CDN in production and that's where the caching issues will appear. Not on my local build which talks straight to S3.

> Reasons I can imagine we have static avatar ids (with possible alternatives):
> 
> 1. we don't want to charge fees
>    
>    * we can exclude profile images in the image fee function
> 2. we want to make sure they aren't deleted
>    
>    * we can exclude active profile images in the delete function
> 
> But I might be missing something.

Yes, the reason for static avatar ids was to keep them free and thus the storage they can occupy should be limited.

That they aren't deleted is handled by only deleting the rows with `paid = 'f'`. Avatars have `NULL` in this row.

To prevent any cache issues, we could give them a new id on every upload but limit the amount of free storage in other, free ways like applying time limits?

PR #579, opened 2023-10-23
Fix TypeError regarding matchAll

ekzyis:
Not tested but pretty sure that's it.

Context: https://stacker.news/items/291566


ekzyis:
Looking at this again, I should have returned `undefined` to make it consistent with the existing return values

PR #585, opened 2023-10-24
Wallet history timestamps

### LINKED ISSUE #583: Surface timestamps in wallet history UI
I often look for when certain events occurred in my wallet history. The data is there in the API call but it's not surfaced in the UI. It would be nice to have an option to see this on the page, especially for invoices and withdrawals 

Timestamps for wallet history events surfaced in some way. If we don't want to add another column to the table, it could be a tooltip that surfaces the info on click. 

The only alternative I've found is to explore the API data directly. 

**Additional context**
This is probably most useful for invoices, withdrawals, and spent events. Earned events are rolled up to the most recent earning per item, so while it's kinda useful, it's not as useful as the others, IMO.

SatsAllDay:
Closes #583 

This PR updates the wallet history page's `Fact` component to include when each wallet fact occurred.

It makes use of the common `timeSince` function that we use in item-related components to show a relative time that a wallet history event occurred.

Users can also hover over the relative time in each table row to see the full ISO timestamp (via `title` attribute).

Sample screenshot:
image

### DIFF
--- a/pages/satistics.js
+++ b/pages/satistics.js
@@ -15,6 +15,7 @@ import { CommentFlat } from '../components/comment'
 import ItemJob from '../components/item-job'
 import PageLoading from '../components/page-loading'
 import PayerData from '../components/payer-data'
+import { timeSince } from '../lib/time'
 
 export const getServerSideProps = getGetServerSideProps({ query: WALLET_HISTORY, authRequired: true })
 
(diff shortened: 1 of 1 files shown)


SatsAllDay:
I wonder if appending ` ago` to the relative times might help folks figure out what this is e.g. `10m ago`. Once it gets to a full date e.g. `24Oct`, it's pretty clear it's a date, but `10m` might not be immediately obvious? Again, just thinking aloud :)

huumn:
Yep, I think appending `ago` or just showing a local date/time is the way to go.

SatsAllDay:
> Yep, I think appending `ago` or just showing a local date/time is the way to go.

Done. I like this much better.

image

rleed:
I've often wished for timestamps...looks great!

PR #591, opened 2023-10-29
Provide option to clear withdrawal invoices

### LINKED ISSUE #201: Provide option to clear withdrawal invoices
https://stacker.news/items/75823

rleed:
To close #201.


api/resolvers/wallet.js
```diff
@@ -206,7 +206,10 @@ export default {
 
       history = history.map(f => {
         if (f.bolt11) {
-          const inv = lnpr.decode(f.bolt11)
+          let inv
+          try {
+            inv = lnpr.decode(f.bolt11)
+          } catch { }
```
ekzyis, line 212: you can just filter by `FORGOTTEN_HASH` instead of `try/catch` no? This could lead to confusion in the future why we're just ignoring errors.

pages/settings.js
```diff
@@ -236,6 +237,26 @@ export default function Settings ({ ssrData }) {
...
+            label={
+              <div className='d-flex align-items-center'>auto forget withdrawal invoices after 7 days
+                <Info>
+                  <ul className='fw-bold'>
+                    <li>use this to protect receiver privacy</li>
+                    <li>applies retroactively, cannot be reversed</li>
+                    <li>withdrawal invoices are kept at least 7 days:
+                      <ul>
+                        <li>for debugging purposes</li>
+                        <li>to prevent a thief from covering his tracks</li>
```
ekzyis, line 250: i would just mention: `withdrawal invoices are kept at least 7 days for security and debugging purposes`

this also doesn't directly refer to the user to be a potential thief, lol

worker/wallet.js
```diff
@@ -118,3 +119,20 @@ export function checkWithdrawal ({ boss, models, lnd }) {
...
 }
+
+export function autoDropWdInvoices ({ models }) {
+  return async function () {
+    console.log('forgetting invoices')
+    try {
+      await serialize(models, models.$executeRaw`
+      UPDATE "Withdrawl" SET
+        hash = '${FORGOTTEN_HASH}',
+        bolt11 = '${FORGOTTEN_HASH}'
```
ekzyis, line 130: Is this tested? I don't think this will work since `Invoice.hash` has an unique index:

```
stackernews=# \d "Invoice.hash_unique"
Index "public.Invoice.hash_unique"
 Column | Type | Key? | Definition
--------+------+------+------------
 hash   | text | yes  | hash
unique, btree, for table "public.Invoice"
```

lib/constants.js
```diff
@@ -77,3 +77,6 @@ export const ITEM_ALLOW_EDITS = [
   // FAQ, privacy policy, changelog, content guidelines
   349, 76894, 78763, 81862
 ]
+
+export const INVOICE_RETENTION_DAYS = 7
+export const FORGOTTEN_HASH = 'f0rg0tten'
```
ekzyis, line 82: I would use 32 zero bytes since that's a common practice for "unused data", I think:

```python-repl
$ python
Python 3.11.5 (main, Sep  2 2023, 14:16:33) [GCC 13.2.1 20230801] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> bytes(32).hex()
'0000000000000000000000000000000000000000000000000000000000000000'
```

But there is another problem that I also mentioned in another comment: There is a unique index on hash and I think we should keep this unique index. Is deleting the hash even relevant for privacy? Afaik, it's just random bytes so there is no PII in there :thinking:

ekzyis (commented):
Left some comments. I am also wondering how bad it would be if users accidentally delete invoices.

Since enabling a simple setting might lead to deletion of a lot of invoices, maybe we should show a warning how many invoices will be deleted if this setting is applied? Currently, we only tell the user that it applies retroactively in the info box. And we don't tell them how big of an impact this can have.

I am basically wondering if enabling this setting should follow a good delete UX to prevent irreversible accidental data loss like in [this case](https://httpie.io/blog/stardust).

api/resolvers/wallet.js
rleed: Good point

worker/wallet.js
rleed: Interesting. I certainly had multiple invoices in my testing...

rleed:
> I am also wondering how bad it would be if users accidentally delete invoices.

I had the same thought... I'm glad I was not alone in that. I will pursue your recommendations on this.

worker/wallet.js
rleed: You are definitely correct that this causes errors. Thinking about the database implementation more, I think the right way to implement this is to null the fields instead of putting in a special string. The reality is that those fields are no longer "required" to be valid, so allowing null would make sense IMO. Let me know if you disagree.

lib/constants.js
rleed: Allowing null per my other comment should still let us keep the unique key. (I think the security issue has to do not only with the randomness of the hash but also with who else might have a copy of it.)

pages/settings.js
```diff
@@ -236,6 +237,26 @@ export default function Settings ({ ssrData }) {
             name='hideInvoiceDesc'
             groupClassName='mb-0'
           />
+          <Checkbox
+            label={
+              <div className='d-flex align-items-center'>auto forget withdrawal invoices after 7 days
```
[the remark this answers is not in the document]
rleed: I've taken your advice to change it everywhere.

prisma/migrations/20231025162230_self_subscription/migration.sql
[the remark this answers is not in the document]
rleed: That gave me a chance to test my git skills 😈

rleed:
I think this is ready now. Thanks for the review!

huumn:
This looks good! I still have a few changes I want to make to the UI and I need to test the job but I'll merge this tomorrow.

PR #595, opened 2023-11-01
Notification badges

### LINKED ISSUE #594: Use Badging API to indicate notification status on PWA
I'd like to know when I have SN notifications from my home screen via native notification badges

Basically I want a non-numeric notification badge on the PWA app icon on my home screen, similar to the red indicator on the notification bell within the SN app, and on the favicon.

None

**Additional context**
[Badging API MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Badging_API)

SatsAllDay:
Closes #594 

This PR adds support for notification badges on the app's icon on the home screen.

I originally was going to mimic the behavior of the notification bell in the UI itself (no count, just an empty badge). However, as I worked on these changes, I changed approaches.

Since badges correspond to push notifications, I thought it made sense to show the badge based on receiving push notifications, instead of just doing some kind of polling on the notes API. This also gives the implicit benefit of only showing badges to users who have enabled push notifications, which is ideal.

The changes work as follows:

* Track the number of push notifications received. Increment the count by 1 for every notification received _that we show_. e.g. notifications that get merged with current notifications don't increment the count. Skipped notifications don't increment the count, etc.
* Every time the count of notifications goes up, update the app's badge w/this new number
* Every time a notification is clicked, decrement the count and update the app's badge accordingly
* If the count is updated to 0, clear the badge altogether instead of displaying 0
* When a user loads the notifications page in the UI itself, clear all open notifications and reset the internal count to 0
* When the notification bell header polls and indicates no new notes, clear all open notifications and reset the internal count to 0

So there's some slight tweaks to the existing behavior here, plus the new badge behavior as well.

As usual, I am open to feedback!

### DIFF
--- a/lib/badge.js
+++ b/lib/badge.js
@@ -1,0 +1,29 @@
+export const badgingApiSupported = (sw = window) => 'setAppBadge' in sw.navigator
+
+export const permissionGranted = async (sw = window, name = 'notifications') => {
+  let permission
+  try {
+    permission = await sw.navigator.permissions.query({ name })
+  } catch (err) {
+    console.error('Failed to check permissions', err)
+  }
+  return permission?.state === 'granted'
+}
+
+export const setAppBadge = async (sw = window) => {
+  if (!badgingApiSupported(sw) || !(await permissionGranted(sw))) return
+  try {
+    await sw.navigator.setAppBadge()
+  } catch (err) {
+    console.error('Failed to set app badge', err)
+  }
+}
+
+export const clearAppBadge = async (sw = window) => {
+  if (!badgingApiSupported(sw) || !(await permissionGranted(sw))) return
+  try {
+    await sw.navigator.clearAppBadge()
+  } catch (err) {
+    console.error('Failed to clear app badge', err)
+  }
+}
--- a/lib/badge.js
+++ b/lib/badge.js
@@ -1,3 +1,7 @@
+export const CLEAR_BADGE_ACTION = 'CLEAR_BADGE'
+
+export const dispatchClearBadgeAction = () => navigator.serviceWorker?.controller?.postMessage({ action: CLEAR_BADGE_ACTION })
+
 export const badgingApiSupported = (sw = window) => 'setAppBadge' in sw.navigator
 
 export const permissionGranted = async (sw = window, name = 'notifications') => {
--- a/lib/badge.js
+++ b/lib/badge.js
@@ -1,10 +1,10 @@
-export const CLEAR_BADGE_ACTION = 'CLEAR_BADGE'
+export const CLEAR_NOTIFICATIONS = 'CLEAR_NOTIFICATIONS'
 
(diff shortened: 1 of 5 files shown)


lib/badge.js
```diff
@@ -0,0 +1,33 @@
+export const CLEAR_BADGE_ACTION = 'CLEAR_BADGE'
+
+export const dispatchClearBadgeAction = () => navigator.serviceWorker?.controller?.postMessage({ action: CLEAR_BADGE_ACTION })
```
ekzyis, line 3: I would rename this function to `clearNotifications` and the action to `CLEAR_NOTIFICATIONS` since it's more about clearing the notifications. It only calls the `clearAppBadge` function defined below in this file. The name confused me since there were two functions which seemed to clear the badge.

sw/eventListener.js
```diff
@@ -155,5 +167,18 @@ export function onMessage (sw) {
     if (event.data.action === 'SYNC_SUBSCRIPTION') {
       return event.waitUntil(onPushSubscriptionChange(sw)(event.oldSubscription, event.newSubscription))
     }
+    if (event.data.action === CLEAR_BADGE_ACTION) {
+      return event.waitUntil((async () => {
+        let notifications = []
+        try {
+          notifications = await sw.registration.getNotifications()
+        } catch (err) {
+          console.error('failed to get notifications')
```
ekzyis, line 176: Missing return? In case of error, you probably don't want to access `notifications`

lib/badge.js
```diff
@@ -0,0 +1,33 @@
...
+
+export const badgingApiSupported = (sw = window) => 'setAppBadge' in sw.navigator
+
+export const permissionGranted = async (sw = window, name = 'notifications') => {
```
ekzyis, line 7: Nitpick: You're not using these functions anywhere else. I would then not export them

ekzyis (commented):
Mhh, looks like I can't test this since [`setAppBadge` is only supported on Safari](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/setAppBadge#browser_compatibility)

2023-11-06-163859_777x306_scrot


I was receiving push notifications but no badge appeared.

SatsAllDay:
> Mhh, looks like I can't test this since [`setAppBadge` is only supported on Safari](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/setAppBadge#browser_compatibility)

Ah, sorry, Chrome on Windows and macOS or Safari! I didn't realize it wasn't on Linux.

lib/badge.js
SatsAllDay: Good call, I can fix that. I probably used them in a previous commit but ended up refactoring quite a bit.

sw/eventListener.js
SatsAllDay: I initialized `notifications` to `[]` to handle this case. I figured even if it failed to fetch notifications, we can clear the badge, but perhaps that's a bad idea.

lib/badge.js
SatsAllDay: Yep, that makes sense. Another case of evolving functionality without renaming. I'll clean it up :)

PR #597, opened 2023-11-02
Resubscribe if service worker lost push subscription

ekzyis:
Potential fix for #411 

> However, I have an idea for a possible workaround (without really fixing the underlying issue since we don't know what is causing this):
> 
> We could delete the push subscription in IndexedDB if the user manually unchecks the push subscription.
>
> In that case, we know that if the service worker has no push subscription but there is a push subscription in IndexedDB during `SYNC_SUBSCRIPTION`, this is probably the bug and we can register a new push subscription in the background and send that one to the server. Currently, we aren't able to distinguish between the bug and manually unsubscribing since in both cases, there would still be a push subscription in IndexedDB.
> 
> So this means that during `unsubscribeFromPushNotifications`, we would add code that also deletes the push subscription in IndexedDB

-- https://github.com/stackernews/stacker.news/issues/411#issuecomment-1790675861

_simulating the bug by unsubscribing but not deleting the subscription in IndexedDB_:

https://github.com/stackernews/stacker.news/assets/27162016/76c11081-8ef6-4753-81b5-c00a0e5bcc93

That it shows `Stacker News notification are now active` on resubscribing may be annoying but I think it would be useful to keep this for now. With this notification, we can see if it's working or not since you will see this notification if you got resubscribed automatically if you lost your subscription for unknown reasons.

Since we delete the push subscription in IndexedDB on manual unsubscription, you won't get automatically resubscribed in this case.

### DIFF
--- a/components/serviceworker.js
+++ b/components/serviceworker.js
@@ -109,24 +109,23 @@ export const ServiceWorkerProvider = ({ children }) => {
       pushManager: 'PushManager' in window
     })
     setPermission({ notification: 'Notification' in window ? window.Notification.permission : 'denied' })
-    // since (a lot of) browsers don't support the pushsubscriptionchange event,
-    // we sync with server manually by checking on every page reload if the push subscription changed.
-    // see https://medium.com/@madridserginho/how-to-handle-webpush-api-pushsubscriptionchange-event-in-modern-browsers-6e47840d756f
-    navigator?.serviceWorker?.controller?.postMessage?.({ action: 'SYNC_SUBSCRIPTION' })
-    logger.info('sent SYNC_SUBSCRIPTION to service worker')
-  }, [])
 
-  useEffect(() => {
-    if (!support.serviceWorker) {
+    if (!('serviceWorker' in navigator)) {
       logger.info('device does not support service worker')
       return
     }
+
     const wb = new Workbox('/sw.js', { scope: '/' })
     wb.register().then(registration => {
       logger.info('service worker registration successful')
       setRegistration(registration)
+      // since (a lot of) browsers don't support the pushsubscriptionchange event,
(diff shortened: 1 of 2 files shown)


SatsAllDay:
Just FYI, I just had push notifications re-enabled on my desktop machine which I had disabled a while ago. I imagine since we didn't use to clear IndexedDB, this PR's code detected this as a lost subscription and re-enabled them for me? It might be good to let people know what's happening if they mysteriously get push notifications re-enabled without their action.

huumn:
I had something similar where it requested push notifications again when I previously denied it.

huumn:
Every time I visit the site after not having been on it in any tab, it asks me to subscribe now. Is this expected?

ekzyis:
> Just FYI, I just had push notifications re-enabled on my desktop machine which I had disabled a while ago. I imagine since we didn't use to clear IndexedDB, _I suspect_ this PR's code detected this as a lost subscription and re-enabled them for me? It might be good to let people know what's happening if they mysteriously get push notifications re-enabled without their action.

Ohh, yes, turns out this change wasn't backwards compatible :sweat_smile:

I forgot about the devices which used to have a push subscription and thus still have a subscription in IndexedDB. This is now detected as the bug from #411 and it tries to resubscribe. I can make this backwards compatible via #616

> Every time I visit the site after not having been on it in any tab, it asks me to subscribe now. Is this expected?

Mhh, no, that wasn't expected. Maybe it's related to the above so that would also be fixed via #616

ekzyis:
> Every time I visit the site after not having been on it in any tab, it asks me to subscribe now. Is this expected?

I hope this is no longer the case?

huumn:
> I hope this is no longer the case?

Yep no longer the case that I get resubscribed every time

PR #620, opened 2023-11-09
Toast on delete bot directive usage

### LINKED ISSUE #614: Visual feedback for successfully using the @delete bot
I've seen some users incorrectly using the `@delete` bot in items. Since we don't offer any feedback that the delete command was successfully interpreted and scheduled, users don't really know that it didn't work.

I'd like to toast when an item is saved if a deletion was successfully scheduled.

We could also do private replies or DMs, but those feel pretty heavy-weight for this, while a toast would be easy and clean.

**Additional context**
See discussion here: https://stacker.news/items/302056/r/WeAreAllSatoshi

**Proposed solution:**
1. Add a `deleteScheduledAt` DateTime nullable field on the `Item` table
2. Update the item resolver to set/clear `deleteScheduledAt` field based on whether the deletion was scheduled/cleared.
3. In the UI, when saving an item, check the `deleteScheduledAt` field on the returned item, and it it's not null, dispatch a success toast indicating when the item will be deleted.

SatsAllDay:
Closes #614 

This PR toasts when an item is created or updated and successfully schedules a deletion job via the `@delete` bot.

It supports all post types and comments, both new and edited.

The toast displays the time at which the item will be deleted, localized to the user's current browser.

If the user mentioned the delete bot but the deletion wasn't scheduled, these changes also toast a warning toast indicating that they may have tried to schedule a deletion but it didn't work, and gives an example of the correct usage.

I also optimized the resolver to only query the DB to find scheduled deletion jobs for your own items, since it's only really necessary for yourself.

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -991,6 +991,10 @@ export default {
       }
       const parent = await models.item.findUnique({ where: { id: item.parentId } })
       return parent.otsHash
+    },
+    deleteScheduledAt: async (item, args, { models }) => {
+      const deleteJobs = await models.$queryRawUnsafe(`SELECT startafter FROM pgboss.job WHERE name = 'deleteItem' AND data->>'id' = '${item.id}'`)
+      return deleteJobs[0]?.startafter ?? null
     }
   }
 }
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -992,7 +992,12 @@ export default {
       const parent = await models.item.findUnique({ where: { id: item.parentId } })
       return parent.otsHash
     },
-    deleteScheduledAt: async (item, args, { models }) => {
+    deleteScheduledAt: async (item, args, { me, models }) => {
+      const meId = me?.id ?? ANON_USER_ID
+      if (meId !== item.userId) {
+        // Only query for deleteScheduledAt for your own items to keep DB queries minimized
+        return null
+      }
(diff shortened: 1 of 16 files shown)


SatsAllDay:
I might try to update this PR to also display a danger or warning toast if the item has a delete bot mention but the delete wasn't scheduled. That might make the experience better, by detecting a failed attempt at the delete directive

SatsAllDay:
> I might try to update this PR to also display a danger or warning toast if the item has a delete bot mention but the delete wasn't scheduled. That might make the experience better, by detecting a failed attempt at the delete directive

@huumn it looks like we don't have a `warning` color in our `globals.scss` file. Is there a hex code you'd prefer for that? Similar to how we have customized success, danger, info, etc. values.

huumn:
> Edit: I guess I could use info instead. I assume warning was omitted since there's so much yellow on SN already?

Yes lol. We use yellow and orange for non-warning uses.

I think either using our orange or yellow might be okay if we don't want to use the red.

SatsAllDay:
> I think either using our orange or yellow might be okay if we don't want to use the red.

Personally, I think the red is overkill in this use case, since the item was successfully upserted, it's just the delete bot usage that didn't work. That's why I investigated the `warning` path. But I'll defer that decision to others, I'm not _that_ picky!

huumn:
I agree. How about using bs-secondary?

SatsAllDay:
Works for me!

PR #629, opened 2023-11-13
Nostr crossposting improvements

AustinKelsay:
This PR allows for crossposting past discussion items to nostr and it also saves the nostr event id for any crossposted items allowing the user to easily link out to it.

## High level summary:
- crosspost old discussion items to nostr
- Save noteId on item when crossposted
- Option to crosspost old discussion is available on item-info
- If it has been crossposted and eventId has been saved then a 'nostr note' dropdown item is presented that will link you to your crosposted note on nostr.band

## User facing changes
- "crosspost to nostr" dropdown item on item-info for old discussions
- "nostr note" dropdown item on item-info for crossposted discussions that links to the note on nostr.band

## Internal code changes:
- Migration to add noteId to Item table
- noteId String added to Item in schema.prisma
- noteId added to typedefs/item
- noteId added to resolvers/item
- noteId added to fragments/item
- Added an Item update mutation in discussion form:
  - In discussion form the Item is now saved to the db, crossposted, and then the noteId from the crosspost is saved to the item
- Added crosspost-item component that lives as a dropdown item under item-info (rendered for discussion items only)

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1066,8 +1066,8 @@ export const updateItem = async (parent, { sub: subName, forward, options, ...it
     await models.item.update({
       where: { id: Number(item.id) },
       data: { noteId: item.noteId }
-    });
-    return { id: item.id, noteId: item.noteId };
+    })
+    return { id: item.id, noteId: item.noteId }
   }
 
   // prevent update if it's not explicitly allowed, not their bio, not their job and older than 10 minutes
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1061,6 +1061,16 @@ export const updateItem = async (parent, { sub: subName, forward, options, ...it
   // in case they lied about their existing boost
   await ssValidate(advSchema, { boost: item.boost }, { models, me, existingBoost: old.boost })
 
+  // Check if we ar ejust updating nEventId
+  if (item.nEventId) {
+    // Update only nEventId and return
+    await models.item.update({
+      where: { id: Number(item.id) },
+      data: { nEventId: item.nEventId }
+    });
+    return { id: item.id, nEventId: item.nEventId }; // Or any other relevant data you want to return
+  }
+
   // prevent update if it's not explicitly allowed, not their bio, not their job and older than 10 minutes
   const user = await models.user.findUnique({ where: { id: me.id } })
   if (!ITEM_ALLOW_EDITS.includes(old.id) && user.bioId !== old.id &&
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1061,14 +1061,13 @@ export const updateItem = async (parent, { sub: subName, forward, options, ...it
   // in case they lied about their existing boost
   await ssValidate(advSchema, { boost: item.boost }, { models, me, existingBoost: old.boost })
 
-  // Check if we ar ejust updating nEventId
-  if (item.nEventId) {
-    // Update only nEventId and return
+  // Check if we are just updating noteId
+  if (item.noteId) {
     await models.item.update({
       where: { id: Number(item.id) },
-      data: { nEventId: item.nEventId }
+      data: { noteId: item.noteId }
     });
-    return { id: item.id, nEventId: item.nEventId }; // Or any other relevant data you want to return
+    return { id: item.id, noteId: item.noteId };
   }
 
   // prevent update if it's not explicitly allowed, not their bio, not their job and older than 10 minutes
@@ -1186,7 +1185,7 @@ const getForwardUsers = async (models, forward) => {
 export const SELECT =
   `SELECT "Item".id, "Item".created_at, "Item".created_at as "createdAt", "Item".updated_at,
   "Item".updated_at as "updatedAt", "Item".title, "Item".text, "Item".url, "Item"."bounty",
-  "Item"."nEventId", "Item"."userId", "Item"."parentId", "Item"."pinId", "Item"."maxBid",
+  "Item"."noteId", "Item"."userId", "Item"."parentId", "Item"."pinId", "Item"."maxBid",
   "Item"."rootId", "Item".upvotes, "Item".company, "Item".location, "Item".remote, "Item"."deletedAt",
   "Item"."subName", "Item".status, "Item"."uploadId", "Item"."pollCost", "Item".boost, "Item".msats,
   "Item".ncomments, "Item"."commentMsats", "Item"."lastCommentAt", "Item"."weightedVotes",
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1176,7 +1176,7 @@ const getForwardUsers = async (models, forward) => {
 export const SELECT =
   `SELECT "Item".id, "Item".created_at, "Item".created_at as "createdAt", "Item".updated_at,
   "Item".updated_at as "updatedAt", "Item".title, "Item".text, "Item".url, "Item"."bounty",
-  "Item"."userId", "Item"."parentId", "Item"."pinId", "Item"."maxBid",
+  "Item"."nEventId", "Item"."userId", "Item"."parentId", "Item"."pinId", "Item"."maxBid",
   "Item"."rootId", "Item".upvotes, "Item".company, "Item".location, "Item".remote, "Item"."deletedAt",
   "Item"."subName", "Item".status, "Item"."uploadId", "Item"."pollCost", "Item".boost, "Item".msats,
   "Item".ncomments, "Item"."commentMsats", "Item"."lastCommentAt", "Item"."weightedVotes",
--- a/components/crosspost-item.js
+++ b/components/crosspost-item.js
@@ -31,8 +31,7 @@ export default function CrosspostDropdownItem({ item }) {
 
                 try {
                     if (item?.id) {
-                        const createdAt = Math.floor(Date.now() / 1000)
-                        const crosspostResult = await crossposter({ ...data.item, createdAt: createdAt })
+                        const crosspostResult = await crossposter({ ...data.item })
                         const eventId = crosspostResult?.eventId;
                         if (eventId) {
                             await upsertDiscussion({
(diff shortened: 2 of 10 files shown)


huumn:
Oh man wasn't expecting this! Awesome!

api/resolvers/item.js
```diff
@@ -1061,6 +1061,15 @@ export const updateItem = async (parent, { sub: subName, forward, options, ...it
...
 
+  // Check if we are just updating noteId
+  if (item.noteId) {
+    await models.item.update({
+      where: { id: Number(item.id) },
+      data: { noteId: item.noteId }
+    })
+    return { id: item.id, noteId: item.noteId }
+  }
+
```
huumn, line 1072: I think it'd be best if adding a noteId to an item were its own mutation.

components/crosspost-item.js
```diff
@@ -0,0 +1,65 @@
...
+              }
+              await router.push(`/items/${item.id}`)
+            }
+          } catch (e) {
+            console.error(e)
+          }
+        }}
+      >
+        crosspost to nostr
+      </Dropdown.Item>
```
huumn, line 58: This gets a lot simpler if adding a noteId is its own mutation.

components/use-crossposter.js
```diff
@@ -53,12 +53,15 @@ export default function useCrossposter () {
   return useCallback(async item => {
     let failedRelays
     let allSuccessful = false
+    let eventId
```
huumn, line 56: If we're calling it `noteId` elsewhere, then we should probably call it that here.

huumn (changes requested):
It hits all the marks functionality-wise!

The only thing preventing the merge is the `upsertDiscussion` hack. This code will get a lot simpler with a `updateNoteId(id, noteId)` mutation on items *and* it will make adding other post types to `crossposter` much easier.

AustinKelsay:
> It hits all the marks functionality-wise!
> 
> The only thing preventing the merge is the `upsertDiscussion` hack. This code will get a lot simpler with a `updateNoteId(id, noteId)` mutation on items _and_ it will make adding other post types to `crossposter` much easier.

Good callouts!

Just added these changes and retested everything

huumn (changes requested):
Looks good. A few final nits

components/crosspost-item.js
```diff
@@ -0,0 +1,65 @@
...
+        upsertNoteId(id: $id, noteId: $noteId) {
+          id
+          noteId
+        }
+      }`
+  )
+
+  const crossposter = useCrossposter()
+
+  return !item?.noteId
```
ekzyis, line 23: I would reverse this condition. The negation + the optional chaining makes this harder to understand than necessary imo. The following should work, too, no?

```js
return item?.noteId
  ? (
    <Dropdown.Item onClick={() => window.open(`https://nostr.band/${item.noteId}`, '_blank')}>
      nostr note
    </Dropdown.Item>
    )
  : (
    <Dropdown.Item
      onClick={async () => {
        try {
          if (!(await window.nostr.getPublicKey())) {
            throw new Error('not available')
          }
        } catch (e) {
          toaster.danger(`Nostr extension error: ${e.message}`)
          return
        }
        try {
          if (item?.id) {
            const crosspostResult = await crossposter({ ...data.item })
            const noteId = crosspostResult?.noteId
            if (noteId) {
              await upsertNoteId({
                variables: {
                  id: item.id,
                  noteId
                }
              })
            }
            toaster.success('Crosspost successful')
          } else {
            toaster.warning('Item ID not available')
          }
        } catch (e) {
          console.error(e)
          toaster.danger('Crosspost failed')
        }
      }}
    >
      crosspost to nostr
    </Dropdown.Item>
    )
```

components/crosspost-item.js
```diff
@@ -0,0 +1,65 @@
...
+            if (!(await window.nostr.getPublicKey())) {
+              throw new Error('not available')
+            }
+          } catch (e) {
+            toaster.danger(`Nostr extension error: ${e.message}`)
+            return
+          }
+          try {
+            if (item?.id) {
+              const crosspostResult = await crossposter({ ...data.item })
```
ekzyis, line 37: If I understand this correctly, this will post _anything_ under my own nostr pubkey as if it's my own content, right?

Shouldn't there at least be some attribution to the OP? Like a link back to the SN post?

components/item-info.js
```diff
@@ -144,6 +145,7 @@ export default function ItemInfo ({
           </Link>}
         {me && !item.meSats && !item.position &&
           !item.mine && !item.deletedAt && <DontLikeThisDropdownItem id={item.id} />}
+        {me && !item.pollCost && !item.url && !item.bounty && <CrosspostDropdownItem item={item} />}
```
ekzyis, line 148: Since this is about sharing data with nostr, I think it would make more sense to put this into the dropdown of the share button.

However, this would mean that you can only crosspost something that is selected as the root item since only then the share button shows up. But I think that's not a problem.

2023-11-23-063359_1920x1080_scrot

2023-11-23-063408_1920x1080_scrot

But then the `nostr note` link would also show up there which is no longer necessarily about "sharing data" :thinking: 

 What do you think?

ekzyis (commented):
_Disclaimer: I haven't tested the crossposting functionality since I don't want to blast old content into nostr, lol. Will try if I can use `socat` for local testing tomorrow._

Looks good overall and good idea with allowing old content to be shared with nostr, too :) Left some comments.

Btw, [Can't comment on lines outside of the diff](https://github.com/refined-github/refined-github/issues/3892), but there is a typo at _components/use-crossposter.js:26_:

```diff
- data?.settings?.nostRelays
+ data?.settings?.nostrRelays
```

AustinKelsay:
> _Disclaimer: I haven't tested the crossposting functionality since I don't want to blast old content into nostr, lol. Will try if I can use `socat` for local testing tomorrow._
> 
> Looks good overall and good idea with allowing old content to be shared with nostr, too :) Left some comments.
> 
> Btw, [Can't comment on lines outside of the diff](https://github.com/refined-github/refined-github/issues/3892), but there is a typo at _components/use-crossposter.js:26_:
> 
> ```diff
> - data?.settings?.nostRelays
> + data?.settings?.nostrRelays
> ```

Sweet thanks for the comments! Here are the fixes I made:

- Now anyone can see nostr note on item-info if it exists:
image

- Only the user who authored the note can crosspost from the share button (if it has not already been crossposted):
image

- Fixed the typo at _components/use-crossposter.js:26_:

components/item-info.js
```diff
@@ -144,6 +144,11 @@ export default function ItemInfo ({
           </Link>}
         {me && !item.meSats && !item.position &&
           !item.mine && !item.deletedAt && <DontLikeThisDropdownItem id={item.id} />}
+        {me && item?.noteId && (
+          <Dropdown.Item onClick={() => window.open(`https://nostr.band/${item.noteId}`, '_blank')}>
```
ekzyis, line 148: Any reason for nostr.band specifically?

I've seen there are nostr wrapper clients (or however they are called) that let you pick which client you want to use. [nostr.com](https://nostr.com/) is one of them (actually don't know any else, lol)

So maybe use nostr.com if there is no reason against it? I don't know much about the nostr ecosystem. But that one lets you pick whatever client you want to use while nostr.band does not afaict:

_nostr.band_
2023-12-07-175942_1920x1080_scrot

_nostr.com_
2023-12-07-175926_1920x1080_scrot

ekzyis: oh, and please use `noopener` as the third argument to the `window.open` call. [seems to be good security practice.](https://stackoverflow.com/questions/40593632/use-window-open-but-block-use-of-window-opener)

AustinKelsay:
> Have taken a good look at everything now.
> 
> Mostly small changes required (see comments), so looks good and should then be ready to get merged :)

Great thanks for the thorough review!

I agree that nostr.com is probably better to link out to for the reasons you stated, I just use nostr.band for dev and their API mostly.

Changes:
- updated upsertNoteId to updateNoteId
  - There's an added check to ensure id is not undefined. If id is undefined, an error is thrown. This is a safety measure to prevent updating all items.
  - The rest of the function remains the same
- changed isOp to mine
- removed old unneeded instances of noteId on upsertDiscussion typedefs/mutations
- Switched out nostr.band for nostr.com for linking out with noteId
- Added noopener arg to window.open

api/resolvers/item.js
```diff
@@ -724,6 +724,18 @@ export default {
         return item
       }
     },
+    updateNoteId: async (parent, { id, noteId }, { models }) => {
+      if (id === undefined) {
+        throw new Error('ID is required for updateNoteId')
```
ekzyis, line 729: Or because you should indeed change this to:

```suggestion
        throw new GraphQLError('id required', { extensions: { code: 'BAD_INPUT' } })
```

you can also fix the above issue, I think :)

Some unsolicited advice: When contributing to any project, try to find similar pieces of code like you're trying to implement in the code base. And then try to imitate those pieces such that the code does not end up looking like a patchwork family over time, lol :)

ekzyis (changes requested):
Great job, two lines left to change but then you definitely get my signature on this :)

api/resolvers/item.js
```diff
@@ -724,13 +724,18 @@ export default {
         return item
       }
     },
-    upsertNoteId: async (parent, { id, noteId }, { models }) => {
+    updateNoteId: async (parent, { id, noteId }, { models }) => {
+      if (id === undefined) {
+        throw new Error("ID is required for updateNoteId");
```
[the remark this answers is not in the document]
ekzyis: huh, guess my comment did indeed not just disappear, did just belong to a different commit for some reason? :)

guess I punched my table for no reason lol :)

AustinKelsay:
> Great job, two lines left to change but then you definitely get my signature on this :)

Ahh sorry I should have caught this, thanks for being patient!  
- Simplified the condition for id 
- Updated to use the GraphQL Error 
- Also updated the description of this PR to reflect all of the changes made throughout the review

ekzyis (approved):

huumn (approved):

PR #634, opened 2023-11-15
Strip EXIF data before upload on client

### LINKED ISSUE #607: Strip sensitive metadata from images before uploading

ekzyis:
Fix #607 

https://github.com/stackernews/stacker.news/assets/27162016/7954f702-56f3-4a07-877c-de28ccf02f80

Forgot to include in video, but image still renders fine so I think this is it. No need for canvas drawing.

Credits go to https://stackoverflow.com/a/77472484

### DIFF
--- a/components/image.js
+++ b/components/image.js
@@ -150,8 +150,9 @@ export const ImageUpload = forwardRef(({ children, className, onSelect, onUpload
         }
       }`)
 
-  const s3Upload = useCallback(file => {
+  const s3Upload = useCallback(async file => {
     const img = new window.Image()
+    file = await removeExifData(file)
     img.src = window.URL.createObjectURL(file)
     return new Promise((resolve, reject) => {
       img.onload = async () => {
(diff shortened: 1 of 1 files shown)


huumn:
I didn't know we could manipulate files like that in the browser! The stuff I was reading must have been super old

ekzyis:
> I didn't know we could manipulate files like that in the browser! The stuff I was reading must have been super old

Oh, I wanted to edit my comment to sound less snarky ("no need for canvas drawing") but I must have forgotten, lol

I also wanted to add some more thoughts I had (since as we discussed, you can't read my mind, haha):

1. even though EXIF data only exists in ["TIFF, JPEG(JPG) & HEIC, but not in png, gif images"](https://stackoverflow.com/a/77472484), this code is run for all images so this might be a problem. But I believe this is not a problem since it searches for EXIF markers (?)
 
But while writing this, I got less confident, so maybe I should indeed test PNG and GIFs. I realized I don't know enough about images to make such a claim without at least some testing, lol

Or just make sure it's only run on TIFF, JPEG & HEIC.

I also tried to find a more authoritative source on which images can contain EXIF data. Why can't PNGs contain EXIF data for example? Also, [a quick search](https://developers.google.com/speed/webp/docs/riff_container) revealed that EXIF data can also be embedded in WEBP images. So maybe the sources claiming "only TIFF, JPEG, HEIC" are just old and don't know about newer image formats like WEBP and that's why they weren't mentioned?

huumn:
Should we put this in draft until you figure this out?

ekzyis:
> Should we put this in draft until you figure this out?

I'll test PNG, GIF and WEBP now. If the images load fine, I'd say we can just ship this

ekzyis:
Done. PNGs, GIFs and WEBPs are not changed at all. Running `diff` on the downloaded images with and without the EXIF filter results in the same exact binary file.

So I think this is safe to ship.

PR #644, opened 2023-11-19
Account Switching

### LINKED ISSUE #489: Account switching

It would be nice if we make it very easy to switch between different accounts.

- stackers could easier maintain multiple nyms 
- companies can easier maintain individual and an official account (like our [official stacker news account](https://stacker.news/sn))

Additionally, it would be nice if we could easily switch to anon without having to log out. But we don't want a real session as anon for privacy reasons (wallet history including invoices would be visible).

Add a "switch account" option to the user dropdown which then opens up a modal to pick or add another account:

2023-09-11-215001_182x372_scrot

After clicking on an account, we swap the existing JWT with the one for that account which should be all that is needed to impersonate a different account.

If you want to add another account, you need to verify ownership of that account first (using any authentication method of that account).

If you want to switch to anon, you just click on it and then it looks like you are logged out but clicking on login would show you all available accounts. Clicking on any of these would instantly log you in without new authentication (since we already have the required JWT stored).

ekzyis:
Closes #489

This PR implements account switching:

2023-11-19-021317_1920x1080_scrot

**Implementation details**

If a user wants to add an account, they are redirected to `/login` with `multiAuth` param set:

_components/switch-account.js:_
```js
const AddAccount = () => {
  const router = useRouter()
  return (
    <div className='d-flex flex-column me-2 my-1 text-center'>
      <Image
        width='135' height='135' src='https://imgs.search.brave.com/t8qv-83e1m_kaajLJoJ0GNID5ch0WvBGmy7Pkyr4kQY/rs:fit:860:0:0/g:ce/aHR0cHM6Ly91cGxv/YWQud2lraW1lZGlh/Lm9yZy93aWtpcGVk/aWEvY29tbW9ucy84/Lzg5L1BvcnRyYWl0/X1BsYWNlaG9sZGVy/LnBuZw' style={{ cursor: 'pointer' }} onClick={() => {
          router.push({
            pathname: '/login',
            query: { callbackUrl: window.location.origin + router.asPath, multiAuth: true }
          })
        }}
      />
      <div className='fst-italic'>+ add account</div>
    </div>
  )
}
```

With this param, a login flow within a session can be initiated. This param is checked in the backend to only set cookies without actually switching the user (`return token` instead of `return user`):

```js
if (multiAuth) {
  // we want to add a new account to 'switch accounts'
  const secret = process.env.NEXTAUTH_SECRET
  // default expiration for next-auth JWTs is in 1 month
  const expiresAt = datePivot(new Date(), { months: 1 })
  const cookieOptions = {
    path: '/',
    httpOnly: true,
    secure: true,
    sameSite: 'lax',
    expires: expiresAt
  }
  const userJWT = await encodeJWT({
    token: {
      id: user.id,
      name: user.name,
      email: user.email
    },
    secret
  })
  const me = await prisma.user.findUnique({ where: { id: token.id } })
  const tokenJWT = await encodeJWT({ token, secret })
  // NOTE: why can't I put this in a function with a for loop?!
  res.appendHeader('Set-Cookie', cookie.serialize(`multi_auth.${user.id}`, userJWT, cookieOptions))
  res.appendHeader('Set-Cookie', cookie.serialize(`multi_auth.${me.id}`, tokenJWT, cookieOptions))
  res.appendHeader('Set-Cookie',
    cookie.serialize('multi_auth',
      JSON.stringify([
        { id: user.id, name: user.name, photoId: user.photoId },
        { id: me.id, name: me.name, photoId: me.photoId }
      ]),
      { ...cookieOptions, httpOnly: false }))
  // don't switch accounts, we only want to add. switching is done in client via "pointer cookie"
  return token
}
return null
```

With these `multi_auth.*` cookies set, we can now show accounts to the user to which they can switch. The `multi_auth` 
cookie is not HTTP only since we want to access it from JS. This shouldn't be a problem since this is only for reading which accounts can be switched to and render this view:

2023-11-19-022224_499x506_scrot

On click, another cookie `multi_auth.user-id` is created via JS. This cookie is read inside a middleware in the backend to replace the session cookie that will be used to determine the user in the backend:

_middleware.js:_
```js
const multiAuthMiddleware = (request) => {
  // switch next-auth session cookie with multi_auth cookie if cookie pointer present
  const userId = request.cookies?.get('multi_auth.user-id')?.value
  const sessionCookieName = '__Secure-next-auth.session-token'
  const hasSession = request.cookies?.has(sessionCookieName)
  if (userId && hasSession) {
    const userJWT = request.cookies.get(`multi_auth.${userId}`)?.value
    if (userJWT) request.cookies.set(sessionCookieName, userJWT)
  }
  const response = NextResponse.next({ request })
  return response
}
```

**Showcase**

https://github.com/stackernews/stacker.news/assets/27162016/49f9ff4f-dc4a-4e46-bcd2-966be20fe66b

**TODO**

- [x] switching to anon
- [x] UI to go from anon back to session
- add `multiAuth` param to all login methods
  - [x] lightning auth
  - [x] nostr auth
  - [ ] ~~email auth~~ _(decided to remove this auth method, see https://github.com/stackernews/stacker.news/pull/644#issuecomment-1822106605)_
  - [ ] ~~github auth~~ _(decided to remove this auth method, see https://github.com/stackernews/stacker.news/pull/644#issuecomment-1822106605)_
  - [ ] ~~twitter auth~~ _(decided to remove this auth method, see https://github.com/stackernews/stacker.news/pull/644#issuecomment-1822106605)_
- [x] testing
- [x] replace hardcoded link for `add account` image
- [x] handle case where account is added which doesn't exist yet _(this will simply create a new account now but current account will never be updated if `multiAuth` is used)_
- [ ] ~~also refresh JWTs in `multi_auth.*`~~ (?) _(how?)_
- [x] think more about if using this "cookie pointer" approach is secure _(found an [article](https://connect2id.com/products/server/docs/guides/select-account) which also mentions "cookie pointers" - and I thought I invented that term, lol)_
- [x] middleware is now applied to all routes ... performance impact? _(nah, i think it's okay)_
- [x] [fix bug when looking at item and switching between anon and user](https://github.com/stackernews/stacker.news/pull/644#issuecomment-1820214123) _("fixed" in 0c8893c)_
- [x] on logout, only delete session token for current logged in account and move user to next available account

### DIFF
--- a/components/fee-button.js
+++ b/components/fee-button.js
@@ -64,7 +64,7 @@ export function postCommentUseRemoteLineItems ({ parentId } = {}) {
 export function FeeButtonProvider ({ baseLineItems = {}, useRemoteLineItems = () => null, children }) {
   const [lineItems, setLineItems] = useState({})
   const [disabled, setDisabled] = useState(false)
-  const me = useMe()
+  const { me } = useMe()
 
   const remoteLineItems = useRemoteLineItems()
 
@@ -115,7 +115,7 @@ function FreebieDialog () {
 }
 
 export default function FeeButton ({ ChildButton = SubmitButton, variant, text, disabled }) {
-  const me = useMe()
+  const { me } = useMe()
   const { lines, total, disabled: ctxDisabled, free } = useFeeButton()
   const feeText = free
     ? 'free'
--- a/components/me.js
+++ b/components/me.js
@@ -8,10 +8,14 @@ export const MeContext = React.createContext({
 })
 
 export function MeProvider ({ me, children }) {
-  const { data } = useQuery(ME, SSR ? {} : { pollInterval: FAST_POLL_INTERVAL, nextFetchPolicy: 'cache-and-network' })
+  const { data, refetch } = useQuery(ME, SSR ? {} : { pollInterval: FAST_POLL_INTERVAL, nextFetchPolicy: 'cache-and-network' })
+  // this makes sure that we always use the fetched data if it's null.
+  // without this, we would always fallback to the `me` object
+  // which was passed during page load which (visually) breaks switching to anon
+  const futureMe = data?.me ?? (data?.me === null ? null : me)
 
   return (
-    <MeContext.Provider value={data?.me || me}>
+    <MeContext.Provider value={{ me: futureMe, refreshMe: refetch }}>
       {children}
     </MeContext.Provider>
   )
--- a/middleware.js
+++ b/middleware.js
@@ -128,8 +128,6 @@ export const config = {
         { type: 'header', key: 'next-router-prefetch' },
         { type: 'header', key: 'purpose', value: 'prefetch' }
       ]
-    },
-    // account switching
-    '/api/graphql', '/_next/data/(.*)'
+    }
   ]
 }
--- a/middleware.js
+++ b/middleware.js
@@ -69,44 +69,6 @@ function referrerMiddleware (request) {
   return response
 }
 
-// TODO: use this middleware to switch between multiple auth accounts
-function multiAuthMiddleware (request) {
-  // switch next-auth session cookie with multi_auth cookie if cookie pointer present
-
-  // is there a cookie pointer?
-  const cookiePointerName = 'multi_auth.user-id'
-  const hasCookiePointer = request.cookies?.has(cookiePointerName)
-  // is there a session?
-  const sessionCookieName = request.secure ? '__Secure-next-auth.session-token' : 'next-auth.session-token'
-  const hasSession = request.cookies?.has(sessionCookieName)
-
-  if (!hasCookiePointer || !hasSession) {
-    // no session or no cookie pointer. do nothing.
-    return NextResponse.next({ request })
-  }
-
-  const userId = request.cookies?.get(cookiePointerName)?.value
-  if (userId === 'anonymous') {
-    // user switched to anon. only delete session cookie.
-    request.cookies.delete(sessionCookieName)
-    return NextResponse.next({ request })
-  }
-
-  const userJWT = request.cookies.get(`multi_auth.${userId}`)?.value
-  if (!userJWT) {
-    // no multi auth JWT found
-    return NextResponse.next({ request })
-  }
-
-  if (userJWT) {
-    // multi auth JWT found in cookie that pointed to by cookie pointer that is different to current session cookie.
-    request.cookies.set(sessionCookieName, userJWT)
-    return NextResponse.next({ request })
-  }
-
-  return NextResponse.next({ request })
-}
-
 export function middleware (request) {
   const resp = referrerMiddleware(request)
 
--- a/middleware.js
+++ b/middleware.js
@@ -69,6 +69,44 @@ function referrerMiddleware (request) {
   return response
 }
 
+// TODO: use this middleware to switch between multiple auth accounts
+function multiAuthMiddleware (request) {
+  // switch next-auth session cookie with multi_auth cookie if cookie pointer present
+
+  // is there a cookie pointer?
+  const cookiePointerName = 'multi_auth.user-id'
+  const hasCookiePointer = request.cookies?.has(cookiePointerName)
+  // is there a session?
+  const sessionCookieName = request.secure ? '__Secure-next-auth.session-token' : 'next-auth.session-token'
+  const hasSession = request.cookies?.has(sessionCookieName)
+
+  if (!hasCookiePointer || !hasSession) {
+    // no session or no cookie pointer. do nothing.
+    return NextResponse.next({ request })
+  }
+
+  const userId = request.cookies?.get(cookiePointerName)?.value
+  if (userId === 'anonymous') {
+    // user switched to anon. only delete session cookie.
+    request.cookies.delete(sessionCookieName)
+    return NextResponse.next({ request })
+  }
+
+  const userJWT = request.cookies.get(`multi_auth.${userId}`)?.value
+  if (!userJWT) {
+    // no multi auth JWT found
+    return NextResponse.next({ request })
+  }
+
+  if (userJWT) {
+    // multi auth JWT found in cookie that pointed to by cookie pointer that is different to current session cookie.
+    request.cookies.set(sessionCookieName, userJWT)
+    return NextResponse.next({ request })
+  }
+
+  return NextResponse.next({ request })
+}
+
 export function middleware (request) {
   const resp = referrerMiddleware(request)
 
@@ -128,6 +166,8 @@ export const config = {
         { type: 'header', key: 'next-router-prefetch' },
         { type: 'header', key: 'purpose', value: 'prefetch' }
       ]
-    }
+    },
+    // account switching
+    '/api/graphql', '/_next/data/(.*)'
   ]
 }
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -137,9 +137,10 @@ function setMultiAuthCookies (req, res, { id, jwt, name, photoId }) {
 }
 
 async function pubkeyAuth (credentials, req, res, pubkeyColumnName) {
-  const { k1, pubkey, multiAuth: multiAuthParam } = credentials
-  // multiAuth query param is a string
-  const multiAuth = typeof multiAuthParam === 'string' ? multiAuthParam === 'true' : !!multiAuthParam
+  const { k1, pubkey } = credentials
+
+  const { body } = req.body
+  const multiAuth = typeof body.multiAuth === 'string' ? body.multiAuth === 'true' : !!body.multiAuth
 
   try {
     const lnauth = await prisma.lnAuth.findUnique({ where: { k1 } })
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -112,7 +112,7 @@ function getCallbacks (req, res) {
   }
 }
 
-function setMultiAuthCookies (req, res, { id, jwt, name, photoId, switchUser }) {
+function setMultiAuthCookies (req, res, { id, jwt, name, photoId }) {
   const b64Encode = obj => Buffer.from(JSON.stringify(obj)).toString('base64')
   const b64Decode = s => JSON.parse(Buffer.from(s, 'base64'))
 
@@ -138,10 +138,8 @@ function setMultiAuthCookies (req, res, { id, jwt, name, photoId, switchUser }) 
   }
   res.appendHeader('Set-Cookie', cookie.serialize('multi_auth', b64Encode(newMultiAuth), { ...cookieOptions, httpOnly: false }))
 
-  if (switchUser) {
-    // switch to user we just added
-    res.appendHeader('Set-Cookie', cookie.serialize('multi_auth.user-id', id, { ...cookieOptions, httpOnly: false }))
-  }
+  // switch to user we just added
+  res.appendHeader('Set-Cookie', cookie.serialize('multi_auth.user-id', id, { ...cookieOptions, httpOnly: false }))
 }
 
 async function pubkeyAuth (credentials, req, res, pubkeyColumnName) {
@@ -184,7 +182,7 @@ async function pubkeyAuth (credentials, req, res, pubkeyColumnName) {
         // update the cookies for multi-authentication.
         const secret = process.env.NEXTAUTH_SECRET
         const userJWT = await encodeJWT({ token: { id: user.id, name: user.name, email: user.email }, secret })
-        setMultiAuthCookies(req, res, { ...user, jwt: userJWT, switchUser: true })
+        setMultiAuthCookies(req, res, { ...user, jwt: userJWT })
         return token
       }
 
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -112,9 +112,10 @@ function getCallbacks (req, res) {
   }
 }
 
-function setMultiAuthCookies (req, res, { id, jwt, name, photoId }) {
+function setMultiAuthCookies (req, res, { id, jwt, name, photoId, switchUser }) {
   const b64Encode = obj => Buffer.from(JSON.stringify(obj)).toString('base64')
   const b64Decode = s => JSON.parse(Buffer.from(s, 'base64'))
+
   // default expiration for next-auth JWTs is in 1 month
   const expiresAt = datePivot(new Date(), { months: 1 })
   const cookieOptions = {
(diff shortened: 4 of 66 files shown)


ekzyis:
There is a bug where you can't switch between anon and a user if you're currently looking at an item:

https://files.ekzyis.com/public/sn/account_switching_item_bug.mp4

ekzyis:
Current state: https://files.ekzyis.com/public/sn/account_switching_002.mp4

Missing (see PR description):

- JWT refresh: Figuring out if I can hook into whatever NextAuth calls to refresh the session token.
- Multi auth support for Github, Email, Twitter: Going to be hard since I can't really test (maybe github) and their flows are very different. But let's see
- some more testing
- replace hardcoded link for add account image

ekzyis:
Mhh, okay, can't find how to access a response to set the required `multi_auth.*` cookies on successful authentication for the [Github](https://next-auth.js.org/providers/github), [Twitter](https://next-auth.js.org/providers/twitter) and [Email](https://next-auth.js.org/providers/email#smtp) provider. Looking at the source also didn't help. I dedcided to just remove these providers from the available auth methods if `multiAuth` is set in 77887ae.

Same problem with refreshing JWTs.

Putting this out of draft since everything else works, so I think we can ship as is.

components/fee-button.js
```diff
@@ -36,8 +36,7 @@ export function postCommentBaseLineItems ({ baseCost = 1, comment = false, me })
   }
 }
 
-export function postCommentUseRemoteLineItems ({ parentId, me } = {}) {
-  if (!me) return () => {}
```
ekzyis, line 40: I hope this doesn't break anything but afaict, it doesn't. It just does unnecessary requests since anon has fixed fees.

huumn: This isn't a hook though. It's a function that returns a hook

ekzyis: > It's a function that returns a hook

Yes but conditionally

ekzyis:
Reminder for myself: check if there is still an error (there probably is) if you're on `/settings` and then you switch to anon

components/me.js
```diff
@@ -8,15 +8,25 @@ export const MeContext = React.createContext({
...
 
 export function useMe () {
-  return useContext(MeContext)
+  const { me } = useContext(MeContext)
+  return me
+}
+
+export function useMeRefresh () {
+  const { refetch } = useContext(MeContext)
+  return refetch
```
huumn, line 31: Just a nit but it might be better to either refactor `useMe` to return the full context object *or* have a `useMeFull` that returns the full context object so that we can avoid creating more getters for the context.

middleware.js
```diff
@@ -13,6 +13,47 @@ export function middleware (request) {
...
+  }
+
+  if (userJWT) {
+    // multi auth JWT found in cookie that pointed to by cookie pointer that is different to current session cookie.
+    request.cookies.set(sessionCookieName, userJWT)
+    return NextResponse.next({ request })
+  }
+
+  return NextResponse.next({ request })
+}
```
huumn, line 51: You say in the PR comment

> On click, another cookie multi_auth.user-id is created via JS. This cookie is read inside a middleware in the backend to replace the session cookie that will be used to determine the user in the backend.

Are we not able to make this switch, replacing the session cookie, on the client? Just want to make sure we *need* this middleware.

pages/api/auth/[...nextauth].js
```diff
@@ -36,6 +38,16 @@ function getCallbacks (req) {
...
 
+      // response is only defined during signup/login
+      if (req && res) {
+        req = new NodeNextRequest(req)
+        res = new NodeNextResponse(res)
+        const secret = process.env.NEXTAUTH_SECRET
+        const jwt = await encodeJWT({ token, secret })
+        const me = await prisma.user.findUnique({ where: { id: token.id } })
+        setMultiAuthCookies(req, res, { ...me, jwt })
+      }
```
Comment (huumn) [inline pages/api/auth/[...nextauth].js:49]: Can you confirm this doesn't run on every request? We don't want every request to incur an additional roundtrip to the database.

components/switch-account.js
```diff
@@ -0,0 +1,155 @@
...
+          <AnonAccount />
+          {
+            accounts.map((account) => <Account key={account.id} account={account} />)
+          }
+          <AddAccount />
+        </div>
+      </div>
+    </>
+  )
+}
```
huumn, line 155: We probably want to reuse the [user list](https://github.com/stackernews/stacker.news/blob/master/components/user-list.js) component without stats here if possible here. Ideally we only have one view for listing users.

huumn (changes requested):
Looks great!

I added some high level remarks. After we talk/work through those, I'll do another more detailed pass.

middleware.js
ekzyis: > Are we not able to make this switch, replacing the session cookie, on the client? Just want to make sure we need this middleware.

No since it's not accessible from JS

components/me.js
ekzyis: Makes sense, will do!

pages/api/auth/[...nextauth].js
Comment (ekzyis) [reply pages/api/auth/[...nextauth].js:49]: Yes, it's only run during login or signup.

components/me.js
ekzyis: Decided to go with the backwards incompatible change since I think a simple search and replace is enough to prevent bugs: bf7ec18

components/switch-account.js
ekzyis: Good idea! We could use the stats bar for other stuff like how many notifications an account has in the future.

Before:

sn ekzyis com_sn_test02(iPhone SE)

Now:

sn ekzyis com_(iPhone SE)

ekzyis:
I'll fix the conflicts, give me a sec

huumn:
No rush. This won't until I ship territories first

components/fee-button.js
ekzyis: > but I believe this is already conditional

Yes, it is. That's what my change is trying to fix since the bug would manifest when switching to anon. The user report shows that the bug _already_ manifests in some cases.  

Example:

This is the full function:

```js
export function postCommentUseRemoteLineItems ({ parentId, me } = {}) {
  if (!me) return () => {}
  const query = parentId
    ? gql`{ itemRepetition(parentId: "${parentId}") }`
    : gql`{ itemRepetition }`

  return function useRemoteLineItems () {
    const [line, setLine] = useState({})

    const { data } = useQuery(query, SSR ? {} : { pollInterval: 1000, nextFetchPolicy: 'cache-and-network' })

    useEffect(() => {
      const repetition = data?.itemRepetition
      if (!repetition) return setLine({})
      setLine({
        itemRepetition: {
          term: <>x 10<sup>{repetition}</sup></>,
          label: <>{repetition} {parentId ? 'repeat or self replies' : 'posts'} in 10m</>,
          modifier: (cost) => cost * Math.pow(10, repetition)
        }
      })
    }, [data?.itemRepetition])

    return line
  }
}
```

`return function useRemoteLineItems ()` is a hook which is only returned if we don't already return `() => {}`. That's what I meant with "but conditionally":

>> It's a function that returns a hook
>
> Yes but conditionally

So yes, this is conditional. That the user reported it means that the bug actually manifests already, too. Not only with my changes, assuming I wouldn't remove this line: `if (!me) return () => {}` (which what this conversation here is about)

ekzyis: Sorry for the detailed and  thus maybe snarky answer, I feel like we've been talking past each other for several replies already now, lol :)

huumn:
For some reason this doesn't work for me:

1. can't switch to anon
2. when I go to login, only some providers are listed

https://github.com/stackernews/stacker.news/assets/34140557/3da75c20-c66b-485f-9414-42c86352b412

ekzyis:
> For some reason this doesn't work for me:
> 
> 1. can't switch to anon

Mhh, I had this error before. I think I fixed it by clearing all my session cookies since it's related to a cookie being HTTP only when it should not have been set as HTTP only (by old code or some other bug?) iirc. Not sure though, need to check.

> 2. when I go to add account only some providers are listed

I wasn't able to hook into the OAuth login process and Email auth (and was hard to test locally). So I thought - at least for the initial version - we can provide only Login with Lightning and Nostr.

> This might need migration logic to be added for stackers that are already logged in if that wasn't tested.

Mhh, I didn't think too much about this but I think this should be 100% backwards compatible. But I will test.

ekzyis:
Did a full test (afaict) again.

https://files.ekzyis.com/public/sn/account_switching_full_test.mp4

1. 00:00 - 00:15 -- Login sets `multi_auth` cookie which is a base64 encoded list of all available accounts and is accessible from JS (HTTP only set to false) and `multi_auth.<userid>` cookie with HTTP only set to true containing the JWT for account switching to this user.

2. 00:27 - 00:32 -- Switching to anon set the pointer cookie `multi_auth.user-id` to `anonymous`. This tells the backend to not use any JWT.

3. 00:33 - 00:56 -- adding a new account updates `multi_auth` cookie and adds a new `multi_auth.<userid>` cookie for that user.

4. 00:59 - 01:06 -- switching between accounts (anon or not anon)

5. 01:07 - 01:13 -- logging out switches to next available account and deletes `multi_auth.<userid>` cookie of completely logs you out (same as before)

6. 01:26 - 02:28 -- testing backwards compatibility by simulating a user that logged in without account switching deployed and then uses account switching

7. 02:28 -- ok there is indeed a bug lol. new account does not show up in the list.

ekzyis:
Putting in draft until bug is fixed

ekzyis:
Sessions that existed before we deployed account switching should not also be able to use account switching immediately. We now sync the cookie from their session if there is no `multi_auth` cookie. See https://github.com/stackernews/stacker.news/pull/644/commits/6839066466c67c33486119c384929445a32abaaf


https://github.com/stackernews/stacker.news/assets/27162016/6dd4d767-b7b7-49dc-8002-7e8939db23df

huumn:
So this is out of draft now?

ekzyis:
> So this is out of draft now?

Going to fix conflicts now, just had a call with someone from Alby

huumn:
This is still not working for me. I even logged out and deleted all cookies.

1. when I click on anon it gets selected but nothing changes
    a.  even after waiting a very long time
    b. my user also stays selected
Screenshot 2023-12-20 at 3 46 15 PM
2. when I attempt to add another account, no account is added and instead I get a `?error=CredentialsSignin`
3. here's my cookie state
Screenshot 2023-12-20 at 3 46 38 PM
4. I'm using brave.
5. When I logout, then it does switch me to anon

ekzyis:
Mhh, interesting, why are your NextJS cookies named differently than mine?

My code assumes that NextJS uses `__Secure-next-auth.session-token` for the session cookie name. See [here](https://github.com/stackernews/stacker.news/pull/644/files#diff-e7145ea3e4f5db1f4bc1c5be397238ee0c522c81afa6a7b8228261f54aa4e704R23) and [here](https://github.com/stackernews/stacker.news/pull/644/files#diff-e2617491e6bcdfec9d6985f1316673b9b7c03ee0ace47aaa8f67a3a0fa748d76R14).

This could cause your problems but I am not 100% sure if that's all there is to it.

Should we have a (quick) call today so we can work through the flow and see where it goes wrong?

ekzyis:
Okay, found out why this doesn't work on localhost.

With 1845db2, I actually completely broke login on non-HTTPS sites since I am forcing the session cookie to have the [`Secure`](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.5) attribute set. Additionally, I am setting the`__Secure-` cookie prefix so the browser can't even read the cookie:

> The __Secure- prefix makes a cookie accessible from HTTPS sites only. A HTTP site can not read or update a cookie if the name starts with __Secure-. This protects against the attack we earlier described, where an attacker uses a forged insecure site to overwrite a secure cookie.

-- https://www.sjoerdlangkemper.nl/2017/02/09/cookie-prefixes/

I am also setting `Secure` on all cookies required for multi authentication. So this explains a lot.

I am going to try to make this work on localhost / HTTP without introducing vulnerabilities.

ekzyis:
Also works on HTTP now:


https://github.com/stackernews/stacker.news/assets/27162016/9c39ad41-2cf2-4896-9e3c-8de7c5358055

ekzyis:
Account switching could also be useful for local development

Need to rebase the code and see if I can remove the code in the middleware and replace it with code similar to #915

ekzyis:
This PR is rebased now and switching between user and anon works. However, adding a new account does not work yet since `multiAuth` isn't set in the backend for some reason.

TODO:

- [ ] fix `multiAuth` not set in backend
- [ ] fix alignment of `selected`
- [ ] use something prettier instead of `+ add account`

ekzyis:
> Update: Unfortunately, this doesn't improve anon UX a lot since you can't pay from your account balance yet.

-- https://github.com/stackernews/stacker.news/pull/644#issuecomment-1822106605

If we allow to share attached wallets between switched accounts (including if you _switched_ to anon but not if you're entirely logged out), that would fix this.

ekzyis:
Current state:

https://github.com/user-attachments/assets/d189bc9d-72b6-42e9-8f9f-2b6a8b861710

Just noticed the flash from "Login" to "Signup" when switching to anon. Investigating.

ekzyis:
Mhh, I should probably refetch specifically the queries that I want to refetch instead of blocking the ones that I don't (whitelist vs blacklist approach).

ekzyis:
The only real thing left here is to add the OAuth providers but I'm having a hard time to get it working.

If it's not linked to any account, it links the auth method to the existing account but I would prefer it to create a new separate account with that provider or throws an error if it's already linked to another account (instead of issuing a JWT jus for that account for switching between).

Essentially, I don't know how to use the [OAuth options](https://next-auth.js.org/configuration/providers/oauth#options) to tell NextAuth to handle the case where `multiAuth` is set differently.

Other than that, this is ready for review.

huumn:
The only way to do something like that with oauth would be to hack around nextjs in some way. To get account linking to work at all, I had to read all the relevant nextjs source. It was designed to link on email address matches between providers iirc.

I haven’t read the code, but perhaps we can “logout” the user (make cookies appear logged out to nextjs) when they go to create a new account so there’s nothing to link to.

ekzyis:
> I haven’t read the code, but perhaps we can “logout” the user (make cookies appear logged out to nextjs) when they go to create a new account so there’s nothing to link to.

Mhh, this could work using something similar to [`credentials: "omit"` for the fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#including_credentials). But the response must still be able to set cookies.

Mhhh ...

ekzyis:
Okay, I didn't really try with email, only looked at [the options](https://next-auth.js.org/providers/email) but I suspect it's going to be a mess like OAuth so let's skip this for now

huumn:
A lot of files have changed in this and afaict the pr description is out of date. Would you mind pointing me to files/functions/components I should be reviewing?

ekzyis:
> A lot of files have changed in this and afaict the pr description is out of date. Would you mind pointing me to files/functions/components I should be reviewing?

I've created #1386 to make the change set here a little smaller.

Another repetitive change is replacing `const me = useMe()` with `const { me } = useMe()`. That's why there are so many files with 2+-.

Here is the list of files sorted by their changes:

```git
components/account.js               | 207 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
pages/api/auth/[...nextauth].js     |  99 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------
components/nav/common.js            |  93 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++----------------------
pages/api/signout.js                |  61 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
components/login.js                 |  55 +++++++++++++++++++++++++++++++++++-----------------
pages/api/graphql.js                |  39 +++++++++++++++++++++++++++++++++++++
pages/_app.js                       |  35 ++++++++++++++++++---------------
package-lock.json                   |  35 ++++++++++++++++++++++++++++++---
components/user-list.js             |  30 ++++++++++++++++++++++++++++-
pages/login.js                      |  24 ++++++++++++++++-------
lib/apollo.js                       |  23 ++++++++++++++++++----
pages/settings/index.js             |  14 ++++++++------
components/item-info.js             |  14 +++++---------
components/lightning-auth.js        |  12 ++++++------
components/item-act.js              |  11 ++++++++---
components/nostr-auth.js            |   9 +++++----
pages/wallet/index.js               |   8 ++++----
components/upvote.js                |   8 ++++----
components/me.js                    |   8 ++++++--
api/ssrApollo.js                    |   8 +++++++-
wallets/index.js                    |   6 +++---
fragments/users.js                  |   6 +++---
components/banners.js               |   6 +++---
api/resolvers/user.js               |   5 +++--
components/user-header.js           |   4 ++--
components/territory-payment-due.js |   4 ++--
components/share.js                 |   4 ++--
components/post.js                  |   4 ++--
components/nav/sticky-bar.js        |   4 ++--
components/nav/mobile/top-bar.js    |   4 ++--
components/nav/mobile/offcanvas.js  |   4 ++--
components/login-button.js          |   4 ++--
components/fee-button.js            |   4 ++--
pages/withdrawals/[id].js           |   2 +-
pages/referrals/[when].js           |   2 +-
pages/[name]/index.js               |   2 +-
components/wallet-logger.js         |   2 +-
components/territory-transfer.js    |   2 +-
components/territory-header.js      |   2 +-
components/territory-form.js        |   2 +-
components/search.js                |   2 +-
components/reply.js                 |   2 +-
components/price.js                 |   2 +-
components/poll.js                  |   2 +-
components/poll-form.js             |   2 +-
components/payment.js               |   2 +-
components/pay-bounty.js            |   2 +-
components/nav/mobile/second-bar.js |   2 +-
components/nav/mobile/footer.js     |   2 +-
components/nav/desktop/top-bar.js   |   2 +-
components/media-or-link.js         |   2 +-
components/logger.js                |   2 +-
components/link-form.js             |   2 +-
components/item.js                  |   2 +-
components/item-full.js             |   2 +-
components/hidden-wallet-summary.js |   2 +-
components/form.js                  |   2 +-
components/discussion-form.js       |   2 +-
components/comment.js               |   2 +-
components/bounty-form.js           |   2 +-
components/autowithdraw-shared.js   |   2 +-
components/adv-post-form.js         |   2 +-
api/typeDefs/user.js                |   2 +-
package.json                        |   1 +
64 files changed, 728 insertions(+), 185 deletions(-)
```

Most important changes are all the files above `package-lock.json` and `pages/login.js` + `api/ssrApollo.js` (related changes) and `lib/apollo.js` + `components/item-act.js` (related changes). You can basically ignore all the files with only 6 changes or less.

huumn:
QA looks good. I'll do a line by line tomorrow.

1. Is it intentional to have to close the modal once an account is picked? I thought it was a bug when I first switched.
2. I can't tell why the last account is white in the list.

Screenshot 2024-09-11 at 8 24 55 PM

huumn:
> I think it's related to if you clicked the link before or not

It didn't change before or after I clicked it strangely.

huumn (changes requested):
I got this when switching while editing a post. 

Keeping the page the same while switching is a pretty difficult requirement. There might be other such instances.

If it were me I'd probably just go nuclear do a refresh when switching ... so that we don't have to think of account switching anytime we add a personalized page. 

Screenshot 2024-09-12 at 11 21 52 AM

huumn:
For reference (not saying this exactly what we should do), when you switch on X it takes you to `/home` regardless of where you were in the other account. It sucks because you lose your place.

For us, I think it might be better to just refresh on switch.

ekzyis:
> For us, I think it might be better to just refresh on switch.

There you go, the modal is now also closed on switch, lol

PR #650, opened 2023-11-22
Create landing-page.css

kerooke:


kerooke:
Added some new styling which I think is cleaner than the existing styling, but will put up another file that re-uses SN colors and styles for consistency.

PR #653, opened 2023-11-23
Unsubscribe on signout

### LINKED ISSUE #625: Receiving push notifications even after logging out of the site
I keep receiving push notifications on my phone even after logging out of the site.

EDIT: I know I could login and disable notifications in the settings page but that should not be necessary because I do want to receive notifications on my other devices.

image

image

ekzyis:
Fix #625 

https://github.com/stackernews/stacker.news/assets/27162016/b10721c7-aaae-498c-bf84-115f5f006adb

### DIFF
--- a/components/header.js
+++ b/components/header.js
@@ -26,6 +26,7 @@ import AnonIcon from '../svgs/spy-fill.svg'
 import Hat from './hat'
 import HiddenWalletSummary from './hidden-wallet-summary'
 import { clearNotifications } from '../lib/badge'
+import { useServiceWorker } from './serviceworker'
 
 function WalletSummary ({ me }) {
   if (!me) return null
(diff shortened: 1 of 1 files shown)


components/header.js
```diff
@@ -84,6 +85,7 @@ function NotificationBell () {
 }
 
 function NavProfileMenu ({ me, dropNavKey }) {
+  const { registration: swRegistration, togglePushSubscription } = useServiceWorker()
```
huumn, line 88: Is it safe to destructure this? Is it always defined?

We have some folks on iphone 7 (ttps://stacker.news/items/328115/r/k00b) which uses a really old safari so just wanted to double check

ekzyis: Yes, should be always defined since `ServiceWorkerProvider` always returns an object with these keys:

```js
const contextValue = useMemo(() => ({
  registration,
  support,
  permission,
  requestNotificationPermission,
  togglePushSubscription
}), [registration, support, permission, requestNotificationPermission, togglePushSubscription])
```

huumn: Sweet thanks!

Issue #640, opened 2023-11-17
Add latest tx fees to page header price carousel


SatsAllDay:
**Is your feature request related to a problem? Please describe.**
No existing problem, I just had the idea of displaying the latest fee rate as one of the price carousel options, alongside price, block height, etc. I thought this would be an interesting metric to see since lately the on-chain fees have been pretty high

**Describe the solution you'd like**
Another option that shows the fee rate for the most recent block. I assume we can get this from SN's node via LND but I have not investigated further.

**Describe alternatives you've considered**
N/A

**Additional context**
https://stacker.news/items/315636/r/WeAreAllSatoshi

huumn:
That'd be fun! They are just so crazy right now

SatsAllDay:
I think this is done, right?

ekzyis:
> I think this is done, right?

_nothing to see here_

PR #658, opened 2023-12-05
Add chain fees to price carousel

st4rgut24:
This PR adds the latest on-chain transaction fees to the price carousel. 

It addresses the issue: https://github.com/stackernews/stacker.news/issues/640


api/resolvers/chainFee.js
```diff
@@ -0,0 +1,37 @@
+import lndService from 'ln-service'
+import lnd from '../lnd'
+
+const cache = new Map()
+const expiresIn = 1000 * 30 // 30 seconds in milliseconds
+
+async function fetchChainFeeRate () {
+  let chainFee = 0
+  try {
+    const fee = await lndService.getChainFeeRate({ lnd })
```
ekzyis, line 10: I believe this is only the estimate since it says "estimate" here:

> **getChainFeeRate**
> Get chain fee rate estimate

-- https://www.npmjs.com/package/ln-service#getchainfeerate

Mentioning this, because when I first saw the ticket, I thought if I am going to work on this, I am going to use the median fee rate of the recent block if possible.

But not sure if possible using LND RPC calls.

So I think this is also fine :)

ekzyis (approved):
Frontend looks good :)

https://github.com/stackernews/stacker.news/assets/27162016/8a2e45ed-bc52-46ee-a795-6928bd2f6ef3

Code looks good :)

All good!

Except we might need to refactor the whole carousel at some point now since everyone is just copying the previous copied implementation which is totally understandable :)

But all good, this is supposed to be a feature PR, not a refactor PR :)

st4rgut24:
Hey! My nym is @souljaboy

Thanks! I'll do that

PR #668, opened 2023-12-07
Use textarea for unintended DMs

### LINKED ISSUE #648: Make LN Addr Withdrawal comments field a textarea instead of an inputtext
It's hard to compose a long message in the comments section of a LN Addr withdrawal using our UI since the comments text area is just an input field, not a text area.

input field -> text area

N/A

**Additional context**
N/A

ekzyis:
2023-12-07-062034_1920x1080_scrot

Close #648

### DIFF
--- a/pages/wallet.js
+++ b/pages/wallet.js
@@ -430,6 +430,7 @@ export function LnAddrWithdrawal () {
                   <>
                     {addrOptions.commentAllowed &&
                       <Input
+                        as='textarea'
                         label={<>comment <small className='text-muted ms-2'>optional</small></>}
                         name='comment'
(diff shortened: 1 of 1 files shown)


ekzyis:
> 🎉

Oh ****

We should disable `Shift+Enter` = Submit there... I already sent some accidental unfinished messages this way ...

ekzyis:
Oh, I wasn't on the latest branch where it's already a textarea.

Interestingly, if you use `as='textarea'`, Shift+Enter no longer works for submitting. But without, it works.

My educated guess would be that textarea component overrides the behavior to explicitly enter line breaks :)

Thank you mysterious HTML spec dude who did this!

Issue #691, opened 2023-12-17
Privacy leak regarding wallet balance


ekzyis:
https://stacker.news/items/355256

_not going to write more here since this is public and I don't want to give more information to attackers how to exploit this than necessary. doesn't Github have a feature to mark some tickets (like vulns) as internal? maybe I just don't have the permissions_

~~I will try to fix asap~~

Update: Ok, I think this is not as severe as I initially thought for reasons I don't want to disclose here.

PR #693, opened 2023-12-17
Mix error messages to hide user balance

ekzyis:
One possible (temporary?) solution for #691 would be to not respond with a error message which tells that creating this invoice would exceed a limit.

By just responding with a different error message, the attacker cannot distinguish between "too many invoices" or "invoice would exceed limit". This isn't ideal but still a lot better imo.

~~I haven't tested this or checked if this would break something but if our code is good, we're not relying on error messages so this change should be fine.~~

edit: Okay, I will test this, don't want to rely on @huumn superhuman skills to know exactly the impact of every line of code changed just by looking at it, lol

TODO:

- [x] test

---

Okay, pretty sure this doesn't break anything. I searched for all kind of substrings of the previous error message, and tested depositing and withdrawing. All good.

### DIFF
--- a/api/resolvers/serial.js
+++ b/api/resolvers/serial.js
@@ -3,8 +3,7 @@ import retry from 'async-retry'
 import Prisma from '@prisma/client'
 import { settleHodlInvoice } from 'ln-service'
 import { createHmac } from './wallet'
-import { msatsToSats, numWithUnits } from '../../lib/format'
-import { BALANCE_LIMIT_MSATS } from '../../lib/constants'
+import { msatsToSats } from '../../lib/format'
 
 export default async function serialize (models, ...calls) {
   return await retry(async bail => {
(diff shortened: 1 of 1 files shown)


kepford:
Fix looks good as a stop gap.

kepford:
Maybe a longer term solution is to show the original message if the user is funding their own SN wallet. Not sure how difficult that is to do but it would be helpful. Maybe even just posting a help message on the wallet funding page stating the limit. That way it is more obvious to a user trying to fund their wallet.

huumn:
Yep agreed. We are already getting people reporting this error and being confused.

ekzyis:
> show the original message if the user is funding their own SN wallet. Not sure how difficult that is to do but it would be helpful.

Mhh, since we can tell from the session who is trying to fund which wallet, I think this should be doable. Good idea!

PR #699, opened 2023-12-20
Add wallet limit banner

ekzyis:
https://stacker.news/items/356071

https://github.com/stackernews/stacker.news/assets/27162016/a016a608-3c23-499e-838a-a4c28631a313

2023-12-20-061636_1920x1080_scrot

2023-12-20-061622_1920x1080_scrot

_ignore actual wallet balance in screenshots, I manually set `limitReached` to true_

So basically, this adds a banner that only shows up if the wallet balance is over the limit. It can be dismissed on the front page but it will always appear in the wallet (if over the limit).


components/banners.js
```diff
@@ -65,3 +67,50 @@ export default function WelcomeBanner () {
     </Alert>
   )
 }
+
+export function WalletLimitBanner ({ dismissible }) {
+  // TODO refactor this since it's mostly the same as the code for the welcome banner
```
ekzyis, line 72: Don't want to mix feature + refactor in one PR.

kepford:
No familiar with the code base but this MR looks good to me. Nice demo video as well @ekzyis!

huumn:
ACK the concept but I'm not sure the banner should be on the homepage. It should definitely be on the wallet pages and maybe show an indicator on the wallet ... banners on the homepage always feel like an assault.

ekzyis:
> banners on the homepage always feel like an assault.

I agree that they are an assault but is this still a problem if you can dismiss them like the welcome banner?

I am open to remove it of course, just wanted to clarify that this isn't a banner that always shows up but only once.

huumn:
Maybe it can work like this:

1. no banner on homepage
2. when they go over the limit, turn their sats in the upper right hand corner orange
3. non-dismissable banner on the wallet page

ekzyis:
> Maybe it can work like this:
> 
> 1. no banner on homepage
> 2. when they go over the limit, turn their sats in the upper right hand corner orange
> 3. non-dismissable banner on the wallet page

Did that in d01164f now:


https://github.com/stackernews/stacker.news/assets/27162016/d13c198f-6e93-45b8-9221-873a6c62f2f5

PR #701, opened 2023-12-20
Request persistent storage

ekzyis:
Related to #411 

TODO:

- [ ] only prompt for persistent storage permission if notifications enabled

### DIFF
--- a/components/serviceworker.js
+++ b/components/serviceworker.js
@@ -99,7 +99,15 @@ export const ServiceWorkerProvider = ({ children }) => {
     if (pushSubscription) {
       return unsubscribeFromPushNotifications(pushSubscription)
     }
-    return subscribeToPushNotifications()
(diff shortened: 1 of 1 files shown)


huumn:
Does this only do this if/when notifications are enabled? I don't want it prompting people *potentially* for no reason. Best to scope it to where we need it.

ekzyis:
> Does this only do this if/when notifications are enabled? I don't want it prompting people _potentially_ for no reason. Best to scope it to where we need it.

It didn't prompt me but I guess I missed the Firefox part:

> Firefox requests permission from the user to use persistent storage. Chromium-based browsers give or deny persistence based on a [heuristic](https://web.dev/articles/persistent-storage#chrome_and_other_chromium-based_browsers) to determine the importance of the content for the user. One criteria for Google Chrome is, for example, PWA installation.

So good catch, the prompt (and other browsers) wasn't on my mind.

PR #704, opened 2023-12-24
Update index.js

meuamigopedro:
Key improvements:

	1.	Destructuring of Environment Variables: I used destructuring to make the code more concise and readable.
	2.	Checking Environment Variables: Added a check to ensure that all required environment variables are present. If any are missing, the script displays an error message and exits.
	3.	Improved Error Handling: Enhanced error handling by displaying detailed error messages and exiting the script if there is a problem during authentication or connection checking with LND.

### DIFF
--- a/api/lnd/index.js
+++ b/api/lnd/index.js
@@ -1,18 +1,25 @@
-import lndService from 'ln-service'
+import lndService from 'ln-service';
 
-const { lnd } = lndService.authenticatedLndGrpc({
-  cert: process.env.LND_CERT,
-  macaroon: process.env.LND_MACAROON,
-  socket: process.env.LND_SOCKET
-})
+// Destructure environment variables
+const { LND_CERT, LND_MACAROON, LND_SOCKET } = process.env;
 
-// Check LND GRPC connection
+// Check if all required environment variables are present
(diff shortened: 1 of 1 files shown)


ekzyis (commented):
> Destructuring of Environment Variables: I used destructuring to make the code more concise and readable.

To be honest, I don't think the code is _significantly_ more concise or readable now. Imho, the comments even add noise. Did you find this code to not be readable enough? If so, why? I want to understand where you are coming from.

> Checking Environment Variables: Added a check to ensure that all required environment variables are present. If any are missing, the script displays an error message and exits.

The script should not exit just because LND is not running imo - at least while developing. A check in production might make sense but afaik, we never needed one so far since these environment variables rarely change.

> Improved Error Handling: Enhanced error handling by displaying detailed error messages and exiting the script if there is a problem during authentication or connection checking with LND.

I don't see how the error messages are _significantly_ better now but I think yours are fine, too.

huumn:
Thanks so much for taking a stab at contributing. I don't think the changes you've added are worth merging, but please don't be discouraged from making more significant changes that add or improve functionality.

meuamigopedro:
Hi, friend! I appreciate the attention and would like to inform you that I've been absent due to Christmas preparations. I will continue to help and I'm available for whatever you need.

PR #707, opened 2023-12-24
Christmas zaps

ekzyis:
This PR adds "christmas zaps": the lightning strike is replaced with snow.

The snowflakes are animated using CSS animations. The javascript codes only creates `<div>` elements with the appropriate CSS class (`snow`) set.

I used this as reference: https://codepen.io/alphardex/pen/dyPorwJ

https://github.com/stackernews/stacker.news/assets/27162016/970fffb9-5f57-47ab-9c7e-008718dc0922

Checked for browser support of [CSS keyframes](https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes#browser_compatibility):

2023-12-24-221625_791x555_scrot

Update: I just realized this isn't visible in light mode, lol. I'll check if a very thin black border around the snow flakes makes sense in light mode.

### DIFF
--- a/components/header.js
+++ b/components/header.js
@@ -19,7 +19,7 @@ import LightningIcon from '../svgs/bolt.svg'
 import SearchIcon from '../svgs/search-line.svg'
 import BackArrow from '../svgs/arrow-left-line.svg'
 import { BALANCE_LIMIT_MSATS, SSR } from '../lib/constants'
-import { useLightning } from './lightning'
+import { useSnow } from './snow'
 import { HAS_NOTIFICATIONS } from '../fragments/notifications'
 import AnonIcon from '../svgs/spy-fill.svg'
 import Hat from './hat'
@@ -172,12 +172,12 @@ function StackerCorner ({ dropNavKey }) {
 
 function LurkerCorner ({ path }) {
   const router = useRouter()
-  const strike = useLightning()
+  const snow = useSnow()
 
   useEffect(() => {
     if (!window.localStorage.getItem('striked')) {
       const to = setTimeout(() => {
-        strike()
+        snow()
         window.localStorage.setItem('striked', 'yep')
       }, randInRange(3000, 10000))
       return () => clearTimeout(to)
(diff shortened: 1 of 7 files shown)


ekzyis:
Using grey snow for light mode now (527a3bd):

https://github.com/stackernews/stacker.news/assets/27162016/5df1478a-c557-4a23-b6c1-5d1b8ed542db

huumn:
Nice work. A few high level generic things I noticed and changed:

1. position should be fixed - we want these animations to display regardless of our scroll position. You were using `absolute`
2. when using keyframes and altering position: always prefer translated[X,Y,Z] over other positioning functions so that we're more likely get run on the GPU. Your falling animation was using `top`.
3. these continue to animating off screen even after the animation is done. We probably don't want to do that but the fastest thing I could add was for their opacity to go back to zero (which has a narrow chance of helping browser optimize them away)

ekzyis:
> 1. position should be fixed - we want these animations to display regardless of our scroll position. You were using `absolute`

Oh, interesting, guess I didn't try that. I was aware of that issue but I didn't find the correct CSS to fix that (and then I must have forgotten about it).

> when using keyframes and altering position: always prefer translated[X,Y,Z] over other positioning functions so that we're more likely get run on the GPU. Your wind animation was right but the falling animation was using top.

interesting, thanks for letting me know about the difference :eyes:

> these continue animating off screen even after the animation is done. We probably don't want to do that but the fastest thing I could add was for their opacity to go back to zero (which has a narrow chance of helping browser optimize them away)

Yeah, I struggled with deleting / "garbage collecting" them. But when I deleted one wave, the other waves got re-rendered which made them start from the top again. [I think the issue was related to this](https://stackoverflow.com/questions/58015502/react-update-element-of-array-without-rerendering-other-array-elements). I also tried using `useMemo` or `React.memo` or `shouldComponentUpdate` but no luck. But with "these continue animating off screen" you mean only the wind animation right? Since only that one is set to be infinite.

huumn:
> But with "these continue animating off screen" you mean only the wind animation right? Since only that one is set to be infinite.

Yes. It's just wind, but for a person who zaps a lot they'll have thousands of particles animating offscreen as they browse the app which will continue to animate as the they navigate.

Issue #524, opened 2023-09-26
Autowithdrawl to lightning address


huumn:
Basically allow lightning addressed people to autowithdrawl their account balance from SN.

This could work a number of ways, but we probably want it to leave some sats in wallet. Options:

A. conditionally, e.g. withdrawal anything over 12k sats, leaving 10k sats, and paying a max fee of 50 sats
B. unconditionally, e.g. always attempt to withdrawal any balance, even if it's just 1 sat, paying a max fee of 10 sats
C? direct funding, e.g. don't hold my sats at all and zap me directly

C is probably a little wacky to do if we want to take our zap sybil fee, but we might be able to cook up something with hold invoices. It's also unclear what to do on failure.

ekzyis:
> C? direct funding, e.g. don't hold my sats at all and zap me directly
> C is probably a little wacky to do if we want to take our zap sybil fee

Maybe we could use routing hints? As if the user uses a private channel with us. Then we could take our zap synbil fee as a routing fee

huumn:
If routing hints can specify a fee explicitly then that would work.

ekzyis:
> The elements contained in the field `r` are commonly referred to as _routing hints_. They allow the receiver to communicate an extra set of edges that may help the sender complete their payment. The hints are usually used when the receiver has some/all private channels, and they wish to guide the sender into this "unmapped" portion of the channel graph. A routing hint encodes effectively the same information that a normal `channel_update` message does. The update is itself packed into a single value with the following fields:
>
> The `pubkey` of the outgoing node in the edge (264 bits)
>
> The `short_channel_id` of the "virtual" edge (64 bits)
>
> The base fee (`fee_base_msat`) of the edge (32 bits)
>
> The proportional fee (`fee_proportional_millionths`) (32 bits)
>
> The CLTV expiry delta (`cltv_expiry_delta`) (16 bits)
>
> The final portion of the data segment is the set of feature bits that communicate to the sender the functionality needed to complete a payment. As an example, if a new payment type is added in the future that isn’t backward compatible with the original payment type, then the receiver can set a required feature bit to communicate that the payer needs to understand that feature to complete the payment.

-- [_Mastering the Lightning Network_, Chapter 15](https://github.com/lnbook/lnbook/blob/develop/15_payment_requests.asciidoc)

So I believe this could work.

However, there is one problem:

[...]

PR #715, opened 2023-12-28
Use bitcoin-connect for self-custodial zaps

### LINKED ISSUE #490: SN wallet integration for seamless payments

We want to let stackers use their SN account more seamlessly with the rest of the Lightning ecosystem.

[Nostr Wallet Connect](https://github.com/nostr-protocol/nips/blob/master/47.md) is one option for this.

LNDHub (#429) is another.

Support [NIP-47](https://github.com/nostr-protocol/nips/blob/master/47.md)

### LINKED ISSUE #533: Nostr Wallet Connect
A part of my dream of a non-custodial stacker news experience. It would be great to be able to zap posts without having to leave a balance on stacker news. This could be done by using [nostr wallet connect](https://github.com/nostr-protocol/nips/blob/master/47.md) where if the user's balance is 0, they would zap from their linked wallet instead.

This in combination with #524 would allow for a complete non-custodial experience for SN

ekzyis:
Close #533 Close #490

With this PR, stackers can connect any wallet that is supported by [Bitcoin Connect](https://github.com/getAlby/bitcoin-connect):

2023-12-28-025720_1920x1080_scrot

When connected, we can add this self-custodial balance to any sats they might hold in their custodial wallet:

2023-12-28-030050_1920x1080_scrot

2023-12-28-030057_1920x1080_scrot

If they now zap people and they don't have enough sats in their custodial wallet (these sats are spent first), they'll get the usual QR code. However, no further action is required by the user. Using the [WebLN](https://www.webln.guide/) API provided by Bitcoin Connect, we can now pay the invoice automatically:

https://github.com/stackernews/stacker.news/assets/27162016/6facb9da-b22d-48ac-9a2d-d0972a424841

TODO:

- [ ] test all connectors
  - [ ] alby
  - [ ] mutiny
  - [ ] umbrel
  - [ ] nwc
  - [x] lnbits
  - [ ] lnc
- [ ] improve payment UX since it's a lot slower than possible: replace invoice status polling via [`getInvoice`](https://www.npmjs.com/package/ln-service#getinvoice) with subscriptions via [`subscribeToInvoice`](https://www.npmjs.com/package/ln-service#subscribetoinvoice)
- [ ] improve payment UX by making it consistent with custodials zaps (don't even show the QR code if possible)
- [ ] improve error handling, for example if payment failed due to channel reserve
- [ ] consider channel reserve when showing balance to user (?)
- [ ] allow mixed payments from custodial wallet AND WebLN provider at the same time
- [ ] improve wallet UI (?)
- [ ] move `me.weblnSats` to `me.privates.weblnSats`
- [ ] is there a more secure way to [persist provider credentials across sessions than in `localStorage`](https://github.com/getAlby/bitcoin-connect/blob/master/src/state/boot.ts)?
- [ ] should we use a [reactive variable](https://www.apollographql.com/docs/react/local-state/reactive-variables/) for `weblnSats` instead of this (possibly hacky) `createRef` + `useImperativeHandle` combination?


components/header.js
```diff
@@ -33,7 +33,8 @@ function WalletSummary ({ me }) {
   if (me.privates?.hideWalletBalance) {
     return <HiddenWalletSummary abbreviate fixedWidth />
   }
-  return `${abbrNum(me.privates?.sats)}`
+  const sats = me.privates?.sats + me.weblnSats
```
rolznz, line 36: if either of these are undefined, I think it will cause a NaN value. Also, does it even make sense to combine the two values? you can only spend from one or the other

components/me.js
```diff
@@ -10,6 +10,10 @@ export const MeContext = React.createContext({
 export function MeProvider ({ me, children }) {
   const { data } = useQuery(ME, SSR ? {} : { pollInterval: 1000, nextFetchPolicy: 'cache-and-network' })
 
+  const futureMe = data?.me || me
+  // weblnSats is initially undefined on page load for some reason
```
rolznz, line 14: is the WebLN provider initialized yet?

components/qr.js
```diff
@@ -1,18 +1,23 @@
 import QRCode from 'qrcode.react'
 import { CopyInput, InputSkeleton } from './form'
 import InvoiceStatus from './invoice-status'
-import { requestProvider } from 'webln'
 import { useEffect } from 'react'
+import { useWebLN } from './webln'
 
 export default function Qr ({ asIs, value, webLn, statusVariant, description, status }) {
   const qrValue = asIs ? value : 'lightning:' + value.toUpperCase()
+  const { provider } = useWebLN()
```
rolznz, line 9: You use `useWebLN()` and also `webLn` is being passed in. Are they both needed? Note: Bitcoin Connect does not inject webln into the window by default.

components/webln.js
```diff
@@ -0,0 +1,106 @@
+import { createContext, createRef, useContext, useEffect, useImperativeHandle, useState } from 'react'
+
+const WebLNContext = createContext({})
+export const WebLNContextRef = createRef()
+
+const fetchWebLNProvider = async () => {
+  // sync provider from local storage
+  // TODO check if there is a better, more secure option
+  const configJson = window.localStorage.getItem('bc:config')
```
rolznz, line 9: If possible you should not directly access the storage like this, it's not part of the Bitcoin Connect public API

components/webln.js
```diff
@@ -0,0 +1,106 @@
...
+  if (configJson) {
+    const WebLNProviders = await import('@getalby/bitcoin-connect').then((mod) => mod.WebLNProviders)
+    const config = JSON.parse(configJson)
+    if (config.connectorType === 'lnbits') return new WebLNProviders.LnbitsWebLNProvider(config.lnbitsInstanceUrl, config.lnbitsAdminKey)
```
rolznz, line 13: why does stacker news need to care about the provider type? this is ideally handled by Bitcoin Connect. Stacker news should just use WebLN.

components/qr.js
rolznz: (and this won't work if `webLn` is undefined)

components/webln.js
```diff
@@ -0,0 +1,106 @@
...
+  const [info, setInfo] = useState(null)
+  const [balance, setBalance] = useState(0)
+
+  const initProvider = async (provider) => {
+    const WebLNProviders = await import('@getalby/bitcoin-connect').then((mod) => mod.WebLNProviders)
+    if (provider instanceof WebLNProviders.LnbitsWebLNProvider) {
+      const connector = 'lnbits'
+      // TODO
+      //   we're reaching into private parts here.
+      //   is there a better way to show to user which connection was used?
```
rolznz, line 29: What is exactly needed here? ideally `webln.getInfo()` returns the necessary info. the website should only interact with the WebLN protocol as much as possible

components/webln.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+}
+
+export function WebLNProvider ({ children }) {
+  const [provider, setProvider] = useState(null)
+  const [info, setInfo] = useState(null)
+  const [balance, setBalance] = useState(0)
+
+  const initProvider = async (provider) => {
+    const WebLNProviders = await import('@getalby/bitcoin-connect').then((mod) => mod.WebLNProviders)
```
rolznz, line 24: The only reason to cast to a specific provider type is if you need underlying functionality (such as being able to make extra Lnbits API calls not supported by WebLN). As far as I know this should not be needed

components/webln.js
```diff
@@ -0,0 +1,106 @@
...
+
+  // event listeners
+  useEffect(() => {
+    let onConnectedUnsub
+    let onDisconnectedUnsub
+    async function effect () {
+      const onConnected = await import('@getalby/bitcoin-connect-react').then((mod) => mod.onConnected)
+      onConnectedUnsub = onConnected(async (provider) => {
+        // TODO
+        //   why is this not fired on page load when there is a provider available in local storage?
```
rolznz, line 62: I think Bitcoin Connect initializes before this functionality executes. So you probably also need to check if Bitcoin Connect is already connected. There is an api function `isConnected` that you can use.

components/webln.js
```diff
@@ -0,0 +1,106 @@
...
+      })
+    }
+    effect().catch(console.error)
+    return () => {
+      onConnectedUnsub?.()
+      onDisconnectedUnsub?.()
+    }
+  }, [])
+
+  // poll balance
```
rolznz, line 82: maybe rather than polling you could just get the balance again after a payment is made? (this won't account for external payments, but I don't really think it is an issue)

fragments/users.js
```diff
@@ -8,6 +8,7 @@ export const ME = gql`
       id
       name
       bioId
+      weblnSats @client
```
rolznz, line 11: does it make sense to expose `weblnSats` differently than `sats`? then you probably need to add more conditionals in the code

rolznz:
@ekzyis for improving the payment UX you could check out the Bitcoin Connect payment component which could potentially replace the existing Stacker News QR component. This will automatically not show the QR if you're already connected. You can try a demo here: https://bitcoin-connect-request-payment-modal--rolznz.repl.co/ (for example - connect a wallet, then close the modal without paying, then re-open the modal and it will not show the QR code)

components/header.js
ekzyis: > if either of these are undefined, I think it will cause a NaN value.

You're right, it would cause a NaN value in that case. But I made sure in _components/me.js:15_ that `me.weblnSats` is never undefined. It's set to 0 in that case.

However, the optional chaining in `me.privates?.sats` makes this look like _it_ might be undefined, yes. I'll double-check that this can actually never be the case, thanks! I think we're just using optional chaining in some cases where it's not really needed.

> Also, does it even make sense to combine the two values? you can only spend from one or the other

Good point, but I think it does make sense. Currently, you can only spend from one or the other. If the user does not have enough sats in their custodial wallet, we show a QR invoice which then get paid from this "WebLN" balance (not really happy with the name).

But with some clever logic, I think we'll be able to spend from both wallets at the same time. For example, imagine someone only has 50 sats in their custodial wallet (`me.privates.sats`) and wants to pay a 100 sat invoice.

We should be able to use 50 sats from their custodial wallet and the remaining via WebLN (assuming they have enough sats there).

So I think it makes sense to keep these balances separate in the frontend code so we can implement such a thing by splitting invoices, for example.

components/me.js
ekzyis: It should be, it's above the `MeProvider` in _\_app.js_:

```js
<WebLNProvider>
  <ApolloProvider client={client}>
    <MeProvider me={me}>
```

I think it's related to how Apollo works since this is a [local-only field](https://www.apollographql.com/docs/react/local-state/managing-state-with-field-policies/#storing-local-state-in-reactive-variables) resolved with a local resolver in _apollo.js_:

```javascript
resolvers: {
  User: {
    weblnSats: (user, args, { cache }) => {
      const balance = WebLNContextRef.current?.balance
      return balance
    }
  }
},
```

But in `WebLNProvider`, I am setting `balance` to 0 by default :thinking: Need to look into this further.

components/qr.js
ekzyis: This isn't relying on `window.webln`. It just confusing naming. We manually pass in a boolean to tell if we should even try to pay the invoice via a WebLN provider. It's set to `false` if we already know the invoice failed (expired, for example). 

See [_components/invoice.js_](https://github.com/stackernews/stacker.news/blob/e20adaebdc617d24958723cf85ba537b64eb8447/components/invoice.js#L17-L51):

```javascript
let webLn = true
if (invoice.confirmedAt || (invoice.isHeld && invoice.satsReceived && !expired)) {
  variant = 'confirmed'
  status = `${numWithUnits(invoice.satsReceived, { abbreviate: false })} ${successVerb || 'deposited'}`
  webLn = false
} else if (expired) {
  variant = 'failed'
  status = 'expired'
  webLn = false
} else if (invoice.cancelled) {
  variant = 'failed'
  status = 'cancelled'
  webLn = false
}

```

ekzyis: > You use `useWebLN()` and also `webLn` is being passed in. Are they both needed? Note: Bitcoin Connect does not inject webln into the window by default.

Oh, didn't see the `webLn` getting passed in, thanks! Will check if that's still needed.

> (and this won't work if webLn is undefined)

Yep, but we don't rely in `window.webln` here. It's just confusing naming. We're manually passing in boolean values. See [_components/invoice.js_](https://github.com/stackernews/stacker.news/blob/e20adaebdc617d24958723cf85ba537b64eb8447/components/invoice.js#L22):

```javascript
let webLn = true
if (invoice.confirmedAt || (invoice.isHeld && invoice.satsReceived && !expired)) {
  variant = 'confirmed'
  status = `${numWithUnits(invoice.satsReceived, { abbreviate: false })} ${successVerb || 'deposited'}`
  webLn = false
} else if (expired) {
  variant = 'failed'
  status = 'expired'
  webLn = false
} else if (invoice.cancelled) {
  variant = 'failed'
  status = 'cancelled'
  webLn = false
}
```

components/webln.js
ekzyis: I am only doing this because I couldn't get `onConnected` to fire on page load. I would love to not do this, haha

That's what I meant with "why do I have to do this myself?" in the TODO in `onConnected`:

> // TODO
//   why is this not fired on page load when there is a provider available in local storage?
//   why do i have to do this myself?

components/webln.js
ekzyis: I am only doing this because I couldn't get `onConnected` to fire on page load. I would love to not do this, haha

That's what I meant with "why do I have to do this myself?" in the TODO in `onConnected`:

> // TODO
//   why is this not fired on page load when there is a provider available in local storage?
//   why do i have to do this myself?

components/webln.js
ekzyis: I agree, we shouldn't need to care. See [this comment](https://github.com/stackernews/stacker.news/pull/715#discussion_r1437560662).

components/header.js
rolznz: how would the combined balance work in practice? you'd need to send some sats to the custodial wallet and then pay the original invoice from the custodial wallet, right? I think this is probably too complicated and also might not even be needed (ideally users either go full self-custodial or full custodial)

components/qr.js
rolznz: Ah ok, thanks for clarifying!

components/header.js
ekzyis: > I think this is probably too complicated and also might not even be needed (ideally users either go full self-custodial or full custodial)

Users can get zapped on SN. So until we also have implemented receiving self-custodial (which is going to be nontrivial since we need to be able to take a fee for ranking and sybil resistance - see discussion in #524), it's going to be natural that users will use both wallets at the same time in the mean time.

Also, we don't intend to drop the custodial wallet anytime soon. It's good for onboarding users.

So in the case users want to start with self-custody, it's highly likely they already have some sats on SN. What to do with these sats if they want to connect a self-custodial wallet?

components/webln.js
ekzyis: `webln.getInfo()` (I use `provider.getInfo()` but I think this makes no difference and I think that's what you meant, too) returns this:

```
{
    "node": {
        "alias": "ek",
        "pubkey": ""
    },
    "methods": [
        "getInfo",
        "getBalance",
        "sendPayment"
    ],
    "version": "1.0",
    "supports": [
        "lightning"
    ]
}
```

I thought it would be nice to show the user which wallet they connected.

2023-12-28-135030_333x245_scrot

components/webln.js
ekzyis: Yep, also related to [this](https://github.com/stackernews/stacker.news/pull/715#discussion_r1437560662).

components/webln.js
ekzyis: Mhh, makes sense. But what do I do if it's already connected? Is it too late then?

Is it not possible to first this event when the event listener was attached and it wasn't fired before? I don't know enough about JS internals to answer this question myself.

components/webln.js
ekzyis: > maybe rather than polling you could just get the balance again after a payment is made?

Ideally, I don't need to poll at all. Is there a way that the connection can tell me when the balance was updated?

> (this won't account for external payments, but I don't really think it is an issue)

Well, I think it would be an issue :smile: We want to show the latest balance on SN, not something that might be stale. We will also probably need to know the _current_ balance for other reasons like [paying from both wallets at the same time.](https://github.com/stackernews/stacker.news/pull/715#discussion_r1437531356)

fragments/users.js
ekzyis: Imo, yes, it does make sense. [Related to this](https://github.com/stackernews/stacker.news/pull/715#discussion_r1437531356).

ekzyis:
> @ekzyis for improving the payment UX you could check out the Bitcoin Connect payment component which could potentially replace the existing Stacker News QR component. This will automatically not show the QR if you're already connected. You can try a demo here: https://bitcoin-connect-request-payment-modal--rolznz.repl.co/ (for example - connect a wallet, then close the modal without paying, then re-open the modal and it will not show the QR code). On mobile it also has buttons like "open in wallet" so should handle all ways for a user to pay.

Ah, right, I didn't check out your payment component yet.

However, ideally, there would be no modal at all. I want to keep the SN UX consistent with custodial zapping. I think one of the best selling points of SN is how easy it is to zap people. Anything that comes between the click and the zap is friction.

components/webln.js
rolznz: @ekzyis I think you just need to execute the same functionality in both cases:
- if it is already connected, you can access the provider with `await requestProvider` - then do something with the provider.
- otherwise, subscribe to the connected event (which again returns the provider), as you are already doing

Does that work for you?

components/header.js
huumn: Until we have self-custodial receives or people stop wanting to receive money, we'll need to custody receives. 

One alternative to combining balances is to *only* display custodial balances on SN and surfacing the self-custodial balance elsewhere.

huumn:
> But ideally, we'd get / stay as close to this UX as possible.

This is the way. At nearly any cost, we don't want to regress UX. 

-----------------

This looks great so far @ekzyis. Nice placement of the connect button. I had a lot of trouble imagining where it could go.

components/webln.js
ekzyis: > if it is already connected, you can access the provider with await requestProvider - then do something with the provider.

Using `requestProvider` prompts to connect a WebLN provider - even if there was no WebLN provider connected before. I only want to access the provider if there was one connected before. And not prompt the user to connect one. I believe this is what `onConnected` would solve.

https://github.com/stackernews/stacker.news/assets/27162016/ed2d8d13-9aec-4402-8afa-c6670616bbad

components/header.js
ekzyis: Since Mutiny will probably not support `get_balance` due to valid privacy concerns (see [this comment](https://github.com/MutinyWallet/mutiny-node/issues/895#issuecomment-1851285461) in https://github.com/MutinyWallet/mutiny-node/issues/895) and since we already have a UI/UX problem with mixing sats from two wallets, I think it makes sense to not show the self-custodial balance at all on SN. Users can always look up their self-custodial balance themselves in the connected wallet.

Also, @rolznz does have a point with this that I didn't really acknowledge in my [previous reply](https://github.com/stackernews/stacker.news/pull/715#discussion_r1437614064). Mixing sats might indeed be more complicated than really necessary:

>  I think this is probably too complicated and also might not even be needed

So this means we would only show the custodial balance on SN (as before). The self-custodial balance is unknown to us but I believe if we implement zapping as follows, this shouldn't be a problem:

Assume we have X sats on SN and we have a WebLN provider connected.

If we now want to pay for something which costs Y sats with Y > X, we first ask the WebLN provider to pay for Y - X sats with a HODL invoice. If that succeeds, we know we can fully pay the invoice with the remaining known X sats. We then use these custodial sats to pay the remaining amount. If that also succeeds (which should be the case in 99.999% of all cases since it would basically use the same code as fully custodial zaps did all this time), we settle the HODL invoice. Payment done and we didn't need to know the self-custodial balance in advance. Using HODL invoices also shouldn't be a problem since all of this should only take a few seconds to settle. So from the PoV of the WebLN provider, it won't even notice it's a HODL invoice since it gets settled (or canceled) pretty quick.

Any thoughts on this @huumn @rolznz ?

components/webln.js
ekzyis: Ignore this, I think we don't need `get_balance` anymore. I've changed my mind after reading the discussion in https://github.com/MutinyWallet/mutiny-node/issues/895. See [this comment](https://github.com/stackernews/stacker.news/pull/715#discussion_r1437934087) from me.

components/header.js
rolznz: I think it might be simpler to not doing any sort of combined payments, and just allow users to withdraw sats from custody to their own wallet (that already works)

Is there any downside there?

components/webln.js
rolznz: @ekzyis `requestProvider` will not prompt if you're already connected. So as long as you check `isConnected` first, there should be no issue.

components/webln.js
rolznz: If you click the connected button it will show where you are connected (via lnbits connector), and actually the LNbits hostname is stored in the bitcoin connect config. But does it help if you know the exact hostname, or would it be enough to know you're connected via LNbits?

I created an issue here: https://github.com/getAlby/bitcoin-connect/issues/177 - maybe we can figure a way to expose some helpful info in a generic way for all connectors.

ekzyis: > But does it help if you know the exact hostname, or would it be enough to know you're connected via LNbits?

Mhh, I think it might be enough to know that I am connected to LNbits. Not entirely sure yet though. When we released the first iteration on this, user feedback will be more valuable than any feedback I can provide you on this as a developer.

components/webln.js
ekzyis: Ah, I see [your comment here](https://github.com/getAlby/bitcoin-connect/issues/176#issuecomment-1871721940) and this makes sense! Will try it out.

components/header.js
ekzyis: > (that already works)

With that, you mean that works for some WebLN providers? Or does it work for all of them already? I mentioned in the call yesterday that I wasn't thinking about Bitcoin Connect for receiving sats yet. So I might not be up to speed regarding this and may have missed some important parts about Bitcoin Connect. I only thought of it for sending sats so far.

> Is there any downside there?

Yes, there is. Even if they would withdraw their sats to their own wallet, they will still receive sats into their custodial wallet. So this problem does not go away with withdrawing sats before connecting.

But we don't even want people to withdraw their sats just to connect a wallet. That would be annoying. Maybe they just want to try out how connecting a wallet works without fully committing to it just yet? Maybe they don't feel comfortable with self-custody yet so they don't feel comfortable with withdrawing their whole balance from SN just yet? etc.

rolznz: > With that, you mean that works for some WebLN providers? Or does it work for all of them already? I mentioned in the call yesterday that I wasn't thinking about Bitcoin Connect for receiving sats yet. So I might not be up to speed regarding this and may have missed some important parts about Bitcoin Connect. I only thought of it for sending sats so far.

What I mean is stacker news already supports paying an invoice for a withdrawal (and it also supports WebLN.makeInvoice). That invoice can come from any wallet.

> So this problem does not entirely go away with withdrawing sats before connecting.
I don't mean before connecting. I just mean if the user wants to withdraw profits then they can withdraw to their own wallet.

I just think that trying to combine self-custodial + custodial to do payments (especially when partially paying from both) is going to be complicated. I think it would be easier if you connect a wallet using Bitcoin Connect, then zaps should come from your connected wallet. Otherwise use the custodial stacker news wallet. If they just want to test out self-custodial zaps and then go back to custodial zaps, they could just disconnect their wallet.

huumn: > If we now want to pay for something which costs Y sats with Y > X, we first ask the WebLN provider to pay for Y - X sats with a HODL invoice.

I think this makes sense. If it's any easier, having WebLN pay Y directly is probably sufficient for MVP when Y > X. But your intuition is correct in that paying for Y as `Y - X non-custodial + X custodial` is the ideal. It's like paying at a store with half cash and half credit.

> I just think that trying to combine self-custodial + custodial to do payments (especially when partially paying from both) is going to be complicated.

Our plan is to eventually do that but we aren't ready to rely on an external wallet 100% yet. When we do that we inherit a user experience that we can't fully control and are inexperienced troubleshooting. Also, if a stacker has a balance on SN, paying with that balance has zero fees and paying with a non-custodial wallet has variable fees. Defaulting to the lower fee option seems like better UX. Further, our code is designed around this custodial model so supplementing it with an external wallet is going be easier to start.

We plan on doing something similar for non-custodial receives. Stackers will receive to their custodial wallets but we'll push out funds to an external wallet if their custodial wallet breaches a certain threshold.

In both cases, once we've nailed supplementing the custodial experience, we'll move to a model where there is no hybrid wallet.

components/webln.js
ekzyis: Using `isConnected` now in e06e862

components/webln.js
ekzyis: I removed the code that is reaching into private parts since https://github.com/getAlby/bitcoin-connect/issues/177 should fix the use case I was solving for.

components/webln.js
```diff
@@ -0,0 +1,59 @@
...
+  useEffect(() => {
+    const unsub = []
+    async function effect () {
+      const [isConnected, onConnected, onDisconnected, requestProvider, launchModal, launchPaymentModal, closeModal] = await import('@getalby/bitcoin-connect-react').then(
+        (mod) => [mod.isConnected, mod.onConnected, mod.onDisconnected, mod.requestProvider, mod.launchModal, mod.launchPaymentModal, mod.closeModal]
+      )
+
+      // if you want to store a function, you need to wrap it with another function because of updater functions
+      // see https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state
+      setLaunchModal(() => launchModal)
```
rolznz, line 23: These functions are imported and then passed as state, could they just be imported where they are actually used? actually I don't think they are even used yet, are they?

components/webln.js
```diff
@@ -0,0 +1,59 @@
...
+  // NOTE these functions are undefined initially - can this be a problem?
+  const [launchModal, setLaunchModal] = useState()
+  const [launchPaymentModal, setLaunchPaymentModal] = useState()
+  const [closeModal, setCloseModal] = useState()
+  const [isConnected, setIsConnected] = useState()
+
+  useEffect(() => {
+    const unsub = []
+    async function effect () {
+      const [isConnected, onConnected, onDisconnected, requestProvider, launchModal, launchPaymentModal, closeModal] = await import('@getalby/bitcoin-connect-react').then(
```
rolznz, line 17: you could import `init` here too and then call it to set Stacker News as the name of the connecting app (this will work for at least NWC and maybe other connectors in the future)

ekzyis:
I am using [our Bitcoin Connect fork](https://github.com/stackernews/bitcoin-connect) now. There were no changes required in this branch. I was able to use `npm link`.

Setup instructions:

1. Clone https://github.com/stackernews/bitcoin-connect
2. Checkout `npm-link` branch because peer dependencies don't work as expected with `npm link`, see https://github.com/npm/npm/issues/5875
3. Run `yarn install`, `yarn build` and `yarn link` in root of cloned fork and in `react/`
4. Run `npm link @getalby/bitcoin-connect` and `npm link @getalby/bitcoin-connect-react` in the SN repo

You should now be able to change code in the fork and after `yarn build` in the root and restart of server, changes should be visible.

ekzyis:
Closing in favor of #749

PR #716, opened 2023-12-28
Use return await in service worker

ekzyis:
Related to #411 

Made sure push notifications still get received - just to be sure.


huumn:
I went ahead and did the whole hog in 9698679

PR #590, opened 2023-10-27
WIP: Disincentivizing the use of all caps in post titles

SatsAllDay:
Closes #584

Issue #584, opened 2023-10-24
Disincentivizing All-Caps Headlines on Stacker News


ghost:
**Is your feature request related to a problem? Please describe.**
The number of posts with all-caps headlines on Stacker News seems to be rising lately, and it is clear by now that the primary purpose of all-caps headlines is to evoke stronger emotions from readers. The capitalization of every headline letter doesn't actually give readers any additional information about a topic... it just makes people feel like they're being yelled at.

**Describe the solution you'd like**
To fix this problem, Stacker News should implement a capital letter threshold for all post headlines (manually-generated links, auto-generated links, discussions, ads, polls, jobs, and bounties).

If a headline includes more than 12 capital letters, the poster will receive a notice directly above the "post" button that says "because you've used more than 12 capital letters in your heading, your post will cost 100x more" or something similar.

This means that if a poster would otherwise be charged 1 sat to post, their all-caps title will cost 100 sats. If they would otherwise be charged 10 sats for their post, their all-caps title will cost 1,000 sats, and on and on.

The number 12 is high enough that people who include acronyms or many short, capitalized words in their titles will not be penalized, but those who insist on capitalizing every letter in their title will. The 100x increase in posting cost is also high enough that it will either disincentivize people from creating all-caps titles, or it will generate far more rewards for other stackers.

**Additional context**
By enforcing this new threshold, we can also remove our current defenses against all-caps titles.

Our current defense system doesn't allow the second letter of the title for manually-generated link posts to be capitalized. However, there are still many all-caps titles that persist across SN despite this rule, and well-intentioned posters that want to refer to acronyms in their link titles have the unfortunate side effect of having the second letter in their acronym turned into a lowercase letter.

SatsAllDay:
> we can also remove our current defenses against all-caps titles.

What are those defenses?

SatsAllDay:
I can work on this one. It seems pretty similar to my most recent PR (#588) in that it's fee-related, so it's probably a good next issue for me to tackle.

m0wer:
Is this still relevant? Any new ideas? Is the proposed approach OK?

huumn:
This in in draft here: https://github.com/stackernews/stacker.news/pull/590 which is blocked because we've been refactoring related code since in was introduced.

Until we have a `receipt` system for recording where upcharge fees come from, I don't want to add more.

energypantry:
[...]

PR #723, opened 2023-12-30
Remove ineffective guard against all caps

ekzyis:
With this PR, I am suggesting to remove the existing guard against all caps titles for the following reasons:

- it's ineffective since it only makes the second letter lower case
- there are ways to not make it trigger and thus still be able to post links with all caps
- it's unexpected and thus confusing
- there are legitimate use cases for using all caps in the first few letters. example: acronyms
- it only applied to link posts (inconsistent UX)

related to https://github.com/stackernews/stacker.news/issues/584, #590

### DIFF
--- a/components/link-form.js
+++ b/components/link-form.js
@@ -146,10 +146,6 @@ export function LinkForm ({ item, sub, editThreshold, children }) {
               variables: { title: e.target.value }
             })
           }
-          if (e.target.value === e.target.value.toUpperCase()) {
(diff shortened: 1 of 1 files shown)


ekzyis:
And can I rename the label "confusing ux" to just "ux"? lol

huumn:
> And can I rename the label "confusing ux" to just "ux"? Then we can use this label more effectively since not every UX ticket will be about confusing UX.

Go for it

PR #724, opened 2023-12-31
Add verbose logging in onPush listener

ekzyis:
This adds more logging so we can trace better where the service worker on iOS shows unexpected behavior.

My educated guess is that `Notification.close()` does not work but [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Notification/close#browser_compatibility), it should work.

2023-12-31-223109_789x531_scrot

But this wouldn't be the first but the second case where [MDN shows invalid data what a browser supports](https://github.com/mdn/browser-compat-data/issues/20305).

Related to #411

### DIFF
--- a/components/serviceworker.js
+++ b/components/serviceworker.js
@@ -58,6 +58,11 @@ export const ServiceWorkerProvider = ({ children }) => {
   })
 
   const subscribeToPushNotifications = async () => {
+    // serviceWorker.controller is null on forced refreshes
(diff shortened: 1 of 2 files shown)


ekzyis:
Regarding asking for clarification if `Notification.close()` is supposed to work in the [WebKit bug report](https://bugs.webkit.org/show_bug.cgi?id=258922): Let's merge this first. Then I can look at the logs and have a better idea of what's going on. Then I can also ask a better question there.

PR #725, opened 2023-12-31
Show territory details in post form

ekzyis:
This reuses the existing code for territory details.

By default, it's open. Not 100% sure if that's a good idea, but I think in most cases, someone would be interested in seeing the territory details before posting.

2024-01-01-004326_1920x1080_scrot

### DIFF
--- a/components/accordian-item.js
+++ b/components/accordian-item.js
@@ -3,7 +3,7 @@ import AccordionContext from 'react-bootstrap/AccordionContext'
 import { useAccordionButton } from 'react-bootstrap/AccordionButton'
 import ArrowRight from '../svgs/arrow-right-s-fill.svg'
 import ArrowDown from '../svgs/arrow-down-s-fill.svg'
-import { useContext, useState } from 'react'
+import { useContext } from 'react'
 
 function ContextAwareToggle ({ children, headerColor = 'var(--theme-grey)', eventKey }) {
   const { activeEventKey } = useContext(AccordionContext)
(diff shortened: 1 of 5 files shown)


huumn:
What if the description is really large tho, like near the 1000 character limit? It probably needs a width/max-width.

I'd also not default it to open. If I'm posting in a territory regularly I'd wind up closing it every time.

huumn:
I mean width. This form needs to be wider if we want to have a long description displayed in it.

fragments/subs.js
```diff
@@ -19,6 +19,13 @@ export const SUB_FIELDS = gql`
     moderated
     moderatedCount
     meMuteSub
+    user {
+      name
+      id
+      optional {
+        streak
+      }
+    }
```
huumn, line 28: Use the fragment below this one instead if you need the user. This fragment is intended to be used without the additional db dip.

ekzyis:
Looks like this now with an example description that takes up all 1000 characters. Some characters are not visible though because they are counted but only used for markdown formatting.

_example with long description:_
2024-01-05-020659_1920x1080_scrot

_example with short description:_
2024-01-05-020154_1920x1080_scrot

components/accordian-item.js
```diff
@@ -33,8 +33,10 @@ export default function AccordianItem ({ header, body, headerColor = 'var(--them
 }
 
 export function AccordianCard ({ header, children, show }) {
+  const [open, setOpen] = useState(show)
```
huumn, line 36: we don't use the changes to this file anymore right? maybe restore it?

PR #726, opened 2024-01-02
Use LND subscriptions

ekzyis:
This PR adds confirming invoices by subscribing to invoice updates using [`subscribeToInvoices`](https://www.npmjs.com/package/ln-service#subscribetoinvoices) since this will be relevant for good UX in #715.

This should increase the speed since the worker only polls LND every 5 seconds but our frontend query polls every second. Difference in speed can also be seen in these two videos:

<details>
<summary>subscription</summary>

https://github.com/stackernews/stacker.news/assets/27162016/bb04bb54-7e3c-4283-8533-3053820124ae

</details>

<details>
<summary>polling</summary>

https://github.com/stackernews/stacker.news/assets/27162016/ad362b8e-ec66-475f-b8ef-281c6829edfd

</details>

We could run this for a few weeks in parallel with the old battle-tested polling as a fallback.

---

TODO since a429bf2:
- test deposits on a429bf2 with following test cases:
  - [x] 1a) worker is running (subscriptions together with polling), create invoice, pay invoice
  - [x] 1b) worker is running (subscriptions standalone), create invoice, pay invoice
  - [x] 1c) worker is not running (subscriptions together with polling), create invoice, start worker, pay invoice
  - [x] 1d) worker is not running (subscriptions together with polling), create invoice, pay invoice, start worker
  - [x] 1e) worker is not running (subscriptions standalone), create invoice, start worker, pay invoice
  - [x] 1f) worker is not running (subscriptions standalone), create invoice, pay invoice, start worker
- test withdrawals on a429bf2 with following test cases:
  - [x] 2a) worker is running (subscriptions together with polling), paste bolt11 invoice into withdrawal form, invoice gets paid using [`payViaPaymentRequest`](https://github.com/stackernews/stacker.news/blob/4d953fc3a98caa77fb3e26ae2e0a046da826dbf8/api/resolvers/wallet.js#L480-L486)
  - [x] 2b) worker is running (subscriptions standalone), paste bolt11 invoice, invoice gets paid using [`payViaPaymentRequest`](https://github.com/stackernews/stacker.news/blob/4d953fc3a98caa77fb3e26ae2e0a046da826dbf8/api/resolvers/wallet.js#L480-L486)
  - [x] 2c) worker is not running (subscriptions together with polling), paste bolt11 invoice, invoice gets paid using [`payViaPaymentRequest`](https://github.com/stackernews/stacker.news/blob/4d953fc3a98caa77fb3e26ae2e0a046da826dbf8/api/resolvers/wallet.js#L480-L486), start worker
  - [x] 2d) worker is not running (subscriptions together with polling), paste bolt11 invoice, start worker, invoice gets paid using `lncli payinvoice`
  - [x] 2e) worker is not running (subscriptions standalone), paste bolt11 invoice, invoice gets paid using [`payViaPaymentRequest`](https://github.com/stackernews/stacker.news/blob/4d953fc3a98caa77fb3e26ae2e0a046da826dbf8/api/resolvers/wallet.js#L480-L486), start worker
  - [x] 2f) worker is not running (subscriptions standalone), paste bolt11 invoice, start worker, invoice gets paid using `lncli payinvoice`

I think these test cases should cover everything :thinking:

And in general, `payViaPaymentRequest` should trigger the same behavior as `lncli payinvoice`. But for 2d) and 2f), I will comment out the `payViaPaymentRequest` code lines and then pay the invoice manually  via `lncli payinvoice` to simulate a successful payment after the worker was restarted.

---

update:

2d) and 2f) results in `unknown failure` since the payment is not known to LND (since I commented out `payViaPaymentRequest`). but this makes sense so is expected behavior. Both work when I simply return in `checkWithdrawal` instead of failing the withdrawal if it wasn't found (this skip queueing a new check). Using `lncli payinvoice` then makes the withdrawal successful.


worker/index.js
```diff
@@ -70,8 +70,29 @@ async function work () {
...
+      } catch (error) {
+        console.error(`error running ${event}`, error)
+        return
+      }
+      console.log(`finished ${event}`)
+    })
+    sub.on('error', () => {
+      // LND connection lost
+      // see https://www.npmjs.com/package/ln-service#subscriptions
+      sub.removeAllListeners()
```
ekzyis, line 87: > Besides the events listed in this documentation, all EventEmitter objects returned by any of the subscribeTo... methods also offer an error event. This event is triggered e.g. when LND is shut down. So, in order to detect and correctly handle such a situation, robust client code should generally also add a listener for the error event.

-- https://www.npmjs.com/package/ln-service#subscribetoinvoices

So iiuc, this should only trigger if LND is shutting down. Afaict, we don't need any "reconnect code". But to be sure, I would run this in parallel with the old code since ~~they don't interfere with each other. If the subscription was faster than the poll, the polling code will just see in the next poll that the invoice was already confirmed.~~

Ok, just realized they _do_ interfere with each other. I tested push notifications and I got three for a single confirmed invoice. Will put this into draft for that reason.

ekzyis:
Ok, we definitely need to make sure that our invoice code is idempotent when we want to run both methods in parallel for a while.

1. `confirm_invoice` is idempotent because it's a noop if `confirmed_at` is anything but `NULL`:

https://github.com/stackernews/stacker.news/blob/240ac3497d13e2da64ea1156d952129a14f0106c/prisma/migrations/20221110224543_msats_funcs/migration.sql#L264-L268

2. Sending user notifications is not idempotent so we need to prevent this line getting hit multiple times:

https://github.com/stackernews/stacker.news/blob/240ac3497d13e2da64ea1156d952129a14f0106c/worker/wallet.js#L34-L39

3. Cancel code is idempotent:

https://github.com/stackernews/stacker.news/blob/240ac3497d13e2da64ea1156d952129a14f0106c/worker/wallet.js#L43-L53

5. Hodl invoice confirmation is idempotent:

https://github.com/stackernews/stacker.news/blob/240ac3497d13e2da64ea1156d952129a14f0106c/worker/wallet.js#L55-L63

6. Scheduling a new job is not idempotent:

https://github.com/stackernews/stacker.news/blob/240ac3497d13e2da64ea1156d952129a14f0106c/worker/wallet.js#L65-L70

7. Canceling hodl invoices is idempotent:

https://github.com/stackernews/stacker.news/blob/240ac3497d13e2da64ea1156d952129a14f0106c/worker/wallet.js#L72-L74

ekzyis:
This is still a draft since I still think I am making too many assumptions here. For example, I am not entirely sure yet when `invoice_updated` gets fired. It might trigger in cases where we don't want to call `checkInvoice` - for example, for withdrawals.

huumn:
If you haven't yet, you can check my Outer Space code which uses subscriptions instead of polling. The main difference IIRC is that you want to store an invoice cursor so that when the worker restarts it can get any invoices it missed while it was off.

We should be able to switch completely off of polling but using both while migrating might make sense.

------------

Withdrawals can be subscribed too as well, but they don't have a cursor iirc so we'll continue to need polling as a fallback.

ekzyis:
> If you haven't yet, you can check my Outer Space code which uses subscriptions instead of polling. The main difference IIRC is that you want to store an invoice cursor so that when the worker restarts it can get any invoices it missed while it was off.

Did that in bb74b5f

> Withdrawals can be subscribed too as well, but they don't have a cursor iirc so we'll continue to need polling as a fallback.

You're right, according to the `ln-service` docs of [`subscribeToPayments`](https://www.npmjs.com/package/ln-service#subscribetopayments), there is no cursor.

However, I believe we don't need a cursor since iiuc, LND already does what we want:

> TrackPayments returns an update stream **for every payment that is not in a terminal state**. Note that if payments are in-flight while starting a new subscription, the start of the payment stream could produce out-of-order and/or duplicate events. In order to get updates for every in-flight payment attempt make sure to subscribe to this method before initiating any payments.

-- https://lightning.engineering/api-docs/api/lnd/router/track-payments

So since we don't need to get updates for every in-flight payment _attempt_, I think this subscription is fine. We only need to be subscribed to "every payment that is not in a terminal state".

And `TrackPayments` is what `ln-service` uses under the hood. See [here](https://github.com/alexbosworth/lightning/blob/8409909987c59d3ce778114e92a4b41609a7a5f6/lnd_methods/offchain/subscribe_to_payments.js#L9C32-L9C32).

So my current conclusion is that we might not even need polling for withdrawals.

huumn:
> So my current conclusion is that we might not even need polling for withdrawals.

We don't need polling, but when the worker starts up, we will need to check the status of any pending payments in case any payments completed while the worker was down.

ekzyis:
> we will need to check the status of any pending payments in case any payments completed while the worker was down.

Ah, makes sense!

huumn:
> Ah, makes sense! I was only thinking about pending payments.

That's what nice about the cursor on receives. We can always resume where we left off.

I asked Bos about why there wasn't a cursor for withdrawals last year, "y no api symmetry?" I forget what he said, but he recommended still using subscriptions for withdrawals but checking for missed payment statuses if the connection is lost/restarted.

ekzyis:
> checking for missed payment statuses if the connection is lost/restarted.

I am doing this on worker startup in a429bf2 now:

```js
// queue status check of all pending withdrawals since they might have been paid by LND while worker was down
await models.$queryRaw`
  INSERT INTO pgboss.job (name, data, retrylimit, retrybackoff, startafter)
  SELECT 'checkWithdrawal', json_build_object('id', w.id, 'hash', w.hash), 21, true, now() + interval '10 seconds'
  FROM "Withdrawl" w WHERE w.status IS NULL`
```
However, since we do polling and subscription in parallel (so we still queue a job during `create_withdrawal`), there will be two jobs queued for every withdrawal that was still pending in the database when the worker is restarted.

I need to add some deduplication since else, there will always be two jobs running in parallel for these withdrawals until they are settled since they requeue themselves.

So basically some JOIN with `pgboss.job` (or `AND NOT EXISTS`) to check if there is already a `checkWithdrawal` job queued for this id.

huumn:
Alternatively, on startup:

1. subscribe to withdrawals
2. loop over withdrawals that are still pending and check them

This should allow you to get rid of the queuing all together.

worker/index.js
ekzyis: > After the last listener has been removed from an EventEmitter object, the subscription is released and subsequent attempts to add a listener will result in an error. **Call subscribeTo... again and add new listeners to the new EventEmitter object.**

-- https://www.npmjs.com/package/ln-service#subscriptions

Since I am not sure if we might lose the connection to LND once in a while, I am going to wrap this code with some reconnect logic to be sure (contrary to what I stated in my previous comment).

prisma/migrations/20221110224543_msats_funcs/migration.sql
```diff
@@ -300,6 +300,7 @@ BEGIN
 
     UPDATE users SET msats = msats - msats_amount - msats_max_fee WHERE id = user_id;
 
+    -- TODO: remove this in a new database migration when we fully migrated from polling to LND subscriptions
```
huumn, line 303: you can't mutate old migrations like this. it changes the hash

worker/index.js
```diff
@@ -70,9 +71,53 @@ async function work () {
     }
   }
 
+  async function subWrapper (subFn, ...eventFns) {
```
huumn, line 74: I don't think we should be mixing wallet stuff into the index file. Separation of concerns and all that

prisma/migrations/20221110224543_msats_funcs/migration.sql
ekzyis: Ah, I see. `npx prisma migrate dev`:

```
- The migration `20221110224543_msats_funcs` was modified after it was applied.
- The migration `20231201014953_territory` was modified after it was applied.
```

worker/index.js
```diff
@@ -71,8 +71,19 @@ async function work () {
   }
 
   await boss.start()
+
+  const [lastConfirmed] = await models.$queryRaw`SELECT "confirmedIndex" FROM "Invoice" ORDER BY "confirmedIndex" DESC NULLS LAST LIMIT 1`
```
huumn, line 75: This still feels like we're failing to encapsulate this wallet code. Ideally, none of this wallet stuff should be in this file.

If you want I can take a stab at it.

ekzyis: Ah, no, I got it. I have an idea.

huumn:
To be honest, this code seems 2-3x more complicated than it should be.

Why that might be:
1. abstractions are used where they aren't *really truly* needed and reduce readability
2. things are named poorly
3. relevant logic is spread across more than one file
4. I don't understand what's migration code and what isn't (and that half this code will be deleted or something)

I'm tempted to check this out and refactor it myself but I know you'd prefer I comment instead.

ekzyis:
> To be honest, this code seems 2-3x more complicated than it should be.
> 
> Why that might be:
> 
> 1. abstractions are used where they aren't _really truly_ needed and reduce readability

Yes, we only need two subscriptions so this `subWrapper` isn't really worth it. Went for an abstraction too early

> 2. things are named poorly

You probably mean the `sub` variable. But doesn't matter, I can try to give everything a better name

> 3. relevant logic is spread across more than one file

Yep, I think only a single import from _worker/wallet.js in _worker/index.js_ should be what's needed

> 4. I don't understand what's migration code and what isn't (and that half this code will be deleted or something)

Mhh, do you think we can just directly migrate to subscriptions then? I feel pretty confident that the subscriptions will work as expected but I also know that real life conditions usually contain unexpected things.

> I'm tempted to check this out and refactor it myself but I know you'd prefer I comment instead.

I'll give this another try for an hour since I think I know what you mean with these points. When this is still not satisfactory, you can do it

huumn:
> Yes, we only need two subscriptions so far and that's probably all we need for a long time so this subWrapper isn't really worth it. Went for an abstraction too early

You have no idea how happy it makes me that you immediately understood what I meant.

> do you think we can just directly migrate to subscriptions then?

You should know, shouldn't you? :) You've been thinking through this code recently. 

The only migration that's needed is for receives. That's because there might be invoices that get confirmed on worker restart, and we don't have a `confirmIndex` yet so we might miss them otherwise. That migration code should, in theory, look a lot like the code for checking withdrawals on startup.

ekzyis:
> You should know, shouldn't you? :) You've been thinking through this code recently.

But you're the one that has access to the LND node, production logs, database etc. so if something breaks in production which isn't really reproducible locally, you will probably be the one who can debug it :)

> The only migration that's needed is for receives.

I think we need to change both postgres functions (`create_invoice` and `create_withdrawal`) since we won't need the job insertion anymore. Or do you mean something else with "the only migration"?

> That's because there might be invoices that get confirmed on worker restart, and we don't have a confirmIndex yet so we might miss them otherwise.

You mean some kind of race condition on worker restart? Not sure I understand you _here_, though. Why wouldn't we have a `confirmedIndex` yet (assuming with _receives_ you mean from the PoV of SN which means _deposits_.)? I am fetching the `confirmedIndex` right before I subscribe to invoices, no?

huumn:
> Or do you mean something else with "the only migration"?

I'm using migration generically - as in we are *migrating* from polling to subscriptions. 

But yes, those db functions will need to be updated in a db migration.

ekzyis:
Did the refactor in 13698e2 now:

> **refactor subscription code some more**
> * remove unnecessary subWrapper abstraction
> * move all wallet related code into worker/wallet.js such that only a single import is needed in worker/index.js

Tested a single deposit and withdrawal and it worked. So I guess this still works as previously without exhaustive testing required. But every code that works on first try makes me always suspicious, lol.

Will work on removing the polling now as discussed in Slack.

huumn:
Also, while you're in the code, I don't think you need the connection retry logic. afaik the grpc client library should reconnect if it experiences issues. I can't remember where I read that but I remember considering this in Outer Space.

ekzyis:
> Also, while you're in the code, I don't think you need the connection retry logic. afaik the grpc client library should reconnect if it experiences issues. I can't remember where I read that but I remember considering this in Outer Space.

Ah, I see. I originally tested my code to check if reconnecting works, but I ran it with `sub.removeAllListeners()` in the error callback so obviously no invoice updates did not trigger any callback, lol. I misinterpreted that.

But now I tested again without `sub.removeAllListeners()` and now I can restart the LND node and it will continue to work. Nice!

Regarding HODL invoices being broken because they don't trigger `invoice_updated` on payment (as mentioned in Slack). I think that's on purpose, see the docs for `SubscribeInvoices`:

> SubscribeInvoices returns a uni-directional stream (server -> client) for notifying the client of **newly added/settled invoices.**

-- https://api.lightning.community/api/lnd/lightning/subscribe-invoices

I'll have to use `SubscribeSingleInvoice`, I believe:

> SubscribeSingleInvoice returns a uni-directional stream (server -> client) to notify the client of state transitions of the specified invoice. Initially the current invoice state is always sent out.

-- https://api.lightning.community/api/lnd/invoices/subscribe-single-invoice

ekzyis:
Will put this in draft until I tested it again.

ekzyis:
> edit: For some reason, the callback does not get triggered if hodl invoices expire. Will take another look at this tomorrow.

Okay, I think I isolated the problem now. `SubscribeToInvoice` _does_ get called if the hodl invoice expires since from "unpaid" -> "expired" is considered to be a state transition.

However, it does not get called if the invoice was already paid and is thus in the `is_held` state.

In that case, "expired" no longer exists. The invoice must be manually settled or canceled now. This does work in our happy path (when there is no error when the client uses the `hash` and `hmac` of the paid invoice to trigger an action) but no in the bad path.

I think I'll just insert a job for that case that gets run when the invoice would expire and then cancels it if it's not settled already.

ekzyis:
Okay, this should be it now.

[video](https://files.ekzyis.com/public/sn/sn_lnd_subscriptions.mp4)

Timestamps:

_00:00 - 00:10_ Deposits

_00:13 - 00:26_ Withdrawals

_00:27 - 00:38_ Successful anon zap / HODL invoice

_00:39 - 01:15_ Expired anon zap / HODL invoice without paying

_01:20 - 02:20_ Simulate error after anon zap payment and wait for expiration_

worker/wallet.js
```diff
@@ -1,36 +1,165 @@
...
-    // on lnd related errors, we manually retry so we don't exponentially backoff
-    await boss.send('checkInvoice', { hash }, walletOptions)
     return
   }
   console.log(inv)
 
-  // check if invoice still exists since HODL invoices get deleted after usage
+  // check if invoice exists since it might just have been created by LND and wasn't inserted into the database yet
+  // but that is not a problem since this function will be called again with the update
+  // FIXME: there might be a race condition here if the invoice gets paid before the invoice was inserted into the db.
```
ekzyis, line 147: I think the probability of users paying invoices before it got inserted in the db is so low, we can basically ignore this case.

However, the chance that `checkInvoice` gets called before the invoice is inserted is pretty high. However, it doesn't really matter since `checkInvoice` on invoices that were just created wouldn't do anything anyway.

I could check if I can skip such `checkInvoices` call though. Maybe via a check if the invoice was created within the last few milliseconds or something. But not sure if that's worth it. Will make the code more messy again without a lot of benefit.

worker/wallet.js
```diff
@@ -106,12 +236,7 @@ export async function checkWithdrawal ({ data: { id, hash }, boss, models, lnd }
     } else if (wdrwl?.failed.is_route_not_found) {
       status = 'ROUTE_NOT_FOUND'
     }
-    await serialize(models, models.$executeRaw`
-      SELECT reverse_withdrawl(${id}::INTEGER, ${status}::"WithdrawlStatus")`)
-  } else {
-    // we need to requeue to check again in 5 seconds
-    const startAfter = new Date(wdrwl.created_at) > datePivot(new Date(), { minutes: -5 }) ? 5 : 60
-    await boss.send('checkWithdrawal', { id, hash }, { ...walletOptions, startAfter })
+    await serialize(models, models.$executeRaw`SELECT reverse_withdrawl(${id}::INTEGER, ${status}::"WithdrawlStatus")`)
```
ekzyis, line 239: Whoops, deleted the `else`. That's probably bug but withdrawals worked :thinking:

huumn:
Oh man! So snappy! I should've done this literally years ago. I'll give it a review first thing tomorrow

worker/wallet.js
huumn: How would they pay an invoice we haven't shown them? Delivering the invoice to the client requires it being stored in the db.

ekzyis: > How would they pay an invoice we haven't shown them? Delivering the invoice to the client requires it being stored in the db.

Ah, good point, haha

ekzyis:
> Oh man! So snappy! I should've done this literally years ago. I'll give it a review first thing tomorrow

It could get even more snappier since we're still polling every second in the frontend but looks snappy enough so far :)

And in production, payments will probably take a lot longer since this is just running locally. So the 1 second poll might be less significant in production.

worker/wallet.js
```diff
@@ -105,13 +235,7 @@ export async function checkWithdrawal ({ data: { id, hash }, boss, models, lnd }
...
     } else if (wdrwl?.failed.is_route_not_found) {
       status = 'ROUTE_NOT_FOUND'
-    }
-    await serialize(models, models.$executeRaw`
-      SELECT reverse_withdrawl(${id}::INTEGER, ${status}::"WithdrawlStatus")`)
-  } else {
-    // we need to requeue to check again in 5 seconds
-    const startAfter = new Date(wdrwl.created_at) > datePivot(new Date(), { minutes: -5 }) ? 5 : 60
-    await boss.send('checkWithdrawal', { id, hash }, { ...walletOptions, startAfter })
+    } else await serialize(models, models.$executeRaw`SELECT reverse_withdrawl(${id}::INTEGER, ${status}::"WithdrawlStatus")`)
```
huumn, line 238: I think this means we aren't reversing withdrawals if we know the status. This looks like a typ.

huumn: This was a typo. I'm working on this locally. I'll push up in a bit.

huumn:
Working on this locally. I noticed we don't have a good strategy for dealing with the `invoice_updated` callback failing. (or the withdrawal callback)

For example:

1. `invoice_updated` is called with `confirm_index = 1` and the db update fails. As is, we just log it.
2. `invoice_update` is called for another invoice `confirm_index = 2` and the update succeeds. Next time the worker restarts, we subscribe at `confirm_index = 2` so the invoice in (1) will never confirm in the db.

This should be rare but it's a critical problem. AFAICT we have a few options:

1. exit the worker process immediately when this happens (but concurrency might still allow the above to happen)
2. fallback to polling in such cases
3. check invoices on startup like with withdrawals (basically always keep our "migration" code) or even have a periodic job that does this
4. record `add_index` and on startup `subscribeToInvoices` with the lowest `add_index` that hasn't confirmed

I think (2) or (3) are our best bets. Because we'll always need (3) for withdrawals, it's probably the best option on the basis of code simplicity.

huumn:
Would you mind explaining why *exactly*:

1. we subscribe to single hodl invoices? The comment there says a lot of things, but I can't tell which case this subscription is handling.
2. we need a the `finalizeHodlInvoice` job? This is where we return the payment, right? Does the `invoice_updated` subscription to the single invoice not tell us when an invoice is expired?

worker/wallet.js
ekzyis: Ah, yes. Didn't read the code properly, just thought I deleted an existing `else` around `reverse_withdrawal` on accident. But I deleted the `else` for requeueing which was correct.

ekzyis:
> Would you mind explaining why exactly:
> 
> 1. we subscribe to single hodl invoices? The comment there says a lot of things, but I can't tell which case this subscription is handling.
> 2. ~we need the `finalizeHodlInvoice` job? This is where we refund the payment possibly, right? Does the `invoice_updated` subscription to the single invoice not tell us when an invoice is expired?~
> 
> _edit_ I think this answers (2):
> 
> > SubscribeToInvoice does get called if the hodl invoice expires since from "unpaid" -> "expired" is considered to be a state transition. LND simply considers the invoice to be "canceled" when it expires (inv.is_canceled is true).
> > However, it does not get called if the invoice was already paid and is thus in the is_held state.
> 
> So if (1) isn't doing that, what is (1) doing?

(1) handles the transition from "unpaid" -> "is_held". `SubscribeToInvoices` does not get called if an HODL invoice was paid since it's not settled yet. It only gets called on invoice creation and invoice settlement.

huumn:
Thanks! That makes a lot of sense.

ekzyis:
> Working on this locally. I noticed we don't have a good strategy for dealing with the `invoice_updated` callback failing. (or the withdrawal callback)
> 
> For example:
> 
> 1. `invoice_updated` is called with `confirm_index = 1` and the db update fails. As is, we just log it.
> 2. `invoice_update` is called for another invoice `confirm_index = 2` and the update succeeds. Next time the worker restarts, we subscribe at `confirm_index = 2` so the invoice in (1) will never confirm in the db.
> 
> This should be rare but it's a critical problem. AFAICT we have a few options:
> 
> 1. _edit: this isn't really a solution_ ~exit the worker process immediately when this happens (but concurrency - especially across the worker cluster - could still allow the above to happen)~
> 2. fallback to polling in such cases
> 3. check invoices on startup like with withdrawals (basically always keep our "migration" code) AND have a periodic job that does this
> 4. record `add_index` and on startup `subscribeToInvoices` with the lowest `add_index` that hasn't confirmed
> 
> (4) doesn't solve this problem for withdrawals and requires a restart, so I think (2) or (3) are among our best bets, but I think (3) is the better bet:
> 
> 1. on the basis of code simplicity because we'll always need (3) for withdrawals
> 2. if the DB connection is failing, we might not be able to insert a job to poll so we'll need (3) anyway

Ah, makes sense. Yes, I think (3) is the best option, too.

huumn:
I made some very small readability changes as I reviewed, updated some of the comments to help someone like me who was less familiar with the code, and made sure we aren't suppressing some of the errors (which I might've overdone but better to be a little overzealous).

This still needs another testing/review pass, but my polar network needs to be updated for TrackPayments I think.

huumn:
I'm not going to be pushing to this anymore tonight fyi. Just wanted to make sure I shared what I had in case you had other mods chambered.

huumn:
BTW you keep force pushing over your prs

ekzyis:
> BTW you keep force pushing over your prs

Oh no, I thought I disabled the mirroring from my repo. Will check after the call, thanks for letting me know

huumn:
Going to merge. I think this is rock solid now. 

@ekzyis I made some more significant changes so please look it over at your convenience and just in case.

ekzyis:
> @ekzyis I made some more significant changes so please look it over at your convenience and just in case.

Understood, will try to break this in 8 hours :)

PR #737, opened 2024-01-07
add link to email signup page for newsletter in newsletter signature

SatsAllDay:
https://stacker.news/items/379497/r/WeAreAllSatoshi

### DIFF
--- a/scripts/newsletter.js
+++ b/scripts/newsletter.js
@@ -152,7 +152,9 @@ Yeehaw,
 Keyan
 A guy who works on Stacker News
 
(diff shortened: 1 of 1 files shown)


ekzyis (commented):
Mhh, I think this will also be included in the newsletter to the people who already signed up. So it might be confusing for these readers to find a signup link in the newsletter that people only receive when they sign up - as if they didn't signup already.

So maybe this is more something to only include in the SN post?

SatsAllDay:
Good point. Maybe I can add a CLI parameter to conditionally include this blurb? I'm not sure how @huumn runs this so I'm not sure the best way to facilitate dynamic content.

huumn:
I'll just delete it when I copy it for the email. All good. Thanks!

PR #738, opened 2024-01-10
Fix missing is_held transition in checkInvoice

ekzyis:
See https://github.com/stackernews/stacker.news/commit/2151323c8d6737cd2a88774c55488b6d05e2d2e6#commitcomment-136652105

### DIFF
--- a/worker/wallet.js
+++ b/worker/wallet.js
@@ -123,6 +123,16 @@ async function checkInvoice ({ data: { hash }, boss, models, lnd }) {
     return await boss.send('nip57', { hash })
   }
 
+  if (inv.is_held) {
+    return await models.invoice.update({
+      where: { hash },
+      data: {
+        msatsReceived: Number(inv.received_mtokens),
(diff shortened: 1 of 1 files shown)


huumn:
I'll do this locally. I have a bunch of changes to this file already.

You were right about needing to retry subscriptions on errors. It's pretty easy to test. Just stop lnd and start it and it'll never reconnect.

I'll push my changes soon.

ekzyis:
> You were right about needing to retry subscriptions on errors. It's pretty easy to test. Just stop lnd and start it and it'll never reconnect.

Huh, I think I did exactly that but I think it just didn't work because we're removing the listener after an error.

I guess that's also the problem here.

huumn:
The authenticated connections reconnects at some rate, but the subscription does not. 

Fixed here: https://github.com/stackernews/stacker.news/commit/c243a6d8bec11c1d46427fc947b6da7eff3f310c

PR #744, opened 2024-01-10
Fix TypeError on item creation if JIT invoicing is used

### LINKED ISSUE #743: Item creation with JIT invoices fails with TypeError
**Description**

JIT invoices for item creation currently fail on c267bf9 because `serializeInvoiceable` returns an empty array if a JIT invoice was used (`hash` provided).

**Steps to Reproduce**

1. Post something without enough sats
2. Pay invoice

HODL invoice gets settled because we immediately settle it after the db transaction.

However, the action still fails because of an `TypeError`:

 > Field Mutation.upsertComment failed with TypeError: Cannot read properties of undefined (reading 'text')

**Expected behavior**

JIT invoices work for posting content + HODL invoices don't get settled if action fails

**Video**

<details>

<summary>video</summary>

https://github.com/stackernews/stacker.news/assets/27162016/e843da46-ff6a-4c5d-9c60-57ec074f5f42

</details>

**Additional context**

Bug is triggered by `results = results.slice(1,-1)` since `results` only contains two elements (see video) and thus the end index of `-1` points to the same start index of `1`:

https://github.com/stackernews/stacker.news/blob/c267bf92fd9de8e9ca2a3f82efdc0297384fbf4e/api/resolvers/serial.js#L85-L91

```
$ node
Welcome to Node.js v21.4.0.
> let x = [1,2]
undefined
> x.slice(1,-1)
[]
> let y = [1,2,3]
undefined
> y.slice(1,-1)
[ 2 ]
```

~~I wonder if we can simply use `results[1]` instead of `results.slice(1,-1)`. But not sure yet, don't want to break anything else.~~ Okay, I guess the intention with this code was to simply remove the first element and don't care about the size of the array. So `results[1]` isn't the same. But something like `results.length > 2 ? results.slice(1,-1) : [results[1]]` would be. _(jeez, is this ugly code but I started with this mess, lol)_

I also wonder where the `1` as the first element of the returned array of `serialize` comes from (see video). `confirm_invoice` returns 0, not 1.

ekzyis:
Fix #743

### DIFF
--- a/api/resolvers/serial.js
+++ b/api/resolvers/serial.js
@@ -85,8 +85,11 @@ export async function serializeInvoicable (query, { models, lnd, hash, hmac, me,
   let results = await serialize(models, ...trx)
 
   if (hash) {
-    if (invoice?.isHeld) await settleHodlInvoice({ secret: invoice.preimage, lnd })
-    results = results.slice(1, -1)
+    if (invoice?.isHeld) {
+      await settleHodlInvoice({ secret: invoice.preimage, lnd });
+      // remove first element since that is the confirmed invoice
+      ([, ...results] = results)
+    }
   }
 
   // if there is only one result, return it directly, else the array
--- a/api/resolvers/serial.js
+++ b/api/resolvers/serial.js
@@ -85,11 +85,9 @@ export async function serializeInvoicable (query, { models, lnd, hash, hmac, me,
   let results = await serialize(models, ...trx)
 
   if (hash) {
-    if (invoice?.isHeld) {
-      await settleHodlInvoice({ secret: invoice.preimage, lnd });
-      // remove first element since that is the confirmed invoice
-      ([, ...results] = results)
-    }
+    if (invoice?.isHeld) { await settleHodlInvoice({ secret: invoice.preimage, lnd }) }
+    // remove first element since that is the confirmed invoice
+    [, ...results] = results
   }
 
   // if there is only one result, return it directly, else the array


api/resolvers/serial.js
```diff
@@ -85,8 +85,11 @@ export async function serializeInvoicable (query, { models, lnd, hash, hmac, me,
   let results = await serialize(models, ...trx)
 
   if (hash) {
-    if (invoice?.isHeld) await settleHodlInvoice({ secret: invoice.preimage, lnd })
-    results = results.slice(1, -1)
+    if (invoice?.isHeld) {
+      await settleHodlInvoice({ secret: invoice.preimage, lnd });
+      // remove first element since that is the confirmed invoice
```
ekzyis, line 90: I actually don't know what the first element is but it should be the confirmed invoice. But it's a 1, not a 0 which is what `confirm_invoice` would return. :thinking:

api/resolvers/serial.js
```diff
@@ -85,8 +85,9 @@ export async function serializeInvoicable (query, { models, lnd, hash, hmac, me,
   let results = await serialize(models, ...trx)
 
   if (hash) {
-    if (invoice?.isHeld) await settleHodlInvoice({ secret: invoice.preimage, lnd })
-    results = results.slice(1, -1)
+    if (invoice?.isHeld) { await settleHodlInvoice({ secret: invoice.preimage, lnd }) }
+    // remove first element since that is the confirmed invoice
+    [, ...results] = results
```
ekzyis, line 90: I actually don't know what the first element is but it should be the confirmed invoice. But it's a 1, not a 0 which is what `confirm_invoice` should return. :thinking:

PR #749, opened 2024-01-14
Expose WebLN interface via React Context

### LINKED ISSUE #75: Feature: Allow connecting self hosted LN wallets / nodes with SN
This is potentially a longer term feature, but it would help decentralization if it would be possible to use self-hosted wallet or even a node with SN. 

Example 1: make it possible to use Alby to vote with sats or receive sats directly into Alby. I'm not sure if there's a good protocol to do this already.  
Example 2: podcasters that register in podcastindex.org provide node id and that lets them receive the streaming sats. It would be great if they could get SN sats using the the same id.

### LINKED ISSUE #490: SN wallet integration for seamless payments

### LINKED ISSUE #533: Nostr Wallet Connect

ekzyis:
Supersedes #715 Closes #533 Closes #490 Closes #75

Based on #753, #752, #777, #778

TODO:
- [x] [feature] LNbits
- [x] [feature] NWC
- [x] [bug] Fix occasional `NaN` in cache update
- [x] [UX] use optimistic cache updates to keep site "snappy" (see 463578c)
- [x] ~[UX] use second lightning strike as "thunder" or "confirmation" (similar to double check marks in messengers) for WebLN zaps by overlapping them~
- [x] ~[UX] adapt first lightning strike based on if payment is pending (WebLN) or done (custodial) - for example, make pending zaps slower, last longer, different color etc.~
- [x] ~[UX] investigate if it makes sense to not immediately generate invoices for async providers (NWC); essentially queueing or batching zaps. For example, we could show a notification on their wallet that when clicked shows that zaps are queued. Only when they are ready to pay, we will initiate a payment which they can approve in their wallet then. If zaps are queued, we could also batch them together and create a single invoice for all of them (needs some updates in the backend to allow hash reuse for acts).~ _not planned, too much for MVP_
- [x] [bug] fix empty fields on hard refresh on config page since configs not initialized on first render
- [x] [UX] show persistent toast while zap is pending
- [x] [bug] toast race condition: invoice cancelation poll vs NWC invoice paid / error on payment
- [x] ~[feature] switch between providers~ not needed yet if we only launch with NWC
- [x] testing
- [x] [bug] cache update is not always undone
- [x] [feature] allow switching between LNbits and NWC
- [x] [UX] warn users about risk with banner
- [ ] [UX] static checks for NWC secret key since it's not validated via info event on save

### DIFF
--- a/components/webln/lnbits.js
+++ b/components/webln/lnbits.js
@@ -53,7 +53,6 @@ export function LNbitsProvider ({ children }) {
         'getInfo',
         'getBalance',
         'sendPayment'
-        // TODO: support makeInvoice and sendPaymentAsync
       ],
       version: '1.0',
       supports: ['lightning']
--- a/components/webln/lnbits.js
+++ b/components/webln/lnbits.js
@@ -82,7 +82,7 @@ export function LNbitsProvider ({ children }) {
 
   const loadConfig = useCallback(() => {
     const config = window.localStorage.getItem(storageKey)
-    if (!config) return null
+    if (!config) return
     const configJSON = JSON.parse(config)
     setUrl(configJSON.url)
     setAdminKey(configJSON.adminKey)
--- a/components/webln/lnbits.js
+++ b/components/webln/lnbits.js
@@ -9,6 +9,7 @@ export function LNbitsProvider ({ children }) {
   const [adminKey, setAdminKey] = useState()
   const [enabled, setEnabled] = useState()
 
+  const name = 'LNbits'
   const storageKey = 'webln:provider:lnbits'
 
   const request = useCallback(async (method, path, args) => {
@@ -122,7 +123,7 @@ export function LNbitsProvider ({ children }) {
 
   useEffect(loadConfig, [])
 
-  const value = { url, adminKey, saveConfig, clearConfig, enabled, sendPayment }
+  const value = { name, url, adminKey, saveConfig, clearConfig, enabled, sendPayment }
   return (
     <LNbitsContext.Provider value={value}>
       {children}
--- a/components/webln/lnbits.js
+++ b/components/webln/lnbits.js
@@ -4,6 +4,62 @@ import { createContext, useCallback, useContext, useEffect, useState } from 'rea
 
 const LNbitsContext = createContext()
 
+const getWallet = async (baseUrl, adminKey) => {
+  const url = baseUrl.replace(/\/+$/, '')
+  const path = '/api/v1/wallet'
+
+  const headers = new Headers()
+  headers.append('Accept', 'application/json')
+  headers.append('Content-Type', 'application/json')
+  headers.append('X-Api-Key', adminKey)
+
+  const res = await fetch(url + path, { method: 'GET', headers })
+  if (!res.ok) {
+    const errBody = await res.json()
+    throw new Error(errBody.detail)
+  }
+  const wallet = await res.json()
+  return wallet
+}
+
+const postPayment = async (baseUrl, adminKey, bolt11) => {
+  const url = baseUrl.replace(/\/+$/, '')
+  const path = '/api/v1/payments'
+
+  const headers = new Headers()
+  headers.append('Accept', 'application/json')
+  headers.append('Content-Type', 'application/json')
+  headers.append('X-Api-Key', adminKey)
+
+  const body = JSON.stringify({ bolt11, out: true })
+
+  const res = await fetch(url + path, { method: 'POST', headers, body })
+  if (!res.ok) {
+    const errBody = await res.json()
+    throw new Error(errBody.detail)
+  }
+  const payment = await res.json()
+  return payment
+}
+
+const getPayment = async (baseUrl, adminKey, paymentHash) => {
+  const url = baseUrl.replace(/\/+$/, '')
+  const path = `/api/v1/payments/${paymentHash}`
+
+  const headers = new Headers()
+  headers.append('Accept', 'application/json')
+  headers.append('Content-Type', 'application/json')
+  headers.append('X-Api-Key', adminKey)
+
+  const res = await fetch(url + path, { method: 'GET', headers })
+  if (!res.ok) {
+    const errBody = await res.json()
+    throw new Error(errBody.detail)
+  }
+  const payment = await res.json()
+  return payment
+}
+
 export function LNbitsProvider ({ children }) {
   const [url, setUrl] = useState('')
   const [adminKey, setAdminKey] = useState('')
@@ -12,38 +68,8 @@ export function LNbitsProvider ({ children }) {
   const name = 'LNbits'
   const storageKey = 'webln:provider:lnbits'
 
-  const request = useCallback(async (method, path, args) => {
-    let body = null
-    const query = ''
-    const headers = new Headers()
-    headers.append('Accept', 'application/json')
-    headers.append('Content-Type', 'application/json')
-    headers.append('X-Api-Key', adminKey)
-
-    if (method === 'POST') {
-      body = JSON.stringify(args)
-    } else if (args !== undefined) {
-      throw new Error('TODO: support args in GET')
-      // query = ...
-    }
-    const url_ = url.replace(/\/+$/, '')
-    const res = await fetch(url_ + path + query, {
-      method,
-      headers,
-      body
-    })
-    if (!res.ok) {
-      const errBody = await res.json()
-      throw new Error(errBody.detail)
-    }
-    return (await res.json())
-  }, [url, adminKey])
-
   const getInfo = useCallback(async () => {
-    const response = await request(
-      'GET',
-      '/api/v1/wallet'
-    )
+    const response = await getWallet(url, adminKey)
     return {
       node: {
         alias: response.name,
@@ -57,70 +83,68 @@ export function LNbitsProvider ({ children }) {
       version: '1.0',
       supports: ['lightning']
     }
-  }, [request])
+  }, [url, adminKey])
 
   const sendPayment = useCallback(async (bolt11) => {
-    const response = await request(
-      'POST',
-      '/api/v1/payments',
-      {
-        bolt11,
-        out: true
-      }
-    )
-    const checkResponse = await request(
-      'GET',
-      `/api/v1/payments/${response.payment_hash}`
-    )
+    const response = await postPayment(url, adminKey, bolt11)
+    const checkResponse = await getPayment(url, adminKey, response.payment_hash)
     if (!checkResponse.preimage) {
       throw new Error('No preimage')
     }
     return { preimage: checkResponse.preimage }
-  }, [request])
+  }, [url, adminKey])
 
-  const loadConfig = useCallback(() => {
+  const loadConfig = useCallback(async () => {
     const config = window.localStorage.getItem(storageKey)
     if (!config) return
     const configJSON = JSON.parse(config)
     setUrl(configJSON.url)
     setAdminKey(configJSON.adminKey)
+
+    try {
+      // validate config by trying to fetch wallet
+      await getWallet(configJSON.url, configJSON.adminKey)
+    } catch (err) {
+      console.error('invalid LNbits config:', err)
+      setEnabled(false)
+      throw err
+    }
+    setEnabled(true)
   }, [])
 
   const saveConfig = useCallback(async (config) => {
+    // immediately store config so it's not lost even if config is invalid
     setUrl(config.url)
     setAdminKey(config.adminKey)
+
     // XXX This is insecure, XSS vulns could lead to loss of funds!
     //   -> check how mutiny encrypts their wallet and/or check if we can leverage web workers
     //   https://thenewstack.io/leveraging-web-workers-to-safely-store-access-tokens/
     window.localStorage.setItem(storageKey, JSON.stringify(config))
+
+    try {
+      // validate config by trying to fetch wallet
+      await getWallet(config.url, config.adminKey)
+    } catch (err) {
+      console.error('invalid LNbits config:', err)
+      setEnabled(false)
+      throw err
+    }
+    setEnabled(true)
   }, [])
 
   const clearConfig = useCallback(() => {
     window.localStorage.removeItem(storageKey)
     setUrl('')
     setAdminKey('')
+    setEnabled(undefined)
   }, [])
 
   useEffect(() => {
-    // update enabled
-    (async function () {
-      if (!(url && adminKey)) {
-        setEnabled(undefined)
-        return
-      }
-      try {
-        await getInfo()
-        setEnabled(true)
-      } catch (err) {
-        console.error(err)
-        setEnabled(false)
-      }
-    })()
-  }, [url, adminKey, getInfo])
+    loadConfig().catch(console.error)
+  }, [])
 
-  useEffect(loadConfig, [])
-
-  const value = { name, url, adminKey, saveConfig, clearConfig, enabled, sendPayment }
+  const value = { name, url, adminKey, saveConfig, clearConfig, enabled, getInfo, sendPayment }
   return (
     <LNbitsContext.Provider value={value}>
       {children}
--- a/components/webln/lnbits.js
+++ b/components/webln/lnbits.js
@@ -1,35 +1,45 @@
-export default {
-  storageKey: 'webln:provider:lnbits',
-  _url: null,
-  _adminKey: null,
-  enabled: false,
-  async load () {
-    const config = window.localStorage.getItem(this.storageKey)
-    if (!config) return null
-    const configJSON = JSON.parse(config)
-    this._url = configJSON.url
-    this._adminKey = configJSON.adminKey
-    await this._updateEnabled()
-    return configJSON
-  },
-  async save (config) {
-    this._url = config.url
-    this._adminKey = config.adminKey
-    await this._updateEnabled()
-    // XXX This is insecure, XSS vulns could lead to loss of funds!
-    //   -> check how mutiny encrypts their wallet and/or check if we can leverage web workers
-    //   https://thenewstack.io/leveraging-web-workers-to-safely-store-access-tokens/
-    window.localStorage.setItem(this.storageKey, JSON.stringify(config))
-  },
-  clear () {
-    window.localStorage.removeItem(this.storageKey)
-    this._url = null
-    this._adminKey = null
-    this.enabled = false
-  },
-  async getInfo () {
-    // https://github.com/getAlby/bitcoin-connect/blob/v3.2.0-alpha/src/connectors/LnbitsConnector.ts
-    const response = await this._request(
+import { createContext, useCallback, useContext, useEffect, useState } from 'react'
+
+// Reference: https://github.com/getAlby/bitcoin-connect/blob/v3.2.0-alpha/src/connectors/LnbitsConnector.ts
+
+const LNbitsContext = createContext()
+
+export function LNbitsProvider ({ children }) {
+  const [url, setUrl] = useState()
+  const [adminKey, setAdminKey] = useState()
+  const [enabled, setEnabled] = useState()
+
+  const storageKey = 'webln:provider:lnbits'
+
+  const request = useCallback(async (method, path, args) => {
+    let body = null
+    const query = ''
+    const headers = new Headers()
+    headers.append('Accept', 'application/json')
+    headers.append('Content-Type', 'application/json')
+    headers.append('X-Api-Key', adminKey)
+
+    if (method === 'POST') {
+      body = JSON.stringify(args)
+    } else if (args !== undefined) {
+      throw new Error('TODO: support args in GET')
+      // query = ...
+    }
+    const url_ = url.replace(/\/+$/, '')
+    const res = await fetch(url_ + path + query, {
+      method,
+      headers,
+      body
+    })
+    if (!res.ok) {
+      const errBody = await res.json()
+      throw new Error(errBody.detail)
+    }
+    return (await res.json())
+  }, [url, adminKey])
+
+  const getInfo = useCallback(async () => {
+    const response = await request(
       'GET',
       '/api/v1/wallet'
     )
@@ -47,9 +57,10 @@ export default {
       version: '1.0',
       supports: ['lightning']
     }
-  },
-  async sendPayment (bolt11) {
-    const response = await this._request(
+  }, [request])
+
+  const sendPayment = useCallback(async (bolt11) => {
+    const response = await request(
       'POST',
       '/api/v1/payments',
       {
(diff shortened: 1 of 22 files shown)


components/webln/lnbits.js
```diff
@@ -0,0 +1,113 @@
...
+      console.error(err)
+    }
+    return configJSON
+  },
+  async save (config) {
+    this._url = config.url
+    this._adminKey = config.adminKey
+    await this._updateEnabled()
+    // XXX This is insecure, XSS vulns could lead to loss of funds!
+    //   -> check how mutiny encrypts their wallet and/or check if we can leverage web workers
```
huumn, line 24: Mutiny just stores secrets in `IndexedDB` ... but they have a much smaller XSS attack surface.

If we are storing secrets on the client, the best we can do afaict:

1. have user encrypt it on client for storing in browser
2. on visit, prompt them to decrypt it, loading it into a web worker where spending conditions are enforced

If this is the best we can do ... we might as well store the encrypted secrets and spending conditions on the server for syncing across multiple devices ... then we also don't have to worry about the browser trashing our storage.

ekzyis: > Mutiny just stores secrets in IndexedDB

There is a password prompt to decrypt the wallet though so there is some encryption going on:

2024-01-14-201452_1920x1080_scrot

but not sure how much encryption would help against XSS anyway. the wallet/credentials would probably be unlocked/decrypted 99% of the time while the user is on the site. having to unlock the wallet regularly would be pretty annoying - but could be an option. like a 10min timer or something. but then again, not sure if even _that_ would help against XSS since the whole client should probably be considered compromised in that case.

> but they have a much smaller XSS attack surface.

Good point

> on visit, prompt them to decrypt it, loading it into a web worker where spending conditions are enforced

yeah, or go the web worker path.

> If this is the best we can do ... we might as well store the encrypted secrets and spending conditions on the server for syncing across multiple devices ... then we also don't have to worry about the browser trashing our storage.

mhh yes, let's see. I tested NWC with Mutiny a little bit and you basically have to approve every spend from your wallet (or configure budgets). so that seems to be pretty safe to use.

ekzyis:
Current state of UX:

https://github.com/stackernews/stacker.news/assets/27162016/81219503-ae61-4475-8c99-516abd14240e

- being optimistic about WebLN payments by wrapping them with cache updates (see [463578c](https://github.com/stackernews/stacker.news/pull/749/commits/463578cfeea026f9036da879411912d79b388368)). In case of failure, the update is undone.
- user feedback via second lightning strike that is thicker.
- if lightning animation was disabled, toasts are used.

However, what I actually intended was to resemble _**lightning and thunder**_. The existing lightning strikes could be used as before. However, for WebLN zaps, they simply mean that the payment was initiated but is still pending.

The second lightning strike is then the "thunder" to confirm the payment. Since [we probably shouldn't play sound](https://www.kirupa.com/ux/sound_ux.htm) since we're no longer in the MySpace era (but maybe it could be a fun setting?), I simply made the second lightning strike thicker. But I imagined the second lightning strike to additionally outline the previous one. This would then also feel more like a confirmation of what happened previously.

This is also similar to how messengers work to indicate the status of messages: One check mark means that the message was sent while two mean that the message was received. 

If we have two strikes (which completely overlap) then I think this could be quite intuitive for users.

components/webln/lnbits.js
huumn: > yeah, or go the web worker path

I think regardless of whether we have a web worker, the secrets will need to be stored somewhere for loading into the web worker in which case we will want to encrypt them. Although you might have something more clever planned already.

We've discussed it a little but anything we allow the app (SN) to do through the web worker, eg payments, will be vulnerable to XSS. We can do some things like making sure the payments are going to SN's node or doing some pre-authorization on the serverside, but afaict we can only add more steps to potential exploits (better than nothing but still).

--------------------

Regardless of these changes even, we should introduce really strict [Content Security Policies](https://nextjs.org/docs/pages/building-your-application/configuring/content-security-policy) to narrow the XSS attack surface.

huumn:
I think your intuition around having two signals is spot on.

however, without the explanation, ie just watching the video, I found the double strike confusing. I couldn't tell why the second strike was happening. 

Another variation:
1. no initial strike
2. persistent toast pops up while zap is pending
    - could have an 'undo'/'cancel'/retry button on it and any other state/progress/control mechanism on it
3. when the zap settles, lightning strike and toast goes away

(2) is something I've been wanting to do for even custodial zaps, giving stackers a chance to cancel zaps.

ekzyis:
Okay, I think this is basically done now from a feature and UX perspective.

Only things left to do:

- [ ] think about security for MVP: NWC allows budget controls so can we ship NWC using unencrypted local storage? Add warning about XSS which users can understand? Wallets contain similar warnings:

2024-01-31-085018_1920x1080_scrot

One idea I had was to only store the decryption key on the server which is bound to having a valid session. But I think the single point of failure is always the client in that case. If we automatically decrypt the wallet credentials (for UX reasons), then XSS is still possible. I believe we can only make XSS harder but never prevent it when we store wallet credentials in the client, as you also mentioned [here](https://github.com/stackernews/stacker.news/pull/749#discussion_r1457813585):

> but afaict we can only add more steps to potential exploits (better than nothing but still).

related to #770 

- [ ] hide LNbits in production since it's 100% vulnerable to XSS

- [ ] consider E2E testing of approval of NWC invoices but that would mean I need to link my local polar network to a wallet that supports NWC. I thought about Alby but not sure that's worth the effort. If they implement NWC correctly, the code should just work.

I have an appointment now and will think more about this when I am back in a few hours.

ekzyis:
Test showcase in multiple parts because I found some bugs and didn't want to re-record everything. Just stopped the capture, fixed the bug and started new capture.

See https://files.ekzyis.com/public/sn/webln

ekzyis:
Ok, will look into #770 but since it's independent of this code, I marked this PR as ready for review.

NWC connections still use local storage in this PR since using IndexedDB is more complicated, it's not a long-term solution and not a real defense layer against XSS like encryption or CSP:

> In a nutshell, XSS means game over!

-- https://pragmaticwebsecurity.com/articles/oauthoidc/localstorage-xss.html

pages/settings/wallets/nwc.js
```diff
@@ -0,0 +1,71 @@
...
+    </CenterLayout>
+  )
+}
+
+export function NWCCard () {
+  const { enabled } = useNWC()
+  return (
+    <WalletCard
+      title='nwc'
+      badges={['send only', 'non-custodialish', 'async']}
```
ekzyis, line 66: Mhh, not sure about using `async` here. Might not be clear what is meant with that. I essentially what to mention that this wallet connection might require confirmation (if no budget is set) and supports budgets (payments within the budget are automatically approved). Some ideas I had: `budgetable` and `requires confirmation`

huumn: `budgetable` might be the most straightforward. I think we'll need to revisit the UI some in the future so we can fit things like `requires confirmation`.

components/webln/nwc.js
```diff
@@ -0,0 +1,224 @@
...
+      }
+    })().catch((err) => {
+      console.error(err)
+      setRelay(null)
+    })
+    return () => {
+      relay?.close()
+      setRelay(null)
+    }
+  }, [relayUrl])
```
ekzyis, line 53: In c91f9f22, I switched from opening a new relay connection for every zap to having a relay connection open the whole time and reuse that one for all zaps. I did this because it's more friendly towards the resources of the relay since if we zap multiple times without confirming, multiple connections would be kept open. It's generally more efficient to reuse an existing connection, for example due to less state that needs to be maintained and less overhead of opening and closing connections.

However, this code might be less stable now since it's missing reconnect code.

I also thought about having it open only for 60 seconds and then closing it so the relay doesn't have to keep a connection open. But might be something to think about later.

ekzyis:
Regarding UX, especially on mobile:

Would be great if wallets like [Mutiny](https://app.mutinywallet.com/) would support [push notifications with actions](https://web.dev/articles/push-notifications-display-a-notification#actions_buttons). Then users don't have to switch apps but can simply click `Approve` or `Deny` in the push notification. Similar to how you can mark messages as read without opening Telegram or Signal.

/cc @benthecarman @TonyGiorgio

benthecarman:
This is a long term plan but requires some architecture changes for us

huumn:
@ekzyis any advice on getting setup to QA this?

ekzyis:
> any advice on getting setup to QA this?

Oh yes, good question haha

For testing approvals in a E2E way, I [setup LNbits](https://github.com/lnbits/lnbits/blob/main/docs/guide/installation.md) (using the recommended option using `poetry`) and connected it to a third node in Polar:

2024-02-04-172531_1918x1059_scrot

For NWC requests, you can use [Mutiny wallet](https://app.mutinywallet.com/) and go to [settings > connections](https://app.mutinywallet.com/settings/connections) but since it's running on a different network, the NWC requests show up but you can't test approvals like this (only rejections). I added some code in _components/webln.js_ that can be used to mock the NWC responses:

```js
// XXX set this to mock NWC relays
const MOCK_NWC_RELAY = false

// need big timeout since NWC is async (user needs to confirm payment in wallet)
const timeout = MOCK_NWC_RELAY ? 3000 : 60000
let timer
const resetTimer = () => {
  clearTimeout(timer)
  timer = setTimeout(() => {
    sub?.close()
    if (MOCK_NWC_RELAY) {
      const heads = Math.random() < 0.5
      if (heads) {
        return resolve({ preimage: null })
      }
      return reject(new Error('mock error'))
    }
    return reject(new Error('timeout'))
  }, timeout)
}
if (MOCK_NWC_RELAY) return resetTimer()
```

I haven't found an easy way to test NWC in a E2E way, too. But the subscription code looks simple enough that it should "just work":

```js
const filter = {
  kinds: [23195],
  authors: [walletPubkey],
  '#e': [request.id]
}
const sub = relay.subscribe([filter], {
  async onevent (response) {
    resetTimer()
    try {
      const content = JSON.parse(await nip04.decrypt(secret, walletPubkey, response.content))
      if (content.error) return reject(new Error(content.error.message))
      if (content.result) return resolve({ preimage: content.result.preimage })
    } catch (err) {
      return reject(err)
    } finally {
      clearTimeout(timer)
      sub.close()
    }
  },
  onclose (reason) {
    clearTimeout(timer)
    reject(new Error(reason))
  }
})
```

@benthecarman, would you mind to review the `sendPayment` function in _components/webln/nwc.js_? That should work as-is, right?

benthecarman:
yeah that looks correct to me

components/webln/nwc.js
```diff
@@ -0,0 +1,224 @@
...
+          }
+        ], {
+          onevent (event) {
+            clearTimeout(timer)
+            const supported = event.content.split()
+            resolve(supported)
+            sub.close()
+          },
+          // some relays like nostr.mutinywallet.com don't support NIP-47 info events
+          // so we simply check that we received EOSE
```
TonyGiorgio, line 144: Is this required anymore? Mutiny has switched to `relay.mutinywallet.com` for all new connections and that should behave properly now. AFAIK, we were the only hacked together relay that didn't respond correctly.

ekzyis: > Mutiny has switched to relay.mutinywallet.com for all new connections and that should behave properly now. 

Yeah, works great now :)

However, I am not sure how other relays will behave in the future. Afaik, there are only two wallets that support NWC so far: Mutiny and Alby.

It's a trade-off between strictly enforcing spec compliance vs. hoping they just didn't implement the info event but do support `pay_invoice`.

But maybe I should indeed remove it for "fail fast" reasons (even if paying invoices might work).

TonyGiorgio: This was mostly a relay implementation problem, not a wallet service problem. NWC wallets _should_ be broadcasting the info event. We were previously but our relay was not responding to them. This was because we were using Blastr as our relay for these and that isn't really a proper relay. I doubt anyone else would be running custom relay software that doesn't do simple event type querying.

ekzyis: Makes sense, removed in 8db4cd1

huumn (changes requested):
A have a few nits too, but this is the real blocker which seems like a bug:

https://github.com/stackernews/stacker.news/assets/34140557/a7d4f0a6-d77c-4917-8518-ee8f341f05dd

I'm using lnbits and all of my payments display as pending despite while seeming to succeed. This might be a UX bug but it immediately confused me so we'll want a solution for it.

------------

Nits:
- we probably shouldn't call this all `webLn` which is a protocol we aren't using here
- my wallet indicator goes red if I put in an incorrect host (good) but I have no way to diagnose (bad)
     - to be fair with lnaddr I don't even make the indicator red when it's failing, but if we know it failed, it should be trivial to store the reason on the client and surface it for stackers
- having an independent context for each wallet source might be okay for now, but it'll be a lot later so maybe begin thinking about how to provide it in one context if not just implement it that way

-------------

We'll want docs on how to set this up using docker. It took me an hour or two to troubleshoot LNBits config problems.

------------

I'll continue testing and see what comes up before giving the code a thorough review.

huumn:
I went back to my tab and it no longer recognizes the setting?

https://github.com/stackernews/stacker.news/assets/34140557/d6c2e0f9-355b-4ca1-89ad-ba32cdf11458

ekzyis:
> I'm using lnbits and all of my payments display as pending despite while seeming to succeed. This might be a UX bug but it immediately confused me so we'll want a solution for it.

Mhh, interesting. The invoices get cancelled for some reason but the toast doesn't update to `zap failed`. So that's a UX bug. Your LNbits instance seems to return 520 for /api/v1/payments (sending the payment) though. What makes you think the payments succeeded?

Did it work before or is this the first time you tried it?

> we probably shouldn't call this all webLn which is a protocol we aren't using here

Mhh, yes, it's not _really_ WebLN since we're not using `window.webln` like in the [WebLN guide](https://www.webln.guide/introduction/readme) but afaict, WebLN is not a protocol, just a standard / interface. Essentially, I called it WebLN in our code since I implemented the same (but incomplete) interface using the React Context API: `enabled`, `getInfo` and `sendPayment` with same return values. So imo, the only difference is `useWebLN` vs `window.webln` and that it's incomplete since we don't need anything else.

This integrates a lot better into our code than attaching something to `window`. [Bitcoin Connect also no longer attaches anything to `window` by default](https://github.com/getAlby/bitcoin-connect?tab=readme-ov-file#webln-global-object). And they mention this in their README:

> This project includes web components for connecting to Lightning wallets and enabling [WebLN](https://webln.guide/). Websites only need to interface with WebLN to connect with multiple wallets, just like the Alby extension.

I would say we're doing the same: we are interfacing with WebLN.

> - my wallet indicator goes red if I put in an incorrect host (good) but I have no way to diagnose (bad)
>    - to be fair with lnaddr I don't even make the indicator red when it's failing, but if we know it failed, it should be trivial to store the reason on the client and surface it for stackers

Good point, that was an oversight. I can add a toast for that.

> having an independent context for each wallet source might be okay for now, but it'll be a lot later so maybe begin thinking about how to provide it in one context if not just implement it that way

Not sure if you already looked into the code, but the consuming code only sees one context: the `WebLNContext`. This context wraps a context per WebLN provider:

```js
export function WebLNProvider ({ children }) {
  return (
    <LNbitsProvider>
      <NWCProvider>
        <RawWebLNProvider>
          {children}
        </RawWebLNProvider>
      </NWCProvider>
    </LNbitsProvider>
  )
}

export function useWebLN () {
  return useContext(WebLNContext)
}
```

All we need to do to switch between providers is to make `provider` a state variable with `useState()` in _components/webln/index.js_ and then use `setProvider` for switching. I think this is a good pattern since I think this exactly when one should use the Context API: We want to _provide_ WebLN providers to the rest of the application and we want to have a nice interface for them (so we can just call `useWebLN` where we need them), so these WebLN (context) providers are wrapped by another context provider.

> We'll want docs on how to set this up using docker. It took me an hour or two to troubleshoot LNBits config problems.

Ah yes, iirc it took me some time to realize that I had to configure `LNBITS_ALLOWED_FUNDING_SOURCES` and `LNBITS_BACKEND_WALLET`, too. Or did you have other problems which I should clarify in the documentation?

> I went back to my tab and it no longer recognizes the setting?

Ah yes, the config gets saved immediately but the status check using `getInfo` is done asynchronously as an effect with `useEffect`. Might make more sense to make it blocking and also use double toasts for pending status check and it it failed or was successful. Related to missing information on red wallet indicator.

huumn:
> Your LNbits instance seems to return 520 for /api/v1/payments (sending the payment) though. What makes you think the payments succeeded?

That was an old log entry. Notice how the error isn't present later.

> Essentially, I called it WebLN in our code since I implemented the same (but incomplete) interface using the React Context API: enabled, getInfo and sendPayment with same return values.

Fair point. I guess so long as the interface is identical we can call it webln.

> Not sure if you already looked into the code, but the consuming code only sees one context: the WebLNContext. 

I see. That's probably the right way to do this. I hadn't looked at the code much.

ekzyis:
> That was an old log entry. Notice how the error isn't present later.

As mentioned in Slack, hopefully this was just because the worker not running. I had an issue that looked very like yours just now and running the worker fixed it. The worker is required to mark HODL invoices as held so we can retry the action.

> I briefly looked at the function. It looks like you're doing that thing where state update 1 -> side effect 1 -> state update 2 -> side effect 2. While you're revisiting the code, consider if this side effect chaining can be avoided. One way I try to trick my brain into avoiding this is to think of useEffect as an antipattern (like goto in C) and to not use it expect where I have no other choice (I still end up using it more than I should but maybe it'll help you notice it more too).

I'll try to integrate this into formik validation. Makes no sense to close the form and _then_ check fors errors. We should keep the form open until the user entered something that works.

huumn:
> As mentioned in Slack, hopefully this happened just because the worker was not running. I had an issue that looked very like yours just now and running the worker fixed it. The worker is required to mark HODL invoices as held so we can retry the action.

Confirmed. I just didn't have the worker running. I'll push in some lnbits docs as I continue testing then do a line by line review.

huumn:
When I post with `0 sats`, the toast says "zap pending" even though I'm just paying posting fees.

huumn:
Bug during commenting: when I submit a comment with fees, it says it failed but it actually succeeded.


https://github.com/stackernews/stacker.news/assets/34140557/cca6a0b7-3565-4488-8948-09b9613457a5

ekzyis:
Wallet configs are now validated on save:

https://github.com/stackernews/stacker.news/assets/27162016/93f270cc-0f6f-4015-a786-dd0b3896e688

However, I realized that using the `info` event for validation does not guarantee that the secret is correct since it's not needed to receive the info event. Only the relay URL and the wallet pubkey must be correct.

I will add some static checks though (correct amount of characters etc).

ekzyis:
> When I post with 0 sats, the toast says "zap pending" even though I'm just paying posting fees.
>
> Edit: same thing on the donation page

> Bug during commenting: when I submit a comment with fees, it says it failed but it actually succeeded.

Oh, I noticed I only tested with zaps and forget to test other things that require payment since I added the toasts. Will do that now

ekzyis:
Tested paying with attached wallet in cc6c460:

- [x] zap
- [x] custom zap (using long press)
- [x] commenting
- [x] donation
- [x] post


https://github.com/stackernews/stacker.news/assets/27162016/7860a634-9044-4d52-ae0f-a6d4d3e02467

huumn:
Will do. 

(Apologies for the stray commit. I didn't check before pushing and now I'd have to force push.)

Other changes are:
- docs for local lnbits setup in docker
- schema allow lnbits on localhost in dev

huumn:
Confirmed. Works really well now! I'll do a pass over the code. I think the only thing we might want to add before shipping is allowing selection of one send-only wallet or the other as default, so that people can add two.

> hide LNbits in production since it's 100% vulnerable to XSS

LNbits has this notion of wallets which are like budgeted accounts so it's not much worse than NWC in that respect. So, I think we can ship with it.

In either case, it might worth adding a warning to either method like 

> Your wallet's credentials are stored in the browser and never go to the server. Absolutely, positively, listen when we say you should **set a budget on the wallet credentials**. Also, for the time being, you will have to reenter your credentials on other devices.

components/webln/nwc.js
```diff
@@ -0,0 +1,247 @@
...
+  const clearConfig = useCallback(() => {
+    window.localStorage.removeItem(storageKey)
+    setNwcUrl('')
+    setRelayUrl(undefined)
+    setWalletPubkey(undefined)
+    setSecret(undefined)
+    setEnabled(undefined)
+  }, [])
+
+  useEffect(() => {
```
huumn, line 84: Do we need the `useEffect` here? Just double checking because you managed to remove the similar logic in the lnbits provider.

huumn (approved):
Fantastic work! The UX is unreal. Stackers are going to love this!

I left one note, but the rest is solid. I think the only other things we might want:

- optional nit: a checkbox to say one send-only provider should be preferred over the other
- a warning on these send only methods cautioning folks to make sure they have limited access to funds.

We don't *need* that as we default to NWC, so I'll leave that up to you.

components/webln/nwc.js
ekzyis: I could also add logic to `loadConfig` and `saveConfig` to open/close relays on NWC url changes. But I think this `useEffect` here is actually pretty nice since it handles closing the relay automatically with the cleanup function and the dependency also automatically handles not closing the relay if the relay URL did not change on a save.

Looking at `loadConfig` and `saveConfig` again, I think they could be refactored even more (since they still have some shared code) and handling the relay connection there would make it more complicated.

So I think this `useEffect` is doing us some favors.

ekzyis:
> LNbits has this notion of wallets which are like budgeted accounts so it's not much worse than NWC in that respect. So, I think we can ship with it.
>
> In either case, it might worth adding a warning to either method like
>
>> Your wallet's credentials are stored in the browser and never go to the server. Absolutely, positively, listen when we say you should set a budget on the wallet credentials. Also, for the time being, you will have to reenter your credentials on other devices.

Mhh, makes sense! Since we can not only pay invoices but also fetch their balance with their LNbits admin key, we could additionally enhance this warning if we detect they have 100k+ sats in their wallet (for example). This would also make it clear that SN can see their balance when they use LNbits. They have to trust us that there is no code[^1] running that stores that somewhere.

I'll do this in another PR so you can easily dismiss that change if you don't think that's a good idea.

> optional nit: a checkbox to say one send-only provider should be preferred over the other
a warning on these send only methods cautioning folks to make sure they have limited access to funds.

:+1: 

[^1]: Even though we're FOSS, "hidden code" could be running which isn't even checked into Github. Verifying which code is actually running on a website is an unsolved problem afaict.

ekzyis:
4b95add adds a `default payment method` checkbox. This required more code (including some `useEffect`'s again :sweat_smile:) than expected. Mostly because there are some cases to consider:

- first enabled payment method -> set default
- different payment method set as default -> unset default for other payment methods
- payment method is no longer enabled -> unset default

But seems to work fine now:

https://github.com/stackernews/stacker.news/assets/27162016/16bed933-1356-42ae-b041-5262c5ad0cd4

Some issues I have with the current implementation:

- [UX] there should probably be an indicator on the card what is the default -> use outline with CSS?
- [UX] should you be able to unset default payment method if that is the only enabled one?
- [code] maybe I can store the default payment method in `WebLNContext` so the individual providers themselves don't need to know if they are the default provider or not :thinking:

I'll think about this more while I add the warning and eat something.

ekzyis:
Added disclaimer in 7a55913:

2024-02-08-184011_1920x1080_scrot

huumn:
Looks great. Ready for merge?

PR #752, opened 2024-01-15
refactor: replace recursion with promise sequence in `useInvoiceable`

ekzyis:
Commit description:

> This commit refactors `useInvoicable`. The hard-to-follow recursion was replaced by awaiting promises which resolve or reject when one step of our JIT invoice flow is done.
>
> Therefore, `onSubmit` is now fully agnostic of JIT invoices. The handler only returns when payment + action was successful or canceled - just like when a custodial zap was successful.

This was done while working on #749 since we will heavily rely on JIT invoices there and thus this was a good time to tackle this TODO:

> // TODO: refactor this so it can be easily understood
// there's lots of state cascading paired with logic
// independent of the state, and it's hard to follow

I think this is much better now. There is no recursion anymore.

Keeping in draft since I want to sleep over this and do another review and test tomorrow.

---

update: this should be stable now. [video](https://files.ekzyis.com/public/sn/invoiceable_refactor.mp4)

### DIFF
--- a/components/invoice.js
+++ b/components/invoice.js
@@ -230,26 +230,25 @@ export const useInvoiceable = (onSubmit, options = defaultOptions) => {
     }
 
     // retry until success or cancel
-    // eslint-disable-next-line no-async-promise-executor, promise/param-names
-    return await new Promise(async (resolveAction, rejectAction) => {
-      const reject = async () => {
+    return await new Promise((resolve, reject) => {
+      const cancelAndReject = async () => {
         await cancelInvoice({ variables: { hash: inv.hash, hmac: inv.hmac } })
-        rejectAction(new Error('invoice canceled'))
+        reject(new Error('invoice canceled'))
       }
       showModal(onClose => {
         return (
           <JITInvoice
             invoice={inv}
             onCancel={async () => {
-              await reject()
+              await cancelAndReject()
               onClose()
             }}
             onRetry={async () => {
-              resolveAction(await retry())
+              resolve(await retry())
             }}
           />
         )
-      }, { keepOpen: true, onClose: reject })
+      }, { keepOpen: true, onClose: cancelAndReject })
     })
   }, [onSubmit, createInvoice, !!me])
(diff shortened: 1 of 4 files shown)


components/modal.js
```diff
@@ -33,6 +33,7 @@ export default function useModal () {
     }
     const previousModalContent = modalStack[modalStack.length - 1]
     setModalStack(modalStack.slice(0, -1))
+    modalOptions?.onClose?.()
```
ekzyis, line 36: This is required to also cancel HODL invoices in case the user clicks on `<-` and already paid:

2024-01-16-023623_524x702_scrot

components/invoice.js
```diff
@@ -117,28 +110,37 @@ const MutationInvoice = ({ id, hash, hmac, errorCount, repeat, onClose, expiresA
...
-                className='mx-1'
-                variant='danger' onClick={async () => {
-                  await cancelInvoice({ variables: { hash, hmac } })
-                  onClose()
+                className='mx-1' variant='info' onClick={async () => {
+                  try {
+                    await onRetry()
+                  } catch (err) {
+                    console.error('retry error:', err)
+                    setRetryError(retryError + 1)
```
huumn, line 135: nit: if you're referencing prior state, you usually want to use the function form of `setState`:

`setRetryError(retryError => retryError + 1)`

huumn (approved):
Dude so impressed! Excellent use of promises and very clean and clear overall.

I have one nit, but we can ignore it if you want.

ekzyis:
> Dude so impressed! Excellent use of promises and very clean and clear overall.

Thank you :)

Key takeaway from this PR for me: Don't ignore the eslint rule `no-async-promise-executor`.

I spent hours debugging why my errors are not getting caught by the `try/catch` in the calling context until I found [this](https://stackoverflow.com/questions/43083696/cant-throw-error-from-within-an-async-promise-executor-function).

Some warnings really make sense.

PR #753, opened 2024-01-16
Fix lightning strike before zap

ekzyis:


ekzyis:
Mhh, there is still a lightning stike before I zap. Not sure what I intended with this then. Going to close this since this might also be are regression. This might have written like this on purpose because of optimistic responses.

Issue #763, opened 2024-01-19
Push notification shows wrong deposit amount


ekzyis:
**Description**

Multiple push messages are sent on deposits. They then get merged to show a wrong amount. See attached video.

**Steps to Reproduce**

Deposit funds

**Expected behavior**

Push notifications shows correct amount

**Video**

https://github.com/stackernews/stacker.news/assets/27162016/d33f47c9-d307-40f5-b170-d7be5fe18721

**Additional context**

Probably related to #726

ekzyis:
What I found out yesterday: It's related to reconnecting but I verified that the `finally` branch does execute. I thought it might not because we rethrow the error in `catch` but it does.

But seems like old subscriptions still continue to fire on reconnection.

ekzyis:
Mhh, can't reproduce locally anymore

huumn:
Wasn't this caused by what you fixed in #764?

ekzyis:
> Wasn't this caused by what you fixed in #764?

No, #764 was only concerned with worker behavior on reconnects: it confirmed already paid invoices again.

This is about multiple push messages being sent per deposit. You can test in prod by deposit 100 sats. You should receive a deposit notification that says you deposited 400 sats.

Since only deposit notifications seem to be broken, I think this is related to the #726. But afaict, we correctly unsubscribe on errors. And as mentioned, wasn't able to reproduce locally.

Maybe we just need to restart the worker? When you deployed https://github.com/stackernews/stacker.news/commit/70aa7dd1ad6f3da7fee10044c284d8f27c7f3cfe, did you also restart the worker?

Or do you only restart it when there are changes to it? (I would guess the latter)

huumn:
Everything rebuilds and restarts on every deployment.

ekzyis:
> Everything rebuilds and restarts on every deployment.

Mhh, I still get "4 sats deposited" when someone sends me 1 sat.

PR #764, opened 2024-01-19
Fix stale confirmedIndex used on reconnect

ekzyis:
On reconnects, we used a stale value for `confirmedIndex`. This means that we would confirm all invoices again since the worker was started.

Found this while debugging #763 but this fixes not #763.

### DIFF
--- a/worker/wallet.js
+++ b/worker/wallet.js
@@ -22,11 +22,16 @@ function subscribeForever (subscribe) {
     let sub
     try {
       return await new Promise((resolve, reject) => {
-        sub = subscribe(resolve, bail)
+        const sub = subscribe(resolve, bail)
         if (!sub) {
-          return bail(new Error('function passed to subscribeForever must return a subscription object'))
+          return bail(new Error('function passed to subscribeForever must return a subscription object or promise'))
         }
-        sub.on('error', reject)
+        if (sub.then) {
+          // sub is promise
+          sub.then(sub => sub.on('error', reject))
+        } else {
+          sub.on('error', reject)
+        }
       })
     } catch (error) {
       console.error(error)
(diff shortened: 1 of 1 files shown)


huumn:
Kind of a nit, but do we need to *require* `subscribeForever` receives a promise just because in one case it's convenient to pass one.

If it's easy to make to the promise optional, I think that's what we should do.

ekzyis:
> Kind of a nit, but do we need to _require_ `subscribeForever` receives a promise just because in one case it's convenient to pass one.
> 
> If it's easy to make to the promise optional, I think that's what we should do.

Good point, my initial fix tried to do this and it looked like this:

```diff
diff --git a/worker/wallet.js b/worker/wallet.js
index 99bd10bd..a91c8edf 100644
--- a/worker/wallet.js
+++ b/worker/wallet.js
@@ -22,11 +22,16 @@ function subscribeForever (subscribe) {
     let sub
     try {
       return await new Promise((resolve, reject) => {
-        sub = subscribe(resolve, bail)
+        const sub = subscribe(resolve, bail)
         if (!sub) {
-          return bail(new Error('function passed to subscribeForever must return a subscription object'))
+          return bail(new Error('function passed to subscribeForever must return a subscription object or promise'))
+        }
+        if (sub.then) {
+          // sub is promise
+          sub.then(sub => sub.on('error', reject))
+        } else {
+          sub.on('error', reject)
         }
-        sub.on('error', reject)
       })
     } catch (error) {
       console.error(error)
@@ -45,13 +50,12 @@ const logEventError = (name, error) => console.error(`error running ${name}`, er
 async function subscribeToDeposits (args) {
   const { models, lnd } = args

-  const [lastConfirmed] = await models.$queryRaw`
+  subscribeForever(async () => {
+    const [lastConfirmed] = await models.$queryRaw`
     SELECT "confirmedIndex"
     FROM "Invoice"
     ORDER BY "confirmedIndex" DESC NULLS LAST
     LIMIT 1`
-
-  subscribeForever(() => {
     const sub = subscribeToInvoices({ lnd, confirmed_after: lastConfirmed?.confirmedIndex })

     sub.on('invoice_updated', async (inv) => {
```

So if you prefer that, I can use that. I don't have a strong opinion about either approach.

huumn:
Seems better to me. I prefer abstractions eat the complexity of using them.

ekzyis:
> Seems better to me. I prefer abstractions eat the complexity of using them.

Done. Tested by starting worker, paying an invoice, restarting LND and paying another invoice (to make sure reconnect worked).

On master, the invoice is confirmed again after restarting LND which means another push notification is sent.

PR #767, opened 2024-01-24
Allow territory founders to pin items

### LINKED ISSUE #61: Post pinning
I want posts to be able to be pinned.

I mostly want this for a daily discussion thread, but if that's what we want there will need to be recurrence. Also, there's a question of whether pinning should be a mod only ability or anyone can pay to get a post pinned. 

It makes sense to give mods this ability, especially for recurrent pins. 

However, perhaps there can be a running auction for pinned posts that non-mods can initiate - basically an ad auction.

ekzyis:
Closes #61 

TODO:
- [x] [bug] update cache for all items to move item to correct position after pin update
- [x] [bug] fix position bug for position > 1
- [x] [feature] implement pinning of comments _(optional)_
- [x] ~[ux] allow pinned items to get zapped _(requesting for comments, optional)_~ _MVP feature creep, let's get some user feedback first_
- [x] [ux] ~use usual sorting (`hot`) for posts pinned by users. currently, it uses `recent` like saloon~ _actually, `recent` makes a lot of sense for pinned posts, not only saloon_
- [x] [bug] fix horizontal layout shift when pinning replies (minimal but still visible)

Current state (b88080c4):

https://github.com/stackernews/stacker.news/assets/27162016/18f21a8f-e531-4f55-b874-63bd9c481000

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -629,51 +629,58 @@ export default {
         throw new GraphQLError('you must be logged in', { extensions: { code: 'FORBIDDEN' } })
       }
 
-      const [old] = await models.$queryRawUnsafe(`${SELECT}, p.position FROM "Item" LEFT JOIN "Pin" p ON p.id = "Item"."pinId" WHERE "Item".id = $1`, Number(id))
-      if (old.subName) {
+      const [item] = await models.$queryRawUnsafe(`${SELECT}, p.position FROM "Item" LEFT JOIN "Pin" p ON p.id = "Item"."pinId" WHERE "Item".id = $1`, Number(id))
+
+      if (!item.subName && !item.parentId) throw new GraphQLError('item must have subName or parentId', { extensions: { code: 'BAD_INPUT' } })
+
+      if (item.subName) {
         // is post since replies don't have subName set
-        const sub = await models.sub.findUnique({ where: { name: old.subName } })
+        // only territory founder can pin posts
+        const sub = await models.sub.findUnique({ where: { name: item.subName } })
         if (Number(me.id) !== sub.userId) throw new GraphQLError('not your sub', { extensions: { code: 'FORBIDDEN' } })
       }
-      if (old.parentId) {
-        if (old.path.split('.').length > 2) throw new GraphQLError('can only pin root replies', { extensions: { code: 'FORBIDDEN' } })
-        const root = await models.item.findUnique({ where: { id: Number(old.parentId) }, include: { pin: true } })
+      if (item.parentId) {
+        // OPs can only pin top level replies
+        if (item.path.split('.').length > 2) throw new GraphQLError('can only pin root replies', { extensions: { code: 'FORBIDDEN' } })
+        const root = await models.item.findUnique({ where: { id: Number(item.parentId) }, include: { pin: true } })
         if (root.userId !== Number(me.id)) throw new GraphQLError('not your post', { extensions: { code: 'FORBIDDEN' } })
       }
 
       let pinId
-      if (old.pinId) {
-        const args = [old.position]
-        if (old.subName) args.push(old.subName)
-        else args.push(old.parentId)
+      if (item.pinId) {
+        // item is already pinned. remove pin
+        const args = [item.position]
+        if (item.subName) args.push(item.subName)
+        else args.push(item.parentId)
         await serialize(
           models,
-          models.item.update({ where: { id: old.id }, data: { pinId: null } }),
-          models.pin.delete({ where: { id: old.pinId } }),
-          // make sure that pins in a sub have no gaps
+          models.item.update({ where: { id: item.id }, data: { pinId: null } }),
+          models.pin.delete({ where: { id: item.pinId } }),
+          // make sure that pins have no gaps
           models.$queryRawUnsafe(`
           UPDATE "Pin"
           SET position = position - 1
           WHERE position > $1 AND id IN (
             SELECT "pinId" FROM "Item" i
-            ${whereClause('"pinId" IS NOT NULL', old.subName ? 'i."subName" = $1' : 'i."parentId" = $1')}
+            ${whereClause('"pinId" IS NOT NULL', item.subName ? 'i."subName" = $1' : 'i."parentId" = $1')}
           )
           `, ...args)
         )
         pinId = null
       } else {
+        // pin item
         const args = []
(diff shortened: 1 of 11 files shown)


ekzyis:
Fixed bugs for post pinning. New state (0031d96f):

https://github.com/stackernews/stacker.news/assets/27162016/1b0afe82-4b7b-4a9a-9c05-e99b027b7517

Will look into comment pinning now

ekzyis:
Top replies can now also be pinned (73a9c4ef): 

https://github.com/stackernews/stacker.news/assets/27162016/d0f4b032-f4f5-459d-9531-358692e5da2e

api/resolvers/item.js
```diff
@@ -495,6 +495,7 @@ export default {
                         FROM "Item"
                         ${whereClause(
                           '"pinId" IS NOT NULL',
+                          '"parentId" IS NULL',
```
ekzyis, line 498: required to exclude pinned comments from query

components/items.js
```diff
@@ -34,15 +34,41 @@ export default function Items ({ ssrData, variables = {}, query, destructureData
...
     <>
       <div className={styles.grid}>
-        {items.filter(filter).map((item, i) => (
-          <Fragment key={item.id}>
-            {pinMap && pinMap[i + 1] && <Item item={pinMap[i + 1]} />}
-            <ListItem item={item} rank={rank && i + 1} siblingComments={variables.includeComments} />
-          </Fragment>
-        ))}
+        {items.filter(filter).map(mergePins)}
+        {remainingPins.map((item, i) => <Item key={item.id} item={item} />)}
```
ekzyis, line 71: This fixes the issue that the saloon only shows up if there are enough items[^1]. If there aren't enough items now, it just shows up at the end even if that means it shows up at a position < 8.

[^1]: I assume this wasn't intended behavior.

huumn:
Overall great work. I didn't want to bother you with the changes I made while debugging given they were minor and didn't review this in a timely way. General feedback:

1. `pinItem` looked more like a rough draft than a final draft
    - many (> 4 instances) of `if (whatever) very long line of code > 100 chars`
        - in general, I'd reserve single line `if`s for very concise things
        - regardless, when lines > 80-100 chars, try to wrap them as a rule
    - redundant logic (eg constructing `args`)
    - deleting pins were atomic, but adding pins weren't
2. avoid using `serialize` when it's not necessary because it's very expensive
    - afaict we should only be using `serialize` when dealing with wallet balances
3. `mergePins` was pretty complicated and hard to read
    - it might've been the right thing to do, but it just tacked on to `pinMap`
    - `mergePins` should've probably returned a new `items` array with `pins` inserted appropriately

ekzyis:
> 1. `pinItem` looked more like a rough draft than a final draft
>    - many (> 4 instances) of `if (whatever) very long line of code > 100 chars`
>        - in general, I'd reserve single line ifs for very concise things

Oh, yes, you mean for example this:

```js
if (!item.subName && !item.parentId) throw new GraphQLError('item must have subName or parentId', { extensions: { code: 'BAD_INPUT' } })
```

Yeah, I agree now. While working on different code afterwards, I realized how it's easier to read if it's on multiple lines and not on one big line. Wasn't obvious to me before. For some reason, I thought one line is easier to read because it's "more concise". But not adding `\n` and `{}` is obviously saving space at the wrong place in hindsight lol

> regardless, when lines > 80-100 chars, try to wrap them as a rule

:+1:

> deleting pins were atomic, but adding pins weren't

Ah yes, that's where I stumbled over not having trivial enough access to the return value of the previous query and I noticed that `UPDATE` queries only allow `sub-SELECT`[^1]. So my first attempt didn't work:

```sql
UPDATE "Item"
SET "pinId" = (INSERT INTO "Pin" ... RETURNING id)
```

Then I went down the easy path for now and forgot to revisit it to fix. Didn't try out using a CTE as you did now.

> avoid using serialize when it's not necessary because it's very expensive

Expensive because it uses `SERIALIZABLE` which is the strictest transaction isolation level? So we're giving up on all concurrency?[^2]

> The Serializable isolation level provides the strictest transaction isolation. This level emulates serial transaction execution for all committed transactions; as if transactions had been executed one after another, serially, rather than concurrently.

_(might be a rhetorical question but I want to make sure I didn't miss something in your argument)_

> `mergePins` was pretty complicated and hard to read

I think I didn't make it clear what the purpose of `mergePins` was. It was supposed to fix this bug which happens if two pins follow each other. The previous (and current) code only allows an unpinned item to follow a pinned item:

https://github.com/stackernews/stacker.news/assets/27162016/eeae3bd7-9c83-43b6-b656-5d4a832e3185

---

[^1]: https://www.postgresql.org/docs/13/sql-update.html. Also mentioning this since I found it interesting. Didn't know that's a thing.
[^2]: https://www.postgresql.org/docs/13/transaction-iso.html#XACT-SERIALIZABLE

huumn:
> Expensive because it uses SERIALIZABLE which is the strictest transaction isolation level? So we're giving up on all concurrency?

Yes exactly. Serializing db changes prevents dirty reads and writes - while generally uncommon, when you're dealing with money you can suffer fund loss.

huumn:
> I think I didn't make it clear what the purpose of mergePins was. It was supposed to fix this bug which happens if two pins follow each other. 

You're right. My fix was broken. [e782bf0](https://github.com/stackernews/stacker.news/pull/767/commits/e782bf0565d3b4fbcdb073b05ecbb270f0c1869a) should fix it.

components/item-info.js
```diff
@@ -174,6 +178,18 @@ export default function ItemInfo ({
             <hr className='dropdown-divider' />
             <MuteSubDropdownItem item={item} sub={sub} />
           </>}
+        {canPin &&
+          <>
+            <hr className='dropdown-divider' />
+            <PinSubDropdownItem item={item} />
+          </>}
```
ekzyis, line 185: Why did you decide to put this into a own section?

I first put it into the territory section (since only founders can pin) but when I added comment pinning, I moved it the "item section".

huumn: Oh there are sections? lol

ekzyis: Ah ok, nevermind. It's in the same section (between two `<hr className='dropdown-divider' />`) as `crosspost to nostr`. I think it makes sense now. I didn't read the code properly.

PR #769, opened 2024-01-28
Fix image deletion batch size

ekzyis:
The job failed if we tried to delete more than 1000 keys at once.

This was mentioned [here](https://docs.aws.amazon.com/cli/latest/reference/s3api/delete-objects.html):

> The request can contain a list of up to 1000 keys that you want to delete.

### DIFF
--- a/api/s3/index.js
+++ b/api/s3/index.js
@@ -24,14 +24,29 @@ export function createPresignedPost ({ key, type, size }) {
   })
 }
 
-export function deleteObjects (keys) {
+export async function deleteObjects (keys) {
   const s3 = new AWS.S3({ apiVersion: '2006-03-01' })
-  return new Promise((resolve, reject) => {
-    s3.deleteObjects({
-      Bucket,
-      Delete: {
-        Objects: keys.map(key => ({ Key: String(key) }))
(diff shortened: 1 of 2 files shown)


api/s3/index.js
```diff
@@ -24,14 +24,29 @@ export function createPresignedPost ({ key, type, size }) {
...
-    s3.deleteObjects({
-      Bucket,
-      Delete: {
-        Objects: keys.map(key => ({ Key: String(key) }))
+  // max 1000 keys per request
+  // see https://docs.aws.amazon.com/cli/latest/reference/s3api/delete-objects.html
+  const batchSize = 1000
+  const deleted = []
+  for (let i = 0; i < keys.length; i += batchSize) {
+    const batch = keys.slice(i, i + batchSize)
```
ekzyis, line 34: No problem here if `i+batchSize > keys.length`:

```
$ node
Welcome to Node.js v21.4.0.
> x = [1,2,3,4,5]
[ 1, 2, 3, 4, 5 ]
> x.slice(5,10)
[]
> x.slice(10,10)
[]
> x.slice(3,10)
[ 4, 5 ]
```

PR #772, opened 2024-01-30
Make pins zappable

ekzyis:
Follow-up to #767 

TODO:
- [x] be less lazy and color pins
- [x] show sats
- [x] saloon is also zappable now but i think this is no problem. we should handle saloon zaps like anon zaps.
- [x] testing (does this break stuff in the backend because of previous assumptions?)
- [x] ~only show pin icon where position is actually fixed~ _mhh, looking at the code, this isn't as trivial to change as I expected. But now I think it might actually be useful to see if an item is pinned even if it's not pinned on that particular page? I mentioned [here](https://stacker.news/items/407067) that I think both cases can be confusing since the pin icon also works as feedback that the item is well... pinned somewhere._
- [ ] add pin fees (100% reward pool if pinnee is OP else 50/50 pool/OP)

### DIFF
--- a/components/item-info.js
+++ b/components/item-info.js
@@ -57,7 +57,7 @@ export default function ItemInfo ({
 
   return (
     <div className={className || `${styles.other}`}>
-      {!item.position && !(!item.parentId && Number(item.user?.id) === AD_USER_ID) &&
+      {!(!item.parentId && Number(item.user?.id) === AD_USER_ID) &&
         <>
           <span title={`from ${numWithUnits(item.upvotes, {
               abbreviate: false,
--- a/components/item.js
+++ b/components/item.js
@@ -41,7 +41,7 @@ export default function Item ({ item, rank, belowTitle, right, full, children, s
         : <div />}
       <div className={`${styles.item} ${siblingComments ? 'pt-3' : ''}`}>
         {item.position
-          ? <Pin width={24} height={24} className={styles.pin} />
+          ? <UpVote item={item} className={styles.upvote} icon={<Pin width={24} height={24} className={styles.pin} />} />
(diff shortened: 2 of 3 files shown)


huumn:
This probably makes pins more subject to gaming. AFAICT pins being unzappable (with the author is pinning them) is a feature not a bug.

ekzyis:
> This probably makes pins more subject to gaming. AFAICT pins being unzappable (when the author is pinning them) is a feature not a bug.

So pins should only be zappable if it wasn't the OP who pinned it?

huumn:
I don't know. It might be best to let the feature breathe more and take reactionary user feedback with a grain of salt.

ekzyis:
> I don't know. It might be best to let the feature breathe more and take reactionary user feedback with a grain of salt.

We can also flip this around: What about adding zaps and then seeing if it gets gamed? I think as it is now, it's indeed kind of weird to use. Pinning is now a trading a fixed position for no zaps which is probably almost always a bad trade for pins.

huumn:
> We can also flip this around: What about adding zaps and then seeing if it gets gamed?

If you feel strong enough about this, go for it. Afaict:

- it makes pinning "useful" for things that aren't notices
- adds more code to maintain
- adds "hidden" functionality to pins

and it's unclear to me whether those add up to a good thing after a couple hours and a handful of pins have been created.

OneOneSeven117:
lightning-pin
lightning-pin2

New to github, I just thought these were a cool idea, if you wanted to have non-zappable and zappable pinned items and be able to differentiate.

huumn:
That looks nice! Maybe we could do something like that.

PR #774, opened 2024-01-31
Microservices

huumn:
This finally makes use of the `capture` microservice I wrote a few months ago. It also adds the microservice definitions to the repo:
- imgproxy
- capture

This is in preparation of running [lnproxy](https://lnproxy.org/) as a microservice.

Eventually we'll want to turn `worker` into a microservice among other things.

Note: There is an unrelated `npm audit fix` commit in here.

I'm going to merge immediately.


huumn:
@ekzyis do you get notified for these ephemeral PRs? Just wondering if I'll need to tag you in them.

ekzyis:
> @ekzyis do you get notified for these ephemeral PRs? Just wondering if I'll need to tag you in them.

Yes, I get a mail and a notification in Github

ekzyis (commented):
lgtm

Issue #768, opened 2024-01-27
Completely unrelated HN links are treated as dupes


kristapsk:
![image](https://github.com/stackernews/stacker.news/assets/4500994/f0675fde-21b0-466e-99b0-4f9a8084e958)

Related to #642?

PR #781, opened 2024-02-02
Fix hacker news and bitcointalk dupes

mzivil:
Fixes https://github.com/stackernews/stacker.news/issues/768

Looks like there was a regression at some point that made all hacker news links and all bitcointalk links appear as dupes. 

Originally, we expected that these links would be caught by the `whitelist.includes(uri)` if-condition, but now `uri` is a regex string (like `(%.)?bitcointalk.org(:[0-9]+)?/index.php`) so we never matched correctly against the whitelist values.

This returns the original behavior of checking the whitelist against the URL's hostname + pathname. It also renames a variable and reassigns the previously mutated variables to new variables for clarity.

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -556,9 +556,9 @@ export default {
       pathname = pathname.replace(/%/g, '\\%')
       pathname = pathname.replace(/_/g, '\\_')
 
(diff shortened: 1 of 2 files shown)


huumn:
Excellent! Figures there was a regression at one point

Issue #776, opened 2024-02-01
Search on URL to disregard www. unless search included www. 


cointastical:
Search phrase: CoinDesk.com

Expected results:
All link posts or description posts with CoinDesk.com  (including www.CoinDesk.com results)

Actual results:
Only link posts or description posts with CoinDesk.com   (i.e., does not include www.CoinDesk.com)

I suppose that might make it impossible for me to EXCLUDE posts with www.CoinDesk.com 

So maybe it would understand a wildcard rather than just dismissing the subdomain.
e..g,  URL:*.CoinDesk.com would include www.CoinDesk.com as well as CoinDesk.com results.  Eh, ... but then the "*." means anything followed by a dot.   And  CoinDesk.com would not match as there is no "." before the CoinDesk.com

Anyway, .... just requesting a feature so that Search works better in this instance of wanting a wider net when searching on a domain name only.

utanapishtim:
@huumn I can take a crack at this one. Will this require setting up semantic search vis-a-vis the docs?

huumn:
You should be able to use search locally without semantic search. It falls back to term matches so long as there isn't a model id set in `env`.

PR #782, opened 2024-02-03
Improved ux for domain only searches

utanapishtim:
Fixes: https://github.com/stackernews/stacker.news/issues/776

Updates `/api/resolvers/search.js` to wildcard match title and text on url passed in `url:` filter when searching.

### DIFF
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -195,7 +195,14 @@ export default {
       const { query, quotes, nym, url } = queryParts(q)
 
       if (url) {
-        whatArr.push({ match_phrase_prefix: { url: `${url.slice(4).toLowerCase()}` } })
+        const domain = url.startsWith('url:www.') ? url.slice(8) : url.slice(4)
+        const value = `*${domain}*`
+        const fields = ['title', 'text']
+
+        const queries = fields
+          .map((field) => ({ wildcard: { [field]: { value, case_insensitive: true } } }))
+
+        whatArr.push({ bool: { should: queries } })
       }
 
       if (nym) {
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -224,7 +224,6 @@ export default {
         }
       })
 
-
       for (const quote of quotes) {
         whatArr.push({
           multi_match: {
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -194,7 +194,7 @@ export default {
 
       const { query, quotes, nym, url } = queryParts(q)
 
-      if (url) {
+      if (url && !query.length) {
         const domain = url.startsWith('url:www.') ? url.slice(8) : url.slice(4)
         const value = `*${domain}*`
         const fields = ['title', 'text']
@@ -203,6 +203,8 @@ export default {
           .map((field) => ({ wildcard: { [field]: { value, case_insensitive: true } } }))
 
         whatArr.push({ bool: { should: queries } })
+      } else if (url) {
+        whatArr.push({ match_phrase_prefix: { url: `${url.slice(4).toLowerCase()}` } })
       }
 
       if (nym) {
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -171,7 +171,8 @@ export default {
     },
     search: async (parent, { q, sub, cursor, sort, what, when, from: whenFrom, to: whenTo }, { me, models, search }) => {
       const decodedCursor = decodeCursor(cursor)
-      let sitems
+      let sitems = null
+      let termQueries = []
 
       if (!q) {
         return {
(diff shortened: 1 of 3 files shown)


api/resolvers/search.js
```diff
@@ -195,7 +195,14 @@ export default {
       const { query, quotes, nym, url } = queryParts(q)
 
       if (url) {
-        whatArr.push({ match_phrase_prefix: { url: `${url.slice(4).toLowerCase()}` } })
+        const domain = url.startsWith('url:www.') ? url.slice(8) : url.slice(4)
```
utanapishtim, line 198: While OP wrote title as "disregard www. unless search included www." I thought widening the scope to other subdomains better satisfied OP's intention of "wanting a wider net when searching on a domain name only".

api/resolvers/search.js
```diff
@@ -194,7 +194,16 @@ export default {
...
+      if (url && !query.length) {
+        const domain = url.startsWith('url:www.') ? url.slice(8) : url.slice(4)
+        const value = `*${domain}*`
+        const fields = ['title', 'text']
+
+        const queries = fields
+          .map((field) => ({ wildcard: { [field]: { value, case_insensitive: true } } }))
+
+        whatArr.push({ bool: { should: queries } })
+      } else if (url) {
```
huumn, line 206: I don't think we need the wildcards on title and text. The analyzer we use on title and text you can perform a normal `match`-like query on a url, e.g. https://stacker.news/search?q=geyser.fund, so maybe we can just add the domain to the `query` variable after it's parsed?

Also more generally, [wildcards are slow](https://opensearch.org/docs/latest/query-dsl/term/wildcard/) and best to avoid them on large fields like `text` if we can.

api/resolvers/search.js
```diff
@@ -194,7 +194,16 @@ export default {
...
         whatArr.push({ match_phrase_prefix: { url: `${url.slice(4).toLowerCase()}` } })
```
huumn, line 207: It might be worth doing a `wildcard` match here instead of `match_phrase_prefix`.

huumn (changes requested):
Welcome to the team! I think these changes are good but I think they can be made better. 

OpenSearch queries are super finicky and dependent on the analyzers used to index things, so the changes I brought up are mostly related to the opaqueness of the system rather than your changes having issues.

api/resolvers/search.js
utanapishtim: I've tried adding the domain back to the query variable after parsing but to no effect, I'm curious if the tokenizer used in production is separate from the default one used for development?

Am I correct that the index is built here:  [`curl -X PUT "http://localhost:9200/item" -ku admin:admin`](https://github.com/stackernews/stacker.news/blob/65cc1dbcc0294f513f4da708889bf907593077f9/docker-compose.yml#L94)

Still investing a more performant fix.

huumn: Ah yes, locally you'll get a default mapping and `standard` analyzer for the index if you don't set a mapping with an `english` analyzer first. The semantic search doc has an example mapping, which if you remove the ml stuff is:

```json
{
  "mappings": {
    "properties": {
      "text": {
        "type": "text",
        "analyzer": "english",
        "fields": {
          "keyword": {
            "type": "keyword",
            "ignore_above": 256
          }
        }
      },
      "title": {
        "type": "text",
        "analyzer": "english",
        "fields": {
          "keyword": {
            "type": "keyword",
            "ignore_above": 256
          }
        }
      }
    }
  }
}
```

huumn (commented):
Sorry for the slow turnaround. I missed that you had updated this. I recommended one small change. If you disagree, let me know and I'll just merge it.

huumn (approved):

api/resolvers/search.js
utanapishtim: Awesome, this helped a lot.

api/resolvers/search.js
utanapishtim: Ended up removing this because the wildcard search on `url` in the filters removed all non-link results.

utanapishtim:
@huumn I think this is ready for a final review. After fiddling around with open search and testing against the english analyzer index was able to eliminate the dependence on wildcard queries.

api/resolvers/search.js
```diff
@@ -192,10 +193,16 @@ export default {
           break
       }
 
-      const { query, quotes, nym, url } = queryParts(q)
+      const { query: _query, quotes, nym, url } = queryParts(q)
+      let query = _query
```
utanapishtim, line 197: We may want to mutate `query` but still leave the other query parts immutable.

api/resolvers/search.js
```diff
@@ -192,10 +193,16 @@ export default {
...
+
+      const isUrlSearch = url && query.length === 0 // exclusively searching for an url
```
utanapishtim, line 199: Because the all the termQueries are nested within a `must` (logical AND) compound query[^1] we need to adjust the `minimum_should_match` param of the nested queries based on whether we are exclusively searching for a url or not. We use this var to indicate the intent of the user.


[^1]: https://opensearch.org/docs/latest/query-dsl/compound/bool/

api/resolvers/search.js
```diff
@@ -206,18 +213,17 @@ export default {
...
-            minimum_should_match: '100%',
-            boost: 1000
-          }
+      termQueries.push({
+        // all terms are matched in fields
+        multi_match: {
+          query,
+          type: 'best_fields',
+          fields: ['title^100', 'text'],
+          minimum_should_match: (isUrlSearch) ? 1 : '100%',
```
utanapishtim, line 222: If we are exclusively searching for an url, we just need to match either the root domain or the fqdn.

api/resolvers/search.js
```diff
@@ -282,7 +288,7 @@ export default {
             fields: ['title^100', 'text'],
             fuzziness: 'AUTO',
             prefix_length: 3,
-            minimum_should_match: '60%'
+            minimum_should_match: (isUrlSearch) ? 1 : '60%'
```
utanapishtim, line 291: If we are exclusively searching for an url, we just need to match either the root domain or the fqdn.

docs/testing-search.md
```diff
@@ -0,0 +1,27 @@
+# Testing Search
```
utanapishtim, line 1: for progeny

huumn:
This is good to merge, but it looks like `npm run lint` is failing. If you're using vscode, you'll want to setup https://marketplace.visualstudio.com/items?itemName=standard.vscode-standard

huumn:
Great work! See you on the next one!

Issue #773, opened 2024-01-30
Territory pins might prevent posts to show up on front page


ekzyis:
Items that were pinned in a territory don't show up on the main front page:

https://github.com/stackernews/stacker.news/assets/27162016/3f3ded4c-5ccd-4a59-8371-8d37f3288448

They should show up as if they weren't pinned (without a pin icon).

PR #783, opened 2024-02-03
allow pins to be zapped but not from pin position

huumn:
This allows pins to be zapped and shown when they aren't pinned in the ranking for the page, eg in top, home, recent, or when visited directly.

This solves some of the issue presented in #773 and some the issues to be solved in #772 without introducing anything too new. It doesn't solve every corner of the issue but I think it might solve a goldilocks amount.

Merging immediately.

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -369,7 +369,6 @@ export default {
               ${relationClause(type)}
               ${joinZapRankPersonalView(me, models)}
               ${whereClause(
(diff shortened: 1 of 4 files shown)


ekzyis:
Mhh, you mentioned [here](https://stacker.news/items/412792) that pinned posts no longer disappear from home but they do:

https://github.com/stackernews/stacker.news/assets/27162016/1b06f297-3d40-421a-9882-ada5b6e962bc

PR #784, opened 2024-02-03
Fix territory form label clicks not toggling correct input

mzivil:
On the territory form, clicking on the labels of the "post types" or "billing" options don't toggle the correct input:

https://github.com/stackernews/stacker.news/assets/158518982/991f675f-5dd8-4b5b-bbeb-02c521baa81d

All the labels within each group were being assigned the same html "id" since we were falling back to the "name" prop. We need to keep the "name" prop the same because that's how the form associates the value with the correct form field. Fix is to assign explicit "id" prop to each checkbox so that the label toggles the correct input.

https://github.com/stackernews/stacker.news/assets/158518982/9fa631b1-9c0c-4264-b865-a1f9286ae143

### DIFF
--- a/components/territory-form.js
+++ b/components/territory-form.js
@@ -172,6 +172,7 @@ export default function TerritoryForm ({ sub }) {
               label='100k sats/month'
               value='MONTHLY'
               name='billingType'
+              id='monthly-checkbox'
               readOnly={!!sub}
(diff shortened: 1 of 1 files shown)


huumn:
Dang how did no one report that!

PR #785, opened 2024-02-04
Fix pinned items don't show up in home

ekzyis:
Related to https://github.com/stackernews/stacker.news/pull/783#issuecomment-1925919942

TODO:
- [x] ~make sure pinned posts don't disappear from recent, too~ _(that was already the case)_


huumn:
Oh nice catch! For some reason I thought my other changes fixed this

Issue #626, opened 2023-11-12
Delete bolt11s from lnd too


huumn:
auto drop bolt11s only deletes them from our db

ekzyis:
> #793

If you include something like `close #793` to your PR description, the ticket will be automatically linked and closed on merge. See https://docs.github.com/en/enterprise/2.13/user/articles/closing-issues-using-keywords

PR #793, opened 2024-02-07
auto canceling bolt11s from lnd when auto dropped from DB

dillon-co:
pretty sure this will work but I'm currently setting up my local env with lnd and lightning to test. Making PR in case y'all want to review the code while I'm getting that all set up. I'll post a comment when I've tested it


worker/wallet.js
```diff
@@ -255,14 +255,24 @@ async function checkWithdrawal ({ data: { hash }, boss, models, lnd }) {
...
     AND now() > created_at + interval '${INVOICE_RETENTION_DAYS} days'
-    AND hash IS NOT NULL;`
+    AND hash IS NOT NULL
+    RETURNING hash, bolt11;)
+    SELECT * FROM updated_rows`
   )
+  // Iterate over each invoice to remove them using ln-service
+  for (const invoice of invoices) {
+    try {
+      await cancelHodlInvoice({ id: invoice.hash, lnd })
```
ekzyis, line 271: You want to use [`deletePayment`](https://www.npmjs.com/package/ln-service#deletepayment). `cancelHodlInvoice` is only to cancel HODL invoice that were paid but not settled yet.

dillon-co: Oh got it. I thought we wanted to cancel open invoices only. Just pushed

ekzyis (commented):
Nice work! I haven't pulled the code and ran some tests yet[^1] but the `try/catch` is definitely essential here. Without it, following invoices would also not get deleted just because `deletePayment` failed for one invoice.

However, it might still be a problem since even a single invoice deletion error means that deletion of this particular invoice will not get retried since the next run of this function will not return this invoice anymore afaict. Then we have a invoice sitting in LND with no trace of it in the DB.

Not sure what the best way to solve this is but this might be similar to what we do with image deletion. There, we fetch all images that should get deleted, then we delete them in AWS S3 and then we delete the ones that were successfully deleted in the database. See _worker/deleteUnusedImages.js_.

Maybe something similar can be done here?

[^1]: Will do so later today

dillon-co:
> However, it might still be a problem since even a single invoice deletion error means that deletion of this particular invoice will not get retried since the next run of this function will not return this invoice anymore afaict. Then we have a invoice sitting in LND with no trace of it in the DB.

Yeah that makes sense. I can update the code to account for that. if there's an error with deleting on LND should I save that in the db somewhere? I see a status enum that I can update to `unknown error` but not sure if I should add a new column for that or something in case we wanted to see what happened with it.


Also: I ran into some docker compose trouble with polar. I had the wrong docker-compose version on my machine so I've been working through that and it's turned out to be a huge headache lol. Still working through that but I'll update the code real quick, push, then fix my env so I can test. I feel lame haha. Onboarding pains are making me code way slower than usual

worker/wallet.js
```diff
@@ -255,14 +255,34 @@ async function checkWithdrawal ({ data: { hash }, boss, models, lnd }) {
...
+    }
+  }
+  if (successfulDeletes.length > 0) {
+    await models.withdrawl.deleteMany({
+      where: { hash: { in: successfulDeletes } }
+    })
+  }
+  if (failedDeletes.length > 0) {
+    await models.withdrawl.updateMany({
+      where: { hash: { in: failedDeletes } }, data: { status: 'UNKNOWN_FAILURE' }
```
ekzyis, line 283: You shouldn't overwrite the `status` column. That's to indicate the status of the withdrawal itself (`ROUTE_NOT_FOUND`, `CONFIRMED` etc.), not the status of a delete.

The `UPDATE` query in the existing code was good as it was.

ekzyis (commented):
I think there is some misunderstanding what #626 was about which makes sense since there was not much description in that ticket. I'll just explain the existing description here since I think you're almost there (pending me pulling the code and running it:

> auto drop bolt11s only deletes them from our db

The existing code only overwrote the hash and bolt11 of withdrawals in the database with `NULL`. We allow stackers to configure this for privacy reasons since a bolt11 payment request does contain the recipient node pubkey (lightning has bad receiver privacy). We also delete the hash since it can be used to verify PII. So the existing code was fine. We don't want to delete the whole withdrawal since it's useful for auditing; only the information that is sensitive should be deleted.

However, since the same information is still available in LND, we also want to delete it there. I don't know if it's possible to only drop the bolt11 and hash from an invoice so we probably have to delete the whole invoice with `deletePayment` there (as you already do).

worker/wallet.js
dillon-co: Ok I reverted it to setting those values to null and then adding the ones back into the database if they failed. Should I add another column to the withdrawl model to show an unsuccessful delete from LND?

ekzyis: > Should I add another column to the withdrawl model to show an unsuccessful delete from LND?

No, I think it's fine to just log the errors and keep retrying. At some point, we should log such errors into some channel that we monitor.

worker/wallet.js
```diff
@@ -255,14 +255,34 @@ async function checkWithdrawal ({ data: { hash }, boss, models, lnd }) {
...
+    try {
+      await deletePayment({ id: invoice.hash, lnd })
+    } catch (error) {
+      console.error(`Error removing invoice with hash ${invoice.hash}:`, error)
+      failedDeletesUpdatePromises.push(models.withdrawl.update({
+        where: { id: invoice.id },
+        data: { hash: invoice.hash, bolt11: invoice.bolt11 }
+      }))
+    }
+  }
```
ekzyis, line 282: `UPDATE ... RETURNING` returns the updated values not the old ones. So this wouldn't work since `invoice.hash` and `invoice.bolt11` is null now:

> In an UPDATE, the data available to RETURNING is the new content of the modified row.

-- https://www.postgresql.org/docs/13/dml-returning.html

I can see how my previous comment was misleading though since I mentioned that "the existing code was fine". I wanted to say that the end result for the withdrawal table should be the same as before.

dillon-co: The `WITH` statement should be making a CTE that returns the values before they were updated

[https://www.postgresql.org/docs/current/queries-with.html](url) But looking at the example at the bottom I might have written it wrong lol

ekzyis: > The `WITH` statement should be making a CTE that returns the values before they were updated

You're right that it creates a CTE and a snapshot of the database and individual CTE's don't see the result of each other but afaict, this doesn't mean that `UPDATE ... RETURNING` inside a CTE does return the original rows. For example, I just tested the behavior on my local machine (using the postgres:13.2 image) with the following queries:

```
stackernews=# SELECT id, hash FROM "Withdrawl" ORDER BY created_at DESC LIMIT 3;
  id   |                               hash
-------+------------------------------------------------------------------
 18326 | 5ed899dc1f0c43c34927e249a4c469a137f4974ce034fa87c9cb38eca7dc6aa6
 18325 | 469b477f103a798d41e7561ad5d0e44f4e1657a78738a1b9546695abe8be7e46
 18324 | 0bcefb7d980e66bad8fd57449a474bf3c813ca28e46dd75c646041dd1ea70110
(3 rows)

stackernews=# WITH updated_rows AS ( UPDATE "Withdrawl" SET hash = NULL WHERE id IN (18326, 18325, 18324) RETURNING id, hash ) SELECT * FROM updated_rows;
  id   | hash
-------+------
 18324 |
 18325 |
 18326 |
(3 rows)
```

> But looking at the example at the bottom I might have written it wrong lol

Which example do you mean?

ekzyis (changes requested):
I pulled the code and ran the function manually by inserting a pgboss job that queues the job for immediate execution by the worker with this query:

```
insert into pgboss.job(name) values ('autoDropBolt11s');
```

For some reason, the invoice update CTE query does not return the modified rows [as I would have expected](https://github.com/stackernews/stacker.news/pull/793/files#r1484725180) but a 0 and then the job fails:

```
running autoDropBolt11s with args null
0
error running autoDropBolt11s TypeError: invoices is not iterable
    at autoDropBolt11s (file:///home/ekzyis/programming/stacker.news/worker/wallet.js:274:25)
    at async file:///home/ekzyis/programming/stacker.news/worker/index.js:67:9
    at resolveWithinSeconds (/home/ekzyis/programming/stacker.news/node_modules/pg-boss/src/manager.js:35:14)
    at <anonymous> (/home/ekzyis/programming/stacker.news/node_modules/p-map/index.js:57:22)
```

worker/wallet.js
dillon-co: the last 2 on that page. I ended up using the wrong query so I updated it and pushed

ekzyis: Ah, I see. Interesting, didn't know about that!

ekzyis (approved):
Looks good now! I had to replace [`$executeRaw`](https://www.prisma.io/docs/orm/prisma-client/queries/raw-database-access/raw-queries#executeraw) with [`$queryRaw`](https://www.prisma.io/docs/orm/prisma-client/queries/raw-database-access/raw-queries#queryraw) though. [It returns the number of affected rows](https://www.prisma.io/docs/orm/prisma-client/queries/raw-database-access/raw-queries#executeraw):


> `$executeRaw` returns the number of rows affected by a database operation, such as UPDATE or DELETE. This function does not return database records.

This also explains [this comment](https://github.com/stackernews/stacker.news/pull/793#pullrequestreview-1873083209) from me.

Tested the code like this:

1. Create some withdrawals for a user using the frontend
2. Enable `autodelete withdrawal invoices` in user settings
3. Backdate withdrawals via database query:

```sql
UPDATE "Withdrawl"
SET created_at = created_at - interval '7 days'
WHERE "userId" = $1;
```
4. Add `console.log` in job:

```diff
diff --git a/worker/wallet.js b/worker/wallet.js
index 1eca5281..1b89194a 100644
--- a/worker/wallet.js
+++ b/worker/wallet.js
@@ -270,6 +270,8 @@ export async function autoDropBolt11s ({ models, lnd }) {
       FROM to_be_updated )
     SELECT * FROM to_be_updated;`)

+  console.log(invoices)
+
   if (invoices.length > 0) {
     const failedDeletesUpdatePromises = []
     for (const invoice of invoices) {
```

5. Run job: `insert into pgboss.job(name) values ('autoDropBolt11s');`

Worker logs:

>running autoDropBolt11s with args null
6
finished autoDropBolt11s
running autoDropBolt11s with args null
0
finished autoDropBolt11s

Since the invoices are updated in the query but the loop failed to run, the invoices were gone and subsequent runs returned 0 for 0 affected rows. Fixed this in 022ed8b2.

Were you able to test the code using job insertion (step 5)?

Tested deletion undo code by manually deleting the payment such that `await deletePayment` will throw an error. Connected to the LND node using polar and ran this command:

```
lnd@alice:/$ lncli deletepayments --payment_hash a00d27b68352a60290899e6ac0eeaf068cb5d0465d213b6e38467e72adea423e
{
        "ok": true
}
```

Then ran step 5. Got this expected error:

```
Error removing invoice with hash a00d27b68352a60290899e6ac0eeaf068cb5d0465d213b6e38467e72adea423e: [
  503,
  'UnexpectedErrorDeletingPayment',
  {
    err: Error: 2 UNKNOWN: non bucket element in payments bucket
        at callErrorFromStatus (/home/ekzyis/programming/stacker.news/node_modules/@grpc/grpc-js/build/src/call.js:31:19)
        at Object.onReceiveStatus (/home/ekzyis/programming/stacker.news/node_modules/@grpc/grpc-js/build/src/client.js:192:56)
        at Object.onReceiveStatus (/home/ekzyis/programming/stacker.news/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:360:141)
        at Object.onReceiveStatus (/home/ekzyis/programming/stacker.news/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:323:181)
        at <anonymous> (/home/ekzyis/programming/stacker.news/node_modules/@grpc/grpc-js/build/src/resolving-call.js:99:78)
        at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
    for call at
        at ServiceClientImpl.makeUnaryRequest (/home/ekzyis/programming/stacker.news/node_modules/@grpc/grpc-js/build/src/client.js:160:32)
        at ServiceClientImpl.<anonymous> (/home/ekzyis/programming/stacker.news/node_modules/@grpc/grpc-js/build/src/make-client.js:105:19)
        at asyncAuto.deletePayment (/home/ekzyis/programming/stacker.news/node_modules/lightning/lnd_methods/offchain/delete_payment.js:43:32)
        at runTask (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:293:13)
        at <anonymous> (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:233:31)
        at processQueue (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:243:13)
        at taskComplete (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:259:9)
        at <anonymous> (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:286:17)
        at <anonymous> (/home/ekzyis/programming/stacker.news/node_modules/async/internal/onlyOnce.js:12:16)
        at cbk (/home/ekzyis/programming/stacker.news/node_modules/lightning/lnd_methods/offchain/delete_payment.js:38:16)
        at runTask (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:295:13)
        at <anonymous> (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:233:31)
        at processQueue (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:243:13)
        at auto (/home/ekzyis/programming/stacker.news/node_modules/async/auto.js:230:5)
        at <anonymous> (/home/ekzyis/programming/stacker.news/node_modules/lightning/lnd_methods/offchain/delete_payment.js:27:12)
        at new Promise (<anonymous>)
        at module.exports (/home/ekzyis/programming/stacker.news/node_modules/lightning/lnd_methods/offchain/delete_payment.js:26:10)
        at autoDropBolt11s (file:///home/ekzyis/programming/stacker.news/worker/wallet.js:279:15)
        at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
        at async file:///home/ekzyis/programming/stacker.news/worker/index.js:67:9
        at resolveWithinSeconds (/home/ekzyis/programming/stacker.news/node_modules/pg-boss/src/manager.js:35:14)
        at <anonymous> (/home/ekzyis/programming/stacker.news/node_modules/p-map/index.js:57:22) {
      code: 2,
      details: 'non bucket element in payments bucket',
      metadata: [Metadata]
    }
  }
]
```

Then tested that a withdrawal with this hash still exists in the database. That was the case :white_check_mark:

dillon-co:
> Were you able to test the code using job insertion (step 5)?

I thought I tested it and the code in the app is working but I'm having some trouble with LNBits. I'm getting this error 
` Error connecting to LndWallet: 'grpc_status'`
`Fallback to VoidWallet, because the backend for LndWallet isn't working properly` 

So it's not connecting to my local LND. I'm pretty sure it's something with my env but it's hard to tell. The logs aren't giving me much to go off of

huumn:
You don't need lnbits to test this PR fyi. LNBits is for sending *to* SN. This PR affects sending *from* SN. 

The point of deleting the Bolt11's is to remove data encoding information about a stacker's lightning node.

.gitignore
ekzyis, line 1: Please remove these changes, they are unrelated to the PR.

docker-compose.yml is also already checked into the repository, so I don't think this has an effect.

ekzyis (changes requested):
As @huumn mentioned, you don't need LNbits for this PR.

This is for withdrawals created for the  custodial wallet at /wallet?type=withdraw:

2024-02-12-232110_1918x1059_scrot

We only need the changes at worker/wallet.js for this PR, so please remove the other changes.

That we use .env.sample as the environment file read by docker-compose.yml might be what confused you, but we need two different env files for running on the host machine (`npm run dev`) and for docker-compose.

Might make sense to rename this file to .env.docker or something but that's not related to this PR.

dillon-co:
Also tested. It's working as expected. 

I was trying to get LNBits up and running to fund the wallet in order to withdrawl lol. Now I've got everything set up so I should be good to work on more stuff without blocks

ekzyis (approved):

huumn (changes requested):
This looks great! All the node/money stuff is pretty gnarly, but the resulting change is super simple. Nice work!

The other reviewer 😉  missed on thing though. We have [a mutation for manually dropping bolt11s](https://github.com/stackernews/stacker.news/blob/1444ff476e61e3b9a4ddb7f45da866a5a87bea76/api/resolvers/wallet.js#L374C1-L390C7) so you'll want to `deleteInvoice` in there too.

huumn:
I ended up having make quite a changes to get this to work: update to Withdrawl did not have a condition on the FROM so it just didn't work. I'm not sure exactly what happened when you and @ekzyis were testing it but for me it failed due to constraint violations. The only thing I can think of is neither of you had an invoice to delete.

I made the changes myself because I figured you'd prefer to move on, but you can review the commit to see what changed.

Also changed (not bugs but less than ideal):

1. we don't want to use `serialize` unless they update stacker's account balances, because serializable transactions are "expensive" generally, ie overkill in any circumstance they aren't needed
2. it's worth not transactionally reversing failed deletes because
    - if a reversal fails we don't want to reverse the successful reversals and not reverse anything
    - if the process crashes, any queued reversals would never be reversed
3. Not from your changes but failed invoice deletion would report both success and failure to the client

Anyway, sorry for the rough time this PR gave you. I'm looking forward to your next one! I'm sure it will be easier now that you know your way around some.

PR #797, opened 2024-02-08
Create Rehype Plugin to parse mentions

### LINKED ISSUE #663: Mention doesn't work on markdown conflicts
https://stacker.news/items/340347

alexlwn123:
Fixes #663 

Took a few tries to figure out, but this strategy seems to be working.

Adds 2 step process for handling mention markup (ex. @username) that prevents unwanted markdown parsing.

1. Replace mentions with placeholder that won't get parsed into markdown. (i.e. an HTML comment)
2. After the text is parsed, rehype the placeholders with mention links.

Before: 
image

After:
image

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -4,7 +4,7 @@ import YouTube from 'react-youtube'
 import gfm from 'remark-gfm'
 import { LightAsync as SyntaxHighlighter } from 'react-syntax-highlighter'
 import atomDark from 'react-syntax-highlighter/dist/cjs/styles/prism/atom-dark'
-import mention from '../lib/remark-mention'
+import { preprocessMentions, rehypeMentions } from '../lib/remark-mention'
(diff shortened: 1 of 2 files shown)


huumn:
I'm sure there's no other way if this is where you ended up, but I don't love that's this requires hackiness in a significant part of the code to handle a corner case. Unless we can come up with a better way to do this, it might be worth leaving this bug in (it's not a huge one) and possibly disallow future nyms from triggering it. 

Frankly, seeing how tricky this is to solve makes me impressed you got anything to work at all.

I'll think on it more and leave this open in the meantime, but consider this done and feel free to move on.

alexlwn123:
@huumn Yeah, I agree this is super hacky and understand wanting to leave this open. Happy to move on from this nightmare. 😂

Issue #712, opened 2023-12-26
Multiselect dropdown replacement for territory crossposting


huumn:
For the UI, We'll probably want something like this: https://react-select.com/home

On the backend:
- we'll need a many-to-many relationship between `Item` and `Sub`
- to avoid additional joins every time we load an `Item`, we'll want to denormalize all the territories `Item` belongs to in trigger that populates an array on `Item` called `subNames`
- `create_item` and `update_item` stored procedures will need to be updated

Economics:
- Posting in multiple territories is cumulative
    - e.g. if posting in `~bitcoin` and `~nostr` and each costs 10 sats to post, the resulting fee is 10+10=20 sats.

m0wer:
Is this still relevant? Does it conflict with the ongoing territory improvements?

huumn:
https://github.com/stackernews/stacker.news/issues/795 should come first because UI/UX is the trickiest part, and we'll want to reuse the component there.

Also, this issue is much more complicated that it sounds because it affects territory revenue.

PR #800, opened 2024-02-09
update for crossposting to territories

byoungdale:
This is the pull request for crossposting to territories. I am starting with the criteria from @huumn in #712 . This is just the changes for the schema so far. I left the old sub field, and added a new `subs` field. I figured this would be better for migration. This migration will update the new table and ‘subs’ field in Item. Then once the rest of the code is pointed toward using the subs and subNames, sub will be unused and ready for cleanup whenever it’s safely in production. That sound like a good path? I’ll update this branch with the rest of the changes as I go.

Here is a quick glance of what I have so far:

### update prisma schema

on Item
```diff
+  subs               Sub[]                 @relation("ItemSub")
+  subNames           String[]
```
on Sub
```diff
+  Items        Item[]         @relation("ItemSub")
```

### updated migration file with trigger on new _ItemSub join table to update Item table
```sql
-- create new update_subnames trigger
CREATE OR REPLACE FUNCTION public.update_subnames()
    RETURNS TRIGGER
    AS $$
BEGIN
    UPDATE "Item"
    SET "subNames" = (
        SELECT ARRAY_AGG("subName")
        FROM "_ItemSub"
        WHERE "itemId" = NEW."id"
    )
    WHERE "id" = NEW."id";

    RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER update_subnames_trigger
AFTER INSERT OR UPDATE ON "_ItemSub"
FOR EACH ROW
EXECUTE FUNCTION update_subnames();
```

### updated create_item and update_item to update _ItemSub table on create and update of an item

create_item
```diff
+    -- Loop over the 'subs' array and insert into "_ItemSub"
+    FOREACH sub_name IN ARRAY subs
+    LOOP
+        INSERT INTO "_ItemSub" ("itemId", "subName")
+        VALUES (item.id, sub_name);
+    END LOOP;
```

update_item
```diff
+    -- Loop over the 'subs' array and insert into "_ItemSub" if it does not already exist
+    FOREACH sub_name IN ARRAY subs
+    LOOP
+        INSERT INTO "_ItemSub" ("itemId", "subName")
+        SELECT item.id, sub_name
+        WHERE NOT EXISTS (
+            SELECT 1 FROM "_ItemSub"
+            WHERE "itemId" = item.id AND "subName" = sub_name
+        );
+    END LOOP;
```

### Update new database fields and tables with existing data

```sql
-- Update new database fields and tables with existing data
-- Insert into ItemSub table
INSERT INTO "ItemSub" ("itemId", "subName")
SELECT "id", "subName"
FROM "Item"
WHERE "subName" IS NOT NULL
AND NOT EXISTS (
    SELECT 1 FROM "ItemSub" WHERE "itemId" = "Item"."id" AND "subName" = "Item"."subName"
);

-- Update subs array in Item table
UPDATE "Item" AS A
SET "subNames" = B."subNames"
FROM (
    SELECT "itemId", ARRAY_AGG("subName") AS "subNames"
    FROM "ItemSub"
    GROUP BY "itemId"
) AS B
WHERE A."id" = B."itemId";
```


huumn:
Cool! I'm going to put this in draft until it's the whole hog.

One thing I've noticed off the jump: if you're going to add a table, we'd prefer it be defined in prisma.schema if it can be. Also, unless there's a convention I'm unaware of "_ItemSub" should probably just be "ItemSub."

byoungdale:
Got it. Sounds good. The underscore style (_Name) is what prisma creates automatically when a many-to-many relationship is defined in the schema. It's an implicit pass-through table creation, which I don't personally like. I'll update that.

PR #805, opened 2024-02-09
Add Content Security Policy headers

### LINKED ISSUE #770: Use Content Security Policy headers
In https://github.com/stackernews/stacker.news/pull/749#discussion_r1457813585, the planned usage of CSP was mentioned:

> Regardless of these changes even, we should introduce really strict [Content Security Policies](https://nextjs.org/docs/pages/building-your-application/configuring/content-security-policy) to narrow the XSS attack surface. For browsers that don't support CSP we should consider preventing storing/using secrets clientside.

Created this separate ticket since it's not necessarily related to #749 since deploying CSP also makes sense on its own.

Also, this ticket can act as a dedicated place for discussion around CSP.

For example, I have read about [`Content-Security-Policy-Report-Only`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#testing_your_policy). This allows us to rollout CSP and monitor the app (via console or via actual reports sent to an endpoint) before we enforce the rules:

> [Testing your policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#testing_your_policy)
To ease deployment, CSP can be deployed in report-only mode. The policy is not enforced, but any violations are reported to a provided URI. Additionally, a report-only header can be used to test a future revision to a policy without actually deploying it.

---

References:

- https://nextjs.org/docs/pages/building-your-application/configuring/content-security-policy
- https://infosec.mozilla.org/guidelines/web_security#content-security-policy
- https://web.dev/articles/csp
- https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html
- https://github.com/vercel/next.js/discussions/54907
- https://reesmorris.co.uk/blog/implementing-proper-csp-nextjs-styled-components
- https://webpack.js.org/guides/csp/
- https://csp-evaluator.withgoogle.com/

ekzyis:
close #770 

Some notes regarding the approach of using a middleware:

We could use meta tags in `<Head>` in app.js but [then we can't use `report-to`](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#3-content-security-policy-meta-tag).

We could add static CSP headers in next.config.js but then we can't enforce a strict CSP with dynamic nonces. We could use hashes but hashes are fragile.

So afaict, middleware is the only viable solution. Even though middlewares feel like third-class citizens in NextJS :/

TODO:

- [x] use nonces to fix usage of `unsafe-inline` for `script-src` and `style-src` ([strict CSP](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#strict-csp))

We could also uses hashes but this can break easily:

> Using hashes can be a risk approach approach. If you change anything inside the script tag (even whitespace) by, e.g., formatting your code, the hash will be different, and the script won't render.

-- https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#note

So I'll try to use nonces first.

- [x] ~use [`report-to`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to) to report errors to Slack channel before enforcing the policy~
- [x] click around and test if everything still works - ~if not, no problem, we can rely on `report-to` before enforcing policy~
- [x] fix or workaround `TypeError: Cannot read properties of undefined (reading 'bind')` which might be related to https://github.com/vercel/next.js/issues/56368

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -66,7 +66,7 @@ export const config = {
     // NextJS recommends to not add the CSP header to prefetches and static assets
     // See https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy
     {
-      source: '/((?!api|_next/static|_next/image|favicon.ico).*)',
+      source: '/((?!api|_next/static|_next/image|_next/webpack-hmr|favicon.ico).*)',
       missing: [
         { type: 'header', key: 'next-router-prefetch' },
         { type: 'header', key: 'purpose', value: 'prefetch' }
--- a/middleware.js
+++ b/middleware.js
@@ -23,16 +23,16 @@ export function middleware (request) {
   const cspHeader = [
     // if something is not explicitly allowed, we don't allow it.
     "default-src 'none'",
-    "font-src 'self'",
+    "font-src 'self' a.stacker.news",
     // we want to load images from everywhere but we can limit to HTTPS at least
-    "img-src 'self' https: data:",
+    "img-src 'self' a.stacker.news m.stacker.news https: data:",
     // Using nonces and strict-dynamic deploys a strict CSP.
     // see https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#strict-policy.
     // Old browsers will ignore nonce and strict-dynamic
     // and fallback to host matching, unsafe-inline and unsafe-eval (no protection against XSS)
     `script-src 'self' 'unsafe-inline' 'unsafe-eval' 'nonce-${nonce}' 'strict-dynamic' https:`,
     // unsafe-inline for styles is not ideal but okay if script-src is using nonces
-    "style-src 'self' 'unsafe-inline'",
+    "style-src 'self' a.stacker.news 'unsafe-inline'",
     "manifest-src 'self'",
     'frame-src www.youtube.com platform.twitter.com',
     "connect-src 'self' https: wss:",
--- a/middleware.js
+++ b/middleware.js
@@ -26,16 +26,20 @@ export function middleware (request) {
     "font-src 'self'",
     // we want to load images from everywhere but we can limit to HTTPS at least
     "img-src 'self' https: data:",
-    // old browsers will ignore nonce and strict-dynamic
+    // Using nonces and strict-dynamic deploys a strict CSP.
+    // see https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#strict-policy.
+    // Old browsers will ignore nonce and strict-dynamic
     // and fallback to host matching, unsafe-inline and unsafe-eval (no protection against XSS)
     `script-src 'self' 'unsafe-inline' 'unsafe-eval' 'nonce-${nonce}' 'strict-dynamic' https:`,
     // unsafe-inline for styles is not ideal but okay if script-src is using nonces
     "style-src 'self' 'unsafe-inline'",
     "manifest-src 'self'",
     "connect-src 'self' https: wss:",
-    // object-src is for plugins
+    // disable dangerous plugins like Flash
     "object-src 'none'",
+    // blocks injection of <base> tags
     "base-uri 'none'",
+    // tell user agents to replace HTTP with HTTPS
     'upgrade-insecure-requests'
   ].join('; ')
 
--- a/middleware.js
+++ b/middleware.js
@@ -1,8 +1,8 @@
 import { NextResponse } from 'next/server'
 
-export function middleware (request) {
-  const regex = /(\/.*)?\/r\/([\w_]+)/
-  const m = regex.exec(request.nextUrl.pathname)
+const referrerRegex = /(\/.*)?\/r\/([\w_]+)/
+function referrerMiddleware (request) {
+  const m = referrerRegex.exec(request.nextUrl.pathname)
 
   const url = new URL(m[1] || '/', request.url)
   url.search = request.nextUrl.search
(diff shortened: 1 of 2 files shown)


huumn:
I don't know if it helps at all, but this is what I did in Outer Space. I can't remember how strict this was or if it was just better than nothing.

```js
const ContentSecurityPolicy = process.env.NODE_ENV === 'development'
  ? `
      default-src 'self';
      connect-src 'self' webpack://* wss://*;
      script-src 'self' 'unsafe-eval';
      worker-src 'self' blob:;
      script-src-elem 'self';
      style-src 'self' 'unsafe-inline';`
  : `
      default-src 'self';
      script-src 'self' 'sha256-/6SBPqW+GW+//4nlXX6Y1nR9dWlh0gsQJ6KK71djH6A=';
      worker-src 'self' blob:;
      connect-src 'self' wss://*;
      style-src 'self' 'unsafe-inline';`

const securityHeaders = [
  {
    key: 'X-DNS-Prefetch-Control',
    value: 'on'
  },
  {
    key: 'X-XSS-Protection',
    value: '1; mode=block'
  },
  {
    key: 'X-Frame-Options',
    value: 'SAMEORIGIN'
  },
  {
    key: 'X-Content-Type-Options',
    value: 'nosniff'
  },
  {
    key: 'Referrer-Policy',
    value: 'origin-when-cross-origin'
  },
  {
    key: 'Content-Security-Policy',
    value: ContentSecurityPolicy.replace(/\s{2,}/g, ' ').trim()
  }
]

```

huumn:
We could also put stuff into nginx if that makes more sense. Just another option!

ekzyis:
> I don't know if it helps at all, but this is what I did in Outer Space. I can't remember how strict this was or if it was just better than nothing.

I think since you used hashes, it is a strict CSP (`self` is ignored in that case)

> We could also put stuff into nginx if that makes more sense. Just another option!

I think I got it working with nonces now in 93d33f3. Will test more tomorrow

ekzyis:
One shouldn't use `X-XSS-Protection` with CSP though [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection):

> Warning: Even though this feature can protect users of older web browsers that don't yet support [CSP](https://developer.mozilla.org/en-US/docs/Glossary/CSP), in some cases, XSS protection can create XSS vulnerabilities in otherwise safe websites. See the section below for more information.

Did set other nice headers like `Referrer-Policy` and `X-Content-Type-Options` in 94ea9de.

Did not add `X-DNS-Prefetch-Control` since not recommended for production sites by [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-DNS-Prefetch-Control).

ekzyis:
CSP looks good according to https://csp-evaluator.withgoogle.com/:

2024-02-10-022629_1920x1080_scrot

Another scan from https://observatory.mozilla.org/:

2024-02-10-023116_1920x1080_scrot

2024-02-10-023214_1920x1080_scrot

Would be nice if we can disable `unsafe-eval` but probably buried in some library.

Going to add the HSTS header for extra points (even though we shouldn't care too much about these points haha)

huumn:
> Would be nice if we can disable unsafe-eval but probably buried in some library.

If you want to find it, let's remove it.

ekzyis:
> If you want to find it, let's remove it.

Good call, `unsafe-eval` is only required in development mode.

Also allowed CDN, media domain where required and fixed blocking of YouTube and twitter embeds via `frame-src`.

ekzyis:
I noticed yesterday that Brave disabled the Reporting API because of its privacy implications: https://github.com/brave/brave-browser/issues/7956

And wasn't able to get it to work on Chrome. Maybe related to the browser deciding when to send reports. Setup should work like this: For reporting, you need to use the CSP directives [`report-uri`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri) (for backwards compatibility) and [`report-to`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to) with the HTTP header `Report-To`. To not enforce the policy, you replace the HTTP header `Content-Security-Policy` with [`Content-Security-Policy-Report-Only`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only).

Will look more into this today. However, I also noticed than one needs to explicitly allow browser extensions (`chrome-extension:` scheme). So this is also related to privacy implications since stackers probably don't like that their browsers sends reports like which include information about which extensions they are using (even though they wouldn't be linked to their account on our side).

Found this nice website related to problems with weird CSP reports: https://csper.io/blog/csp-report-filtering

It linked to this summary of "WTF" reports: https://github.com/nico3333fr/CSP-useful/blob/master/csp-wtf/explained.md

I would like to have CSP reports but if it's too much work, I think we can ship it enforced and rely on user reports etc. and I can look into this at some other point in the future (when it feels more important).

I think the current CSP policy is not interfering with most functionality on the site.

ekzyis:
Regarding `TypeError: Cannot read properties of undefined (reading 'bind')`:

This error only happened if the dev build in the browser was trying to connect to wss://sn.ekzyis.com/_next/webpack-hmr with the server built in prod mode running.

If I refreshed the page the errors stopped showing up since the prod build does not use hot reload.

Ignoring that route in c134c561 now.

PR #810, opened 2024-02-12
Validate pubkey, relay URL and secret of NWC URL

ekzyis:
Mhh, wasn't able to use `Yup.transform` so we can keep using `validationSchema`. I tried stuff along these lines:

```js
export const nwcSchema = object({
  nwcUrl: string().required('required').trim().matches(/^nostr\+walletconnect:\/\//),
  relayUrl: string().required('relay url required').trim().wss('relay must use wss://'),
  walletPubkey: string().required('pubkey required').trim().matches(NOSTR_PUBKEY_HEX, 'pubkey must be 64 hex chars'),
  secret: string().required('secret required').trim().matches(/^[0-9a-z]{64}$/, 'secret must be 64 hex chars')
}).transform(url => ({ url, ...parseNwcUrl(url) }))
```

but in any case, it threw errors for fields that don't exist (`relayUrl`, `walletPubkey`, `secret`) which makes sense. So not sure if we can use `transform` to transform a string into an object but still pretend it's the same string we are validating?

https://github.com/stackernews/stacker.news/assets/27162016/543ee34a-6faf-4029-939f-43932d314944

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -749,7 +749,7 @@ export function CheckboxGroup ({ label, groupClassName, children, ...props }) {
 const StorageKeyPrefixContext = createContext()
 
 export function Form ({
-  initial, validate, schema, onSubmit, children, initialError, validateImmediately,
+  initial, schema, onSubmit, children, initialError, validateImmediately,
   storageKeyPrefix, validateOnChange = true, invoiceable, innerRef, ...props
 }) {
   const toaster = useToast()
@@ -812,7 +812,6 @@ export function Form ({
     <Formik
       initialValues={initial}
       validateOnChange={validateOnChange}
-      validate={validate}
       validationSchema={schema}
       initialTouched={validateImmediately && initial}
       validateOnBlur={false}
--- a/components/form.js
+++ b/components/form.js
@@ -749,7 +749,7 @@ export function CheckboxGroup ({ label, groupClassName, children, ...props }) {
 const StorageKeyPrefixContext = createContext()
 
 export function Form ({
-  initial, schema, onSubmit, children, initialError, validateImmediately,
+  initial, validate, schema, onSubmit, children, initialError, validateImmediately,
   storageKeyPrefix, validateOnChange = true, invoiceable, innerRef, ...props
 }) {
   const toaster = useToast()
@@ -812,6 +812,7 @@ export function Form ({
     <Formik
       initialValues={initial}
       validateOnChange={validateOnChange}
+      validate={validate}
       validationSchema={schema}
       initialTouched={validateImmediately && initial}
       validateOnBlur={false}
--- a/components/webln/nwc.js
+++ b/components/webln/nwc.js
@@ -2,6 +2,7 @@
 
 import { createContext, useCallback, useContext, useEffect, useState } from 'react'
 import { Relay, finalizeEvent, nip04 } from 'nostr-tools'
+import { parseNwcUrl } from '../../lib/url'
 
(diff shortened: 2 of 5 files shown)


huumn:
I'd guess you'll want something like this but I haven't used their transform yet so not sure:

```js
export const nwcSchema = object({
  nwc: string().required('required').transform(url => ({ url, ...parseNwcUrl(url) })).object({
  nwcUrl: string().required('required').trim().matches(/^nostr\+walletconnect:\/\//),
  relayUrl: string().required('relay url required').trim().wss('relay must use wss://'),
  walletPubkey: string().required('pubkey required').trim().matches(NOSTR_PUBKEY_HEX, 'pubkey must be 64 hex chars'),
  secret: string().required('secret required').trim().matches(/^[0-9a-z]{64}$/, 'secret must be 64 hex chars')
  })
})
```

ekzyis:
Was able to use transform but still had to cast validation errors into the shape formik expects. So couldn't use `Formik.validationSchema` which works well with `Yup` and is what we use everywhere for validation.

However, even then, I still had problems with the order of validation (see https://github.com/jquense/yup/issues/851), so I decided to not use `transform` at all and instead do this:

```js
export const nwcSchema = object({
  nwcUrl: string()
    .required('required')
    .test(async (nwcUrl, context) => {
      // run validation in sequence to control order of errors
      // inspired by https://github.com/jquense/yup/issues/851#issuecomment-1049705180
      try {
        await string().required('required').validate(nwcUrl)
        await string().matches(/^nostr\+?walletconnect:\/\//, { message: 'must start with nostr+walletconnect://' }).validate(nwcUrl)
        let relayUrl, walletPubkey, secret
        try {
          ({ relayUrl, walletPubkey, secret } = parseNwcUrl(nwcUrl))
        } catch {
          // invalid URL error. handle as if pubkey validation failed to not confuse user.
          throw new Error('pubkey must be 64 hex chars')
        }
        await string().required('pubkey required').trim().matches(NOSTR_PUBKEY_HEX, 'pubkey must be 64 hex chars').validate(walletPubkey)
        await string().required('relay url required').trim().wss('relay must use wss://').validate(relayUrl)
        await string().required('secret required').trim().matches(/^[0-9a-fA-F]{64}$/, 'secret must be 64 hex chars').validate(secret)
      } catch (err) {
        return context.createError({ message: err.message })
      }
      return true
    })
})
```

I think this is also clearer now and we can stay consistent and also use `validationSchema` here.

huumn:
Thread explaining why transforms don't work: [How does yup's trim work?](https://github.com/jaredpalmer/formik/issues/473)

Formik doesn't use the returned object from yup which contains the transforms. According to that thread, we'd just need to add this line to `onSubmit`:

```js
onSubmit={(values) => {
  const castValues = mySchema.cast(values)
  ...
})
```

ekzyis:
> Thread explaining why transforms don't work: [How does yup's trim work?](https://github.com/jaredpalmer/formik/issues/473)
> 
> Formik doesn't use the returned object from yup which contains the transforms. According to that thread, we'd just need to add this line to `onSubmit`:
> 
> ```js
> onSubmit={(values) => {
>   const castValues = mySchema.cast(values)
>   ...
> })
> ```

But we would still need to "cast back" manually to properly show error messages for the input field named `nwcUrl`. This casting back to properly recognize errors was mostly the issue I was having iirc. And transforms only one-way.

huumn:
Ah I see. Still, might be useful for transforms like `trim()`

Issue #419, opened 2023-08-17
Offline page is not styled


huumn:
Pleb Lab's internet has been super wonky and we lose connectivity and regain it every minute or so. It's letting me experience our offline behavior.

![Screenshot 2023-08-17 at 12 34 20 PM](https://github.com/stackernews/stacker.news/assets/34140557/a2e29085-4b6c-4091-94f9-7a11618d24a9)

I'd guess that despite nextjs statically rendering the offline page and us precaching it in the service worker, the offline page references other assets that aren't precached. 

The solve would be to precache the contents of `next/static` or whatever the actual path of the static build output is.

ekzyis:
Hopefully fixing all these failed network requests when going into offline mode will fix this:

https://github.com/stackernews/stacker.news/assets/27162016/0e7d7011-1a42-46e8-9227-b524eea7e9cf

huumn:
This is related to #612

ekzyis:
I looked again into this because it is somewhat related to #770 (webpack also has a [section about CSP](https://webpack.js.org/guides/csp/)).

I have noticed following things:

1. Missing CDN CORS headers might prevent some assets to get cached[^1]
2. We're loading different assets when we're offline vs online (stale offline page cache)?

Maybe they are related.

What I did is the following:

I noticed that my browser was fetching different assets for the [offline page](https://stacker.news/offline) when I was online vs offline. The assets that are fetched when I was offline didn't exist (different hashes). For example, currently, we use this CSS file: https://a.stacker.news/_next/static/css/c53707eabe8d902f.css. But in the service worker cache, there was a different URL with a different hash. This looks as if the `workbox-offline-fallbacks` cache does not refresh the cache for the offline page on deployments. Maybe it can't detect if it needs to refresh that page. However, we do include CORS headers for everything that matches `_next/:asset` so CORS can't be the reason for this.

I then deleted the cache and forced a new service worker registration which refreshed the cache. Then I got this page when I am offline:

![2024-02-06-111951_1918x1059_scrot](https://github.com/stackernews/stacker.news/assets/27162016/0d7fdec4-a6ac-49ec-82b6-4a82780b51c1)

See how it tries to fetch the asset from `a.stacker.news/waiting.gif` and not from `stacker.news/waiting.gif` which is what got cached:

[...]

PR #814, opened 2024-02-13
Fix local prod builds assuming CDN

ekzyis:
Prod builds assumed that we're running in an AWS environment and use a CDN. This commit changes that.

Now, if the AWS way to fetch the commit failed _and_ the normal git command fails and only if, we assume we're running the prod build locally and don't configure the CDN.


ekzyis:
However, I am thinking about adding one more subdomain to sn.ekzyis.com so I can be even closer to prod environment while developing. If assets are fetched from a subdomain or not is relevant for caching[^1] and CSP. 

[^1]: see https://github.com/stackernews/stacker.news/issues/419#issuecomment-1929289197)

PR #826, opened 2024-02-14
Add Opt-in to Display Linked Accounts in Profile

### LINKED ISSUE #759: Opt-in display of twitter, github, nostr auth
This would be nice for helping people authenticate an identity.

context: https://stacker.news/items/385220/r/k00b

alexlwn123:
Closes #759 

* Adds opt-in "Connected Accounts" to profile screen 
* Adds settings to "hide my linked (twitter|nostr|github) profile"
  * defaults to true
* Stores connected twitter & Github ids in the User table
   * Removes the handle if unlinked

image

image


huumn:
Nice changes overall!

To be used for existing stacker it looks like one needs to: 
1. set the setting
2. link the account or signin

And in that order, right?

alexlwn123:
> To be used for existing stacker it looks like one needs to:
> 
> 1. set the setting
> 2. link the account or signin
> 
> And in that order, right?

Ah, yeah... for existing stackers they would need to link the account/sign-in again. However, the setting can be toggled before or after the account is linked.

huumn:
It makes sense. Just wanted to understand. 

I think we'll need a way to communicate this to existing stackers. I've learned to consider the customer service cost of these things lol. Perhaps if the id is not present:

1. disable the toggle
2. display an `i` info icon next to the toggle that explains why it's disabled and how to enable it.

alexlwn123:
Absolutely, your suggestion makes a lot of sense and addresses important user experience concerns. 

I'll add the change disable the toggle and add an info for when the ID is missing. Thanks for highlighting this!

huumn (changes requested):
Marking as changes requested is all

components/user-header.js
```diff
@@ -180,8 +184,35 @@ function HeaderNym ({ user, isMe }) {
     : <NymView user={user} isMe={isMe} setEditting={setEditting} />
 }
 
+function SocialLink ({ name, id }) {
+  if (name === 'Nostr') {
+    const npub = hexToBech32(id)
+    return (
+      <Link className={styles.social} target='_blank' href={`https://snort.social/${npub}`} rel='noreferrer'>
+        <NostrIcon width={20} height={20} className='me-1' />
+        {npub.slice(0, 20)}...
```
ekzyis, line 193: I think it's more common to show both ends for such strings. So I would suggest to use first ten characters and last ten characters.

```suggestion
        {npub.slice(0, 10)}...{npub.slice(-10)}
```

components/user-header.js
```diff
@@ -180,8 +184,35 @@ function HeaderNym ({ user, isMe }) {
     : <NymView user={user} isMe={isMe} setEditting={setEditting} />
 }
 
+function SocialLink ({ name, id }) {
+  if (name === 'Nostr') {
+    const npub = hexToBech32(id)
+    return (
+      <Link className={styles.social} target='_blank' href={`https://snort.social/${npub}`} rel='noreferrer'>
```
ekzyis, line 191: For nostr notes, we use https://nostr.com/ already since it asks the user which client they want to use.

So I think we should also use it here.
```suggestion
      <Link className={styles.social} target='_blank' href={`https://nostr.com/${npub}`} rel='noreferrer'>
```

lib/nostr.js
```diff
@@ -16,6 +16,11 @@ export function hexToBech32 (hex, prefix = 'npub') {
   return bech32.encode(prefix, bech32.toWords(Buffer.from(hex, 'hex')))
 }
 
+export const truncateNpub = (string = '', maxLength = 140) =>
+  string.length > maxLength
+    ? `${string.substring(0, maxLength)} […]`
+    : string
+
```
ekzyis, line 23: Doesn't seem to be used

_didn't pull the code, only used CTRL+F in the browser_

ekzyis (commented):
Nice! This basically works as verification, too. So companies like ACINQ that have asked for verification in the past and have official twitter, nostr or github accounts can use this then :)

lib/nostr.js
alexlwn123, line 23:
```suggestion
```

alexlwn123:
Thanks for the review! All great suggestions. 

Implemented feedback.
* Added info bubble on settings
* Disables setting if account id is not set

image

huumn (approved):
Great job! Code is clear and everything.

I'm going to pull to QA but I suspect this is good for merge.

PR #830, opened 2024-02-14
Fix settings save requires relay set

ekzyis:


ekzyis:
This change was so trivial, I decided this can be my first merge

ekzyis:
Oh, I think you usually use "squash and merge" since I see no merge commits on the master branch

huumn:
My fix was a little more to the point. 

For future reference: these yup `test`s imply `required()` unless we allow `value === undefined` to return true

PR #834, opened 2024-02-15
Fix QR code interaction with WebLN provider

### LINKED ISSUE #819: Race condition between WebLN provider init and sendPayment
**Description**

If you're on a page that would trigger `sendPayment`, it might get called before the config was loaded from local storage.

For example, for LNbits, it would make a request to https://stacker.news/api/payments (404 Not Found) instead of \<lnbits url\>/api/payments since `url` is still a empty string.

**Steps to Reproduce**

Refresh page on /invoices/:id often enough

**Expected behavior**

Waits until provider is initialized

### LINKED ISSUE #820: Wrong value passed to sendPayment on QR payments
**Description**

In _invoice.js_, we pass `invoice.bolt11` to the `<Qr>` component which then calls `sendPayment` with that value. However, in `sendPaymentWithToast` we expect an object with the keys bolt11, hash, hmac. So bolt11 is undefined since a string was passed.

https://github.com/stackernews/stacker.news/blob/2bbcbfbb268c658c6b6297b2267f91c995370400/components/invoice.js#L48-L54

https://github.com/stackernews/stacker.news/blob/2bbcbfbb268c658c6b6297b2267f91c995370400/components/qr.js#L14-L26

https://github.com/stackernews/stacker.news/blob/2bbcbfbb268c658c6b6297b2267f91c995370400/components/webln/index.js#L84-L111

**Steps to Reproduce**

Visit /invoices/:id with WebLN provider enabled (LNbits or NWC attached)

**Expected behavior**

bolt11 is passed and payment is attempted

**Additional context**

Fix should be to just pass object in shape `{ bolt11 }` since hash and hmac should only be required for HODL invoices. HODL invoices are only used for anon zaps and there we specifically disable WebLN since we only show a QR code if the WebLN provider failed before.

ekzyis:
fix #819 fix #820

_Wow, it's suddenly 5am here. Will take another look at this when I slept lol_

### DIFF
--- a/components/qr.js
+++ b/components/qr.js
@@ -20,7 +20,7 @@ export default function Qr ({ asIs, value, webLn, statusVariant, description, st
       }
     }
     effect()
-  }, [])
+  }, [provider])
 
   return (
     <>
--- a/components/qr.js
+++ b/components/qr.js
@@ -3,13 +3,11 @@ import { CopyInput, InputSkeleton } from './form'
 import InvoiceStatus from './invoice-status'
 import { useEffect } from 'react'
 import { useWebLN } from './webln'
-import { useToast } from './toast'
 
 export default function Qr ({ asIs, value, webLn, statusVariant, description, status }) {
   const qrValue = asIs ? value : 'lightning:' + value.toUpperCase()
 
   const provider = useWebLN()
-  const toaster = useToast()
 
   useEffect(() => {
     async function effect () {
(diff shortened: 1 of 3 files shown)


components/qr.js
```diff
@@ -3,27 +3,24 @@ import { CopyInput, InputSkeleton } from './form'
...
+          await provider.sendPayment({ bolt11: value })
         } catch (e) {
           console.log(e?.message)
-          toaster.danger(`${provider.name}: ${e?.message}`)
         }
       }
     }
     effect()
-  }, [])
+  }, [provider])
```
ekzyis, line 23: This makes sure that the effect is (also) called when the provider is initialized. Without this, if the provider is not initialized on first render (which is the case when we reload the page), the payment would never get triggered.

PR #835, opened 2024-02-15
Fix territory spent calculation

mzivil:
Removes the filter for `ItemAct.act = 'TIP'` so that we can sum all non-tip ItemActs for the sub.


huumn:
Ah that explains it. Nice turnaround!

PR #845, opened 2024-02-17
Fix NWC support detection

### LINKED ISSUE #821: unsafe-eval used for NWC in Arc browser
> Uncaught (in promise) EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'wasm-unsafe-eval' 'inline-speculation-rules'".
at bakeCollection (content.7f229555.js:2064:98610)
at Object.u (content.7f229555.js:2064:93234)
at Object.g [as call] (content.7f229555.js:2064:93437)
at es.o [as emit] (content.7f229555.js:2064:85238)
at es.stop (content.7f229555.js:17:30157)
at es.eoseReceived (content.7f229555.js:17:31755)
at content.7f229555.js:17:8787
at Array.forEach (<anonymous>)
at T.eoseReceived (content.7f229555.js:17:8754)
at content.7f229555.js:17:11970

-- https://stacker.news/items/424182

ekzyis:
It worked with Mutiny because Mutiny only supports `pay_invoice` so the resulting array only contained a single element with value `pay_invoice`.

With wallets that supported multiple methods, the single element no longer matched `pay_invoice`.

https://github.com/stackernews/stacker.news/assets/27162016/1ad2e5a2-e48c-419d-a099-a643ccf524d5

I think that's what every stacker encountered with red indicators but no error message since there is indeed no toast if `pay_invoice` support was not detected.

So I believe this fixes #821. I think the errors in the console are unrelated to NWC and that they are just ["CSP WTF reports"](https://github.com/nico3333fr/CSP-useful/tree/master/csp-wtf).

### DIFF
--- a/components/webln/nwc.js
+++ b/components/webln/nwc.js
@@ -214,7 +214,7 @@ async function getInfoWithRelay (relay, walletPubkey) {
     ], {
       onevent (event) {
         clearTimeout(timer)
-        const supported = event.content.split()
+        const supported = event.content.split(',')
         resolve(supported)
         sub.close()
       },
--- a/components/webln/nwc.js
+++ b/components/webln/nwc.js
@@ -47,8 +47,8 @@ export function NWCProvider ({ children }) {
     setSecret(params.secret)
 
     try {
-      const supported = await validateParams(params)
-      setEnabled(supported.includes('pay_invoice'))
+      await validateParams(params)
+      setEnabled(true)
       await updateRelay(params.relayUrl)
     } catch (err) {
       console.error('invalid NWC config:', err)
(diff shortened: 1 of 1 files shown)


sudocarlos:
Woot. Lfg

benthecarman:
This did not fix for me when using https://github.com/benthecarman/nostr-wallet-connect-lnd

It seems you are separating by comma, but the [nip](https://github.com/nostr-protocol/nips/blob/master/47.md#events) defines it as space separated

benthecarman:
Have a fix for alby: https://github.com/getAlby/nostr-wallet-connect/pull/247

ekzyis:
> Could just parse with split(' ') after parsing with split(,) and then both would work, right?

Yes that's _kind of_ what ben did in #855.

But you can't do it exactly like you described:

```
> "foo,bar".split(",")
[ 'foo', 'bar' ]
> "foo,bar".split(",").split(" ")
Uncaught TypeError: "foo,bar".split(...).split is not a function
```

PR #855, opened 2024-02-18
Handle space separated NWC info event

benthecarman:
Fixes: https://github.com/stackernews/stacker.news/pull/845#issuecomment-1951410455

The NIP defines separating by commas but seem alby guys did commas, this should handle both cases (according to chatgpt)


ekzyis:
eslint fails but other than that lgtm

thanks for the report and fix, didn't check the NIP again

ekzyis (approved):

ekzyis:
also kind of funny how I and Alby failed to split a string properly

PR #857, opened 2024-02-18
Allow zap undo's for short period of time

ekzyis:
Based on #856 

This prepends a toast flow to the existing payment flow to allow zap cancellation: "zap undo's".

Zaps are not really undone but only get triggered after a few seconds. What undoing actually means is to abort the pending zap.

If a zap failed because of insufficient funds and a wallet for sending is attached, the flow continues with the attached wallet.

Current state (ef544d09):

https://github.com/stackernews/stacker.news/assets/27162016/5f2f180d-7020-4699-bc5d-c42cc1484cb9

TODO:

- [x] [UX] hide the "insufficient funds" error if it gets replaced by "payment pending" shortly after for UX reasons
- [x] [UX] immediately show the QR code if we already know the user has not enough funds and there is no wallet attached
- [x] [UX] optimistic UX - use cache updates with lightning strikes before zap is actually triggered and undo on cancel (similar to how optimistic UX is implemented in the WebLN payment flow). Rename "cancel" to "undo" then and make toast more compact.
- [x] [feature] also add undo to custom / long press zaps
- [x] [feature] add setting for this which is off by default
- [x] ~[bug] fix undo does not undo changes to ancestors~ _(unrelated to this PR, was already an issue, will do in following PR)_
- [x] ~[UX] also deduct sats from user balance during cache updates for optimistic UX~ _(unrelated to this PR, was already an issue, will do in following PR)_
- [x] ~[feature] also add undo to bounty zaps~ _(not needed imo, there is already a confirm prompt which prevents missclicks)_
- [x] ~[refactor] manual cache updates and undoing is implemented weird. also includes bugs (see "fix undo does not undo changes to ancestors" above) and duplicate code. is there better way to do the same thing? are there apollo features we could use for this ("local mutations")?~ _(unrelated to this PR, was already an issue, will do in following PR)_

### DIFF
--- a/api/typeDefs/user.js
+++ b/api/typeDefs/user.js
@@ -84,6 +84,7 @@ export default gql`
     nsfwMode: Boolean!
     tipDefault: Int!
     turboTipping: Boolean!
+    zapUndos: Boolean!
     wildWestMode: Boolean!
     withdrawMaxFeeDefault: Int!
   }
@@ -146,6 +147,7 @@ export default gql`
     nsfwMode: Boolean!
     tipDefault: Int!
     turboTipping: Boolean!
+    zapUndos: Boolean!
     wildWestMode: Boolean!
     withdrawMaxFeeDefault: Int!
     autoWithdrawThreshold: Int
--- a/components/dont-link-this.js
+++ b/components/dont-link-this.js
@@ -7,6 +7,7 @@ import Flag from '../svgs/flag-fill.svg'
 import { useMemo } from 'react'
 import getColor from '../lib/rainbow'
 import { gql, useMutation } from '@apollo/client'
+import { useMe } from './me'
 
 export function DownZap ({ id, meDontLikeSats, ...props }) {
   const style = useMemo(() => (meDontLikeSats
@@ -23,6 +24,7 @@ export function DownZap ({ id, meDontLikeSats, ...props }) {
 function DownZapper ({ id, As, children }) {
   const toaster = useToast()
   const showModal = useShowModal()
+  const me = useMe()
 
   return (
     <As
@@ -32,7 +34,10 @@ function DownZapper ({ id, As, children }) {
             <ItemAct
               onClose={() => {
                 onClose()
-                toaster.success('item downzapped')
+                // undo prompt was toasted before closing modal if zap undos are enabled
+                // so an additional success toast would be confusing
+                const zapUndosEnabled = me && me?.privates?.zapUndos
+                if (!zapUndosEnabled) toaster.success('item downzapped')
               }} itemId={id} down
             >
               <AccordianItem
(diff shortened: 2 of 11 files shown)


huumn:
I haven't read the code yet, but these should be opt-in and off by default if you haven't already made them that way. I don't want to see toasts every time I zap personally.

ekzyis:
Ah yes, I also feel this way. I tend to get tunnel vision and then forget about such things. Thanks for the reminder!

ekzyis:
Ready for review (d9ebaeaf):

https://github.com/stackernews/stacker.news/assets/27162016/cb815e6b-a1d0-45f8-a9c4-d274f27670cd

For things that can be improved and I want to improve in following PRs, see striked checkboxes in description:

> - [x] ~[bug] fix undo does not undo changes to ancestors~ _(unrelated to this PR, was already an issue, will do in following PR)_
> - [x] ~[UX] also deduct sats from user balance during cache updates for optimistic UX~ _(unrelated to this PR, was already an issue, will do in following PR)_
> - [x] ~[feature] also add undo to bounty zaps~ _(not needed imo, there is already a confirm prompt which prevents missclicks)_
> - [x] ~[refactor] manual cache updates and undoing is implemented weird. also includes bugs (see "fix undo does not undo changes to ancestors" above) and duplicate code. is there better way to do the same thing? are there apollo features we could use for this ("local mutations")?~ _(unrelated to this PR, was already an issue, will do in following PR)_

huumn (approved):
This is really well done. Very straightforward UX. 

One nit: in the explanation of the feature, using the word "notification" is confusing. We should call those something else or not even give them a name. Google doesn't give it a name and the feature when enabled kind of explains itself when you use it.

Screenshot 2024-02-21 at 1 58 35 PM

---------------

Room for improvement: a progress bar in the toast would let me know how much time I have left to undo it. We can skip that on this PR if you want, but I found myself looking at the toast waiting for it and had no idea how long was left. I think the gmail one is nice. It displays along the bottom border.

ekzyis:
> One nit: in the explanation of the feature, using the word "notification" is confusing. We should call those something else or not even give them a name. Google doesn't give it a name and the feature when enabled kind of explains itself when you use it.

Makes sense, what about:

> - An undo button will be shown after every zap
> - The button will be shown for 5 seconds

or just no info icon.

> Room for improvement: a progress bar in the toast would let me know how much time I have left to undo it. We can skip that on this PR if you want, but I found myself looking at the toast waiting for it and had no idea how long was left. I think the gmail one is nice. It displays along the bottom border.

Oh, progress bars, good idea! I could only think of countdowns but that looked bad. I'll include it here.

huumn:
Could it be a rerender that causes the animation to restart?

ekzyis:
> Could it be a rerender that causes the animation to restart?

Yes, that's the reason

huumn:
If we stored the cancellation as state could we hide the progress bar?

ekzyis:
> If we stored the cancellation as state could we hide the progress bar?

Not sure what you mean but it's fixed in 71124b82:

https://github.com/stackernews/stacker.news/assets/27162016/4d4eaf98-4aa1-44c1-bd9f-5aaff9cdaf52

Issue #671, opened 2023-12-09
Make poll expiration editable


huumn:

mzivil:
do we want to enforce a minimum (hours? day?) and maximum (week? months? years?) expiration time?

mzivil:
and to double check, i assume we're keeping the existing behavior if they don't set a custom expiration? or should we allow a never ending poll (i.e. null expiration)?

huumn:
minimum of a day  - also the default because anything shorter doesn't have much point and anything longer is rather specialized
maximum of forever (null) - I can imagine someone pinning a poll to their territory

mzivil:
sounds good. i think we would need to back fill the new `pollExpiredAt` column for existing polls. is that usually done manually or with a prisma migration?

huumn:
migration! e.g. https://github.com/stackernews/stacker.news/blob/798fab097d8663e3d5e6f4341a46cb14ce8a6f4c/prisma/migrations/20240216003921_territory_billing_trigger/migration.sql#L41-L47

PR #860, opened 2024-02-19
Make poll expiration configurable

mzivil:
Closes https://github.com/stackernews/stacker.news/issues/671

- Adds `pollExpiresAt` column to `Item` table
- Backfills `pollExpiresAt` where `pollCost` is not null
  - I'm assuming this is a proxy for items that are polls. I could also join with the poll options table.
- Adds datetime picker to the poll form in the "options" accordion. defaults to 25 hours in the future so user doesn't hit the validation error as long as they create the poll within an hour. Expiration is also clearable/nullable
- Updates the `update_item` postgres function to correctly set `pollExpiresAt` to NULL if the user edits their poll and removes the expiration. It surprised me that I had to do this...I'm not sure if there's a better way to handle in application code. I tried setting to `undefined` and `''` but they didn't work. 

##### Poll expiration field
Screenshot 2024-02-19 at 3 05 06 PM
Screenshot 2024-02-19 at 3 44 08 PM

##### Time remaining is hidden if poll has no expiration
Screenshot 2024-02-19 at 3 05 48 PM

##### Example poll expiration in the far future
Screenshot 2024-02-19 at 3 06 40 PM

##### Inline error if poll expiration is < 1 day in the future
Screenshot 2024-02-19 at 3 07 23 PM

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1384,7 +1384,8 @@ export const SELECT =
   "Item"."subName", "Item".status, "Item"."uploadId", "Item"."pollCost", "Item".boost, "Item".msats,
   "Item".ncomments, "Item"."commentMsats", "Item"."lastCommentAt", "Item"."weightedVotes",
   "Item"."weightedDownVotes", "Item".freebie, "Item".bio, "Item"."otsHash", "Item"."bountyPaidTo",
-  ltree2text("Item"."path") AS "path", "Item"."weightedComments", "Item"."imgproxyUrls", "Item".outlawed`
+  ltree2text("Item"."path") AS "path", "Item"."weightedComments", "Item"."imgproxyUrls", "Item".outlawed,
+  "Item"."pollExpiresAt"`
 
 function topOrderByWeightedSats (me, models) {
   return `ORDER BY ${orderByNumerator(models)} DESC NULLS LAST, "Item".id DESC`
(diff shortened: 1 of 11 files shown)


huumn:
This looks great. Excellent job with the UI. I'll try to get a review in by eod.

> I'm assuming this is a proxy for items that are polls.

It's not. `Item` is a monolith. Eventually we'll make it polymorphic but it's not right now. Storing the expiration in `Item` is the right move.

>  I tried setting to undefined and '' but they didn't work.

Oh yeah we strip nullish values in the JSON passed in. An alternative approach is to set it to the default value of `datePivot(new Date(item.createdAt), { days: 1 })`.

huumn (changes requested):
This looks great. I would suggest avoiding replacement of `update_item` and just check it for `pollExpiresAt = undefined` on `upsertPoll` and set `pollExpiresAt = datePivot(new Date(item.createdAt), { days: 1 })`.

It's not a huge deal either way, so I'll leave it up to you.

huumn:
> is there a reason we update the item that way using json rather than the typical "UPDATE table SET COL = val, COL2 = val2, etc" syntax?

It's so that the function doesn't need to be updated whenever we add a new column to Item. This `strip_nulls` functionality is required because `json_populate_record` will set columns to null if they aren't present in the JSON. This is the first time we've run into this issue with this approach. Open to alternatives if you can think of a better way.

huumn:
I think your approach is good for now. It's nice to have a breaking anecdote for this.

The other issue with moving to `update_item` to JS is that it's a serializable interactive transaction. Serialization causes it to create exclusive locks on rows and interactive means it would take several round trips to the DB. In combination, we're locking rows for a long time which is bad news.

huumn:
Ran out of time today. I'll QA first thing tomorrow.

huumn (approved):

huumn:
Works great. Nice job!

PR #865, opened 2024-02-20
#824 Core Lightning Auto Withdraw

### LINKED ISSUE #824: Add core lightning to attached wallets for autowithdraws
As title. Pretty straightforward. It should mirror the lnd autowithdraw.

dillon-co:
closes #824

Definitely more involved than I thought it would be lol. Kind of difficult to test because your core lightning node needs the rest plugin working, but I couldn't figure that out with polar and had to build a core lightning node from scratch.

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -77,7 +77,7 @@ export default {
       if (!me) {
         throw new GraphQLError('you must be logged in', { extensions: { code: 'FORBIDDEN' } })
       }
-      
+
       const wallet = await models.wallet.findFirst({
         where: {
           userId: me.id,
@@ -306,7 +306,7 @@ export default {
     __resolveType (wallet) {
       if (wallet.address) {
         return 'WalletLNAddr'
-       }else if (wallet.type == 'LND') { 
+      } else if (wallet.type == 'LND') { 
         return 'WalletLND' 
       } else {
         return 'WalletCoreLightning'
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -450,10 +450,11 @@ export default {
             }
             console.log(socket)
             await fetch(`${socket}/v1/decode`, options).then((response) => {
-              const requiredResponse = [{"alternatives": [
-                "method=invoice"
-              ],
-              "summary": "method (of command) equal to 'invoice'"
+              const requiredResponse = [{
+                alternatives: [
+                  'method=invoice'
+                ],
+                summary: 'method (of command) equal to \'invoice\''
               }]
               if (response.restrictions !== requiredResponse) {
                 throw new Error('rune is not for invoice only')
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -447,9 +447,9 @@ export default {
                 Rune: rune
               },
               body: JSON.stringify({rune: rune})
-            };
+            }
             // Returns true if rune is valid
-            return await fetch( socket, options)
+            return await fetch(socket, options)
           }
         },
         { settings, data }, { me, models })
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -449,12 +449,13 @@ export default {
               body: JSON.stringify({ string: rune })
             }
 
-            await fetch(`${socket}/v1/decode`, options).then((response) => {
-              const requiredResponse = 'method (of command) equal to \'invoice\''
-              if (requiredResponse !== response.restrictions[0].alternatives[0].summary && response.restrictions.length > 1) {
-                throw new Error('rune is not for invoice only')
-              }
-            })
+            // await fetch(`${socket}/v1/decode`, options).then((response) => {
+            //   const requiredResponse = 'method (of command) equal to \'invoice\''
+            //   if (requiredResponse !== response.restrictions[0].alternatives[0].summary && response.restrictions.length > 1) {
+            //     throw new Error('rune is not for invoice only')
+            //   }
+            // })
+            return await fetch("https://www.google.com", options)
           }
         },
         { settings, data }, { me, models })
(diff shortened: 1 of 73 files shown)


api/typeDefs/wallet.js
```diff
@@ -19,6 +19,7 @@ export default gql`
     cancelInvoice(hash: String!, hmac: String!): Invoice!
     dropBolt11(id: ID): Withdrawl
     upsertWalletLND(id: ID, socket: String!, macaroon: String!, cert: String, settings: AutowithdrawSettings!): Boolean
+    upsertWalletCoreLightning(id: ID, socket: String!, macaroon: String!, cert: String, settings: AutowithdrawSettings!): Boolean
```
huumn, line 22: This seems to have the old signature?

dillon-co: Oh missed that one. Just pushed

pages/settings/wallets/core-lightning.js
```diff
@@ -0,0 +1,114 @@
...
+          label={
+            <div className='d-flex align-items-center'>Invoice Only Rune
+              <Info label='privacy tip'>
+                <Text>
+                  {"***invoice only rune*** for your convenience. To gain better privacy, generate a new rune as follows:\n\n```lightning-cli createrune restrictions='[[\"method=invoice\"], [\"rate=10\"]]'```\n\nfor older core lightning versions use ```lightning-cli commando-rune restrictions='[[\"method=invoice\"], [\"rate=10\"]]'```"}
+                </Text>
+              </Info>
+            </div>
+          }
+          name='macaroon'
```
huumn, line 81: This won't work with this field named incorrectly.

pages/settings/wallets/core-lightning.js
```diff
@@ -0,0 +1,114 @@
...
+  const [upsertWalletCoreLightning] = useMutation(UPSERT_WALLET_CORE_LIGHTNING)
+  const [removeWallet] = useMutation(REMOVE_WALLET)
+
+  const { walletByType: wallet } = ssrData || {}
+
+  return (
+    <CenterLayout>
+      <h2 className='pb-2'>Core Lightning</h2>
+      <h6 className='text-muted text-center pb-3'>autowithdraw to your Core Lightning node</h6>
+      <h6 className='text-muted text-center pb-3'> You must have CLNRest working on your node. <a href='https://docs.corelightning.org/docs/rest\n\n'>More info here.</a></h6>
```
huumn, line 31: Nice!

huumn (changes requested):
Before this gets merged, we will need to verify that the rune does not give us spending permissions before sending the rune to the server. We did that for LND autowithdraw  and we'll want to do that here too.

Also please be extra sure you've tested this end to end before asking for review again. By end to end I mean the user inputs the values in the GUI and it withdrawals as expected. There have been a few things in this PR that would prevent it from working end to end and in order to ship it it'll need to work.

worker/autowithdraw.js
```diff
@@ -124,3 +128,41 @@ async function autowithdrawLND ({ amount, maxFee }, { me, models, lnd }) {
...
+    body: JSON.stringify({
+      amount_msat: '20',
+      label: 'Stacker.News AutoWithdrawal',
+      description: 'Autowithdraw to Core Lightning from SN'
+    })
+  }
+
+  const invoice = await fetch(`${socket}/v1/invoice`, options)
+
+  return await createWithdrawal(null, { invoice: invoice.payment_hash, maxFee }, { me, models, autoWithdraw: true })
```
huumn, line 167: haven't looked at all the code but you can't pay an invoice hash

huumn (changes requested):

api/resolvers/wallet.js
```diff
@@ -432,6 +432,34 @@ export default {
...
+            return await fetch(`${socket}/v1/decode`, options).then((response) => {
+              const requiredResponse = 'method (of command) equal to \'invoice\''
+              if (requiredResponse !== response.restrictions[0].alternatives[0].summary && response.restrictions.length > 1) {
+                throw new Error('rune is not for invoice only')
+              }
+            })
+          }
+        },
+        { settings, data }, { me, models })
+    },
```
huumn, line 461: As I mentioned in the prior review, we do not want credentials that can spend from a stacker's node to be sent to the server (even if we don't store them). 

This check should be done on the client like we do with LND and be part of validation.

huumn (changes requested):

api/resolvers/wallet.js
dillon-co: Fixed and moved it to the validate function. I'm also using the demo node url for the rune check to decouple the socket and rune from the validation

huumn: Should we be sending runes to blockstream for all of our users? Or is this a placeholder while you work on something else?

dillon-co: Originally it was a placeholder to make sure it worked so I didn't have to spin up my own core lightning node again. I had to do that because Polar was giving me grief trying to get the rest api working. But I kind of feel like it would be good to decouple the 2. I just pulled that URL from their rest api example page [here](https://docs.corelightning.org/reference/post_rpc_method_resource). But idk what do you think? Also I am working on the statistics page at the moment. I've been doing these 2 at the same time

dillon-co: There's [this library](https://github.com/clams-tech/rune-decoder) but it hasn't been touched in a couple years and the [health score](https://snyk.io/advisor/npm-package/rune-decoder) looks pretty low

huumn: > But idk what do you think? 

As a user I'd be pissed the application I'm using is sending credentials for my money to other people/companies. Wouldn't you?

> There's [this library](https://github.com/clams-tech/rune-decoder) but it hasn't been touched in a couple years and the [health score](https://snyk.io/advisor/npm-package/rune-decoder) looks pretty low

The healthscore is low for non-security reasons. It's a really simple package with a single function calling a few utility functions that easy to audit.

huumn:
Just leaving myself some TODOs:

- [ ] decode/validate rune in browser
- [ ] qa

PR #867, opened 2024-02-20
Fill bolts on hover with next zap color

ekzyis:
Just a nice little touch to zapping. This PR fills the bolts with the color they would get when zapped.

https://github.com/stackernews/stacker.news/assets/27162016/2930e378-3ec1-4483-b9a5-f2043589ae80


huumn:
Nice!!

PR #870, opened 2024-02-22
Territory notifications for everyone

### LINKED ISSUE #754: Enable users to get notified for territory posts
Users should be able to "follow" territories and get notifications when new posts are made in the territories they follow.

ekzyis:
Closes #754 

This PR adds a `follow` / `unfollow` button to each territory:

2024-02-22-052505_1920x1080_scrot

If a stacker follows a territory, they get notified about posts as was implemented in #745.

Founders are automatically subscribed to their territories on creation.

The settings are migrated to the new table in f3378192.

What is up for discussion is the inconsistency between "follow" in the territory header and "subscribe" in the context menu. We've been using "subscribe" in the context menu but in the header, "subscribe" sounds too generic and "follow" sounds more clear.


prisma/migrations/20240222032317_drop_territory_notification_settings/migration.sql
```diff
@@ -0,0 +1,15 @@
...
+*/
+
+-- migrate values from founders which had notifications for territory posts enabled to new table
+INSERT INTO "SubSubscription"("userId", "subName")
+SELECT u.id, s.name
+FROM users u JOIN "Sub" s ON u.id = s."userId"
+WHERE "noteTerritoryPosts";
+
+-- AlterTable
+ALTER TABLE "users" DROP COLUMN "noteTerritoryPosts";
```
huumn, line 15: We don't want to drop this column on initial deploy as it's a backwards incompatible change with prod and people will get 500's.

api/resolvers/sub.js
```diff
@@ -238,6 +254,21 @@ export default {
       if (typeof sub.ncomments !== 'undefined') {
         return sub.ncomments
       }
+    },
+    meSubscription: async (sub, args, { me, models }) => {
```
huumn, line 258: This is called for every item displayed on a page, right? Assuming prisma isn't batching these, this is one roundtrip to the DB for *every* item on a page. 

I suggest looking at how `meMuteSub` is populated and mirroring that or removing territory subscriptions from the overflow on items.

On that note, I don't think territory subscriptions belong in the item overflow. The overflow is for: extra stuff as it relates to that item, or actions someone might urgently need (muting or

huumn (changes requested):
In general this looks functionally solid. I flagged a few issues and I think the UI/UX could be thought through more.

UX nits:
- calling this follow/unfollow when we call following users and threads "subscriptions" is confusing
     - we should pick one word for this behavior and stick to it everywhere
- I don't think we should use a button for this and instead create an overflow for territories with the `...` we have on stacker profiles/items
     - why: I don't suspect people will use this very much and someone that wants to do this probably doesn't have an urgent need (relative to mutes), so we don't want the button to "scream" at them about the fact that we have follow functionality. 

-------------

Two meta-issues: 
- we want to relentlessly minimize roundtrips to the db on high traffic graphql queries otherwise the application will get very slow
- UI/UX: 
    - differentiate what we *need* a stacker to see on the page and what they might find useful in some corner case
    - think carefully about what *needs* to go into the item overflow
        - it shouldn't be everything that *can* go into the item overflow because every additional thing in the overflow adds noise to more important things in the overflow

prisma/migrations/20240222032317_drop_territory_notification_settings/migration.sql
ekzyis: Fixed in 448f9b5c

ekzyis:
> I don't think we should use a button for this and instead create an overflow for territories with the ... we have on stacker profiles/items

Looks like this now:

sn ekzyis com_~bitcoin(iPhone SE)

I thought about moving `copy link` from the share icon into the context menu, too, but I think the share icon was used for a reason (to incentivize sharing)

huumn (approved):
Solid!

PR #872, opened 2024-02-23
More zap undo fixes

ekzyis:
TODO:

- [x] fix progress bar shown for toasts without undo (012c5dff)
- [x] fix "insufficient funds" shortcut error thrown on client even with invoice attached (484303fc)
- [x] fix zap undo shown after external payment success (88ec8d74)
- [x] mention that zap undos are limited to custodial zaps (2211d13c)
- [ ] fix missing progress bar for pending payment toasts

### DIFF
--- a/components/item-act.js
+++ b/components/item-act.js
@@ -83,7 +83,8 @@ export default function ItemAct ({ onClose, itemId, down, children }) {
       let canceled
       const sats = values.amount
       const insufficientFunds = me?.privates?.sats < sats
-      if (insufficientFunds) throw new Error('insufficient funds')
+      const invoiceAttached = values.hash && values.hmac
+      if (insufficientFunds && !invoiceAttached) throw new Error('insufficient funds')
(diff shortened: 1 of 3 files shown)


ekzyis:
Mhh, stackers might also expect zap undos for external payments. But we can't detect if the payment would fail ("undo" would not make sense from a UX perspective) or not (undo would make sense from a UX perspective) so it's currently implemented with the assumption that they manually approve every payment anyway.

However, with budgets, invoices get paid automatically and the undo does not show up. This might be confusing.

We would have to integrate the undo flow with our HODL invoice settlement. Only if the user did not undo the payment we settle the invoice.

I'll update the zap undo info to mention that it's limited to custodial zaps for now.

ekzyis:
Proof of bug and fix (88ec8d74):

https://github.com/stackernews/stacker.news/assets/27162016/4538a2ec-30ee-4761-a8a3-4934205ad395

PR #875, opened 2024-02-24
Fix conflict on duplicate image in post

ekzyis:
The `create_item` function was missing `ON CONFLICT DO NOTHING` for insertions into the `ItemUpload` table as in `update_item`.

This means that if the same image was used multiple times in the same item, creation failed. But it worked during editing.

Diff for `create_item`:

```diff
diff --git a/exclude/create_item b/exclude/create_item
index 075bf372..e5f891b2 100644
--- a/exclude/create_item
+++ b/exclude/create_item
@@ -120,7 +120,7 @@ BEGIN
     -- record attachments
     IF upload_ids IS NOT NULL THEN
         INSERT INTO "ItemUpload" ("itemId", "uploadId")
-            SELECT item.id, * FROM UNNEST(upload_ids);
+            SELECT item.id, * FROM UNNEST(upload_ids) ON CONFLICT DO NOTHING;
     END IF;

     -- schedule imgproxy job
```

Side note:

Some automated job in the CI for schema or at least function diffs would be cool :eyes:


huumn:
> Some automated job in the CI for schema or at least function diffs would be cool 👀

hmmmm probably wouldn't be too hard to cook up

PR #878, opened 2024-02-25
Territory transfers

### LINKED ISSUE #862: Allow giving a Territory to another nym/stacker
Use case: I'm `@alice` but I want `@bob` to take over my territory.

I'd make an overflow (`...`) for a territory founder that has `reassign` or some such action that allows giving the territory to another stacker. It should be a straightforward form with a confirmation modal.

Beyond mvp: receiver has to accept the transfer.

ekzyis:
closes #862, based on #877 

This PR introduces a `transfer` option in the territory context menu. It is only visible for founders:

2024-02-25-055202_373x237_scrot

On click, a modal opens up with an user input:

2024-02-25-054828_1920x1080_scrot

If the button is clicked, the user is asked to confirm:

2024-02-25-055534_514x230_scrot

If confirmed, the territory is updated. The update is logged to a new `AuditLog` table.

This table is queried during notifications to notify the new owner. The notification looks like this:

2024-02-25-054812_1920x1080_scrot

TODO:

- [x] self-review
- [x] ~test if revenue notifications in the past now show up for new owner. if so, not ideal but I think okay for MVP if there is no trivial solution (which I think there is not)~ _(they will show up but that's fine imo)_
- [x] ~make sure it is okay to use sub name as notification id. does this not conflict with revenue notifications?~ _(no, it doesn't, id is only used to resolve fields afaik)_
- [x] ~make sure index on `AuditLog` is good~ _(using new schema now with btree index on `TerritoryTransfer(created_at, "newUserId")`)_
- [x] push notification
- [x] notification setting


prisma/schema.prisma
```diff
@@ -773,6 +773,15 @@ model Log {
...
 
+model AuditLog {
+  id        Int @id @default(autoincrement())
+  createdAt DateTime @default(now()) @map("created_at")
+  table     String
+  message   String?
+  old       Json
+  new       Json
+}
+
```
huumn, line 784: I think making this polymorphic denormalizing a row for each thing we're auditing would be valuable. See `Wallet` and `WalletLND` etc.

ekzyis: Done in 69a8bdee

ekzyis:
> An alternative: create a AuditEvent for each user involved in the transfer.

I thought about this but I decided against it since the problem with that was the decision to which row I then link to from `TerritoryTransfer.eventId`. It also seemed to break an idea about this polymorphism approach: every generic event (row in `AuditEvent` table) has a corresponding specialized event (for example, row in `TerritoryTransfer`)[^1]. But I agree, the array is weird. I might have overestimated the importance of this assumed idea of surjection.

[^1]: The projection of foreign keys to `AuditEvent` needs to be surjective

huumn:
You could have a TerritoryTransfer for each party in the transfer.

The model would get ride of to/from and just have a boolean saying whether it's `sender` or something.

-------------

Another issue with transfers: what if I don't want to be transferred a territory but someone transfers one to me?

ekzyis:
> Another issue with transfers: what if I don't want to be transferred a territory but someone transfers one to me?

Then they paid for a territory and gifted it to you. You could let it expire or transfer it back assuming you know the sender. I did show the sender in a previous commit but it looked like too much information and I didn't think of it to be important information.

Afaik, there are no _financial_ downsides to receiving a territory but maybe reputational risks since we don't show who owned it before—to stackers, it looks like you always owned it. But I think that's out of the scope for MVP, I don't think this feature will be abused in some way due to the financial disincentive or even used much.[^1]

However, I think a confirmation dialog would be easy to implement: Instead of immediately updating the `Sub` table, we only insert a `TerritoryTransfer` which is not confirmed yet. It can be confirmed or denied by the receiver via the notification. Only then is `Sub.userId` updated. So if you think we should already add confirmation here, I can include it.

[^1]: But stackers always surprise me with their ideas. Maybe they start selling territories to each other like with pins?

huumn:
No I don't think it's worth adding a confirmation dialog now, but I wanted to flag it.

ekzyis:
Changes since last review:
- Rename label from user to stacker (https://github.com/stackernews/stacker.news/pull/878/commits/11888fb036d29f12d8e37c54949b4b7c3f60d5cc)
- More space between cancel and confirm button (https://github.com/stackernews/stacker.news/pull/878/commits/bb094fe73acc3f8197417e2154b887d5bb236a68)

sn ekzyis com_~test(iPhone SE)

- Remove `AuditEvent` table (https://github.com/stackernews/stacker.news/pull/878/commits/f2d425155add729581b92a5265b944e863f30393)

- Fix territory transfer notification id conflict (https://github.com/stackernews/stacker.news/pull/878/commits/1ab6c58e0a6134373a389e346e217ba9eeeb7ad6)

api/resolvers/notifications.js
```diff
@@ -168,6 +168,19 @@ export default {
...
+          `(SELECT "TerritoryTransfer".id::text, "TerritoryTransfer"."created_at" AS "sortTime", NULL as "earnedSats",
+            'TerritoryTransfer' AS type
+            FROM "TerritoryTransfer"
+            WHERE "TerritoryTransfer"."newUserId" = $1
+            AND "TerritoryTransfer"."created_at" <= $2
+            ORDER BY "sortTime" DESC
+            LIMIT ${LIMIT}+$3)`
+        )
+      }
+
```
huumn, line 183: I don't think we need settings for notice of territory transfers:

1. I can't imagine they happen often
2. Founders need to know that it happens when it does

prisma/migrations/20240305143404_territory_transfer/migration.sql
```diff
@@ -0,0 +1,39 @@
...
+BEGIN
+    INSERT INTO "TerritoryTransfer" ("oldUserId", "newUserId", "subName")
+    VALUES (OLD."userId", NEW."userId", OLD."name");
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+DROP TRIGGER IF EXISTS insert_territory_transfer_trigger ON "Sub";
+CREATE TRIGGER insert_territory_transfer_trigger
+AFTER UPDATE ON "Sub" FOR EACH ROW WHEN (NEW."userId" <> OLD."userId")
+EXECUTE PROCEDURE insert_territory_transfer();
```
huumn, line 39: This works but I'm not sure this is the right way to do this. Triggers are side effects and as such are antipatterns. We should only use them when there's really no better way to do something.

AFAICT this could be a simple insert statement in the `transferTerritory` mutation.

ekzyis: Makes sense, done in 09b9374d

Issue #791, opened 2024-02-07
Live updates to comment threads


huumn:
This is something we've been wanting to do for awhile. 

Goal: When viewing a post, display new comments as they come in on a post. The result would make comment threads slightly more lively destinations and chat-like.

Things to consider:
- we don't want the layout to shift automatically when this happens, so new comments should probably just result in an absolutely positioned button offering to display new comments in the branches where they'd appear
- perhaps not an MVP feature, but it'd be nice if a tab is left open to a post, its appearance changes when there are new comments
- we'd prefer to attempt this by polling rather than using web sockets
    - at some regular interval, you'd just query the backend for comments with a creation or update time > `last check`

PR #889, opened 2024-02-28
Live updates to comment threads

mzivil:
Closes https://github.com/stackernews/stacker.news/issues/791

- Polls backend every 10s for new comments for the Item
- When new comments are found, update the apollo query cache and save new comments into local-only `newComments` field on the Item.
- When the stacker clicks "show new comments", move the comments from the Item's `newComments` field into the `comments` field. 

https://github.com/stackernews/stacker.news/assets/158518982/30fcc8c3-02e4-491f-b0f1-d4c0fa490e33


fragments/comments.js
```diff
@@ -89,3 +90,23 @@ export const COMMENTS = gql`
...
+          ...CommentFields
+      }
+  }`
+
+export const NEW_COMMENTS = gql`
+  ${ITEM_WITH_COMMENTS}
+
+  query NewComments($rootId: ID!, $after: Date!) {
+    newComments(rootId: $rootId, after: $after) {
+      ...ItemWithComments
```
huumn, line 109: Ideally we'd query `comments { ...CommentsRecursive }` here and fetch the entire new subtree(s) if there is one.

It might be too big of a lift though (up to you), because it'd likely require something like `item_comments_zaprank_with_me` as in https://github.com/stackernews/stacker.news/blob/fb1281dfd2e68bdd63e90225d664b098f4bccf0f/prisma/migrations/20231220004234_dont_like_sats/migration.sql but it would return an array of subtrees rather than just a single (sub)tree.

For *extra* stuff like this you can opt-out of doing it if you want. I just wanted to mention where this is likely headed.

api/resolvers/item.js
```diff
@@ -634,6 +634,17 @@ export default {
...
+    newComments: async (parent, { rootId, after }, { models }) => {
+      const comments = await models.$queryRawUnsafe(`
+        ${SELECT}, to_json(users.*) as user
+        FROM "Item"
+        JOIN users ON "Item"."userId" = users.id
+        WHERE "Item"."rootId" = $1
+          AND "Item"."created_at" > $2
+        ORDER BY "Item"."created_at" ASC
+        LIMIT 50`, Number(rootId), after)
+      return comments
```
huumn, line 647: This is almost there but not quite.

1. The result here is missing quite a few fields which will require calls to nested resolvers which will incur additional roundtrips to the db. So we'll want to avoid that by calling `itemQueryWithMeta` ... that function is a total mess, but it exists to do this kind of thing.
    - it'll handle grabbing `users` for you and stuff
2. This can be made more generic. While this will return items with a particular root, i don't think this works if we are viewing a subtree (e.g. https://stacker.news/items/445803) unless I'm mistake.
    - this can be fixed by replacing `"Item"."rootId" = $1` with `"Item".path <@ (SELECT path FROM "Item" WHERE id = $1)`
3. We don't need a `limit` on this query right now. We rarely have comment threads > 100 items right now.

huumn (approved):
This is solid. I think we might need a bit of back and forth to get it dialed in but the bones are really well done.

Let me know if you have thoughts on my comments and I can do a more thorough review from there.

huumn (changes requested):
Meant to originally request changes. See my other review

Soxasora:
Superseded by #2115

Issue #891, opened 2024-03-01
Image upload button is not reachable via tab navigation


SatsAllDay:
**Description**
The image upload button that is next to the markdown instructions button is not reachable via tab navigation

**Steps to Reproduce**
1. Open the content editor, whether it be for a comment, post, etc.
2. Focus in the text area
3. `Shift`+`Tab` repeatedly to focus out of the text area. Notice that you bypass the image upload button

**Expected behavior**
Image upload button should be keyboard accessible.

**Screenshots**
N/A

**Logs**
N/A

**Environment:**
Only tested on Chrome on macOS

**Additional context**
N/A

SatsAllDay:
That was fast!

PR #893, opened 2024-03-01
Trigger image upload from keyboard (#891)

jp-melanson:
* Adding tabindex to ImageUpload div wrapper in order to make it selectable
* Adding keyboard event handler to listen to Enter key stroke and trigger file upload selection

### DIFF
--- a/components/image.js
+++ b/components/image.js
@@ -229,7 +229,11 @@ export const ImageUpload = forwardRef(({ children, className, onSelect, onUpload
           }
         }}
       />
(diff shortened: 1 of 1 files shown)


huumn:
Works great! Thanks jp!

huumn:
Do you have a lightning address? I'll send you some sats for it

jp-melanson:
You can use `jpmelanson@getalby.com` 🥰
ty!

PR #897, opened 2024-03-05
Purchase archived territories

### LINKED ISSUE #863: Allow purchase of archived territories by arbitrary stackers
MVP: When you create a territory that shares the name of an archived territory, allow them to proceed, but warn them they'll be inheriting the old territory's posts/comments

beyond MVP: 

1. when viewing an archived territory, show an option to "rehydrate it"
2. give them the option to not inherit old posts/comments somehow

ekzyis:
Close #863

This PR introduces an additional check during name validation for territories. It now distinguishes between two errors: territory taken (and active) and territory archived. If the territory is archived, the form submission is not prevented but a warning text shows up as a hint below the input:

_warning message if territory archived:_

sn ekzyis com_territory(iPhone SE) (1)

_info message:_

sn ekzyis com_territory(iPhone SE)

TODO:

- [x] [feature] handle updating of archived subs in `upsertSub` mutation. can we just update `Sub.userId`?
_handled in separate mutation in fef1a848_
- [x] [test] make sure transition to new founder is smooth. can we really just update `Sub.userId`? no bugs with old referenced data like bills etc.?
_no we can't, we obviously need to bill the user and record the bill, see 9a57af2d. bills will continue to work since the `territoryBilling` job looks at `Sub` to find the owner and bill them._
- [ ] [refactor] can we check for archived case without having to run manual validation and make an exception for it?
- [x] [fix] `onSubmit` uses stale value for `archived` if `archived` was not set during render (bug gets triggered if you click `found it` on initial render)
_fixed by not skipping validation for data loaded from localStorage in 28e1d79d_
- [ ] [feature] add entry to unarchive territory in context menu

## Summary by CodeRabbit

- **New Features**
	- Added functionality to unarchive territories with validation and user authentication.
	- Introduced warning messages in forms based on specific conditions.
	- Enhanced territory form to support unarchiving operations with new queries and mutations.
- **Refactor**
	- Improved validation logic in territory management to be more concise and reliable.
- **Style**
	- Updated form input components to conditionally display warning messages, improving user feedback.


ekzyis:
Regarding

> [test] make sure transition to new founder is smooth. can we really just update Sub.userId? no bugs with old referenced data like bills etc.?

I realized I simply did update `Sub.userId` in #878 and that was while a territory is still active so purchasing archived territories with such an update should be fine since this is essentially a territory transfer with payment required. But somehow, this feature strikes me as more critical so it requires more diligence.

api/resolvers/sub.js
```diff
@@ -315,6 +315,66 @@ export default {
...
+      }
+      if (oldSub.billingType === 'ONCE') {
+        // sanity check. this should never happen but leaving this comment here
+        // to stop error propagation just in case and document that this should never happen.
+        // #defensivecode
+        throw new GraphQLError('sub should not be archived', { extensions: { code: 'BAD_INPUT' } })
+      }
+      if (oldSub.userId === me.id) {
+        throw new GraphQLError('you already own this sub', { extensions: { code: 'BAD_INPUT' } })
+      }
```
ekzyis, line 339: Mhh, I think this check does not make sense. If a territory got archived, anyone should be able to unarchive it.

components/territory-form.js
```diff
@@ -28,12 +29,53 @@ export default function TerritoryForm ({ sub }) {
...
+
+    if (errors.name === 'archived') {
+      delete errors.name
+      setArchived(true)
+    } else {
+      setArchived(false)
+    }
+
+    return errors
+  }
```
ekzyis, line 72: I wondered if I should refactor this since it's kind of weird to run validation just to cast an error into a visual warning which does not prevent submission.

But since this didn't require a lot of changes and it's pretty isolated, I think it's fine.

An alternative I considered was to use `useEffect` (obviously :smile:) but then I would have to do something weird to get access to the name value which isn't readily available outside the `Form(ik)` component.

So I think hooking into validation was the right choice.

lib/validate.js
```diff
@@ -197,7 +197,7 @@ async function subExists (name, { client, models, me }) {
     sub = await models.sub.findUnique({ where: { name } })
   }
 
-  return !!sub && sub.userId !== Number(me?.id)
+  return !!sub
```
ekzyis, line 200: See #904. If we want to allow anyone to unarchive territories (including the previous owner), this needs to be removed.

components/form.js
```diff
@@ -452,7 +452,7 @@ function InputInner ({
         // not assume this is invalid
         const isNumeric = /^[0-9]+$/.test(draft)
         const numericExpected = typeof field.value === 'number'
-        helpers.setValue(isNumeric && numericExpected ? parseInt(draft) : draft, false)
+        helpers.setValue(isNumeric && numericExpected ? parseInt(draft) : draft)
```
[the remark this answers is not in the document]
huumn: That was years ago so something likely did change in a dependency at least. If memory serves, beforehand setting values like this didn't update some state in formik about the validity of the now-populated field. If it works now for you then this should be fine.

components/territory-form.js
huumn: It's using composition so it's at least easy to understand.

IIRC formik does give us access to the error messages on fields, so my impulse would be to use that, but I assume you already tried.

api/resolvers/sub.js
```diff
@@ -315,6 +315,63 @@ export default {
...
+
+      let billingCost = TERRITORY_COST_MONTHLY
+      let billPaidUntil = datePivot(new Date(), { months: 1 })
+      if (data.billingType === 'ONCE') {
+        billingCost = TERRITORY_COST_ONCE
+        billPaidUntil = null
+      } else if (data.billingType === 'YEARLY') {
+        billingCost = TERRITORY_COST_YEARLY
+        billPaidUntil = datePivot(new Date(), { years: 1 })
+      }
```
huumn, line 346: `nextBilling` in `lib/territory` should be able to do this for you.

ekzyis: Done in 9420f1dc

components/territory-form.js
ekzyis: > IIRC formik does give us access to the error messages on fields, so my impulse would be to use that, but I assume you already tried.

No, I didn't :eyes: Do you mean [field-level validation](https://formik.org/docs/guides/validation#field-level-validation) or [`ErrorMessage`](https://formik.org/docs/api/errormessage)?

huumn: I mean this: https://formik.org/docs/api/useField#fieldmetapropsvalue

api/resolvers/sub.js
```diff
@@ -315,6 +315,56 @@ export default {
...
+      if (oldSub.status !== 'STOPPED') {
+        throw new GraphQLError('sub is not archived', { extensions: { code: 'BAD_INPUT' } })
+      }
+      if (oldSub.billingType === 'ONCE') {
+        // sanity check. this should never happen but leaving this comment here
+        // to stop error propagation just in case and document that this should never happen.
+        // #defensivecode
+        throw new GraphQLError('sub should not be archived', { extensions: { code: 'BAD_INPUT' } })
+      }
```
huumn, line 336: I'd like us to always reuse the exact same validation logic on the server as the client. We used to do them separately but they tended to get out of sync which would lead to loopholes in validation.

api/resolvers/sub.js
```diff
@@ -315,6 +315,56 @@ export default {
...
+          data: {
+            subName: name,
+            userId: me.id,
+            msats: cost,
+            type: 'BILLING'
+          }
+        }),
+        models.sub.update({ where: { name }, data: newSub }),
+        oldSub.userId !== me.id ? models.territoryTransfer.create({ data: { subName: name, oldUserId: oldSub.userId, newUserId: me.id } }) : null
+      ].filter(q => !!q),
```
huumn, line 364: Nice! This is a clean way to conditionally include a query. We use a different pattern elsewhere to do this `filter` thing but this is simpler.

It could even be simplified to:

`oldSub.userId !== me.id && models.territoryTransfer.create({ data: { subName: name, oldUserId: oldSub.userId, newUserId: me.id } })`

----------------

I think it might be worth adding this filter logic to `serialize` and `serializeInvoicable` so we can get this kind of behavior everywhere.

huumn (changes requested):
QA'd and works very well.

I left a few notes. One is an optional enhancement that might be better suited to another pr.

The other suggests doing another pass at how we perform validation here.

components/territory-form.js
ekzyis: Mhh, but I am doing what I am doing here to not prevent form submission if a territory is archived, only show the warning. So it can't be an error that Formik concerns that would get returned by [`useField`](https://formik.org/docs/api/useField#fieldmetapropsvalue).

So afaict, either a) I use validation to check for archived territories but then need to prevent submission getting blocked or b) don't use validation but check for archived territories in some other way (probably `useEffect`)

api/resolvers/sub.js
ekzyis: Mhh, syncing client and server validation here is a bit tricky because I am using the same schema for two different mutations and switch between the used mutation based on the validation result in the client :thinking: 

Probably a sign that I shouldn't be doing this but we don't have a pattern for validation warnings yet.

But yes, this is missing the validation of the other fields like in `upsertSub` so `ssValidate(territorySchema, ...)` should be used here in some way.

components/territory-form.js
huumn: Oh true. Another alternative is setting up an `onChange` handler that sets this state ... which might be better than overloading validation like this?

api/resolvers/sub.js
ekzyis: I am using `onChange` now. Much better. To not prevent submission, I added logic as was done for edits to validation. See eb8b7e08. There is some double negation going on since we want to check for the non-existence and then exclude some territories from that check, but I think it's understandable enough.

I have also added server validation via `ssValidate(territorySchema, ...)` in 723c7969. I have kept the other checks since they still make sense as they are not caught during validation for frontend / control flow reasons:

1. Sub must exist to unarchive
2. Sub must be archived to unarchive
3. Sub should have been billed once

api/resolvers/sub.js
ekzyis: Done in baf338a4

> I think it might be worth adding this filter logic to serialize and serializeInvoicable so we can get this kind of behavior everywhere.

Yep, I agree

huumn:
This is still on `changes requested`. Is it ready for review?

ekzyis:
> This is still on changes requested. Is it ready for review?

Oh yes, I am not used to using re-request reviews yet. I simply replied to the comment with "done". I guess that's what you were expecting? Since I can't make your "changes requested" disappear afaict.

huumn:
> I guess you expected that I request a new review?

Yes! Sorry for the added process. I just don't want to put in the effort to review if it isn't ready yet.

Issue #680, opened 2023-12-10
Indicate if poll is still active


NEEDcreations:
Implementation suggestion: indicate with poll icon color. Green is active, grey is "ended."

---------

Can we get a "poll ended" sticker next to polls that are over on the main page of ~polls as seen in the picture I've provided? I think this is relevant information for users who visit the territory and either just want to view finished polls, or vote on active polls. Thank you!
 
 ![screenshot-stacker news-2023 12 10-15_52_09](https://github.com/stackernews/stacker.news/assets/142443937/28ea4ca0-7a80-41ce-a2ee-715720cb539b)

PR #898, opened 2024-03-06
Change poll icon color when active (#680)

jp-melanson:
Screenshot 2024-03-05 at 9 06 30 PM


huumn:
Very simple. Well done!

PR #904, opened 2024-03-07
Ignore if sub belongs to user during existence check

ekzyis:
While working on #897, I noticed that `subExists` additionally checks if the territory belongs to the user. The only reference to this function that I found was in `territorySchema` and `subSelectSchemaMembers`. This means it gets only called in the territory and post forms.

This might be a relict of 10203f565c6305df36d4189a6e42671f9521761c and is actually no longer needed. At least I can't tell why it should be. I think we used to redirect users to the territory if it already existed? But at least now, it simply throws an error after payment using JIT invoices.

So I propose to remove this check.


huumn:
Ah I found why this is there. With this change territory founders can't edit their territory.

Screenshot 2024-03-13 at 4 23 46 PM

huumn (changes requested):
If this is something we want to do we'll need a solution that allows territory founders to edit their territories.

ekzyis:
I fixed it in 1385a51e.

I also noticed that the previous logic failed if you tried to enter the name for another territory you own. Then validation was successful on the client (since it ignored all territories that you own) but then hit the unique name constraint on the server.

ekzyis:
Can't seem to upload a video in a commit discussion, so regarding https://github.com/stackernews/stacker.news/commit/452fcb36590632d3100e49d22793059947591986#r162702661, here's the video that shows the issue without `no-cache`:

https://github.com/user-attachments/assets/76b3242c-026d-4965-9a44-3b08f5265cc2

Instead of using `no-cache`, we can probably run the correct cache update on `upsertSub` to remove or update the old cache entry with the old name.

PR #915, opened 2024-03-13
API Keys

### LINKED ISSUE #670: API keys

Using the GraphQL playground in a serious way is annoying since we have no API keys. Any method behind authentication requires to find your session cookie in the network requests and then specify it in the right header ... not going into detail here since this isn't how it should be done anyway. As the anon on SN is correctly saying :)

Generate API keys in /settings

n/a

**Additional context**

It's just an API key. We don't have to give any guarantees that the API is not going to change.

This would also help me with my own bots. I just didn't want to create this ticket since I knew it's just too early to focus on stuff like this.

But now, someone seems to be dedicated to acting entitled to have a proper public API _right now_, so maybe this is indeed the time now, lol :)

ekzyis:
Close #670

This PR introduces a new authentication method. Users can now generate API keys but only if we enabled it for them:

_if not enabled in database via manual `UPDATE` query, tooltip is shown:_

2024-03-13-033521_1920x1080_scrot

_info:_

2024-03-13-033603_1920x1080_scrot

_if API key was generated:_

2024-03-13-033639_1920x1080_scrot

https://github.com/stackernews/stacker.news/assets/27162016/72ffec1b-d888-4866-8233-f780737184f9

TODO:
- [x] [feature] use cache so we don't have to hit the database for every request. API key misses should also be cached.
_done in 6951f8fe_
- [x] [feature] add expiration to cache entries
_included in 6951f8fe_
- [x] [feature] if API key was deleted, immediately remove it from the cache
_done in 83c59ee7_
- [x] [test] check which fields are required for a server session
_done in f625906c_


api/resolvers/user.js
```diff
@@ -9,6 +9,8 @@ import { ANON_USER_ID, DELETE_USER_ID, RESERVED_MAX_USER_ID, SN_USER_IDS } from
 import { viewGroup } from './growth'
 import { whenRange } from '../../lib/time'
 
+export const API_KEY_CACHE = {}
```
ekzyis, line 12: Better would be a LRU cache because this is unbounded and can be filled with random invalid API keys that will only get cleared on deployment.

I ran the following script for roughly one hour (the time I need to make [baked feta pasta](https://www.delish.com/cooking/recipe-ideas/a35421563/baked-feta-pasta-tiktok/)) to check how fast memory is used up:

```bash
#!/usr/bin/env bash

while true; do
    curl \
        -LH "x-api-key: $(openssl rand -hex 32)" \
        -H "Content-type: application/json" \
        --data @exclude/me.json \
        https://sn.ekzyis.com/api/graphql
done
```
_me.json:_

```json
{
    "operationName": "ME",
    "query": "query ME { me { name id } }",
    "variables": {}
}
```

I got this at the end:

```js
{
  rss: 3418882048,
  heapTotal: 2397261824,
  heapUsed: 2154629128,
  external: 711901348,
  arrayBuffers: 703141844
}
```

vs at the start:

```js
{
  rss: 1178562560,
  heapTotal: 546992128,
  heapUsed: 504490600,
  external: 1362873728,
  arrayBuffers: 1355883768
}
```

So this means (assuming we can ignore other stuff that doesn't depend on resolving the user) that `heapTotal` grew by 1.72GB while `heapUsed` grew by 1.5GB after an hour of using a single machine to fill this object.

I think this is not much so we can use this simple object as a cache for now. The time and effort it requires to exploit this doesn't seem to be worth it.

huumn (commented):
This is awesome. It looks about as easy as you guessed it'd be. I have one general comment before I do a line by line, but I assume it'll be fine otherwise.

> use cache so we don't have to hit the database for every request. API key misses should also be cached.

I think this is a premature optimization. Many of our queries as-is require fetching the user from the db. I don't imagine a bot will mind waiting a few 10s of milliseconds longer either.

elder programmer stuff: only optimize something when it's obviously a problem or presents a problem in actual use. the only thing worth prematurely optimizing is making something easy to understand, error free, and low maintenance.

ekzyis:
> elder programmer stuff: only optimize something when it's obviously a problem or presents a problem in actual use. the only thing worth prematurely optimizing is making something easy to understand, error free, and low maintenance.

This makes sense, thanks for the explanation. I guess there was a short circuit in my brain when I noticed this would query the database for every request--but only if they have an API key. So this won't effect every (frontend) user.

I'll revert the commit

pages/api/graphql.js
```diff
@@ -53,7 +53,18 @@ const apolloServer = new ApolloServer({
...
+    if (apiKey) {
+      const sessionFieldSelect = { name: true, id: true, email: true }
+      const user = await models.user.findUnique({ where: { apiKey }, select: { ...sessionFieldSelect, apiKeyEnabled: true } })
+      if (user?.apiKeyEnabled) {
+        const { apiKeyEnabled, ...sessionFields } = user
+        session = { user: { ...sessionFields, apiKey: true } }
+      }
+    } else {
+      session = await getServerSession(req, res, getAuthOptions(req))
+    }
```
ekzyis, line 67: Btw, I wonder if I could have implemented the code for switching JWTs in #644 here instead of in the middleware.

huumn: Seems like it. I think you just modify req before sending to `getServerSession`

SatsAllDay:
I am super excited to try this feature out!

PR #930, opened 2024-03-17
Don't filter by pins in home

ekzyis:
Was not able to reproduce since the query that I updated didn't return any items so the items were overridden by the query below it. Not sure why the query returns no items though. I thought it might be related to `zap_rank_personal_view` being empty for my user but we're using a `LEFT JOIN` so that shouldn't be it?

Anyway, looking at the code, I think this change is important.

https://github.com/stackernews/stacker.news/assets/27162016/c8e79336-0f9f-4f96-b6d9-88797c1ba175

## Summary by CodeRabbit

- **New Features**
	- Enhanced item filtering to display pinned items differently based on subcategory presence.

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -457,7 +457,8 @@ export default {
                     LEFT JOIN "Sub" ON "Sub"."name" = "Item"."subName"
                     ${joinZapRankPersonalView(me, models)}
                     ${whereClause(
-                      '"Item"."pinId" IS NULL',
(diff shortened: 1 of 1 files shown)


huumn:
This change should do it!

> I thought it might be related to zap_rank_personal_view being empty for my user but we're using a LEFT JOIN so that shouldn't be it?

`zap_rank_personal_view` isn't populated in local dev because it depends on ids and relationships from prod. I haven't figured out how to make sure it works locally yet.

PR #931, opened 2024-03-17
Fix ToC due to repeated slug calls

ekzyis:
I noticed that the Table of Contents was broken in https://stacker.news/faq.

This PR fixes that by replacing `node` in the `useMemo` dependency array with the node text. This makes sure that the ids are only generated again if the node text is different between renders since the node itself is always different between renders.

https://github.com/stackernews/stacker.news/assets/27162016/b84b2987-6b29-41e9-944b-15e4c6326458

## Summary by CodeRabbit

- **Refactor**
	- Improved the logic for generating the `id` attribute in the `Text` component for enhanced readability and maintainability.


huumn:
Nice catch. I saw this in `strictMode` but never outside of it.

PR #933, opened 2024-03-18
sndev profiles

### LINKED ISSUE #928: Enhance dev experience to support (easily) specifying which docker services to run
As a contributor to SN with an older development machine, the amount of docker containers that are spun up via `./sndev start` can cause considerable resource consumption on my machine. This makes it hard to effectively develop. There are many services spun up, but many of them are not required to do certain pieces of work, depending on the changes you're looking to make.

In my perspective, there are a few options:
1. Update `sndev` to provide a way to opt-out of spinning up certain docker services via `./sndev start`. This could be done by adding a new CLI flag like `--exclude <service>` so the user could execute e.g. `./sndev start --exclude opensearch`. It could also be done by supporting some `.gitignore`d configuration file that is consulted when `./sndev start` is executed. 

2. Similar to number 1 above, `sndev` could be reduced to the minimum required services by default (`app`, `db`?), and developers could opt-in to other services, based on what they're trying to do. I'm not sure if this is better than having to opt-out (option 1).

3. As @huumn mentioned in the related comment thread (linked below), there could also be some cleanup performed on the `depends_on` entries in `docker-compose.yml` to reduce which services are automatically spun up, if you were to just spin up the `app` service (as an example) using `./sndev start app`.  I like the idea of a config-driven approach more, I think, but this option certainly provides a workable solution. I guess it comes down to whether you want `depends_on` to reflect a prod-like configuration, or just the bare minimums for a functional development environment.

Enumerated above.

**Additional context**
[Context comments where this was discussed](https://stacker.news/items/466556/r/WeAreAllSatoshi?commentId=467727)

huumn:
fixes #928 by using [docker compose profiles](https://docs.docker.com/compose/profiles/).

By default all will be run. If you only want to run specific services for specific features, set `COMPOSE_PROFILES` to use one or more of `minimal|images|search|payments`. To only run mininal services without images, search, or payments:

```txt
$ COMPOSE_PROFILES=minimal ./sndev start
```

To run with images and payments services:

```txt
$ COMPOSE_PROFILES=images,payments ./sndev start
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced the development environment setup by introducing the ability to run specific services (`minimal`, `images`, `search`, `payments`) using `COMPOSE_PROFILES`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 Launch a fully featured SN development environment in a single command.
 
-```txt
+```sh
 $ ./sndev start
 ```
 
@@ -38,25 +38,32 @@ Go to [localhost:3000](http://localhost:3000).
 
 Start the development environment
 
-```txt
+```sh
 $ ./sndev start
 ```
 
-By default all will be run. If you only want to run specific services for specific features, set `COMPOSE_PROFILES` to use one or more of `minimal|images|search|payments`. To only run mininal services without images, search, or payments:
+By default all services will be run. If you want to exclude specific services from running, set `COMPOSE_PROFILES` to use one or more of `minimal|images|search|payments`. To only run mininal services without images, search, or payments:
 
-```txt
+```sh
 $ COMPOSE_PROFILES=minimal ./sndev start
+```
+
+Or, as I would recommend:
+
+```sh
+$ export COMPOSE_PROFILES=minimal
+$ ./sndev start
 ```
 
 To run with images and payments services:
 
-```txt
+```sh
 $ COMPOSE_PROFILES=images,payments ./sndev start
 ```
 
 View all available commands
 
-```txt
+```sh
 $ ./sndev help
 
                             888
--- a/README.md
+++ b/README.md
@@ -42,6 +42,18 @@ Start the development environment
 $ ./sndev start
 ```
 
+By default all will be run. If you only want to run specific services for specific features, set `COMPOSE_PROFILES` to use one or more of `minimal|images|search|payments`. To only run mininal services without images, search, or payments:
+
+```txt
+$ COMPOSE_PROFILES=minimal ./sndev start
+```
+
+To run with images and payments services:
+
+```txt
+$ COMPOSE_PROFILES=images,payments ./sndev start
+```
+
 View all available commands
 
 ```txt
--- a/sndev
+++ b/sndev
@@ -74,12 +74,30 @@ EOF
     fi
   fi
 
-  if [ $# -eq 0 ]; then
-    docker__compose up --build
-    exit 0
+  profile=''
+  args='--build'
+  while test $# -gt 0; do
+    case "$1" in
+      --profile)
+        profile="$profile --profile $2"
+        shift
+        shift
+        ;;
+      *)
+        args="$args $1"
+        shift
+        ;;
+    esac
+  done
+
+
+  if [ -z "$profile" ]; then
+    if [ -z "$COMPOSE_PROFILES" ]; then
+      profile="--profile images --profile search --profile payments"
+    fi
   fi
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -42,6 +42,18 @@ Start the development environment
 $ ./sndev start
 ```
 
+By default all will be run. If you only want to run specific services for specific features, set `COMPOSE_PROFILES` to use one or more of `minimal|images|search|payments`. To only run mininal services without images, search, or payments:
+
+```txt
```
SatsAllDay, line 47:
```suggestion
```bash
```

README.md
```diff
@@ -42,6 +42,18 @@ Start the development environment
...
+$ COMPOSE_PROFILES=minimal ./sndev start
+```
+
+To run with images and payments services:
+
+```txt
```
SatsAllDay, line 53:
```suggestion
```bash
```

sndev
```diff
@@ -9,7 +9,11 @@ docker__compose() {
     exit 0
   fi
 
-  CURRENT_UID=$(id -u) CURRENT_GID=$(id -g) command docker compose --env-file .env.development "$@"
+  if [ -z "$COMPOSE_PROFILES" ]; then
+    COMPOSE_PROFILES="images,search,payments"
+  fi
+
+  CURRENT_UID=$(id -u) CURRENT_GID=$(id -g) COMPOSE_PROFILES=$COMPOSE_PROFILES command docker compose --env-file .env.development "$@"
```
SatsAllDay, line 16: How does `minimal` work here? I don't see `minimal` referenced in the changes to `docker-compose.yml`, so I assume that means any services with no profiles are started, and since no services are tagged with `minimal`, then you just get those without profiles at all?

SatsAllDay (commented):
Did you intentionally not apply a profile to `worker`? So that `COMPOSE_PROFILES=minimal` maps to `db`, `app`, `worker`? I am good with that, just wanted to confirm since in the other discussion we identified that `worker` is not truly required for a minimum setup.

I was also going to say that I thought it would be nice to bake this support into `sndev` natively, but I think using the `COMPOSE_PROFILES` env var which is already natively supported by docker compose provides a cleaner, more scalable solution long term, so I am good with it. I guess the alternative would be to supply the same string e.g. `payments,images` to `./sndev start` as an arg e.g. `./sndev start payments,images`. It just depends on whether you want to go with ENV VARs or CLI args for `./sndev` long-term, I suppose.

Disclaimer: I have not pulled this to try it out locally, but the changes do look good to me.

SatsAllDay:
> I was also going to say that I thought it would be nice to bake this support into sndev natively, but I think using the COMPOSE_PROFILES env var which is already natively supported by docker compose provides a cleaner, more scalable solution long term, so I am good with it. I guess the alternative would be to supply the same string e.g. payments,images to ./sndev start as an arg e.g. ./sndev start payments,images. It just depends on whether you want to go with ENV VARs or CLI args for ./sndev long-term, I suppose.

Looking at the commits, I see you first explored making this an argument-driven config, and opted for the env var approach for simplicity. All good then!

sndev
huumn: Exactly. It's just a way to communicate that you don't want all or any other profile. Because I want the default to be all.

huumn:
> Looking at the commits, I see you first explored making this an argument-driven config, and opted for the env var approach for simplicity. All good then!

It was a bit of a nightmare trying to manipulate command line args in a posix way. This setting also doesn't seem like something anyone would want to type out every time they called start/stop/status/etc. 

If I were to use this I'd probably just export it. 

```bash
export COMPOSE_PROFILES=minimal
./sndev start
./sndev stop
```

Thanks for the review!

PR #948, opened 2024-03-20
Add MailHog container to provide login with email without extra config

felipebueno:
EDIT: The GIF is a little bit weird... I uploaded it to Imgur: https://imgur.com/a/mWWqKKR

mailhog

## Summary by CodeRabbit

- **Documentation**
	- Added instructions for email testing using MailHog.
	- Included guidance on using Sendgrid by creating an account.


huumn:
Cool! Please be sure to use https://docs.docker.com/compose/profiles/ so this can be prevented from running should someone want. See the README and our other uses of profiles to see how profiles are being used.

felipebueno:
I didn't know what would be the proper profile name for the MailHog container so I went with ```email```

Let me know if you have a better name

README.md
```diff
@@ -61,6 +61,12 @@ To run with images and payments services:
 $ COMPOSE_PROFILES=images,payments ./sndev start
 ```
 
+To run with the option to sign up/login using [MailHog](#MailHog):
+
+```sh
+$ COMPOSE_PROFILES=email ./sndev start
+```
+
```
huumn, line 69: This will run by default so I think adding this to the README is a bit confusing. We already suggest how to start only specific resources above.

huumn (changes requested):
Looking good! 

The only other thing missing afaict is updating this line https://github.com/stackernews/stacker.news/blob/4b4a5361ef29fd0c51325a1e8a44d20b8195b019/sndev#L13

So that the email profile is started by default.

README.md
felipebueno: Yeah, makes sense. I'm going to remove it

felipebueno:
> Looking good!
> 
> The only other thing missing afaict is updating this line
> 
> https://github.com/stackernews/stacker.news/blob/4b4a5361ef29fd0c51325a1e8a44d20b8195b019/sndev#L13
> 
> So that the email profile is started by default.

Done!

README.md
felipebueno: eef9b3de4d0c0ae3dbab9671a81737cbef108332

README.md
```diff
@@ -325,6 +325,19 @@ You can login to test features like posting, replying, tipping, etc with `./snde
 
 ### Login with Email
 
+#### MailHog
+
+- The app is already prepared to send emails through [MailHog](https://github.com/mailhog/MailHog) so no extra configuration is needed
+- Click "sign up" and enter any email address (remember, it's not going anywhere beyond your workstation)
+- Access MailHog's web UI on http://localhost:1025
```
huumn, line 332: I think you mean to say port `8025`.

`1025` is the smtp port.

docker-compose.yml
```diff
@@ -377,6 +377,23 @@ services:
...
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock:ro
+  mailhog:
+    image: mailhog/mailhog:latest
+    container_name: mailhog
+    profiles:
+      - email
+    restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8025"]
```
huumn, line 387: This healthcheck fails for me because the container does not have curl in its PATH.

huumn (changes requested):
This is almost there. One typo and the healthcheck needs to pass.

README.md
felipebueno: oops. Fixed!

Now I'm going to fix the healthcheck

docker-compose.yml
felipebueno: Fixed!

I changed it to use wget instead. I checked and it's available

image

huumn:
Works great! I'll merge now. Where would you like your sats sent?

felipebueno:
> PS In the future please re-request review after you've made the suggested changes. Otherwise I'll assume it isn't ready.

Oh, I forgot I had to do that. Will do

> Works great! I'll merge now. Where would you like your sats sent?

Awesome!

You can send the sats here:

lnbc1pnq90rzpp5236n8kknq268sgg5p0jpf33e4k7djmv0wh077ym5qsm8l5c37tlsdqswdhzqurjypekzarncqzzsxqyz5vqsp52z4kw6mhz6ln8skf3ytwdufej654t7f3jl7r7dnxfprq5edq5wzq9qyyssq0aqfm9mpxy3z8pvttgd6dvdtvydd0a5y0wla6wgwarlpnejy2krj8xx7ed8ep75n7yqlq32jma96yvvs7lmyu5pjzsu7ln9e4qe952cpdj9pz2

huumn:
Do you have a static payment method I can reuse and list in awards.csv?

felipebueno:
> awards.csv

Sure!

You can use my main SN account:

```felipe@stacker.news```

:)

PR #960, opened 2024-03-24
Gracefully handle errors when fetching lnurlp .well-known info for ln addr withdrawals

### LINKED ISSUE #735: Improve the error message when LNURL Pay request fails due to reasons
When trying to withdraw to an invalid Lightning Address, it seems the HTTP responses from some servers are not handled well by SN, when the response is not .json or the response is not one that SN is already expecting.

Examples:

Correct, expected behavior:

Lightning Address for an invalid domain: test@totally123invalid456domain.com  <-- SN flash message: "fetch failed"
https://totally123invalid456domain.com/.well-known/lnurlp/test   

Lightning Address for an lnurl pay server, but for an invalid user: notk00b@stacker.news  <-- SN flash message: "user @notk00b does not exist"
https://stacker.news/.well-known/lnurlp/notk00b   <-- HTTP/1.1 400 Bad Request [w/ application/body response, and message is the value for the "reason" key]

Undesired (?) behavior:

Lightning Address that produces an unexpected JSON response: test123@bitrefill.me    <-- SN flash message: "Error message not found."
https://bitrefill.me/.well-known/lnurlp/test123   <-- HTTP/1.1 400 Bad Request [w/ application/body response, but message is the value for the "**message**" key]

Lightning address that produces an HTML response: test@gmail.com  <-- SN flash message: "Unexpected token < in JSON at position 0"
https://gmail.com/.well-known/lnurlp/test   <-- HTTP/1.1 404 Not Found [w/ text-html response]

SatsAllDay:
Closes #735 

If `fetch` or `req.json` fails, catch those errors and return a default error to the user

If the res payload indicates error but doesn't return a `reason`, also return the same default error message to the user

Sample shown in the UI for various error reasons:
image

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for fetching Lightning Address metadata, ensuring more robust responses to fetch failures and JSON parsing errors.

### DIFF
--- a/lib/lnurl.js
+++ b/lib/lnurl.js
@@ -33,10 +33,19 @@ export async function lnAddrOptions (addr) {
     // support HTTP and HTTPS during development
     protocol = process.env.PUBLIC_URL.split('://')[0]
   }
-  const req = await fetch(`${protocol}://${domain}/.well-known/lnurlp/${name}`)
-  const res = await req.json()
+  const unexpectedErrorMessage = `An unexpected error occurred fetching the Lightning Address metadata for ${addr}. Check the address and try again.`
+  let res
+  try {
+    const req = await fetch(`${protocol}://${domain}/.well-known/lnurlp/${name}`)
+    res = await req.json()
+  } catch (err) {
+    // If `fetch` fails, or if `req.json` fails, catch it here and surface a reasonable error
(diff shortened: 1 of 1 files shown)


huumn:
Very nice!

PR #961, opened 2024-03-24
add `node_modules` volume to decouple node_modules dir between host and containers

### LINKED ISSUE #959: worker docker container fails to start in local dev
**Description**
When running the local docker setup, I've found that the worker container fails to start, citing the following issue:
```
node:internal/process/esm_loader:97
    internalBinding('errors').triggerUncaughtException(
                              ^
Error: The package "@esbuild/linux-x64" could not be found, and is needed by esbuild.
If you are installing esbuild with npm, make sure that you don't specify the
"--no-optional" or "--omit=optional" flags. The "optionalDependencies" feature
of "package.json" is used by esbuild to install the correct binary executable
for your current platform.
    at generateBinPath (/app/node_modules/esbuild/lib/main.js:1899:15)
    at esbuildCommandAndArgs (/app/node_modules/esbuild/lib/main.js:1969:33)
    at ensureServiceIsRunning (/app/node_modules/esbuild/lib/main.js:2133:25)
    at startSyncServiceWorker (/app/node_modules/esbuild/lib/main.js:2336:19)
    at Object.<anonymous> (/app/node_modules/esbuild/lib/main.js:2377:3)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
```

**Steps to Reproduce**
`COMPOSE_PROFILES=minimal ./sndev start`

**Expected behavior**
Worker container should start

**Screenshots**
N/A

**Logs**
```
node:internal/process/esm_loader:97
    internalBinding('errors').triggerUncaughtException(
                              ^
Error: The package "@esbuild/linux-x64" could not be found, and is needed by esbuild.
If you are installing esbuild with npm, make sure that you don't specify the
"--no-optional" or "--omit=optional" flags. The "optionalDependencies" feature
of "package.json" is used by esbuild to install the correct binary executable
for your current platform.
    at generateBinPath (/app/node_modules/esbuild/lib/main.js:1899:15)
    at esbuildCommandAndArgs (/app/node_modules/esbuild/lib/main.js:1969:33)
    at ensureServiceIsRunning (/app/node_modules/esbuild/lib/main.js:2133:25)
    at startSyncServiceWorker (/app/node_modules/esbuild/lib/main.js:2336:19)
    at Object.<anonymous> (/app/node_modules/esbuild/lib/main.js:2377:3)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
```

**Environment:**

**Additional context**
I've done some research and the issue appears to be caused by the following. We mount the local copy of `node_modules` into our docker containers. This means we share `node_modules` between our host machine and the docker containers. When one runs `npm install` on the host machine, certain dependencies like `@esbuild` will install based on the host platform. In my case, `darwin` versions are installed, since I am running on macOS on an intel chip. However, since the docker container is based on the linux platform, when you go to invoke certain programs like `worker`, for example, the node runtime will look for different versions of dependencies that are platform dependent.

A hacky solution is to force install the platform-specific version of the package you need for the docker container to work i.e. `npm install --force @esbuild/linux-x64`. This isn't ideal since it adds an entry to `package.json` and requires anyone who runs on a different platform to run `npm install` with `--force` to succeed.

A better solution is probably to install the platform-specific dependency within the docker container. I am trying that out locally.

SatsAllDay:
Closes #959 

This PR removes the dependency on `tsx` from `package.json` and `package-lock.json`, which pairs to the platform upon which the install is performed. It instead runs the worker script using `npx tsx` which installs within the docker container's context, so the right platform is guaranteed.


huumn:
We run `npm run worker` in prod, so I'm hesitant to change that. I responded in the issue. I think we might find a more generic solution.

SatsAllDay:
Closing based on discussion [here](https://github.com/stackernews/stacker.news/pull/964#issuecomment-2018654832)

PR #962, opened 2024-03-24
More zap undo fixes

ekzyis:
I am currently torturing myself with zap undos enabled in prod. I think zap undos can be a nice feature but we're not there yet.[^1]

This PR fixes some existing issues with it:

toasts due to zaps to the same item are not merged if not same method is used (default vs custom amount):

https://github.com/stackernews/stacker.news/assets/27162016/2e947122-d6ff-44e8-97c1-134361945630

_fixed in 24a89135_

progress bar jumped if an underlying toast finished:

https://github.com/stackernews/stacker.news/assets/27162016/66f6f433-4931-4b01-a7b2-7a0024ded9ee

_fixed in 3c6bf3af_

custom zap modal is always closed after any custom zap is finished. this can get REALLY annoying if you use custom zaps a lot:

https://github.com/stackernews/stacker.news/assets/27162016/960ba7e2-6f5a-43ef-8fee-6dc0385a3939

_not fixed yet_

[^1]: For example, it would make more sense to make the setting numeric so the undo flow would only get triggered if your zap is above X sats. I think that would make this feature a lot more useful, less annoying and hence more used. I will do this in another PR though, it should only require very few changes.


huumn:
Just a heads up to be *extra* careful with causing the entire component tree to rerender in by changing the value of context providers, eg make sure your hook deps are airtight. Currently, both the modal provider and toast provider each cause an additional rerender on every page load according to the React profiler.

PR #964, opened 2024-03-24
reinstall tsx without node_modules present

### LINKED ISSUE #959: worker docker container fails to start in local dev

huumn:
My suspicion is this fixes #959 (assuming tsx is the only issue).

There's a bug in npm https://github.com/npm/cli/issues/4828 that only adds the platform `npm install` is run on to `package-lock.json` when `node_modules` is present. I can at least confirm this change resulted in all other platform dependencies being added to `package-lock.json`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2353,6 +2353,51 @@
       "integrity": "sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==",
       "dev": true
     },
+    "node_modules/@esbuild/android-arm": {
+      "version": "0.18.20",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm/-/android-arm-0.18.20.tgz",
+      "integrity": "sha512-fyi7TDI/ijKKNZTUJAQqiG5T7YjJXgnzkURqmGj13C6dCqckZBLdl4h7bkhHt/t0WP+zO9/zwroDvANaOqO5Sw==",
+      "cpu": [
+        "arm"
+      ],
+      "optional": true,
+      "os": [
+        "android"
+      ],
+      "engines": {
+        "node": ">=12"
+      }
+    },
+    "node_modules/@esbuild/android-arm64": {
+      "version": "0.18.20",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm64/-/android-arm64-0.18.20.tgz",
+      "integrity": "sha512-Nz4rJcchGDtENV0eMKUNa6L12zz2zBDXuhj/Vjh18zGqB44Bi7MBMSXjgunJgjRhCmKOjnPuZp4Mb6OKqtMHLQ==",
+      "cpu": [
+        "arm64"
+      ],
+      "optional": true,
+      "os": [
+        "android"
+      ],
+      "engines": {
+        "node": ">=12"
+      }
+    },
+    "node_modules/@esbuild/android-x64": {
+      "version": "0.18.20",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-x64/-/android-x64-0.18.20.tgz",
+      "integrity": "sha512-8GDdlePJA8D6zlZYJV/jnrRAi6rOiNaCC/JclcXpB+KIuvfBN4owLtgzY2bsxnx666XjJx2kDPUmnTtR8qKQUg==",
+      "cpu": [
+        "x64"
+      ],
+      "optional": true,
+      "os": [
+        "android"
+      ],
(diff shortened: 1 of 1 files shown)


SatsAllDay:
Will test it locally later tonight 👍

SatsAllDay (approved):
OK, yes, this does appear to fix the issue 🎉 

I was making some headway with the `node_modules` volume approach discussed [here](https://github.com/stackernews/stacker.news/issues/959#issuecomment-2016902716), though I don't think it was going to solve this issue on its own.

If you're still interested in decoupling the `node_modules` install within the docker containers from that on the host machine, I can repurpose #961 to check that in. Let me know, and thank you for working on a fix for this!

huumn:
> If you're still interested in decoupling the node_modules install within the docker containers from that on the host machine, I can repurpose https://github.com/stackernews/stacker.news/pull/961 to check that in.

Would it still be useful to do that with this issue solved? I guess it might allow us to cache the modules during the container build. If you've figured it out already, I'd be happy to merge it, but if it's too tricky maybe we just move on.

huumn:
If it adds extra steps to getting things like intellisense I don't think it's worth it. Learning experience it is I guess.

I'll send you sats for the help on the issue and stuff though.

Issue #199, opened 2022-09-28
Show invitees to inviter


huumn:
https://stacker.news/items/74801

it's sufficient to just do this in the notification

PR #969, opened 2024-03-25
[WIP] feat: enchance markdown links

### LINKED ISSUE #93: Smarter URL detection, not every @ in the middle makes it an e-mail address

benalleng:
Adds an enhancement to markdown links because unfortunately remark-gfm is not very exact and will pick up on almost any link that includes an @ and try to turn it into an email, although a niche problem stackers like perfection

As of now all the links I've gone through work nicely, though I did bring up a question in the issue that we could add some github-like functionality to let typing `#199` act the same way it does on github if we want item numbers to be something the users interact with more, in the future we might also add the search like in mentions but i digress

here is a picture of some links you can try and test
image

Closes #93 

### TODO

- clean up regex
- seeing as i removed the larger remark-gfm I need to see all the possible matches that got broken
- I think I'd prefer if `example.com` redirected to an external link as it redirects to `stacker.news/items/example.com` ATM

@huumn I would like a quick sanity check if we think moving away from gfm and using some custom remark + individual remark modules stuff is worth it? (Or just a sanity check on this PR in general, my brain has been officially fried by my docker hell yesterday)

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -2,10 +2,13 @@ import styles from './text.module.css'
 import ReactMarkdown from 'react-markdown'
 import YouTube from 'react-youtube'
 import gfm from 'remark-gfm'
+import remarkFootnotes from 'remark-footnotes'
 import { LightAsync as SyntaxHighlighter } from 'react-syntax-highlighter'
 import atomDark from 'react-syntax-highlighter/dist/cjs/styles/prism/atom-dark'
 import mention from '@/lib/remark-mention'
 import sub from '@/lib/remark-sub'
+import email from '@/lib/remark-email'
+import url from '@/lib/remark-url'
 import React, { useState, memo, useRef, useCallback, useMemo, useEffect } from 'react'
 import GithubSlugger from 'github-slugger'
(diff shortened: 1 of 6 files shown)


huumn:
> I would like a quick sanity check if we think moving away from gfm and using some custom remark + individual remark modules stuff is worth it?

In general, I'd discourage doing something that has broad implications to fix a specific issue.

I welcome fixing this ticket with a specific change, in an ideally minimal way, first. 

After you've submitted a few other PRs and/or opened issues for related things, and your feet are a little more wet, then I think *proposing* something more ambitious with adequate specification is warranted.

huumn:
> very enlightening on the inner workings of sn

Great! Once you have a little more of knowledge it should be easier to be more surgical with changes. 

We need general changes too, but it's going to be difficult without more experience on the project. So, to start, I recommend thinking about changes like "what's the most minimal way of doing this that closes the issue." If something minimal isn't desirable or achievable, I recommend proposing the non-minimal thing for feedback before starting.

Issue #884, opened 2024-02-27
Enable readerView compatibility


h0dlr:
**Is your feature request related to a problem? Please describe.**
Yes, in the majority of browsers I've tested, accessing the contents on readingView is not compatible with the HTML structure generated by Stacker.News

**Describe the solution you'd like**
I believe it's just question to use the appropriated HTML tags to format the article: ie. `<article>`

**Describe alternatives you've considered**
The alternative is to install an extension provide data to third parties about the contents I'm reading!

**Additional context**
N/A

huumn:
H0dlr do you have a lightning address I can send the sats for this issue to?

h0dlr:
I'm glad you implemented this feature, thank you very much! you can send to `0xe14b9b5981c729a3@ln.tips`

h0dlr:
I also found the HTML structure isn't still right on Chrome and Safari and the _Reader View_ still unavailable

This has helped me better understand https://stackoverflow.com/questions/30730300/optimize-website-to-show-reader-view-in-firefox

Just submitted https://github.com/stackernews/stacker.news/pull/1005 to add a `<p>` and including the title inside the `<article>`

Have been not able to test

felipebueno:
Hmm I can´t reproduce the bug. What device are you using on your tests?

h0dlr:
Mac OS

PR #974, opened 2024-03-26
Enable readerView compatibility (issue #884)

felipebueno:
This PR implements the feature from issue https://github.com/stackernews/stacker.news/issues/884

fullItemContainer from ```div``` to ```article``` to make it compatible with browser's reader view

image

image

## Summary by CodeRabbit

- **Refactor**
	- Updated the HTML structure of the `TopLevelItem` component for better semantic clarity.

### DIFF
--- a/components/item-full.js
+++ b/components/item-full.js
@@ -140,7 +140,7 @@ function TopLevelItem ({ item, noReply, ...props }) {
       belowTitle={item.forwards && item.forwards.length > 0 && <FwdUsers forwards={item.forwards} />}
       {...props}
     >
-      <div className={styles.fullItemContainer} ref={textRef}>
+      <article className={styles.fullItemContainer} ref={textRef}>
         {item.text && <ItemText item={item} />}
         {item.url && <ItemEmbed item={item} />}
         {item.poll && <Poll item={item} />}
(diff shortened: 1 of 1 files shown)


benalleng:
Looks great on discussions! What happens to the content in polls and bounties?

felipebueno:
> Looks great on discussions! What happens to the content in polls and bounties?

Good question!

I just confirmed it. They stay the same:

image

image

h0dlr:
The stay the same because there's not enough content to trigger the readView. I assume these screenshots are on fireFox.

From what I've tested, the feature only works on FF and comments, however, are not loaded in ReadView. ReaderView in Chrome and Safari does not get enabled. I presume it is because of the html structure. `<article>` should be the fist tag after `<main>`

felipebueno:
@h0dlr 

>  I assume these screenshots are on fireFox

Hmm yeah, you are right, those screenshots are on firefox, both mobile and desktop

I just tried it on Chrome (on Linux) and it really doesn't get enabled

> <article> should be the fist tag after <main>

Hmm I think that would be a problem. There are many divs before the article tag. I'm not sure how to fix that

PR #980, opened 2024-03-26
Hash API keys with SHA-256 before storing them

ekzyis:
As reported by @SatsAllDay, I stored API keys in plaintext in #915. This is bad since before #915, database leaks could not result in account takeovers. With API keys in plaintext, this is the case.

This PR fixes this by only storing SHA-256 hashes of API keys.

Additionally, the API key is only shown once to the user (obviously, since we literally cannot show it again).

https://github.com/stackernews/stacker.news/assets/27162016/fd5b8844-ec60-4fee-a8d4-3ee842e7c6a6

## Summary by CodeRabbit

- **New Features**
	- Introduced an enhanced API key management system, including generation, deletion, and verification of API keys using hashes for improved security.
	- Added new modal design adjustments for better alignment and presentation.

- **Bug Fixes**
	- Fixed user authentication flow to directly query the database for user information using API key hashes, enhancing the reliability of user sessions.

- **Refactor**
	- Updated the API key field in the user model from a string to a boolean to reflect its enabled status.
	- Improved the modal component's styling for better user interface consistency.

- **Database Changes**
	- Implemented database schema changes to support API key hashing, including dropping the old `apiKey` column and introducing `apiKeyHash` with a unique index.

- **Style**
	- Enhanced global modal styles for improved visual coherence across the application.

### DIFF
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -47,7 +47,7 @@ async function authMethods (user, args, { models, me }) {
     twitter: oauth.indexOf('twitter') >= 0,
     github: oauth.indexOf('github') >= 0,
     nostr: !!user.nostrAuthPubkey,
-    apiKey: user.apiKeyEnabled ? user.apiKey : null
+    apiKey: user.apiKeyEnabled ? !!user.apiKeyHash : null
   }
 }
 
@@ -541,7 +541,14 @@ export default {
         throw new GraphQLError('you are not allowed to generate api keys', { extensions: { code: 'FORBIDDEN' } })
       }
 
-      const [{ apiKey }] = await models.$queryRaw`UPDATE users SET "apiKey" = encode(gen_random_bytes(32), 'base64')::CHAR(32) WHERE id = ${me.id} RETURNING "apiKey"`
+      // I trust postgres CSPRNG more than the one from JS
+      const [{ apiKey, apiKeyHash }] = await models.$queryRaw`
+      SELECT "apiKey", encode(digest("apiKey", 'sha256'), 'hex') AS "apiKeyHash"
+      FROM (
+        SELECT encode(gen_random_bytes(32), 'base64')::CHAR(32) as "apiKey"
+      ) rng`
+      await models.user.update({ where: { id: me.id }, data: { apiKeyHash } })
+
       return apiKey
     },
     deleteApiKey: async (parent, { id }, { models, me }) => {
@@ -549,7 +556,7 @@ export default {
         throw new GraphQLError('you must be logged in', { extensions: { code: 'UNAUTHENTICATED' } })
       }
 
-      return await models.user.update({ where: { id: me.id }, data: { apiKey: null } })
+      return await models.user.update({ where: { id: me.id }, data: { apiKeyHash: null } })
     },
     unlinkAuth: async (parent, { authType }, { models, me }) => {
       if (!me) {
--- a/api/typeDefs/user.js
+++ b/api/typeDefs/user.js
@@ -104,7 +104,7 @@ export default gql`
     github: Boolean!
     twitter: Boolean!
     email: String
-    apiKey: String
+    apiKey: Boolean!
   }
 
   type UserPrivates {
--- a/pages/api/graphql.js
+++ b/pages/api/graphql.js
@@ -56,8 +56,11 @@ export default startServerAndCreateNextHandler(apolloServer, {
     const apiKey = req.headers['x-api-key']
     let session
     if (apiKey) {
-      const sessionFieldSelect = { name: true, id: true, email: true }
-      const user = await models.user.findUnique({ where: { apiKey }, select: { ...sessionFieldSelect, apiKeyEnabled: true } })
+      const [user] = await models.$queryRaw`
+      SELECT id, name, email, "apiKeyEnabled"
+      FROM users
+      WHERE "apiKeyHash" = encode(digest(${apiKey}, 'sha256'), 'hex')
+      LIMIT 1`
       if (user?.apiKeyEnabled) {
         const { apiKeyEnabled, ...sessionFields } = user
         session = { user: { ...sessionFields, apiKey: true } }
--- a/prisma/migrations/20240326181608_api_key_sha256/migration.sql
+++ b/prisma/migrations/20240326181608_api_key_sha256/migration.sql
@@ -1,0 +1,16 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `apiKey` on the `users` table. All the data in the column will be lost.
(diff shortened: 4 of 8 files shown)


SatsAllDay:
I'll review this evening!

pages/api/graphql.js
```diff
@@ -56,8 +56,11 @@ export default startServerAndCreateNextHandler(apolloServer, {
     const apiKey = req.headers['x-api-key']
     let session
     if (apiKey) {
-      const sessionFieldSelect = { name: true, id: true, email: true }
-      const user = await models.user.findUnique({ where: { apiKey }, select: { ...sessionFieldSelect, apiKeyEnabled: true } })
+      const [user] = await models.$queryRaw`
+      SELECT id, name, email, "apiKeyEnabled"
+      FROM users
+      WHERE "apiKeyHash" = encode(digest(${apiKey}, 'sha256'), 'hex')
```
SatsAllDay, line 62: I suppose this is not susceptible to SQLI due to using `models.$queryRaw`, vs `models.$queryRawUnsafe`, right?

api/typeDefs/user.js
```diff
@@ -104,7 +104,7 @@ export default gql`
     github: Boolean!
     twitter: Boolean!
     email: String
-    apiKey: String
+    apiKey: Boolean!
```
SatsAllDay, line 107: What is the purpose of this boolean, just to indicate whether you've configured an API key?

api/resolvers/user.js
```diff
@@ -541,15 +541,22 @@ export default {
         throw new GraphQLError('you are not allowed to generate api keys', { extensions: { code: 'FORBIDDEN' } })
       }
 
-      const [{ apiKey }] = await models.$queryRaw`UPDATE users SET "apiKey" = encode(gen_random_bytes(32), 'base64')::CHAR(32) WHERE id = ${me.id} RETURNING "apiKey"`
+      // I trust postgres CSPRNG more than the one from JS
+      const [{ apiKey, apiKeyHash }] = await models.$queryRaw`
+      SELECT "apiKey", encode(digest("apiKey", 'sha256'), 'hex') AS "apiKeyHash"
+      FROM (
+        SELECT encode(gen_random_bytes(32), 'base64')::CHAR(32) as "apiKey"
+      ) rng`
```
SatsAllDay, line 549: what does `rng` do here?

SatsAllDay (approved):
Just a couple minor comments, but looks great!

api/typeDefs/user.js
ekzyis: Yes. With this, the frontend can tell if the button is disabled because API key generation was not enabled or because there already is an API key which they explicitly have to delete first.

pages/api/graphql.js
ekzyis: Yes, `$queryRaw` is safe to use with tagged templates.

It's not safe if called as a function with parentheses since you could have already injected SQL while building your string but that's not what I do here.[^1]

[^1]: https://www.prisma.io/docs/orm/prisma-client/queries/raw-database-access/raw-queries#in-queryraw-and-executeraw

api/resolvers/user.js
ekzyis: It's the alias for the subquery and I picked "rng" for "random number generator":

```
stackernews=# SELECT "apiKey", encode(digest("apiKey", 'sha256'), 'hex') AS "apiKeyHash"
stackernews-#       FROM (
stackernews(#         SELECT encode(gen_random_bytes(32), 'base64')::CHAR(32) as "apiKey"
stackernews(#       );
ERROR:  subquery in FROM must have an alias
LINE 2:       FROM (
                   ^
HINT:  For example, FROM (SELECT ...) [AS] foo.
```

SatsAllDay: Ah ok I wasn't thinking the alias was required. Thanks!

prisma/migrations/20240326181608_api_key_sha256/migration.sql
```diff
@@ -0,0 +1,16 @@
...
+
+  - You are about to drop the column `apiKey` on the `users` table. All the data in the column will be lost.
+  - A unique constraint covering the columns `[apiKeyHash]` on the table `users` will be added. If there are existing duplicate values, this will fail.
+
+*/
+-- DropIndex
+DROP INDEX "users.apikey_unique";
+
+-- AlterTable
+ALTER TABLE "users" DROP COLUMN "apiKey",
```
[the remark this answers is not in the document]
SatsAllDay: Yea I guess my point was, if users had generated keys, this would break their existing keys , which isn't ideal. Yes they can make a new one but their existing one would suddenly stop working.

It's probably more of an academic exercise to migrate the existing keys instead of just wiping them out, since the feature isn't widely adopted.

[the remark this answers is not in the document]
ekzyis: Yes, I definitely would have reconsidered this if at least one stacker which wasn't me generated new keys

huumn (approved):
This is a nit because most people won't use this but I wanted to flag it anyway.

The button + delete icon button are weird UX. I think it'd be better for there to be one button that changes depending on the state: disabled create, enabled create, delete.

I'm going to merge because polishing a UI for 3 people isn't a priority, but wanted to point this out.

ekzyis:
> The button + delete icon button are weird UX. I think it'd be better for there to be one button that changes depending on the state: disabled create, enabled create, delete.

Mhh, ok, noted!

> I'm going to merge because polishing a UI for 3 people isn't a priority, but wanted to point this out.

*2 people, the request from remindme was also me :eyes:

huumn:
> *2 people, the request from remindme was also me 👀

lol I didn't know! I was bummed we hadn't reserved that nym. Now it makes sense.

ekzyis:
Hey @coderabbitai, can you catch backwards-incompatible database migrations?

For example, a column was dropped in this PR.

styles/globals.scss
huumn, line 1: I think these changes made the modal invoice skeleton weird.

Screenshot 2024-04-16 at 2 24 33 PM

huumn: @ekzyis do you remember what they were for? I'd like to remove them but make sure I don't break what they were put in for.

ekzyis: Yes, I added screenshots in https://github.com/stackernews/stacker.news/pull/985 to explain what these changes were for.

This also caused a bug for the QR code scanner which I fixed in https://github.com/stackernews/stacker.news/pull/997

So I think it's a width issue since width is now set to `fit-content` for `.modal-body`

huumn: Okay, I'll make sure to qa those a little extra. Thanks!

PR #989, opened 2024-03-28
Limit scope of API Keys

### LINKED ISSUE #984: Limit scope of api keys to reduce account takeover risks
With API Keys being introduced, I thought about the potential ramifications if an API Key got into the wrong hands (not from a SN leak, but if the user accidentally exposed it).

Obviously there is always some risk, such as draining the SN wallet funds, etc. But maybe there's some APIs that we can restrict access to via API keys?

I think it might make sense to disallow certain API actions to be performed by API keys, most notably:
1. `unlinkAuth` - should an API Key be compromised, the attacker could use this API to remove existing auth methods, locking the account owner out of their account
2. Whatever APIs are used to register new auth methods, which when used with (1) above would allow a complete account takeover (I think)?
3. Since we're limited to a single API key right now, I don't think there's any reason to limit access to the `generateApiKey` and `deleteApiKey` APIs, since the attacker would basically be locking themself out if they deleted the only API key.

In the long term, following a model like OAuth scoping is probably a good idea, but I imagine we should not roll our own, or if we do, it is a much larger effort to:
1. Enumerate different scopes/roles
2. Assign APIs to each scope/role
3. Generically/abstractly enforce API Key scopes on each API 

This is more of a stop-gap/band-aid to give a little more security for us clumsy stackers ;) 

**Additional context**
N/A

As always, I'm open to input. Maybe I'm making a bigger deal out of this than is warranted.

SatsAllDay:
Fixes #984 

Disallow the following APIs when invoked via API Key:
* item.act (zap)
* create withdrawal
* unlink auth method
* link unverified email
* donate to rewards <-- This one wasn't originally discussed, but I thought it sort-of fit the description since it's a way to remove sats from the user's account. I would be ok leaving this in place, though. Open to discussion


SatsAllDay:
Quick update: I think I have most of this done. I am trying to dig into how to prevent adding new auth methods via API Key now.

ekzyis:
> @ekzyis I think this is ready for review now!

Going to review today because I didn't yesterday

ekzyis (changes requested):
Afaict, the motivation behind preventing some mutations is to prevent account takeovers and sending funds directly to an attacker if an API key was compromised while not making the API key useless. So posting etc. should still work and donating to rewards is fine since an attacker does not own the reward pool.

With this motivation in mind, I went over all mutations by searching for `extend type Mutation` and think the following mutations should also be blocked:

- `createInvite`: attacker can invite themself with a lot of sats attached
- `upsertWalletLND`: attacker can configure autowithdrawal to LND and withdraw all current and future funds like this
- `upsertWalletLNAddr`: same as above

_you can catch both current and future autowithdrawals by doing the check in `upsertWallet`_

> As far as I can tell, the only "auth method" that can be added via API is the unverified email. All the rest seem to depend on an authenticated UI session, since we only check for the API Key in the graphql API page. If I am mistaken please let me know!

I _think_ you are right. [NextAuth](https://github.com/nextauthjs) verifies CSRF tokens on login and sign up and the same code is used to link another auth method.

_If_ I understand how this works correctly, the CSRF tokens are only given out via SSR to a browser. As you mentioned, the API key is only checked for the GraphQL API so one shouldn't be able to use API keys to trick SSR into giving out CSRF tokens to workaround this.

**_However_**, since I am not _entirely_ sure and I think linking another auth method is not a common use case for API keys anyway, I would simply block everything that is related to user authentication. This would also make sure that we don't get a false sense of security around API keys where we think account takeovers are no longer possible. So I would err on the side of caution and would fill any hole that might look like a possible account takeover (however much convolution might be required) in mutations in this PR at once.

Many words to simply say that I would also block following mutation:

- `createAuth`

Requesting changes because updating autowithdrawal settings should definitely be blocked via API key (considering the motivation to block anything at all).

SatsAllDay:
Thank you for the detailed feedback! It's clear that there are some features that I haven't explored, so I didn't think to block them with this PR. I'll push some updates!

SatsAllDay:
OK, I updated the PR to incorporate your feedback!

I think you can see it in the commit history on the PR, but I did actually have `createAuth` included in one of my original [commits](https://github.com/stackernews/stacker.news/pull/989/commits/77c2d7d82b9407fd1af9685ba445a3f0c17cd732), but I later changed my mind and [undid it](https://github.com/stackernews/stacker.news/pull/989/commits/7f32abc15f76bd91cc0fb189ae931807548b322c), because there was no usage of `me` in the `createAuth` resolver prior to my change, so I figured it was safe to not add the API Key assertion because it seemed that there was no association between the created auth challenge and any known user. I did include that change in my latest commit, just to cover our bases as you suggested. But I thought I'd at least explain my prior rationale.

Also, I shared a test script in the PR description. I have that just floating around in my local git repo. I know there isn't an existing pattern for tests in the repo right now, but I was wondering if there is any value in checking something like that in, even if it just serves as a reference? No worries either way, but I figured I'd ask.

ekzyis (approved):
> I think you can see it in the commit history on the PR, but I did actually have createAuth included in one of my original [commits](https://github.com/stackernews/stacker.news/pull/989/commits/77c2d7d82b9407fd1af9685ba445a3f0c17cd732), but I later changed my mind and [undid it](https://github.com/stackernews/stacker.news/pull/989/commits/7f32abc15f76bd91cc0fb189ae931807548b322c), because there was no usage of me in the createAuth resolver prior to my change, so I figured it was safe to not add the API Key assertion because it seemed that there was no association between the created auth challenge and any known user. I did include that change in my latest commit, just to cover our bases as you suggested. But I thought I'd at least explain my prior rationale.

Oh, you are right, I did not notice that (your previous commit and that it didn't use `me`) :eyes: In that case, I don't really mind it anymore. Sorry for requesting a change regarding this! You can remove it or keep it now.

> Also, I shared a test script in the PR description. I have that just floating around in my local git repo. I know there isn't an existing pattern for tests in the repo right now, but I was wondering if there is any value in checking something like that in, even if it just serves as a reference? No worries either way, but I figured I'd ask.

The test script was helpful and I used it but I don't think we need to check it in :)

SatsAllDay:
> Oh, you are right, I did not notice that (your previous commit and that it didn't use me) 👀 In that case, I don't really mind it anymore. Sorry for requesting a change regarding this! You can remove it or keep it now.

No worries! I'll keep it in.

> The test script was helpful and I used it but I don't think we need to check it in :)

Sounds good!

PR #990, opened 2024-03-28
Add PR template

ekzyis:
With the checklist in this PR template, we should no longer be caught off by backwards incompatible changes like in #980 which cause downtime during deployments. The checklist is formulated intentionally in a way that if the box is not checked, the answer to the question is probably no.

This was partly inspired by the [PR template for the Alby browser extension](https://github.com/getAlby/lightning-browser-extension/blob/master/.github/pull_request_template.md). Are there other good templates that we could use as a reference?

I also added some examples of PRs that had good descriptions in my opinion. This list might be biased / opinionated though. Are there better examples? Are these examples useful?

This PR template also adds other useful information about the review process. Anything to add? For example, we could mention that we mark PRs as drafts ourselves when we request changes but we did not do this so far afaict. I think we simply mentioned to request a new review.

## Summary by CodeRabbit

- **Documentation**
	- Introduced a pull request template to standardize the descriptions for changes, enhancing clarity and consistency in project documentation.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,54 @@
+**Description**
+
+<!--
+
+A clear and concise description of what you changed and why.
+
+Bullet points can be enough.
+
+You can use the following PRs as inspiration:
+  - https://github.com/stackernews/stacker.news/pull/227 (feature)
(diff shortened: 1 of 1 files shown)


huumn:
Cool! I think this is stellar start

Issue #718, opened 2023-12-29
Title provided by user is overridden in link form


ekzyis:
> If creating a 'link' post, changing the link's auto-generated title, then moving away from the page; upon returning the title has reverted back to the original auto-generated one.
To recreate:
• Click on yellow 'post' button
• Choose 'link' button
• Paste a URL into the URL field
• Visibly change the title after it has been auto generated
• Click away your link post page
• Return back to the link post page that you were creating.
Problem:
This could prove frustrating for users who take their time tweaking a link's creation before posting it.

-- https://stacker.news/items/370068

> It's probably the function that fetches the title overriding the existing value.

-- https://stacker.news/items/370100

ekzyis:
_proof of bug_

https://github.com/stackernews/stacker.news/assets/27162016/72087b1b-b974-476f-915a-b1d3a60b89c9

ekzyis:
There is also a bug with the title query which uses `useLazyQuery`. It might be related to [this](https://github.com/apollographql/apollo-client/issues/9338) but `onCompleted` does fire in our case. It's just empty.

This can be fixed by setting the fetch policy to `network-only` but this seems like a bandaid. I will see how to fix this bug without tampering with the cache:


https://github.com/stackernews/stacker.news/assets/27162016/a896d2e1-e123-4eed-abdb-f06ac1abdf09

PR #991, opened 2024-03-28
Fix title from draft being dismissed (#718)

jp30566347:
Add a way to bypass the title generation query when the url change is from local storage (draft) and not user interaction.

## Summary by CodeRabbit

- **New Features**
	- Enhanced form input handling to recognize draft changes, improving the user experience for data auto-saved from local storage.
- **Refactor**
	- Improved logic in link form validation to accurately detect empty titles, especially for draft changes, ensuring more reliable form submissions.

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -453,7 +453,8 @@ function InputInner ({
         const isNumeric = /^[0-9]+$/.test(draft)
         const numericExpected = typeof field.value === 'number'
         helpers.setValue(isNumeric && numericExpected ? parseInt(draft) : draft)
-        onChange && onChange(formik, { target: { value: draft } })
+        // draft means change event is from local storage
+        onChange && onChange(formik, { target: { value: draft, draft: true } })
       }
     }
   }, [overrideValue])
--- a/components/form.js
+++ b/components/form.js
@@ -453,8 +453,7 @@ function InputInner ({
         const isNumeric = /^[0-9]+$/.test(draft)
         const numericExpected = typeof field.value === 'number'
         helpers.setValue(isNumeric && numericExpected ? parseInt(draft) : draft)
-        // draft means change event is from local storage
-        onChange && onChange(formik, { target: { value: draft, draft: true } })
+        onChange && onChange(formik, { target: { value: draft } })
       }
     }
   }, [overrideValue])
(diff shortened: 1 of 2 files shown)


jp30566347:
Suggestion fix for #718

ekzyis (changes requested):
This fixes that the draft value is overridden:

https://github.com/stackernews/stacker.news/assets/27162016/254c0217-a04a-4011-b1f8-c967f3563ab8

But it breaks fetching the title on page load:

https://github.com/stackernews/stacker.news/assets/27162016/16c50a35-3415-4705-ada3-93040f1cbb5e

The title is still fetched if the URL is changed after page load.

jp30566347:
Oh, I missed that requirement, I will take a closer look!

jp30566347:
Updated fix, I'm open to suggestions because it's a bit smelly.
The difficulty here is getting the draft value of the title from the onChange event of the url field.
Note that if there is no draft for a specific field, the onChange is not fired, so it's impossible from the parent component to determine if a draft exists or not. 

I was also wondering if we should regenerate a title (override existing one) if the url is changed? To me it sounds like it but right now it's not.

ekzyis:
> I was also wondering if we should regenerate a title (override existing one) if the url is changed? To me it sounds like it but right now it's not.

I agree with you. If you change the URL, the old title will most likely no longer make sense. So I think regenerating the title via fetch would make sense.

However, unless it's a regression (like the page load fetch), you don't need to care about it in this PR. Currently not deep in that part of the code right now so not sure if that is already a thing, will try to review the changes later today.

But if the code gets too messy, maybe we can relax requirements a bit. Never overriding draft values might be more important. Not sure yet though.

components/form.js
```diff
@@ -453,7 +453,8 @@ function InputInner ({
         const isNumeric = /^[0-9]+$/.test(draft)
         const numericExpected = typeof field.value === 'number'
         helpers.setValue(isNumeric && numericExpected ? parseInt(draft) : draft)
-        onChange && onChange(formik, { target: { value: draft } })
+        // draft means change event is from local storage
+        onChange && onChange(formik, { target: { value: draft, draft: true } })
```
huumn, line 457: It doesn't look like we are using `draft: true` anymore, are we?

jp30566347: I removed it. Good catch.

huumn:
I agree it's smelly but it's smelly in an obvious way at least. If this is something we want, it's either this or a new abstraction which this change doesn't justify alone.

I don't think we have this kind of field interdependence anywhere else.

PR #994, opened 2024-03-29
Wallet Logs

### LINKED ISSUE #823: Add user-exposed logs for attached wallets
Problem: currently we only display an indicator light, green for good and red for bad, on wallets to indicate the wallet health. While this is fine for healthy wallets, when the wallet isn't working, stackers can't see what's wrong. 

Solution: create a log for each attached wallet. These will need to work differently because some wallets are run on the server (receives) and some in the browser (sends), but the log should look the same in either case.

More background: most of the opaque problems with wallets happen when we connect to them and we don't report those errors. Errors with paying invoices are reported in `satistics` but errors otherwise aren't surfaced. We need to record and report these.

Implementation hints: For server-side receive wallets, create a `WalletLog` table and record connection errors/successes there. You'll also want to create a `autowithdrawWalletId` column on the `Withdrawl` table that replaces the `autowithdraw` column so invoice errors can also be displayed in the log. When on a wallet attach page display or link to the log for the wallet.

For client-side send wallets, you'll want to record the log in indexeddb or localstorage and similarly show the log on the wallet attach page.

---------

A step beyond MVP, color the indicator light orange if the wallet is having intermittent issues but is occasionally healthy.

ekzyis:
## Description

Close #823 

This PR adds a new React context: `WalletLoggerContext`

This context can be used with `useWalletLogger` to show wallet logs at /wallet/logs.

## Screenshots

<details>
<summary>desktop (72ca7417)</summary>

2024-04-03-175341_1920x1080_scrot

2024-04-03-175328_1920x1080_scrot

</details>

<details>
<summary>mobile (72ca7417)</summary>

localhost_3000_wallet_logs(iPhone SE) (4)

localhost_3000_wallet_logs(iPhone SE) (5)

</details>

<details>
<summary>UX video (72ca7417)</summary>

https://github.com/stackernews/stacker.news/assets/27162016/83c796a0-994f-41e0-943e-9af78c7f7858

</details>

### Sending via NWC

1. Clone and build [`nostr-wallet-connect-lnd`](https://github.com/benthecarman/nostr-wallet-connect-lnd)
2. Copy TLS certificate and admin macaroon from `stacker_lnd` container:
```
$ docker cp stacker_lnd:/home/lnd/.lnd/tls.cert tls.cert
$ docker cp stacker_lnd:/home/lnd/.lnd/data/chain/bitcoin/regtest/admin.macaroon admin.macaroon
```
3. Run `nostr-wallet-connect-lnd`:
```
$ RUST_LOG=info nostr-wallet-connect-lnd \
  --relay wss://relay.damus.io --network regtest \
  --lnd-host 127.0.0.1 --lnd-port 20010 \
  --cert-file stacker_lnd_tls.cert --macaroon-file stacker_lnd_admin.macaroon \
  --keys-file $(pwd)/keys.json
```

4. Copy and paste `nostr+walletconnect://` URL

### Receiving via LND

Use `stacker_lnd:10009` and `xxd -p -c0` on invoice.macaroon and tls.cert from container.

### 1. Where should link to full logs be?

It's currently at /wallet below `wallet history`:
<details>
<summary>click to show screenshot</summary>

localhost_3000_wallet_logs(iPhone SE) (2)

</details>

I thought about putting it into `attached wallets` since it's logs about attached wallets but that's too many clicks imo

**update:** Just saw that you mentioned it to be on the wallet attach page in #823:

> When on a wallet attach page display or link to the log for the wallet.

However, I planned to have all logs for every wallet at the same page with an optional dropdown filter to filter for specific wallets instead of having individual pages with no option to see all logs together. I think it's useful to see all interaction with attached wallets at one place (send + receive). When we have wallet fallbacks, that will be even more useful. What do you think @huumn?

**update:** I think I could _also_ link to /wallet/logs on every wallet attach page with a query param to immediately filter the logs by that wallet.

**update:** Added embedded logs and skipped adding a filter. You can just go to the individual page.

### 2. Show full payment hash in logs + more info like payment preimage?

The hash was truncated because of overflow issues and not sure how relevant the full hash is.

**update:** I think it's useful for debugging to be able to easily copy-paste payment hashes etc.

I now show the full hash + preimage in the logs with a horizontal scrollbar on x-overflow.

### 3 . Font size too small?

I am using `x-small` as the font size since it makes it easy to look at a lot of logs at the same time (good overview)

### 4. Expose secret in logs but don't persist it?

The original wallet logs screenshots show that the first and last 6 characters of the secret are shown in the log. I originally planned to not persist the secret in the logs so for loaded logs, the secret would show `******`.

**update:** I decided against that now since it's probably confusing. Secret is now never shown in the logs.

### 5. Delete logs on server for privacy reasons?

I shortly considered to delete all logs about a wallet if it was unattached but that might be confusing behavior.

I think we can keep this out of this PR and do it later.

---

TODO:

- [x] [feature] log statements for LNbits
- [x] [feature] fetch autowithdrawal logs from server
- [x] [feature] persist logs in local storage or IndexedDB | only store last 24h or more if unbounded client-side storage amount is actually not a problem
- [x] [feature] ~ability to sort logs by recent: if we store a lot of logs, I don't think one is interested in the oldest logs first.~ autoscroll is better
- [x] [refactor] fetching all logs from IndexedDB at once and then using `Array.filter` for pagination should make the code less complex

## Summary by CodeRabbit

- **New Features**
	- Introduced a new component for displaying log messages in a styled manner.
	- Added comprehensive logging functionality across various components and pages, including service workers and wallets.
	- Implemented a new `WalletLogs` component for real-time log following, filtering, and display in wallet settings.
	- Enhanced error handling and logging for wallet transactions and settings adjustments.
	- Added a new GraphQL query for fetching wallet logs.
- **Enhancements**
	- Improved the logging mechanism by introducing context providers for service workers and wallets.
	- Modified existing components to utilize the new logging contexts for better clarity and error reporting.
- **Bug Fixes**
	- Removed unnecessary validation check for `wss://` protocol in relay URLs to prevent false errors.
- **Refactor**
	- Updated various components to use the enhanced logging functionalities and contexts.
- **Style**
	- Added new styles for log message display, navigation, and table formatting.
- **Chores**
	- Updated the database schema to support wallet logs.
	- Enhanced the auto withdrawal process with better logging.

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -529,7 +529,7 @@ export default {
   }
 }
 
-const addWalletLog = async ({ wallet, level, message }, { me, models }) => {
+export const addWalletLog = async ({ wallet, level, message }, { me, models }) => {
   try {
     await models.walletLog.create({ data: { userId: me.id, wallet, level, message } })
   } catch (err) {
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -304,6 +304,20 @@ export default {
         cursor: history.length === LIMIT ? nextCursorEncoded(decodedCursor) : null,
         facts: history
       }
+    },
+    walletLogs: async (parent, args, { me, models }) => {
+      if (!me) {
+        throw new GraphQLError('you must be logged in', { extensions: { code: 'FORBIDDEN' } })
+      }
+
+      return await models.walletLog.findMany({
+        where: {
+          userId: me.id
+        },
+        orderBy: {
+          createdAt: 'asc'
+        }
+      })
     }
   },
   WalletDetails: {
@@ -415,10 +429,11 @@ export default {
       data.macaroon = ensureB64(data.macaroon)
       data.cert = ensureB64(data.cert)
 
+      const wallet = 'walletLND'
       return await upsertWallet(
         {
           schema: LNDAutowithdrawSchema,
-          walletName: 'walletLND',
+          walletName: wallet,
           walletType: 'LND',
           testConnect: async ({ cert, macaroon, socket }) => {
             const { lnd } = await authenticatedLndGrpc({
@@ -426,24 +441,35 @@ export default {
               macaroon,
               socket
             })
-            return await createInvoice({
-              description: 'SN connection test',
-              lnd,
-              tokens: 0,
-              expires_at: new Date()
-            })
+            await addWalletLog({ wallet, level: 'SUCCESS', message: 'connected to LND node' }, { me, models })
+            try {
+              const inv = await createInvoice({
+                description: 'SN connection test',
+                lnd,
+                tokens: 0,
+                expires_at: new Date()
+              })
+              await addWalletLog({ wallet, level: 'SUCCESS', message: 'created test invoice' }, { me, models })
+              return inv
+            } catch (err) {
+              await addWalletLog({ wallet, level: 'ERROR', message: `could not create test invoice: ${err.message || err.toString?.()}` }, { me, models })
+              throw err
+            }
           }
         },
         { settings, data }, { me, models })
     },
     upsertWalletLNAddr: async (parent, { settings, ...data }, { me, models }) => {
+      const wallet = 'walletLightningAddress'
       return await upsertWallet(
         {
           schema: lnAddrAutowithdrawSchema,
-          walletName: 'walletLightningAddress',
+          walletName: wallet,
           walletType: 'LIGHTNING_ADDRESS',
           testConnect: async ({ address }) => {
-            return await lnAddrOptions(address)
+            const options = await lnAddrOptions(address)
+            await addWalletLog({ wallet, level: 'SUCCESS', message: 'fetched payment details' }, { me, models })
+            return options
           }
         },
         { settings, data }, { me, models })
@@ -453,7 +479,23 @@ export default {
         throw new GraphQLError('you must be logged in', { extensions: { code: 'UNAUTHENTICATED' } })
       }
 
-      await models.wallet.delete({ where: { userId: me.id, id: Number(id) } })
+      const wallet = await models.wallet.findUnique({ where: { userId: me.id, id: Number(id) } })
+      if (!wallet) {
+        throw new GraphQLError('wallet not found', { extensions: { code: 'BAD_INPUT' } })
+      }
+
+      // determine wallet name for logging
+      let walletName = ''
+      if (wallet.type === 'LND') {
+        walletName = 'walletLND'
+      } else if (wallet.type === 'LIGHTNING_ADDRESS') {
+        walletName = 'walletLightningAddress'
+      }
+
+      await models.$transaction([
+        models.wallet.delete({ where: { userId: me.id, id: Number(id) } }),
+        models.walletLog.create({ data: { userId: me.id, wallet: walletName, level: 'SUCCESS', message: 'wallet deleted' } })
+      ])
 
       return true
     }
@@ -487,6 +529,14 @@ export default {
   }
 }
 
+const addWalletLog = async ({ wallet, level, message }, { me, models }) => {
+  try {
+    await models.walletLog.create({ data: { userId: me.id, wallet, level, message } })
+  } catch (err) {
+    console.error('error creating wallet log:', err)
+  }
+}
+
 async function upsertWallet (
   { schema, walletName, walletType, testConnect }, { settings, data }, { me, models }) {
   if (!me) {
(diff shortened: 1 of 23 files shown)


huumn:
> update: I think I could also link to /wallet/logs on every wallet attach page with a query param to immediately filter the logs by that wallet.

If this component is composable it should be trivial to add an accordian of the wallet logs to every wallet while still having a single superlog. That's the route I'd take at least.

ekzyis:
> If this component is composable it should be trivial to add an accordian (or a tab) of the wallet logs to every wallet while still having a single superlog on another page. That's the route I'd take at least.

Makes sense! I forgot about the use case of using the logs _while_ configuring the wallet. I mostly had in mind to use the logs for payment failures etc.

Current state (6368d4db):

https://github.com/stackernews/stacker.news/assets/27162016/a7d4b353-78ba-41e1-b1fd-d9aad3d8cb8b

Only thing left to do is to have logs for the autowithdrawal methods.

I was also thinking about having a dropdown to filter by wallet in the superlog but maybe that's MVP feature creep. It shouldn't be difficult to add but I probably shouldn't judge features by "how difficult they would be to add", lol.

**Additional context**

In cb1f1a66, I started to simply load all logs from IndexedDB at once (instead of only the last 5m) and remove pagination that looked like this:

2024-04-01-031541_1920x1080_scrot

> **Simply load all logs and remove navigation**
>
> I realized the code for navigation was most likely premature optimization which even resulted in worse UX:
Using the buttons to load logs from 5m, 1h, 6h ago sometimes meant that nothing happened at all since there were no logs from 5m, 1h, 6h ago.
That's why I added a time string as "start of logs" so it's at least visible that it changed but that looked bad so I removed it.
>
> But all of this was not necessary: I can simply load all logs at once and then the user can scroll around however they like.
>
> I was worried that it would be bad for performance to load all logs at once since we might store a lot of logs but as mentioned, that's probably premature optimization.
>
> WHEN a lot of logs are stored AND this becomes a problem (What problem even? Slow page load?), THEN we can think about this.
>
> If page load ever becomes slow because of loading logs, we could probably simply not load the logs at page load but only when /wallet/logs is visited.
>
> But for now, this works fine.

-- cb1f1a66, commit message

Deciding against any form of navigation and "performance optimization" and simply relying on a scrollbar in all cases made it possible to delete a lot of complicated code:

```
 components/logger.js | 51 +++++----------------------------------------------
 pages/wallet/logs.js | 30 ++++--------------------------
 2 files changed, 9 insertions(+), 72 deletions(-)
```

I am still a little bit worried what happens if one has a lot of logs (since storage is currently unbounded) but I am pretty sure it's not that big of a problem as I initially thought.

prisma/schema.prisma
```diff
@@ -157,6 +158,18 @@ model Wallet {
...
+  createdAt DateTime @default(now()) @map("created_at")
+  userId    Int
+  user      User   @relation(fields: [userId], references: [id], onDelete: Cascade)
+  wallet    String
+  level     LogLevel
+  message   String
+
+  @@index([userId, createdAt])
+}
+
```
ekzyis, line 172: I had to remove a foreign key to the wallet table since on errors, we don't create wallets.

Instead of doing some mental gymnastics to have a wallet that doesn't work in the wallet table, I simply abandoned the idea to have a link between the wallet table and the logs.

This also makes it possible to persist logs about wallets that have been detached.

api/resolvers/wallet.js
```diff
@@ -304,6 +304,20 @@ export default {
...
+      }
+
+      return await models.walletLog.findMany({
+        where: {
+          userId: me.id
+        },
+        orderBy: {
+          createdAt: 'asc'
+        }
+      })
```
ekzyis, line 320: We are wasting a lot of bandwidth per wallet save here since we always fetch all logs instead of passing the last id the frontend already knows.

But this is as simple as it can be and maybe it's as unnecessary as trying to optimize log performance by not loading all logs from IDB at once but have pagination but that made the code a lot more complicated (see https://github.com/stackernews/stacker.news/commit/cb1f1a66eda0a6378c1e17930b118ac62f028e19).

huumn:
I don't see any glaring issues and I haven't QA'd it. Is there anything you'd like feedback on? 

If not, can I assume you are happy and confident enough in it that I can just merge it and deploy it?

ekzyis:
> I don't see any glaring issues and I haven't QA'd it. Is there anything you'd like feedback on?
> 
> If not, can I assume you are happy and confident enough in it that I can just merge it and deploy it?

I see what you're doing :eyes: 

I am happy with the UX, the code is simple enough and I tested everything except logs for successful autowithdrawals.

But I also tested that now and had to fund my wallet for it. Then I noticed that there is a pretty serious render bug where `sendPayment` gets called over and over again. Probably related to context usage and what you mentioned with re-renders in another PR. Will fix that now then it should be good to go!

huumn:
> I see what you're doing 👀

🥸😆

ekzyis:
Mhh, 2b632cb6 fixes it but for a proper fix which does no longer make `provider` change just because I logged something, I need to split the context value for `WalletLoggerContext` in two, I think.

Then I can use `logger` in `NWCProvider` etc. without making it change because logging something changes the logs :thinking: 

I am also recreating the context value on every render currently but that's the smaller issue.

ekzyis:
Ok, 2b4e634 fixes that the provider context value changes due to logging.

Tested that logs still show up fine and logs are still appended fine.

We can merge this now.

PR #1008, opened 2024-04-02
Add founding date to territory header

ekzyis:
## Description

This PR adds the founding date to the territory header.

## Screenshots

_desktop:_

2024-04-03-011756_909x153_scrot

_mobile:_

localhost_3000_~Design(iPhone SE)

## Additional Context

I often wonder if a territory is new or I just haven't noticed it yet.

I think knowing which territories are new and which are old is interesting for all kind of reasons.

## Checklist

verified that `createdAt` is a required column in the database

### DIFF
--- a/components/territory-header.js
+++ b/components/territory-header.js
@@ -43,6 +43,8 @@ export function TerritoryInfo ({ sub }) {
           <Link href={`/${sub.user.name}`}>
             @{sub.user.name}<span> </span><Hat className='fill-grey' user={sub.user} height={12} width={12} />
           </Link>
+          <span>on </span>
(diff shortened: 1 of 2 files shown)


benalleng:
tACK! Works great! 
It could be cool in the future to add the "founding" item like we do for users with the "stacking since:" ui could be cool to tie in the items to this as well since we don't have sorting by oldest in territories

ekzyis:
Note for myself and anyone who reads this regarding this and #1026:

I must have not refreshed my browser since in the screenshot, I could select ~Design. So I didn't see that it would break the query to fetch all subs.

[^1]:

PR #1015, opened 2024-04-04
less confusing expression for the hint of when the autowithdraw will be initiated

### LINKED ISSUE #1010: Auto withdrawal help text confusing?
**Description**
The tooltip/helptext/whatever it's called for the auto withdrawal desired balance field doesn't seem to quite make sense?

> attempts to keep your balance no greater than 1000 sats of this amount

What does this mean?
**Steps to Reproduce**
Nav to wallet settings
Attach a LN Address

**Expected behavior**
More clear help text for this input field 

**Screenshots**
image

**Logs**
N/A

**Environment:**
N/A

**Additional context**

SouthKoreaLN:
## Description

As hint, instead of giving the relative change, it gives the absolute value for the threshold at which the transfer will be initiated. I use the original value + this relative change by multiplying by showing 11/10 instead of 1/10 of the threshold value chosen by the user.

closes #1010

## Screenshots

I tested in works sndev mode and updates dynamically when the user changes the desired value:

Screenshot 2024-04-04 at 5 00 15 PM

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted the calculation of the auto-withdrawal trigger threshold for accuracy.
	- Updated hint text to clarify auto-withdrawal trigger conditions based on the balance exceeding the threshold.

### DIFF
--- a/components/autowithdraw-shared.js
+++ b/components/autowithdraw-shared.js
@@ -43,9 +43,9 @@ export function AutowithdrawSettings ({ priority }) {
             name='autoWithdrawThreshold'
             onChange={(formik, e) => {
               const value = e.target.value
-              setSendThreshold(Math.max(Math.floor(value / 10), 1))
+              setSendThreshold(Math.max(Math.floor(value / 10 * 11), 1))
             }}
-            hint={isNumber(sendThreshold) ? `attempts to keep your balance no greater than ${numWithUnits(sendThreshold)} of this amount` : undefined}
+            hint={isNumber(sendThreshold) ? `will attempt auto-withdraw when your balance exceeds ${sendThreshold} sats` : undefined}
             append={<InputGroup.Text className='text-monospace'>sats</InputGroup.Text>}
           />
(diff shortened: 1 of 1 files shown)


SouthKoreaLN:
Actually, this will not work. It should only change the display value for the hint, not the actual threshold which is used in other part of the code. I'll take care of it.

SouthKoreaLN:
Now it should be correct.

huumn:
This is good to merge after the linting errors are fixed.

Where should I send the sats for this?

SouthKoreaLN:
south_korea_ln@stacker.news is fine. This bug fix almost feels too silly to get paid for though. At least it's helping me get familiar with the code base, next.js, etc... so hopefully I'll be able to contribute more sizeable next time.

huumn:
> This bug fix almost feels too silly to get paid for though.

No it's not. Every bit like this is something else we don't have to do. This kind of stuff goes a long way.

PR #1019, opened 2024-04-04
Fix effect dependencies

ekzyis:
## Description

I forgot to add some effect dependencies in #1014

## Checklist

## Summary by CodeRabbit

- **Refactor**
	- Updated triggering logic in `WalletLoggerProvider` for enhanced performance.
	- Improved parameter validation and relay URL handling in `NWCProvider`.

### DIFF
--- a/components/logger.js
+++ b/components/logger.js
@@ -228,7 +228,7 @@ const WalletLoggerProvider = ({ children }) => {
     const log = { wallet, level, message, ts: +new Date() }
     saveLog(log)
     setLogs((prevLogs) => [...prevLogs, log])
-  }, [setLogs, saveLog])
+  }, [saveLog])
 
   return (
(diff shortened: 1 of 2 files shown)


components/logger.js
```diff
@@ -228,7 +228,7 @@ const WalletLoggerProvider = ({ children }) => {
     const log = { wallet, level, message, ts: +new Date() }
     saveLog(log)
     setLogs((prevLogs) => [...prevLogs, log])
-  }, [setLogs, saveLog])
+  }, [saveLog])
```
ekzyis, line 231: State setters never change so they don't need  to be added to effect dependencies. Just wanted to make this dependency array consistent with other dependency arrays I am currently working on.

Probably too annoying to make it consistent everywhere (add state setters as dependencies or not). I think it's just opinion.

Issue #1023, opened 2024-04-05
sndev: 317: Syntax error: redirection unexpected


felipebueno:
**Screenshots**
![image](https://github.com/stackernews/stacker.news/assets/241271/3b8205c0-241a-47f5-9c2a-aec1ff2f04d2)

![image](https://github.com/stackernews/stacker.news/assets/241271/a5696e2f-3632-49cb-8596-9b42a285ceab)

**Environment:**
Ubuntu 22.04
zsh

PR #1024, opened 2024-04-05
Fix for #1023 sndev: 317: Syntax error: redirection unexpected

felipebueno:
Quick fix for https://github.com/stackernews/stacker.news/issues/1023

TBH, I'm not sure if this is the right fix because I don't know what's the purpose of that ```print ... cut``` line... I didn't put to much thought on that. I just changed to echo so I could start my env :smile: 

## Summary by CodeRabbit

- **Refactor**
	- Improved the method for extracting the remote name in the deployment script for enhanced reliability.

### DIFF
--- a/sndev
+++ b/sndev
@@ -314,7 +314,7 @@ __sndev__pr_track() {
 
   push=$(git remote -v | grep -e "$url .*push" | head -n1) || true
   if [ -n "$push" ]; then
-    remote=$(echo "$push" | cut -d' ' -f1)
+    remote=$(echo "$push" | cut -f 1)
   else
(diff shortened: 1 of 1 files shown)


huumn:
Nice catch. I forgot to [shell check](https://www.shellcheck.net/) it ... herestrings are a bash thing.

sndev
```diff
@@ -314,7 +314,7 @@ __sndev__pr_track() {
 
   push=$(git remote -v | grep -e "$url .*push" | head -n1) || true
   if [ -n "$push" ]; then
-    remote=$(printf "%s" $(cut -f 1 <<<"$push"))
+    remote=$(echo "$push" | cut -d' ' -f1)
```
huumn, line 317:
```suggestion
    remote=$(echo "$push" | cut -f 1)
```

Issue #727, opened 2024-01-02
Cannot login PWA (iOS) with e-mail or Nostr


Sebastix:
**Description**
I'm not able to login with the PWA installed on my iOS device. I've linked my Nostr and email as auth methods.
* E-mail does not work, because the login link will not open the PWA and I'm not able to past the login URL in the PWA.
* Login with Nostr is displaying the error message no nostr extension is found.

**Environment:**
If you only experience the issue on certain devices or browsers, provide that info.
 - Device: iPhone 12 Mini
 - OS: iOS
 - Browser: PWA (Safari)

Using Github Auth works for logging in (my least preferable way of loggin in).
The Lightning auth method also doesn't work on my side. When I copy-paste the lnurld url in my Thunderhub client I get a success message but SN is returning `Auth failed. Try again or choose a different method.`

huumn:
Thanks for creating an issue! Both the email (an ios only problem) and nostr requiring an extension are known to us but we haven't put them here to track yet afaik.

Lightning Auth should work. When it fails to link, it's usually because that lightning auth has been linked to another account.

Sebastix:
@huumn I tried to login with Lightning in the past, but is was not connected to my account so I disconnected it.
So yes, it could be linked to another SN account now (which I'm not using). How could I restore this?

ekzyis:
> Both the email (an ios only problem)

I wonder if we could and should fix that by not only having a [magic link as provided by Next-Auth](https://next-auth.js.org/providers/email) but also a login code which iOS users can copy and manually paste into the PWA.

huumn:
> I wonder if we could and should fix that by not only having a [magic link as provided by Next-Auth](https://next-auth.js.org/providers/email) but also a login code which iOS users can copy and manually paste into the PWA.

We can and we should yes. :)

> Not sure if this is fixable by us on mobile.

We can do this by sending a DM with a magic link!

ekzyis:
> We can do this by sending a DM with a magic link!

Oof, DMs on nostr, don't get me started haha. But yes, I think this could indeed be a very nice way to login without compromising too much on security.

Sebastix:
![image](https://github.com/stackernews/stacker.news/assets/582669/b8c49a3e-1e67-4e4b-82a3-f96f03d8b7c4)

Too bad that a Safari extension (Nostore above in the screenshot) is not working in a PWA for that

huumn:
[...]

PR #1030, opened 2024-04-05
Embed the login url as a qrcode into the html email

felipebueno:
I own quite a few sn accounts (😅) and I use the *login with email* feature **a lot** to log on many devices both on sn's pwa and on my native client and this is a feature I always thought would be really nice to have :)

This makes it much easier to logon on different devices without needing to setup my email account to click the link or copy & paste it somewhere else so I can use it.

## Screenshots
image

image

I covered the qrcode on purpose so I could take the screenshot

image

## Summary by CodeRabbit

- **New Features**
	- Enhanced the authentication process by including a QR code in the HTML email template for improved security and user convenience.

### DIFF
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -11,6 +11,8 @@ import { getToken } from 'next-auth/jwt'
 import { NodeNextRequest } from 'next/dist/server/base-http/node'
 import { schnorr } from '@noble/curves/secp256k1'
 import { notifyReferral } from '@/lib/webPush'
+import { renderToString } from 'react-dom/server'
+import QRCode from 'qrcode.react'
 
 /**
  * Stores userIds in user table
@@ -343,6 +345,8 @@ const newUserHtml = ({ url, site, email }) => {
   const mainBackgroundColor = '#ffffff'
   const buttonBackgroundColor = '#FADA5E'
 
+  const urlQrCode = renderToString(<QRCode value={url} renderAs='svg' size={256} />)
+
   return `
 <!doctype html>
 <html xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office">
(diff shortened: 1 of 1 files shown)


huumn:
amazing!

felipebueno:
> it adds a dominant visual element to something a new stacker will encounter... I think it might confuse people even if it's cool

Can't we make it less dominant then? Maybe reducing the size and/or placing it at the footer of the email

> few people will use

Today there might be few people and, by your comment, mostly on iOS devices but once SN grows outside of the US, many people will be using it on Android devices. Most of the world use Android

Also boomers like email :smile: 

I'm not sure if this is a convincing argument to add this feature though

ekzyis:
My thoughts:

It's possible to embed logos inside QR codes without breaking them. For example, see https://qr.io/. So it should be possible to do this here, too, and merge the login button with the QR code like this:

2024-04-09-031117_1600x900_scrot

So you either click the QR code to login or scan it. The text could therefore be:

> Click or scan this QR code

I think this would be a good compromise between "adds a dominant visual element to something a new stacker will encounter but few people will use" and it being a very nice and thoughtful feature for the ones who will use it. I think this would make it less of a dominant element that most people won't use since button + QR code is now the same element so everyone will use it.

felipebueno:
@ekzyis that's a good idea

What do you think, @huumn?

huumn:
I think separate is less confusing personally.

felipebueno:
> how would your parents react to this email?

With or without a qrcode, I'm certain my father would call me asking for help on "how to start start stackin' them sats" :smile:

felipebueno:
@huumn @ekzyis

What about something like this?

The yellow box is a qrcode. I just covered it because I don't know if it contains any info about my account

image

huumn:
I think what we want to do is as stated in this comment: https://github.com/nextauthjs/next-auth/issues/709#issuecomment-699496122

This would allow us to get magic link login on iOS PWA + give you your nice to have.

Basically, the link click says "yes, this is me" rather than "take me to the application and log me in."

---------------

In isolation, I wouldn't mind having this option, but people have so many issues with email logins on mobile already that I want to solve those issues first. So, we'll close this for now.

huumn:
@felipebueno if you'd like to solve this with either

1. an OTP (magic code rather than link) ... this is probably the easiest option
2. changing the meaning of the links

#727 is `difficulty:hard` with `priority:high`

PR #1032, opened 2024-04-06
Add a ctrl + click eventListener to open notifications in new tab

benalleng:
## Description

Adds a new `ctrl + click` event listener to the entire notification div to allow users to decide whether theywant to go directly to the notification or for it to be opened in a new tab

## Additional Context

There was some discussion about playing with `a` tags but only a select few notification types have `a` tags in the main body at the moment so to implement this would require redesigning all the notification objects. Also, to allow link-nesting we would have to really play with the css as html does not allow an `a` tag inside of an `a` tag. As a result this current change will not allow users to go the `right click` -> `open in new tab` route

*** Because this is adding a popup with `window.open()` this does create a pop-up permission since we are working within a `div` and it does not have traditional `a` tag functionality. I'm not totally sure if there is a way around this as it is not technically "clicking" a link (I tried with creating an anchor tag in [3d87012](https://github.com/stackernews/stacker.news/pull/1032/commits/3d8701226d4de2885e85034ec220555e6483703c) but this has the same problem)

## Checklist

* I did test with several notification types but not all of them!
is there any situation where we would want a different query type than `?commentId=` and is there a situation where this ternary `n.item.parentId ? as + '?commentId=' + n.id : as` breaks down?

* I don't think this is a behavior we really want on mobile, while some users might* want this on a longpress or something I think we should avoid having stacker news be open in 100+ mobile tabs, also mobile browsers are pretty strict security wise so this would result in users having to check pop-up boxes regularly

### DIFF
--- a/components/notifications.js
+++ b/components/notifications.js
@@ -79,10 +79,7 @@ function NotificationLayout ({ children, n, href, as, fresh }) {
         }, router.asPath, { ...router.options, shallow: true })
         if (e.ctrlKey) {
           e.preventDefault()
-          const anchor = document.createElement('a')
-          anchor.href = n.item.parentId ? as + '?commentId=' + n.id : as
-          anchor.target = '_blank'
-          anchor.click()
+          window.open(n.item.parentId ? as + '?commentId=' + n.id : as, '_blank')
         } else {
           router.push(href, as)
         }
--- a/components/notifications.js
+++ b/components/notifications.js
@@ -30,11 +30,13 @@ import { LongCountdown } from './countdown'
 import { nextBillingWithGrace } from '@/lib/territory'
 import { commentSubTreeRootId } from '@/lib/item'
 
+const nid = n => n.__typename + n.id + n.sortTime
+
 function Notification ({ n, fresh }) {
   const type = n.__typename
 
   return (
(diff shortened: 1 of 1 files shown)


components/notifications.js
```diff
@@ -57,24 +59,30 @@ function Notification ({ n, fresh }) {
...
         nid && await router.replace({
           pathname: router.pathname,
           query: {
             ...router.query,
-            nid
+            nidQuery
           }
         }, router.asPath, { ...router.options, shallow: true })
-        router.push(href, as)
+        if (e.ctrlKey) {
```
SatsAllDay, line 80: I'd also consider checking `e.metaKey` for mac users using Cmd+Click, though in general I think it's better to handle it via `a` tags instead of event listeners, since `a` tags are accessible. It sounds like you explored this in the PR description but opted against it for now.

benalleng: Yeah, I tried this route but tbh I'm not sure if I'm satisfied by the experience, it works ctrl + click opens in a new tab but it feels lacking like who is going to actually use this. 
Right now this doesn't pass the vibe check.

benalleng:
Closing in favor of trying to use anchor tags to achieve this Mvp

PR #1040, opened 2024-04-08
Merge serializeInvoiceable with serialize

ekzyis:
## Description

* merged `serializeInvoiceable` with `serialize`
* renamed `checkInvoice` to `verifyPayment` since it can be confused with `checkInvoice` from the worker.
* added support for conditional queries to `serialize` by filtering the input array with `.filter(q => !!q)`

## Checklist

### DIFF
--- a/api/resolvers/serial.js
+++ b/api/resolvers/serial.js
@@ -6,13 +6,30 @@ import { createHmac } from './wallet'
 import { msatsToSats, numWithUnits } from '@/lib/format'
 import { BALANCE_LIMIT_MSATS } from '@/lib/constants'
 
-export default async function serialize (models, ...calls) {
-  return await retry(async bail => {
+export default async function serialize (trx, { models, lnd, me, hash, hmac, fee }) {
+  // wrap first argument in array if not array already
+  const isArray = Array.isArray(trx)
+  if (!isArray) trx = [trx]
+
+  // conditional queries can be added inline using && syntax
+  // we filter any falsy value out here
+  trx = trx.filter(q => !!q)
+
+  let invoice
+  if (hash) {
+    invoice = await checkInvoice(models, hash, hmac, fee)
+    trx = [
+      models.$executeRaw`SELECT confirm_invoice(${hash}, ${invoice.msatsReceived})`,
+      ...trx
+    ]
+  }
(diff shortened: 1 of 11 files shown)


api/resolvers/serial.js
```diff
@@ -64,38 +81,20 @@ export default async function serialize (models, ...calls) {
...
+    if (invoice?.isHeld) {
+      await settleHodlInvoice({ secret: invoice.preimage, lnd })
+    }
     // remove first element since that is the confirmed invoice
-    [, ...results] = results
+    results = results.slice(1)
   }
 
-  // if there is only one result, return it directly, else the array
-  results = results.flat(2)
```
ekzyis, line 94: Did not include this line in the new merged code since I couldn't find why we needed it. It was added in 10203f565. I funded my wallet, founded a territory, changed billing from 100k/month to 1m/year, zapped around using NWC, everything worked fine.

I understand what it's supposed to do (unflat arrays) but seems unnecessary or even like a source for bugs. It might have been related to our weird array input/output parsing which no longer exists. If you put in an array, you now get back an array and vice versa. Before, the output did not directly depend on the input but on the output length which was weird.

api/resolvers/serial.js
```diff
@@ -64,38 +81,20 @@ export default async function serialize (models, ...calls) {
...
-  return results.length > 1 ? results : results[0]
+  // if first argument was not an array, unwrap the result
+  return isArray ? results : results[0]
 }
 
-export async function checkInvoice (models, hash, hmac, fee) {
+async function verifyPayment (models, hash, hmac, fee) {
   if (!hash) {
     throw new GraphQLError('hash required', { extensions: { code: 'BAD_INPUT' } })
   }
```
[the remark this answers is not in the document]
ekzyis: Interesting, one should use [`crypto.timingSafeEqual`](https://nodejs.org/dist/latest-v6.x/docs/api/crypto.html#crypto_crypto_timingsafeequal_a_b).

api/resolvers/serial.js
ekzyis: Did everything except create a simple comment, tunnel vision was strong with this one haha

PR #1041, opened 2024-04-08
Move code to @/lib/invoice

ekzyis:
## Description

Code in components/invoice should be limited to React components. The `useInvoiceable` hook etc. belongs more in lib/.

## Checklist


huumn:
I agree it seems weird they're in components, but most of the lib code is helper stuff that can run on client/server. 

Shouldn't we create a hooks dir?

ekzyis:
Mhh, I didn't see `useInvoiceable` as a hook but I think you're right, it's a hook that composes other hooks.

However, then it gets confusing again: Are there other hooks we would put in a hooks/ folder? Would we put the `useMe` hook in there? It's a hook that uses the `useContext` hook but imo it makes sense to keep it close to the context defined in components/me.js which provides the "context component" (= context provider).

This gets annoying to think about fast. If we can put more into a hooks/ folder than just `useInvoiceable` without making it more confusing, then I am fine with it.

However, thinking more about this, `useToaster` is similar to `useMe` since both provide access to a context. Additionally, I will probably refactor all or parts of this code anyway and I was planning to use context providers. I really like the Context API since it seems to capsulate stuff well which is currently the biggest problem with the zap undo + webln + invoiceable + optimistic UX mess: everything is mixed together which requires to keep a lot of things in mind at once. So I thought about using the Context API more anyway.

**Therefore, afaict, this discussion hints that `useInvoiceable` should _indeed_ also be used as a context (renamed as `useInvoice` for example) or that at least our current invoice flow should make more use of contexts. This would also make sense from a natural language view: invoices are always created with some context. The context is the thing we're paying for.**

In general, I think we're not using some React features enough and/or missing out on some like [`useReducer`]((https://react.dev/reference/react/useReducer#usereducer)). I think we're reimplementing the pattern for it in our toast code.

Maybe there is more stuff that we're currently doing that would be better solved with "advanced React features"?

Conclusion: instead of moving it out of components/, I should refactor the code so it makes sense to keep it there! lol

ekzyis:
Mhh, I didn't see `useInvoiceable` as a hook but I think you're right, it's a hook that composes other hooks.

However, then it gets confusing again: Are there other hooks we would put in a hooks/ folder? Would we put the `useMe` hook in there? It's a hook that uses the `useContext` hook but imo it makes sense to keep it close to the context defined in components/me.js which provides the "context component" (= context provider).

This gets annoying to think about fast. If we can put more into a hooks/ folder than just `useInvoiceable` without making it more confusing, then I am fine with it.

However, thinking more about this, `useToaster` is similar to `useMe` since both provide access to a context. Additionally, I will probably refactor all or parts of this code anyway and I was planning to use context providers. I really like the Context API since it seems to capsulate stuff well which is currently the biggest problem with the zap undo + webln + invoiceable + optimistic UX mess: everything is mixed together which requires to keep a lot of things in mind at once. So I thought about using the Context API more anyway.

**Therefore, afaict, this discussion hints that `useInvoiceable` should _indeed_ also be used as a context (renamed as `useInvoice` for example) or that at least our current invoice flow should make more use of contexts. This would also make sense from a natural language view: invoices are always created with some context. The context is the thing we're paying for.**

In general, I think we're not using some React features enough and/or missing out on some like [`useReducer`]((https://react.dev/reference/react/useReducer#usereducer)). I think we're reimplementing the pattern for it in our toast code.

Maybe there is more stuff that we're currently doing that would be better solved with "advanced React features"?

Conclusion: instead of moving it out of components/, I should refactor the code so it makes sense to keep it there! lol

huumn:
afaict most react projects have 3 directories (and they make sense imo):

- components
- hooks
- providers

ekzyis:
I can definitely see providers/ being useful for us. I am just unsure about the distinction between hooks vs providers, components. Most of the hooks we use are related to contexts or make sense to be close to components like `useZap` or `useAct`.

I'll keep your comment in mind though but will close this PR because I have better plans for this code now.

PR #1042, opened 2024-04-08
CLN autowithdrawal

### LINKED ISSUE #824: Add core lightning to attached wallets for autowithdraws

ekzyis:
## Description

Close #824 

## Screenshots

2024-04-12-020615_1920x1080_scrot

<details>

<summary>video</summary>

https://github.com/stackernews/stacker.news/assets/27162016/ea25b5ce-48c4-4a0c-a674-9196d6c1d6ff

</details>

## Additional Context

**Issue with funding new addresses**

For testing, you might need to reset your local bitcoin network. I noticed that mining blocks to an address did not increase the balance of that address. Deleting all volumes fixed that. I suspect that all coins were already generated.

**CLN connection**

There are multiple ways to connect to a CLN node:

- [REST API](https://docs.corelightning.org/docs/rest)
- [gRPC APIs](https://docs.corelightning.org/docs/grpc)
- [Commando](https://docs.corelightning.org/docs/commando)

(See https://docs.corelightning.org/docs/app-development.)

I am planning to use gRPC or Commando.

<details>

<summary>more context</summary>

Commando might be the easiest to setup since it only requires a rune and no certificate etc. + there is already a library one can use: [`lnsocket`](https://github.com/jb55/lnsocket).

For gRPC with CLN, stackers only need to enable gRPC via `--grpc-port` and we already use gRPC for LND so it's a consistent experience. However, there does not seem to be a library like [`ln-service`](https://github.com/alexbosworth/ln-service) for CLN we could use. So I would have to first generate language-specific bindings which I never did before but I am looking forward to learn about it.

I don't think I'll use REST since stackers would need to install python dependencies separately.

</details>

~update: I am using gRPC now.~

update: gRPC does not seem to support runes and couldn't get commando with [`lnsocket`](https://github.com/jb55/lnsocket) or [`lnmessage`](https://github.com/aaronbarnardsound/lnmessage) to work. With `lnsocket`, it simply returns an undefined error (it rejects at [this line](https://github.com/jb55/lnsocket/blob/0fad35b1352d2f1986d044bdab870a9d43433f62/dist/node/lnsocket.js#L204)) and with `lnmessage`, the `commando` call simply stalls. Not sure what I was doing wrong.

Will try to use [CLNRest](https://docs.corelightning.org/docs/rest) now.

update: done.

## TODO

- [x] create invoice with rune
- [x] validate rune permissions
- [x] autowithdrawal to CLN wallets in worker
- [x] ~check if protobuf is specific to versions: are there breaking changes between CLN versions?~ _(no longer using gRPC)_
- [x] ~strip protobuf to only what we need~ _(no longer using gRPC)_

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new CLN wallet management interface, allowing configuration and management of Core Lightning Node wallets.
	- Added auto withdrawal functionality for CLN wallets.
	- Implemented mining block functionality for `stacker_cln` service in the Docker environment.
- **Enhancements**
	- Updated GraphQL to support new mutations and queries for CLN wallets.
	- Enhanced logging to display 'ok' for success levels in UI components.
- **Bug Fixes**
	- Adjusted renaming logic in the logger to handle `walletCLN` correctly.
- **Documentation**
	- Added environment variable documentation for `STACKER_CLN_ADDR`.
- **Refactor**
	- Renamed job execution configurations for clarity and better maintenance.


docker/cln/hsm_secret
```diff
@@ -0,0 +1 @@
+$ÀDMBÆÔäî$«bã†y5úÞ$E0”Þ°P·§)Ì
```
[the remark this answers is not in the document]
ekzyis: it's okay, it's only for local development

docker-compose.yml
```diff
@@ -253,14 +253,17 @@ services:
...
             bitcoin-cli -chain=regtest -rpcport=${RPC_PORT} -rpcuser=${RPC_USER} -rpcpassword=${RPC_PASS} generatetoaddress 100 ${LND_ADDR}
             bitcoin-cli -chain=regtest -rpcport=${RPC_PORT} -rpcuser=${RPC_USER} -rpcpassword=${RPC_PASS} generatetoaddress 100 ${STACKER_LND_ADDR}
+            bitcoin-cli -chain=regtest -rpcport=${RPC_PORT} -rpcuser=${RPC_USER} -rpcpassword=${RPC_PASS} generatetoaddress 100 ${STACKER_CLN_ADDR}
           else
             echo "Mining a block to sn_lnd... ${LND_ADDR}"
             bitcoin-cli -chain=regtest -rpcport=${RPC_PORT} -rpcuser=${RPC_USER} -rpcpassword=${RPC_PASS} generatetoaddress 1 ${LND_ADDR}
             echo "Mining a block to stacker_lnd... ${STACKER_LND_ADDR}"
             bitcoin-cli -chain=regtest -rpcport=${RPC_PORT} -rpcuser=${RPC_USER} -rpcpassword=${RPC_PASS} generatetoaddress 1 ${STACKER_LND_ADDR}
+            echo "Mining a block to stacker_cln... ${STACKER_CLN_ADDR}"
+            bitcoin-cli -chain=regtest -rpcport=${RPC_PORT} -rpcuser=${RPC_USER} -rpcpassword=${RPC_PASS} generatetoaddress 1 ${STACKER_CLN_ADDR}
```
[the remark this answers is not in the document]
ekzyis: Where did I hardcode an address?

lib/cln.js
```diff
@@ -0,0 +1,161 @@
...
+    })
+  })
+  const inv = await res.json()
+  if (inv.error) {
+    throw new Error(inv.error.message)
+  }
+  return inv
+}
+
+// https://github.com/clams-tech/rune-decoder/blob/57c2e76d1ef9ab7336f565b99de300da1c7b67ce/src/index.ts
```
ekzyis, line 33: Copied code from https://github.com/clams-tech/rune-decoder since `npm install` from Github lead to import issues (relative import `./base64` wasn't found for some reason). However, there is no license provided for this code :sweat_smile: 

Opened a PR to be fair: https://github.com/clams-tech/rune-decoder/pull/1

lib/cln.js
```diff
@@ -0,0 +1,161 @@
...
+      summary
+    }
+  })
+
+  return {
+    id,
+    hash,
+    restrictions
+  }
+}
```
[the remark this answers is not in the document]
ekzyis: Resolved, see https://github.com/clams-tech/rune-decoder/pull/1

sndev
```diff
@@ -441,6 +478,9 @@ COMMANDS
     fund          pay a bolt11 for funding
     withdraw      create a bolt11 for withdrawal
 
+  cln:
```
huumn, line 481: Maybe this should be a flag to `fund` instead, and the help section renamed to `wallet` rather than `lnd`?

sndev
```diff
@@ -305,6 +333,15 @@ sndev__help_stacker_lncli() {
   docker__stacker_lnd --help
 }
 
+sndev__stacker_lightning-cli() {
```
huumn, line 336: This isn't a valid identifier in posix shell because of the `-` and the script fails to load for me.

I've been running my script changes through https://www.shellcheck.net/ just in case and heeding important warnings.

huumn: I fixed this locally already

sndev
huumn: We'll just leave this for now I think.

huumn:
Did you test autowithdraws to cln on this? I keep getting route not found even though manual withdrawals work to the cln node?

huumn:
Nice work!

sndev
ekzyis: > Maybe this should be a flag to fund instead, and the help section renamed to wallet rather than lnd

Ah yes, this would be better

PR #1048, opened 2024-04-09
Add anchor box for notification links

### LINKED ISSUE #202: Weird hitbox to open notifications in new tab
Hey, I noticed that I can only open notifications in a new tab if I click directly on the title. See screenshots.

Only when the title is selected, the context menu provides an option to "open in new tab".

This means that CTRL+MouseLeft also does only work when clicking on titles, not on the whole row as one may expect.

2022-10-04-183956_1920x1080_scrot
2022-10-04-183950_1920x1080_scrot

benalleng:
## Description

This is attempt `#2` to Close #202 now adds an anchor link box and adjusts some positioning for a smooth experience. Allows the user to use the entire nitification card as a link and it no longer uses `onCLick` functionality. Did require quite a bit of css so there may be some optimizations there.  inspired by this link from the issue https://bdwm.be/html5-alternative-nested-anchor-tags/ and looked at how reddit organizes their front page cards as these are very similar in structure.

## Screenshots

A minor adjustment was made such that all notifications are read or unread
image

## Checklist

Notifications tested
  - [ ] Streak
  - [ ] EarnNotification
  - [ ] RevenueNotification
  - [ ] SubStatus
  - [ ] Invitification
  - [ ] NostrZap
  - [ ] InvoicePaid
  - [ ] WithdrawlPaid
  - [x] Referral
  - [ ] Votification
  - [ ] ForwardedVotification
  - [x] Mention
  - [x] JobChanged
  - [x] Reply
  - [x] FollowActivity
  - [ ] TerritoryPost
  - [ ] TerritoryTransfer

I would like to test this with ALL notification types as I run a minimal env I haven't gotten a chance to create all possible notifications. (Getting a new computer soon as my current workstation is from 2016)

## Summary by CodeRabbit

- **New Features**
	- Enhanced notification system to include direct links within notifications for quicker access.
	- Improved text components to support hyperlinks, enhancing navigation and user interaction.
	- Updated comment display to conditionally show more details, improving content relevance and user engagement.
  
- **Style Updates**
	- Introduced new CSS classes for better layout and visibility of notifications, text, and upvote features.
	- Adjusted existing styles to accommodate new functionalities and improve user interface aesthetics.


benalleng:
@huumn looks like the tests for `parseInternalLinks` are failing after 81d3212 but for the life of me I can't figure out where the tests are getting their environment variables from as I've tried setting things to https://stacker.news but the tests still fail

components/comment.js
```diff
@@ -207,18 +207,31 @@ export default function Comment ({
...
+                <div ref={textRef}>
+                  {item.searchText
+                    ? <SearchText text={item.searchText} />
+                    : (
+                      <Link className={styles.commentLink} href={href} rel={item.rel ?? UNKNOWN_LINK_REL}>
+                        {item.outlawed && !me?.privates?.wildWestMode
+                          ? '*stackers have outlawed this. turn on wild west mode in your [settings](/settings) to see outlawed content.*'
+                          : truncate ? truncateString(item.text) : item.text}
+                      </Link>)}
+                </div>
```
benalleng, line 221: I think I need to reconsider turning the entire comment into a link as I dind't consider links being in a comment therby coming back around to the problem we started with.

image

benalleng: unfortunately this breaks the comment notifications being visually "read".

components/notifications.js
```diff
@@ -77,6 +81,12 @@ function NotificationLayout ({ children, nid, href, as, fresh }) {
         router.push(href, as)
       }}
     >
+      {!!href && (
+        <Link
+          className={styles.linkBox}
+          href={typeof href === 'string' ? href : n.item?.parentId ? as + '?commentId=' + n.id : as}
```
huumn, line 87: We want to hide the `?commentId` from the url for magicness.

`Link`'s `href` can be an object. https://nextjs.org/docs/pages/api-reference/components/link

I think we should be able to do `<Link href={{ pathname: href, asPath: as }} />`. You might've tried that though. The magicness isn't required but I like it personally.

components/notifications.js
```diff
@@ -320,13 +331,13 @@ function Votification ({ n }) {
...
       </small>
-      <div>
+      <div className={styles.detailsOverlay}>
         {n.item.title
           ? <Item item={n.item} />
           : (
             <div className='pb-2'>
               <RootProvider root={n.item.root}>
-                <Comment item={n.item} noReply includeParent clickToContext />
+                <Comment href={href} as={as} item={n.item} noReply includeParent clickToContext />
```
huumn, line 340: Curious. Why does this bleed into the comment/text component?

benalleng: When a comment contains links or youtube videos I want to still be able to click on those items while the box still contains the link but I think I have another solution whereby I can do this by adding some z-index to those child links specifically.

benalleng: sad I messed up the rebase :cry:

huumn: revert and just merge. I squash 'em anyway.

benalleng: on a normal comment like this 
image
the higlighted area is the text and therefor is not a link and cannot be right-clicked from here without adding a linnkBox.

I am trying to see if I can add some zIndex to not need these links

benalleng: Right, I've run into this probelm before but It looks like playing with the z-index gets be to this problem.
image

even after adding a higher z-index to the dropdown it persists so I think it might be a dead end with z-index alone

components/notifications.js
benalleng: I think i resolved this too soon, I was fooled by the standard `router.push(href, as)` behavior but with the `<Link>` being primarily used for opening in a new tab (in my testing the `as` doesn't work when opening in a new tab) we should actually keep these query params

benalleng: We can keep the magicness when just regularly navigating but for opening in a new tab we will want these params

components/notifications.js
huumn: I see. You need the linkbox to get click events but it can't if it's behind `Text`. Hmmm I wonder if there's a `pointer-events: none` thing that can be done.

components/notifications.js
huumn: That makes sense. I think the new window won't have the window state to read.

benalleng:
what is the purpose of this code
```
if (ignoreClick(e)) return
nid && await router.replace({
  pathname: router.pathname,
  query: {
    ...router.query,
    nid
  }
}, router.asPath, { ...router.options, shallow: true })
router.push(href, as)
```

vs this code that I've replaced it with in the latest commit?
```
e.preventDefault()
router.push(href, as)
```

I understand that it seems like if nid is not null then use router.replace but it seems that it doesn't work properly, when clicking notifications the history is never replaced and the ignoreClick(e) always seems to run but maybe I'm misunderstanding something.

huumn:
Oh, it's so that when you: click notification -> hit back button, the notification you clicked is highlighted ... when you have a lot of notification it can help you keep your place.

components/notifications.js
benalleng: Good idea! that worked great, took a little bit of restoring pointers with `pointer-events: auto` on some of the children elements but its much cleaner.

benalleng:
I think this is ready for some testing, I want to make a list of the notifications to add a `<div className={styles.detailsOverlay}>` which adds the `pointer-events: none;` to and then test where there are some children that may need `pointer-events: auto;` to restore clicking back to those child links like in the zap or post metadata links.

huumn:
I've pulled to qa and I think the small thing you were missing when you were using `z-index` was instead of using `z-index` you should've reordered the html nodes so they stack naturally. 

That is, I suspect placing `.linkBox` as the last child in the notification `div` makes it work pretty cleanly.

benalleng:
Wow look at that, ok i think that can simplify this even more.

PR #1051, opened 2024-04-10
Merge serializeInvoiceable with serialize without bug

ekzyis:
## Description

#1040 contained a bug that broke item creation and updating and was thus reverted.

Turns out that was at least one reason why `results.flat(2)` was needed (see https://github.com/stackernews/stacker.news/pull/1040#discussion_r1555123391).

This PR contains the same changes + ffbbcad0f3e75ef86f2f517cc58cce1d31b72df5 which fixes item creation and updating without flattening results which was also something I wanted to get rid of.

## Additional Context

I now searched for `serialize` and tested everything that depends on the return value of it + zaps.

Therefore, following functions were tested:

- `createItem`
- `updateItem`
- `act` via zaps and custom zaps
- `donateToRewards`
- `paySub`
- `createSub`
- `updateSub`
- `createInvoice`
- `cancelInvoice`
- `createWithdrawal`

Also tested invoice and withdrawal confirmations.

## Checklist

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1,6 +1,6 @@
 import { GraphQLError } from 'graphql'
 import { ensureProtocol, removeTracking, stripTrailingSlash } from '@/lib/url'
-import serialize, { serializeInvoicable } from './serial'
+import serialize from './serial'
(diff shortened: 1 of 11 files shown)


huumn:
>  fixes item creation and updating without flattening results

Much nicer than my `flatten()` hack!

Issue #771, opened 2024-01-30
Zaps not registered on first click if modal was closed before


ekzyis:
**Description**

The first short press on a zap icon is not registered if a modal was closed before.

**Steps to Reproduce**

1. Open modal
2. Close modal
3. Click zap icon -> Zap not registered
5. Click zap icon again -> Zap registered

**Expected behavior**

Zaps are always registered

**Video**

https://github.com/stackernews/stacker.news/assets/27162016/31c48306-2052-4a41-abdc-38408fefea96

**Additional context**

I thought it might be related to the document losing focus on modal open and the modal not giving back focus to the document, but `document.activeElement` returned the same before and after a modal was closed.

Also, if you add `onClick` to a parent element (see video), it does get registered. So I think it's something wrong with `LongPressable` and not related to missing focus. Using raw HTML works as expected.

I think this was always the case but it annoyed me enough while working on #749 to create a ticket now for this and I found easy steps to reproduce now.

huumn:
Can confirm this has been a bug since we added `LongPressable`

jp30566347:
This is a hard one. Couldn't find a root cause, but have a workaround.

PR #1055, opened 2024-04-11
Fix first zap when modal closed (#771)

jp30566347:
## Description

Fixes the onShortPress callback not being invoked after the long press callback is.

 - ~Assign a new key to force remounting of LongPressable component when modal is closed from long press~
 - Update LongPressable component with custom version with a fix (See https://github.com/stardust66/react-longpressable/pull/14) 
 - Extract handlers
 - Remove unnecessary async keyword from callback
 - Remove hover state when closing modal, otherwise it stays colored

## Screenshots

https://github.com/jp30566347/stacker.news/assets/1215807/e96be5ce-eab6-4bbc-9ba3-f48f4faf11dd

## Additional Context

~Could not find the root cause of this, but this workaround seems to trigger proper callback registration.~

I tried:
 - created a sandboxed environment from the react-longpressable and could NOT reproduce
 - removed children entirely (`<ActionTooltip />`) and this was still happening~

Turns out @huumn had good instinct on the modal messing with LongPressable event handlers.

## Checklist


components/upvote.js
```diff
@@ -142,43 +143,51 @@ export default function UpVote ({ item, className }) {
...
+
+  const handleLongPress = (e) => {
+    if (!item) return
+
+    // we can't tip ourselves
+    if (disabled) {
+      return
+    }
+
+    // force remount to fix onClick handler being ignored
```
ekzyis, line 158: +1 for the comment

ekzyis (approved):
Can confirm this fixes the issue. Don't see anything wrong with this approach except that we don't fully understand where it's coming from.

However, imo, bug is definitely annoying enough to justify fixing it this way instead of digging deeper. +1 for the additional context provided.

Additional context from me:

- Interesting that this only happens on long press. If I don't have enough sats, the QR code for payment pops up on short press. If I close it, short press still works as expected on first try.
- It's only broken for the item that was clicked. If you long press and then close, short press works for any other item on first try except the item was long pressed before.

huumn:
I remember debugging this when I first noticed it after it was introduced (years ago). 

IIRC it's caused by the modal popping up when the long press threshold is reached. The modal prevents `onPointerUp` from running which sets `isLongPressing = false`. So when the modal closes `isLongPressing = true` still.

jp30566347:
Good hint!

I wonder what is the reason behind this line: https://github.com/stardust66/react-longpressable/blob/99e33d25e7c720a91eb259eb1b4b72a2bec8837d/src/index.js#L76

I think it should be turned off now that the callback is getting invoked for long press?

huumn:
> I think it should be turned off now that the callback is getting invoked for long press?

Definitely! I think it's just to control whether the shortpress or the longpress handler is called `onPointerUp`.

jp30566347:
I'll confirm and get back to you

jp30566347:
Opened a PR to their repo: https://github.com/stardust66/react-longpressable/pull/14

In the meantime, happy to fork under stacker.news org and use that version.

It seems to fix it for me 🤞

huumn:
If you want to just move your version to components and uninstall the package, that'd be great! (It's a pretty tiny package.)

Issue #1056, opened 2024-04-11
make a homepage for the foss side of sn (sndev.team)


huumn:
This should be clientside-only hostable in a static way, e.g. with github pages.

It doesn't need to be too fancy, but it'd be nice if it showed off all the contributors, the rewards they've earned (using awards.csv), and made the README and other docs a little easier to navigate. Basically, it'd wrap all the stuff available on github, and aggregate it.

SouthKoreaLN:
So if I understand well, it'd be writing a script that generates the `index.html` file to be hosted on github pages parsing the info contained in the current versions of awards.csv, README.md and the files inside the docs folder? And beautifying it all.

I'd like to assign this to myself, but it'll take some time to finish as I am doing this at night and I've only just started learning frontend. Anyone else, feel free to speak up if you feel like you're the more apt person for this.

SouthKoreaLN:
As a first test, tried my hand at processing the `awards.csv` file. I parse the data using this tool
`<script src="https://cdnjs.cloudflare.com/ajax/libs/PapaParse/5.3.0/papaparse.min.js"></script>`
then add up all rewards for each contributor and organize it in a table ranking by biggest earner first to give something like this...
<img width="868" alt="Screenshot 2024-04-12 at 10 34 52 PM" src="https://github.com/stackernews/stacker.news/assets/11783268/a08fff4b-98ab-4e73-8078-359fb2910248">

SouthKoreaLN:
Updated some progress as a draft: https://github.com/stackernews/stacker.news/pull/1058.
Far from finished.

PR #1058, opened 2024-04-12
Foss home page

SouthKoreaLN:
## Description

- [x] Navigation bar
- [x] Parse info from awards.csv for awards tab
- [x] Convert README.md to html
- [ ] Populate docs tab
- [ ] Make everything look nicer
- [ ] Make sure I don't use any non-standard packages or break SN conventions 

## Screenshots
Screenshot 2024-04-12 at 11 23 37 PM

Screenshot 2024-04-12 at 11 23 26 PM

## Additional Context

Really just getting familiar with all of this.

## Checklist

### DIFF
--- a/components/autowithdraw-shared.js
+++ b/components/autowithdraw-shared.js
@@ -43,9 +43,9 @@ export function AutowithdrawSettings ({ priority }) {
             name='autoWithdrawThreshold'
             onChange={(formik, e) => {
               const value = e.target.value
-              setSendThreshold(Math.max(Math.floor(value / 10), 1))
+              setSendThreshold(Math.max(Math.floor(value / 10 * 11), 1))
             }}
-            hint={isNumber(sendThreshold) ? `attempts to keep your balance no greater than ${numWithUnits(sendThreshold)} of this amount` : undefined}
+            hint={isNumber(sendThreshold) ? `will attempt auto-withdraw when your balance exceeds ${sendThreshold} sats` : undefined}
             append={<InputGroup.Text className='text-monospace'>sats</InputGroup.Text>}
           />
           <Input
--- a/components/autowithdraw-shared.js
+++ b/components/autowithdraw-shared.js
@@ -3,7 +3,6 @@ import { Checkbox, Input } from './form'
 import { useMe } from './me'
 import { useEffect, useState } from 'react'
 import { isNumber } from 'mathjs'
-import { numWithUnits } from '@/lib/format'
 
 function autoWithdrawThreshold ({ me }) {
   return isNumber(me?.privates?.autoWithdrawThreshold) ? me?.privates?.autoWithdrawThreshold : 10000
@@ -45,7 +44,7 @@ export function AutowithdrawSettings ({ priority }) {
               const value = e.target.value
               setSendThreshold(Math.max(Math.floor(value / 10), 1))
             }}
-            hint={isNumber(sendThreshold) ? `will attempt auto-withdraw when your balance exceeds ${sendThreshold*11} sats` : undefined}
+            hint={isNumber(sendThreshold) ? `will attempt auto-withdraw when your balance exceeds ${sendThreshold * 11} sats` : undefined}
             append={<InputGroup.Text className='text-monospace'>sats</InputGroup.Text>}
           />
           <Input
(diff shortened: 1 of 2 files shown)


SouthKoreaLN:
This relates to https://github.com/stackernews/stacker.news/issues/1056.

huumn:
Looking great so far!

SouthKoreaLN:
Screenshot 2024-04-13 at 8 56 54 PM

SouthKoreaLN:
Parsing logic should be pretty much ok now. I'll play around with CSS next time to make things look nice. Probably should try to make it look a bit like the header of stacker.news...

SouthKoreaLN:
Screenshot 2024-04-14 at 6 56 57 AM
Screenshot 2024-04-14 at 6 57 18 AM
Screenshot 2024-04-14 at 6 57 27 AM

SouthKoreaLN:
I've put this as ready for review, as I am not fully sure if this is what you had in mind. Let me know if this is not the proper way to get feedback, for instance, if it is better to discuss this at the `issue` level rather than here with a PR for review.

I can make additional changes as required, just let me know :) 
You can also say you don't like it and not use it at all, no worries.

Depending on your feedback, I can eventually also proceed with uploading it to github pages (I did create the project already for that, but have kept it empty for now).

SouthKoreaLN:
Last update properly takes care of code blocks when converting markdown to html.
Screenshot 2024-04-14 at 10 45 01 PM

huumn:
How are you previewing this?

SouthKoreaLN:
> How are you previewing this while developing it? It doesn't load the files for me when I open the file in the browser.
> 
> I think it might be worth `fetch`ing the files from github regardless, but I don't know.

`python -m http.server` and then I just visualize it with `http://localhost:8000/index.html`.
I did that as my browser seemed to be blocking the CVS file from being loaded. 
Let me think how to resolve this in a more flexible way.

huumn:
> I've put this as ready for review, as I am not fully sure if this is what you had in mind. Let me know if this is not the proper way to get feedback, for instance, if it is better to discuss this at the issue level rather than here with a PR for review.

I'd say either in the issue or mentioning me with a question while this is in draft is probably the best "I'm not sure" move.

I'm going to pop this into draft and monitor it in case you have any questions.

SouthKoreaLN:
Caught up with my job until an hour ago.

I've replaced the FETCH calls with direct readings (using iframe) of the HTML files generated from the Markdown files. That way, you can open the index.html file in your browser and have a first look at it. I did NOT convert the `awards.csv` file to HTML though, as this requires more logic, and that is probably not worth it just for development purposes (also, see next reason).

Indeed, my suggestion, in case you're ok with what you see in this browser viewable HTML version, is that I create a new PR, but for the directory associated with Github pages where I push the `fetch`-version of the file. As it will be hosted on Github, there won't be such "FETCH" restrictions, and we can include the `README.md`, `docs/*md` and `awards.csv` file there and generate them on the fly. If it were only for the `md` files, it would probably make sense to make stand-alone HTML versions and use something like `iframe` to load them (like I did here). However, because of the awards.csv file that will be modified much more often, I presume, it'll be easier for you guys to just copy the csv file to that directory and update it there.

What do you think?

SouthKoreaLN:
Seems like the `GitGuardian Security Checks" warning got triggered when I asked here to update my branch with the most recent changes from `master`. Not sure what those specific secrets entail...

huumn:
> What do you think?

My only "concern" is that I might not want to host `sndev.team` on github. If we can't fetch them from github due to CORs, then it might be worth "building" them into a static page.

The simplest way to do this might be to create a script in `scripts` that when run, copies all the markdown file and the csv file into your `index.html` file.

If making the build script is too big of a lift, feel free to proceed with the github pages version, and I'll open up another issue for the "build" phase.

SouthKoreaLN:
@huumn I forgot about this one. Is this still something you're interested in, or has it become void due to changes in how you manage the awards?

huumn:
> @huumn I forgot about this one. Is this still something you're interested in, or has it become void due to changes in how you manage the awards?

Definitely still interested in this! I was considering reminding you about it after I saw your search pr.

PR #1063, opened 2024-04-13
anchor box for notifications

### LINKED ISSUE #202: Weird hitbox to open notifications in new tab

huumn:
closes #202

@benalleng I think this does it by adding a `linkBox` child and a bunch of css covering the `ignoreClick` function's logic. 

This link was really informative as `postition: static` is always below `position: absolute` in the stacking context: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_without_z-index

The big aha for me here was making `linkBox`'s siblings relative. No `z-index` was neessary.

Couldn't have done it without all the ground you covered in #1048 

## Summary by CodeRabbit

- **Style**
	- Added styling rules for improved visual layout of notification elements.
	- Adjusted positioning for better alignment of layout elements.
	- Updated pointer-events for enhanced interactivity within notifications.


benalleng (changes requested):
Looks great! I've learned a ton about css from this issue alone! One small issue is Iframes, possible to add it back with the global css but we will want to restore pointer events for the youtubeContainer or just all Iframes

components/notifications.module.css
```diff
@@ -26,4 +26,36 @@
...
+}
+
+.notificationLayout * {
+    pointer-events: none;
+}
+
+.notificationLayout :global(.upvoteParent),
+.notificationLayout :global(.upvoteParent) *,
+.notificationLayout :global(.pointer),
+.notificationLayout :global(.pointer) *,
```
benalleng, line 50: maybe just something like 
```
.notificationLayout :global(iframe),
```

huumn: Nice catch!

Issue #858, opened 2024-02-19
Make NWC zaps more responsive


benthecarman:
I keep accidentally double zapping posts because it doesn't provide immediate feedback when I click the lightning bolt. Would be nice if it went to a pending state immediately so I'd know that I did click it

ekzyis:
I believe most latency comes from attempting to pay from the custodial wallet first so maybe [this line](https://github.com/stackernews/stacker.news/pull/857/files#diff-00f337183418936435f48bbe59c72903633a7edb34bca37f6002c467362531a8R314) (already throwing the error on the client instead of waiting for a server response) in #857 might already help but I will check. Thanks for the feedback!

benthecarman:
> I believe most latency comes from attempting to pay from the custodial wallet first

My balance is displayed so you should have this locally and should be able to tell if it will work or not

ekzyis:
> My balance is displayed so you should have this locally and should be able to tell if it will work or not

Yeah, that's what I am doing with the line I was referring to in #857.

huumn:
This was fixed awhile ago.

PR #1071, opened 2024-04-15
Replace useInvoiceable with usePayment

ekzyis:
### Problem

This PR is the first step of refactoring the layers of code for external payments.

Essentially, we have three layers: 

1. Optimistic UX via manual Apollo cache updates
2. Zap undo: delay zapping and show option to "undo" (which is technically a cancel)
3. Payment code (custodial vs attached)

These layers are currently a mess since code for them is all over the place + zap undo, external payments and toast flows for both are implemented by wrapping functions. We don't have _distinct_ layers that don't leak into each other.

### Solution

This PR focuses on the payment layer. It replaces the function wrapper `useInvoiceable` with `usePayment` which hooks into the new context `PaymentContext`.[^1] This implements what I described in https://github.com/stackernews/stacker.news/pull/1041#issuecomment-2043959494:

> I really like the Context API since it seems to capsulate stuff well which is currently the biggest problem with the zap undo + webln + invoiceable + optimistic UX mess: everything is mixed together which requires to keep a lot of things in mind at once. So I thought about using the Context API more anyway.
>
> **Therefore, afaict, this discussion hints that `useInvoiceable` should indeed also be used as a context (renamed as `useInvoice` for example) or that at least our current invoice flow should make more use of contexts. This would also make sense from a natural language view: invoices are always created with some context. The context is the thing we're paying for.**

[^1]: This is somewhat related to "composition over inheritance" since wrapping functions can be seen as inheritance imo.

All `usePayment` does is to provide a function which returns a hash and hmac for anything that requires an external payment (QR or WebLN). The context is aware of the fee and freebies via the `FeeButtonContext` as well as the user balance so it knows when a payment is required and when not. This way, we can simply always call `usePayment` in `Form::onSubmit` if `invoiceableV2` is set without suffering latency overhead as described in #858.

`invoiceableV2` is a feature flag to gradually move from the old code used by `invoiceable` to the new code. I didn't want to change the old code but completely rewrite it while being able to easily look at it while doing so. When the old code is no longer needed, we can rename `invoiceableV2` to `invoiceable`. `invoiceableV2` is currently only used for comments.

I didn't want to create one big PR for all layers since I prefer to focus on one layer at a time. Make them do one thing but do it well.

This is also why this PR does not merge into `master` but `wallet-ux`. When refactor of all layers is merged into `wallet-ux` + real UX improvements on top, we can merge `wallet-ux` into `master`.

So consider this PR more as a way for early feedback before we actually deploy this. Merging it does not mean the code will not change later on when it's in `wallet-ux` but ideally, it wouldn't if I created good abstractions.

## Checklist

- [ ] Are your changes backwards compatible?

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -891,7 +891,7 @@ export default {
              WHERE act IN ('TIP', 'FEE')
              AND "itemId" = ${Number(id)}::INTEGER
              AND "userId" = ${me.id}::INTEGER)::INTEGER)`,
-          { models }
+          { models, lnd, hash, hmac }
         )
       } else {
         await serialize(
--- a/components/bounty-form.js
+++ b/components/bounty-form.js
@@ -104,7 +104,8 @@ export function BountyForm ({
         ...SubSelectInitial({ sub: item?.subName || sub?.name })
       }}
       schema={schema}
-      invoiceable={{ requireSession: true }}
+      requireSession
+      invoiceable
       onSubmit={
         handleSubmit ||
         onSubmit
--- a/components/fee-button.js
+++ b/components/fee-button.js
@@ -11,6 +11,7 @@ import AnonIcon from '@/svgs/spy-fill.svg'
 import { useShowModal } from './modal'
 import Link from 'next/link'
 import { SubmitButton } from './form'
+import { PaymentProvider } from './payment'
 
 const FeeButtonContext = createContext()
 
@@ -64,6 +65,7 @@ export function postCommentUseRemoteLineItems ({ parentId } = {}) {
 export function FeeButtonProvider ({ baseLineItems = {}, useRemoteLineItems = () => null, children }) {
   const [lineItems, setLineItems] = useState({})
   const [disabled, setDisabled] = useState(false)
+  const me = useMe()
 
   const remoteLineItems = useRemoteLineItems()
 
@@ -76,18 +78,24 @@ export function FeeButtonProvider ({ baseLineItems = {}, useRemoteLineItems = ()
 
   const value = useMemo(() => {
     const lines = { ...baseLineItems, ...lineItems, ...remoteLineItems }
+    const total = Object.values(lines).reduce((acc, { modifier }) => modifier(acc), 0)
+    // freebies: there's only a base cost and we don't have enough sats
+    const free = total === lines.baseCost?.modifier(0) && lines.baseCost?.allowFreebies && me?.privates?.sats < total
     return {
       lines,
       merge: mergeLineItems,
-      total: Object.values(lines).reduce((acc, { modifier }) => modifier(acc), 0),
+      total,
       disabled,
-      setDisabled
+      setDisabled,
+      free
     }
-  }, [baseLineItems, lineItems, remoteLineItems, mergeLineItems, disabled, setDisabled])
+  }, [me, baseLineItems, lineItems, remoteLineItems, mergeLineItems, disabled, setDisabled])
 
   return (
     <FeeButtonContext.Provider value={value}>
-      {children}
+      <PaymentProvider>
+        {children}
+      </PaymentProvider>
     </FeeButtonContext.Provider>
   )
 }
@@ -111,9 +119,7 @@ function FreebieDialog () {
 
 export default function FeeButton ({ ChildButton = SubmitButton, variant, text, disabled }) {
   const me = useMe()
-  const { lines, total, disabled: ctxDisabled } = useFeeButton()
-  // freebies: there's only a base cost and we don't have enough sats
-  const free = total === lines.baseCost?.modifier(0) && lines.baseCost?.allowFreebies && me?.privates?.sats < total
+  const { lines, total, disabled: ctxDisabled, free } = useFeeButton()
   const feeText = free
     ? 'free'
     : total > 1
--- a/components/form.js
+++ b/components/form.js
@@ -18,7 +18,6 @@ import { gql, useLazyQuery } from '@apollo/client'
 import { USER_SUGGESTIONS } from '@/fragments/users'
 import TextareaAutosize from 'react-textarea-autosize'
 import { useToast } from './toast'
-import { useInvoiceable } from './invoice'
 import { numWithUnits } from '@/lib/format'
 import textAreaCaret from 'textarea-caret'
 import ReactDatePicker from 'react-datepicker'
@@ -32,6 +31,15 @@ import Thumb from '@/svgs/thumb-up-fill.svg'
 import Eye from '@/svgs/eye-fill.svg'
 import EyeClose from '@/svgs/eye-close-line.svg'
 import Info from './info'
+import { InvoiceCanceledError, usePayment } from './payment'
+import { useMe } from './me'
+
+export class SessionRequiredError extends Error {
+  constructor () {
+    super('session required')
+    this.name = 'SessionRequiredError'
+  }
+}
 
 export function SubmitButton ({
   children, variant, value, onClick, disabled, nonDisabledText, ...props
@@ -793,11 +801,14 @@ const StorageKeyPrefixContext = createContext()
 
 export function Form ({
   initial, schema, onSubmit, children, initialError, validateImmediately,
-  storageKeyPrefix, validateOnChange = true, invoiceable, innerRef, ...props
+  storageKeyPrefix, validateOnChange = true, invoiceable, requireSession, innerRef, ...props
 }) {
   const toaster = useToast()
   const initialErrorToasted = useRef(false)
   const feeButton = useFeeButton()
+  const payment = usePayment()
+  const me = useMe()
+
   useEffect(() => {
     if (initialError && !initialErrorToasted.current) {
       toaster.danger('form error: ' + initialError.message || initialError.toString?.())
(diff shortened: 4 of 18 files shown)


ekzyis:
> `invoiceableV2` is a feature flag to gradually move from the old code used by `invoiceable` to the new code. I didn't want to change the old code but completely rewrite it while being able to easily look at it while doing so.

I decided that it's easier to simply rip out all old code after I rewrote `useInvoiceable` as `usePayment`. So this PR currently only implements using WebLN and QR codes for external payments but does not implement optimistic UX or zap undos. The code for optimistic UX and zap undos was simply removed.

I also removed the `JITInvoice` component. It complicated the flow since it seemed to require passing `onSubmit` to the payment code which was previously solved via wrapping and is the very thing I wanted to avoid in the new code. I also think it wasn't necessary and lead to confusing UX when the payment was successful but the action failed. On error, the frontend will immediately cancel the invoice now (or the worker will after the shortened timeout on payment) and a common error toast will be shown.

Since polling used to be done in `JITInvoice`, I added a new `poll` property to `Invoice` for polling support. We can probably reuse `Invoice` now in `FullInvoice` (pages/invoices/[id].js).

## Quick overview of new `usePayment` API

Common invoice control flow code is made available via `useInvoice` as a low-level API:

<details>
<summary>
Code for `useInvoice`
</summary>

```js
const useInvoice = () => {
  const client = useApolloClient()

  const [createInvoice] = useMutation(gql`
    mutation createInvoice($amount: Int!) {
      createInvoice(amount: $amount, hodlInvoice: true, expireSecs: 180) {
        id
        bolt11
        hash
        hmac
        expiresAt
        satsRequested
      }
    }`)
  const [cancelInvoice] = useMutation(gql`
    mutation cancelInvoice($hash: String!, $hmac: String!) {
      cancelInvoice(hash: $hash, hmac: $hmac) {
        id
      }
    }
  `)

  const create = useCallback(async amount => {
    const { data, error } = await createInvoice({ variables: { amount } })
    if (error) {
      throw error
    }
    const invoice = data.createInvoice
    return invoice
  }, [createInvoice])

  const isPaid = useCallback(async id => {
    const { data, error } = await client.query({ query: INVOICE, fetchPolicy: 'no-cache', variables: { id } })
    if (error) {
      throw error
    }
    const { hash, isHeld, satsReceived, cancelled } = data.invoice
    // if we're polling for invoices, we're using JIT invoices so isHeld must be set
    if (isHeld && satsReceived) {
      return true
    }
    if (cancelled) {
      throw new InvoiceCanceledError(hash)
    }
    return false
  }, [client])

  const waitUntilPaid = useCallback(async id => {
    return await new Promise((resolve, reject) => {
      const interval = setInterval(async () => {
        try {
          const paid = await isPaid(id)
          if (paid) {
            resolve()
            clearInterval(interval)
          }
        } catch (err) {
          reject(err)
          clearInterval(interval)
        }
      }, FAST_POLL_INTERVAL)
    })
  }, [isPaid])

  const cancel = useCallback(async ({ hash, hmac }) => {
    return await cancelInvoice({ variables: { hash, hmac } })
  }, [cancelInvoice])

  return { create, isPaid, waitUntilPaid, cancel }
}
```

</details>

_It might make sense to export this and use this low-level API elsewhere but I will check this out later._

This API is currently only consumed by the payment context:

<details>

<summary>Code for `PaymentProvider`</summary>

```js
export const PaymentProvider = ({ children }) => {
  const me = useMe()
  const feeButton = useFeeButton()
  const invoice = useInvoice()
  const waitForWebLnPayment = useWebLnPayment()
  const waitForQrPayment = useQrPayment()

  const waitForPayment = useCallback(async (invoice) => {
    try {
      return await waitForWebLnPayment(invoice)
    } catch (err) {
      if (err instanceof InvoiceCanceledError) {
        // bail since qr code payment will also fail if invoice was canceled
        throw err
      }
      // ignore any other error and fallback to QR code
    }
    return await waitForQrPayment(invoice)
  }, [waitForWebLnPayment, waitForQrPayment])

  const request = useCallback(async (amount) => {
    amount ??= feeButton?.total
    const free = feeButton?.free

    // if user has enough funds in their custodial wallet or action is free, never prompt for payment
    // XXX this will probably not work as intended for deposits < balance
    //   which means you can't always fund your custodial wallet with attached wallets ...
    //   but should this even be the case?
    const insufficientFunds = !me || me.privates.sats < amount
    if (free || !insufficientFunds) return [{ hash: null, hmac: null }, null]

    const inv = await invoice.create(amount)

    await waitForPayment(inv)

    const cancel = () => invoice.cancel(inv).catch(console.error)

    return [inv, cancel]
  }, [me, feeButton?.total, invoice, waitForPayment])

  const cancel = useCallback(({ hash, hmac }) => {
    if (hash && hmac) {
      invoice.cancel({ hash, hmac }).catch(console.error)
    }
  }, [invoice])

  const value = useMemo(() => ({ request, cancel }), [request, cancel])
  return (
    <PaymentContext.Provider value={value}>
      {children}
    </PaymentContext.Provider>
  )
}
```

</details>

### Form integration

We still set `invoiceable` on `<Form>` components. If `invoiceable` is set, we use `payment.request(amount)` to request the hash + hmac of a paid invoice and pass it to the mutation. We use custom errors now for control flow:

```js
const payment = usePayment()
...
const onSubmitInner = useCallback(async ({ amount, ...values }, ...args) => {
  let cancel
  try {
    if (onSubmit) {
      if (requireSession && !me) {
        throw new SessionRequiredError()
      }
      let hash, hmac
      if (invoiceable) {
        [{ hash, hmac }, cancel] = await payment.request(amount)
      }
      await onSubmit({ hash, hmac, amount, ...values }, ...args)
      if (!storageKeyPrefix) return
      clearLocalStorage(values)
    }
  } catch (err) {
    if (err instanceof InvoiceCanceledError) {
      return
    }
    const msg = err.message || err.toString?.()
    toaster.danger('submit error: ' + msg)
    cancel?.()
  }
}, [me, onSubmit, feeButton?.total, toaster, clearLocalStorage, storageKeyPrefix, payment])
```

I think it's way more clear what the above code does than how we used `useInvoiceable` before. We also no longer compare error messages but use `instanceof`:

```js
// if `invoiceable` is set,
// support for payment per invoice if they are lurking or don't have enough balance
// is added to submit handlers.
// submit handlers need to accept { satsReceived, hash, hmac } in their first argument
// and use them as variables in their GraphQL mutation
if (invoiceable && onSubmit) {
  const options = typeof invoiceable === 'object' ? invoiceable : undefined
  onSubmit = useInvoiceable(onSubmit, options)
}

const onSubmitInner = useCallback(async (values, ...args) => {
  try {
    if (onSubmit) {
      // extract cost from formik fields
      // (cost may also be set in a formik field named 'amount')
      const cost = feeButton?.total || values?.amount
      if (cost) {
        values.cost = cost
      }
      await onSubmit(values, ...args)
      if (!storageKeyPrefix) return
      clearLocalStorage(values)
    }
  } catch (err) {
    const msg = err.message || err.toString?.()
    // ignore errors from JIT invoices or payments from attached wallets
    // that mean that submit failed because user aborted the payment
    if (msg === 'modal closed' || msg === 'invoice canceled') return
    toaster.danger('submit error: ' + msg)
  }
}, [onSubmit, feeButton?.total, toaster, clearLocalStorage, storageKeyPrefix])
```

### Imperative API

Using the `usePayment` API without `<Form>` is basically the same: We simply call `payment.request` when we need a hash + hmac and are able to cancel the payment with the returned `cancel` function. We could also use `payment.cancel({ hash, hmac })` which is a simple wrapper around `invoice.cancel`. Calling `payment.cancel` will always be possible in the same scope as calling the mutation since the mutation needs hash + hmac already. Therefore, no props drilling is needed. Here is zapping via short press:

```js
return useCallback(async ({ item, me }) => {
  const meSats = (item?.meSats || 0)
  // add current sats to next tip since idempotent zaps use desired total zap not difference
  const sats = meSats + nextTip(meSats, { ...me?.privates })
  let hash, hmac, cancel
  try {
    [{ hash, hmac }, cancel] = await payment.request(sats - meSats)
    const variables = { id: item.id, sats, act: 'TIP', hash, hmac }
    const optimisticResponse = { act: { path: item.path, ...variables } }
    await zap({ variables, optimisticResponse })
    strike()
  } catch (error) {
    if (error instanceof InvoiceCanceledError) {
      return
    }
    console.error(error)
    toaster.danger('zap: ' + error?.message || error?.toString?.())
    cancel?.({ hash, hmac })
  }
}, [strike, payment])
```

Before, we use a separate wrapper `useInvoiceModal` to popup a modal after zapping failed due to no session or insufficient funds:

```js
const invoiceableAct = useInvoiceModal(
  async ({ hash, hmac }, { variables, ...apolloArgs }) => {
    await act({ variables: { ...variables, hash, hmac }, ...apolloArgs })
    strike()
  }, [act, strike])

  ...

return useCallback(async ({ item, me }) => {
  const meSats = (item?.meSats || 0)
  // add current sats to next tip since idempotent zaps use desired total zap not difference
  const sats = meSats + nextTip(meSats, { ...me?.privates })
  const amount = sats - meSats
  const variables = { id: item.id, sats, act: 'TIP', amount }
  const insufficientFunds = me?.privates.sats < amount
  const optimisticResponse = { act: { path: item.path, ...variables } }
  const flowId = (+new Date()).toString(16)
  const zapArgs = { variables, optimisticResponse: insufficientFunds ? null : optimisticResponse, update, flowId }
  try {
    if (insufficientFunds) throw new Error('insufficient funds')
    strike()
    if (zapUndosThresholdReached(me, amount)) {
      await zapWithUndos(zapArgs)
    } else {
      await zap(zapArgs)
    }
  } catch (error) {
    if (payOrLoginError(error)) {
      // call non-idempotent version
      const amount = sats - meSats
      optimisticResponse.act.amount = amount
      try {
        await invoiceableAct({ amount }, {
          variables: { ...variables, sats: amount },
          optimisticResponse,
          update,
          flowId
        })
      } catch (error) {}
      return
    }
    console.error(error)
    toaster.danger('zap: ' + error?.message || error?.toString?.())
  }
})
```



I tested the code and it seems to work quite well. I think I am going to merge this into `wallet-ux` now and work on adding optimistic UX and zap undos in a way that is hopefully similarly straight forward to reason about.

PR #1076, opened 2024-04-16
Use password field to NWC connection

bumi:
## Description

The NWC connection string contains sensitive data and thus should be handled like a password. 
Setting the input type to password and preventing auto complete prevents the user from accidentally exposing the connection string and it prevents the browser from saving the field value.

## Additional Context

I don't have a running stacker.news instance and made this change using the GitHub editor.  🙈
But I hope it's a bit better than a plain issue :) 

## Summary by CodeRabbit

- **New Features**
	- Enhanced security for the NWC URL input field in settings by adding password-type entry and disabling autocomplete.

### DIFF
--- a/pages/settings/wallets/nwc.js
+++ b/pages/settings/wallets/nwc.js
@@ -47,6 +47,8 @@ export default function NWC () {
           initialValue={nwcUrl}
           label='connection'
           name='nwcUrl'
+          type='password'
+          autoComplete='new-password'
           required
           autoFocus
         />


pages/settings/wallets/nwc.js
```diff
@@ -47,6 +47,8 @@ export default function NWC () {
           initialValue={nwcUrl}
           label='connection'
           name='nwcUrl'
+          type='password'
+          autoComplete='new-password'
```
ekzyis, line 51: [interesting, didn't know about `new-password`!](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#preventing_autofilling_with_autocompletenew-password)

makes sense to use it here

ekzyis (approved):

PR #1081, opened 2024-04-16
Prevent autowithdrawals from reversing deposits

ekzyis:
## Description

Before posting in ~mostly_harmless, I deposit funds manually so I can pay the fee.

However, autowithdrawals then reverse my deposit if I have forgotten to update the threshold before.

This PR fixes that by updating `autoWithdrawThreshold` on confirmed deposits if autowithdrawals are enabled.

## Additional Context

n/a

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced the autowithdrawal process to prevent reversal of deposits by dynamically adjusting the withdrawal threshold based on specific conditions.

### DIFF
--- a/worker/wallet.js
+++ b/worker/wallet.js
@@ -119,13 +119,19 @@ async function checkInvoice ({ data: { hash }, boss, models, lnd }) {
   }
 
   if (inv.is_confirmed) {
+    const user = await models.user.findUnique({ where: { id: dbInv.userId } })
     // NOTE: confirm invoice prevents double confirmations (idempotent)
     // ALSO: is_confirmed and is_held are mutually exclusive
     // that is, a hold invoice will first be is_held but not is_confirmed
     // and once it's settled it will be is_confirmed but not is_held
     const [[{ confirm_invoice: code }]] = await serialize([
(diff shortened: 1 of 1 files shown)


huumn:
hmmm this doesn't go back to the user set `autoWithdrawThreshold` after the action completes. This will affect more than posting/commenting too, won't it? Like if I zap a post 100k sats, then my `autoWithdrawThreshold` is that high from now on?

huumn:
I'm not sure what the answer to the above questions are, but if they are true, the better hack here might be to delay the autowithdraw job by 1 minute or something by altering the `startat` time when the job is inserted.

ekzyis:
> hmmm this doesn't go back to the user set autoWithdrawThreshold after the action completes. This will affect more than posting/commenting too, won't it? Like if I zap a post 100k sats, then my autoWithdrawThreshold is that high from now on?

Oh, you're right, this was only meant for manual deposits to workaround being a freebie. I basically only shifted the problem from having to update my threshold before deposits to updating it after completing what I wanted to do with this.

> but if they are true

tested, they are true

> the better hack here might be to delay the autowithdraw job by 1 minute or something by altering the startat time when the job is inserted.

Makes sense, I'll do that!

PR #1082, opened 2024-04-17
follow superior nextjs env var pattern

huumn:
I finally read the [nextjs env docs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables) and whowudaguessd they have a better pattern for prod/dev/local env files.

We needed to do this for prod because our env configuration was so large cloudformation barfed when I went to add our new env vars. Many of those env vars are not secrets and could be move to `.env.production`, which I did, giving us more room for new secrets. This should also make it easier for not-me to add new not-secret env variables (and for anyone curious about prod to see more of how that works).

Anyway, this will f'up anyone using `.env.development` for secrets so heads up @stackernews/eng. From now on, you'll want to put secrets/overrides in `.env.local`. (This also allowed us to get rid of the (likely) buggy `.env.sample` merge in `sndev` ... which was dumb.)

## Summary by CodeRabbit

- **Documentation**
	- Enhanced readability and organization of environment configuration files.
	- Clarified and streamlined instructions for setting email configurations and Docker setups.
- **Chores**
	- Updated `.gitignore` to better manage environment files.
- **Refactor**
	- Simplified startup logic in the `sndev` script to enhance the initial setup process.


ekzyis:
> From now on, you'll want to put secrets/overrides in `.env.local.`

Seems like the env vars in `.env.development` override any variables in `.env.local` (environment > env file).

Therefore, I have to set secrets like VAPID keys in `.env.development` or remove the env vars there.

And since `.env.development` is checked into version control now, my working tree is dirty.

I'll fix this in a PR

huumn:
I think this might be a docker thing. .env.local should take priority if we let nextjs load them

SatsAllDay:
Ah maybe this is the source of [this issue](https://github.com/stackernews/stacker.news/commit/dd6e921e2e00950f19da5bd6d76769c89268d1bb#r141179262) I am seeing.

I'll move to .env.local and see how things go.

ekzyis:
> I think this might be a docker thing. .env.local should take priority if we let nextjs load them

My thinking with #1085 was that `env_file` puts env vars into the environment of the container (`env` in bash or `process.env` in nodejs) so due to the [environment variables load order](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#environment-variable-load-order), `.env.development` in `env_file` becomes `process.env` and thus effectively overrides any `.env.local` that is loaded during runtime. At least that explained [the behavior that I saw](https://github.com/stackernews/stacker.news/pull/1082#issuecomment-2061724213).

I then also used `--env-file .env.local` so env var expansion would work as expected inside the docker-compose file.

However, I think `env_file` is redundant now if `sndev` is used since [`--env-file` overrides `env_file`](https://docs.docker.com/compose/environment-variables/set-environment-variables/#additional-information-3).

Since an array for `env_file` is only supported in the latest release of docker-compose (v2.24 from January this year), I think we can simply remove `.env.local` from `env_file` to also support older versions of docker-compose and rely on `--env-file` fixing the env var load order.

PR #1087, opened 2024-04-17
Hidden price sizing

benalleng:
## Description

The fixed width for the hidden price stars was not properly determining the expected width of the number "underneath" (note if your cursor is near the edge of the numers it "flashes") this change ensures that the width is actually fixed and since this span uses monospace we can hard code the width based on the width of the 6 `******` characters at `0.6em` each. Ths the total width when hidden is `3.6em`  

This will break down at > 1000.0t sats as the value will be longer than the stars but if you have that much in stacker news god help you (also it will overflow to the right,not really moving anything)

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed layout shifts in the wallet summary component by setting a constant width.

### DIFF
--- a/components/hidden-wallet-summary.js
+++ b/components/hidden-wallet-summary.js
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect } from 'react'
+import { useState, useRef } from 'react'
 import { abbrNum, numWithUnits } from '@/lib/format'
 import { useMe } from './me'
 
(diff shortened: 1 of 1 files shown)


huumn:
This broke the display on `/wallet`. I went another way in a77f778 to fix it. I'll grant an award for it still.

components/hidden-wallet-summary.js
```diff
@@ -1,22 +1,20 @@
...
-  }, [])
+  // prevent layout shifts when hovering by fixing width to initial rendered width of '******'
+  // We can simply multiply 6 by '0.6em' since monopsace has a fixed width for all characters
+  const width = '3.6em'
 
   return (
     <span
-      ref={ref} style={{ width: fixedWidth ? width : undefined }}
-      className='text-monospace' align='right' onPointerEnter={() => setHover(true)} onPointerLeave={() => setHover(false)}
+      ref={ref} style={{ display: 'block', width: fixedWidth ? width : undefined, textAlign: 'right' }}
```
benalleng, line 16: Ahh, didn't catch that this should have been `display: 'inline-block'`

huumn: I tried that too but then it doesn't work in the navbar! It's a tricky mistress.

benalleng: hmm couldn't break it like that did but yours works just the same :+1:

PR #1088, opened 2024-04-18
Add 200ms delay to scroll in capture worker

benalleng:
## Description

@huumn pointed out to me that the capture worker unfortunately doesn't work in prod, I tried initially to not add a 200ms delay thinking that fixed timeouts can be flaky but alas my initial solution did not work.

In this Pr I keep the puppeteer native `await page.waitForSelector('.outline-it')` but I also add a `200ms` timeout to help ensure consistency in actually scrolling to the screenshot.

## Screenshots

<details>

image

image

image

image

</details>

## Checklist

  - Tested on a few screenshots but I have a relatively small sample size still

### DIFF
--- a/capture/index.js
+++ b/capture/index.js
@@ -92,8 +92,10 @@ app.get('/*', async (req, res) => {
     console.timeLog(timeLabel, 'page loaded')
 
     if (commentId) {
-      console.timeLog(timeLabel, 'scrolling to comment')
       await page.waitForSelector('.outline-it')
+      setTimeout(() => {
+        console.timeLog(timeLabel, 'scrolling to comment')
+      }, 200)
     }
 
     const file = await page.screenshot({ type: 'png', captureBeyondViewport: false })


capture/index.js
```diff
@@ -92,8 +92,10 @@ app.get('/*', async (req, res) => {
     console.timeLog(timeLabel, 'page loaded')
 
     if (commentId) {
-      console.timeLog(timeLabel, 'scrolling to comment')
       await page.waitForSelector('.outline-it')
+      setTimeout(() => {
+        console.timeLog(timeLabel, 'scrolling to comment')
+      }, 200)
```
huumn, line 98: `setTimeout` runs async so this will return immediately then proceed to screenshot.

What you want to do here is something like `await new Promise((resolve, reject) => setTimeout(resolve, 100))` to block the function execution.

I think 100ms should be plenty of time too btw.

huumn (changes requested):

capture/index.js
benalleng: Ok! thanks

benalleng: caught a lightning bolt in this one

image

PR #1090, opened 2024-04-18
Add PasswordInput component

### LINKED ISSUE #1077: Add button to reveal value for inputs with type="password"

It's not possible to see which values were entered into an input with `type="password"`.

A simple button next to the input like we have to scan invoices for withdrawals:

2024-04-16-130554_738x381_scrot

on click, it would toggle between hiding and revealing the secret.

the icon could be the open and closed eye icon we use for collapsing comments.

n/a

**Additional context**

got this idea while reviewing #1076

benalleng:
## Description

Adds a PasswordHider component input group that can be appended on the end of password inputs

Closes #1077 

## Screenshots

Screenshot_2024-04-18_10-18-54

Screenshot_2024-04-18_10-18-35

## Checklist

  - Only added on the NWC input are there other Password inputs in sn?

## Summary by CodeRabbit

- **New Features**
	- Enhanced user experience by adding password visibility toggle feature in settings, allowing users to switch between hiding and showing password inputs.
	- Introduced a new `PasswordInput` component for handling password inputs in the LNBits settings page.
	- Improved security and user interaction by replacing `ClientInput` with `PasswordInput` in the NWC settings.

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -1054,7 +1054,7 @@ function Client (Component) {
   }
 }
 
-export function PasswordHider ({ onClick }) {
+function PasswordHider ({ onClick }) {
   const [showPass, setShowPass] = useState(false)
 
   const handleClick = () => {
@@ -1075,6 +1075,23 @@ export function PasswordHider ({ onClick }) {
             fill='var(--bs-body-color)' height={20} width={20}
           />}
     </InputGroup.Text>
+  )
+}
+
+export function PassworInput ({ initialValue, label, name, newPass, autoFocus, required = true }) {
+  const [showPass, setShowPass] = useState(false)
+
+  return (
+    <ClientInput
+      initialValue={initialValue}
+      label={label}
+      name={name}
+      type={showPass ? 'text' : 'password'}
+      autoComplete={newPass ? 'new-password' : 'current-password'}
+      autoFocus={autoFocus}
+      required={required}
+      append={<PasswordHider onClick={() => setShowPass(!showPass)} />}
+    />
   )
 }
 
--- a/components/form.js
+++ b/components/form.js
@@ -1054,11 +1054,8 @@ function Client (Component) {
   }
 }
 
-function PasswordHider ({ onClick }) {
-  const [showPass, setShowPass] = useState(false)
-
+function PasswordHider ({ onClick, showPass }) {
   const handleClick = () => {
-    setShowPass(!showPass)
     onClick()
   }
 
@@ -1090,7 +1087,7 @@ export function PasswordInput ({ initialValue, label, name, newPass, autoFocus, 
       autoComplete={newPass ? 'new-password' : 'current-password'}
       autoFocus={autoFocus}
       required={required}
-      append={<PasswordHider onClick={() => setShowPass(!showPass)} />}
+      append={<PasswordHider showPass={showPass} onClick={() => setShowPass(!showPass)} />}
     />
   )
 }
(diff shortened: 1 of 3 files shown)


pages/settings/wallets/nwc.js
```diff
@@ -47,10 +49,11 @@ export default function NWC () {
           initialValue={nwcUrl}
           label='connection'
           name='nwcUrl'
-          type='password'
+          type={showPass ? 'text' : 'password'}
           autoComplete='new-password'
           required
           autoFocus
+          append={<PasswordHider onClick={() => setShowPass(!showPass)} />}
```
ekzyis, line 56: I think it makes sense to also pass in `showPass`. Then you don't need to keep any state inside `PasswordHider` and it would be less confusing than having two `setShowPass` functions which need to be synchronized.

ekzyis (commented):
Didn't pull the code and test but looks good!

Added a small comment about state.

Can you also add this to wherever else we use `type='password'`?

Also, if you feel like it: maybe it's possible to do this all inside a component named `PasswordInput` which encapsulates all state? As is, the logic for hiding a password is dealt outside of `PasswordHider` and thus needs to be duplicated wherever it's used.

I imagine something that can be used like this:

```jsx
<PasswordInput
  initialValue={nwcUrl}
  label='connection'
  name='nwcUrl'
  // type={showPass ? 'text' : 'password'} <-- this is automatically handled inside the component
  autoComplete='new-password'
  required
  autoFocus
  // append={<PasswordHider onClick={() => setShowPass(!showPass)} />} <-- this too
/>
```

pages/settings/wallets/lnbits.js
```diff
@@ -51,12 +51,11 @@ export default function LNbits () {
           required
           autoFocus
         />
-        <ClientInput
+        <PasswordInput
```
benalleng, line 54: Is the lnbits admin key required? I couldn't parse whether this needs to be required or not from the validation

components/form.js
```diff
@@ -1052,5 +1054,43 @@ function Client (Component) {
   }
 }
 
+function PasswordHider ({ onClick, showPass }) {
+  const handleClick = () => {
+    onClick()
+  }
```
ekzyis, line 1060: Looks like an artifact and no longer needed. You can simply pass `onClick` to `onClick`, no?

components/form.js
```diff
@@ -1052,5 +1054,43 @@ function Client (Component) {
...
+    <ClientInput
+      initialValue={initialValue}
+      label={label}
+      name={name}
+      type={showPass ? 'text' : 'password'}
+      autoComplete={newPass ? 'new-password' : 'current-password'}
+      autoFocus={autoFocus}
+      required={required}
+      append={<PasswordHider showPass={showPass} onClick={() => setShowPass(!showPass)} />}
+    />
```
ekzyis, line 1091: I think using `...props` in the function signature and then `{...props}` here  for anything that we simply pass through to `<ClientInput>` would be better

pages/settings/wallets/lnbits.js
ekzyis: Oh, must have been an oversight when I implemented this. It's definitely required.

ekzyis (commented):
Nice! Just two nitpicks but to make it even more really good.

_Still haven't pulled the code but will try today, I'm travelling currently_

ekzyis:
I just noticed that the UI/UX is reversed for comments. When the comment is uncollapsed, we show the closed eye. When it's collapsed, we show the collapsed eye.

So I think we should make it consistent here. A classic case of [this](https://dev.to/ben/should-a-button-communicate-the-current-state-the-intended-behavior-or-both-50na).

huumn:
Also true of dark/light mode. It's weird having eyes on all the comments though.

PR #1091, opened 2024-04-18
Enhanced the cursor to be stateful

### LINKED ISSUE #147: Eliminate duplication that can happen after clicking More when ranking changes over time
After an initial load of the front page, an item might move from a position in the top 21 into a lower position before the More button is pressed. Then when More is pressed, that item likely will appear again in the newly added results.  

A solution is to exclude from the results for pressing more any items that already appeared in the results above that existed before More was pressed.

crodas:
## Description

Fixes #147

To avoid duplicated results push some state towards the cursor.

The cursor uses a Bloom Filter[1], a probabilistic data structure. In this data structure, False positive matches are possible, but false negatives are not – in other words, a query returns either "possibly in set" or "definitely not in set". Elements can be added to the set, but not removed. The more items added, the larger the probability of false positives.

[1] https://en.wikipedia.org/wiki/Bloom_filter

### DIFF
--- a/capture/index.js
+++ b/capture/index.js
@@ -94,6 +94,7 @@ app.get('/*', async (req, res) => {
     if (commentId) {
       console.timeLog(timeLabel, 'scrolling to comment')
       await page.waitForSelector('.outline-it')
+      await new Promise((resolve, _reject) => setTimeout(resolve, 100))
     }
 
     const file = await page.screenshot({ type: 'png', captureBeyondViewport: false })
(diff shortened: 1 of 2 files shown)


huumn:
This is going to require using a bloomfilter in postgres queries. It doesn't look like you've gotten to that part in the PR yet so I wanted to be sure to flag that.

For a nice to have thing, my "heads up" advice is that it's going to be easy to overengineer a solution here. The requirement this is going to need to meet is "is it worth a probabilistic data structure, another package and new code to solve this minor inconvenience" and "is this the simplest way to solve this problem or merely the best solution that can be tacked on"?

PR #1093, opened 2024-04-19
Crosspost replies to nostr

AustinKelsay:
## Description

Allows for crossposting of replies IF both root and parent have been crossposted already

## Screenshots

[walkthrough](https://www.youtube.com/watch?v=72zzhwMQBsg)

## Additional Context

- I needed to add a few queries to refetch parent/root items to check noteId in the crossposting flow and also determine if the "crosspost" option should render or not
- Had to handle the two different kinds of replies (replies to a short text note and replies to a parameterized replaceable note)
- Had to add noteId to COMMENT_FIELDS 

## Checklist

- [ ] Are your changes backwards compatible?

### DIFF
--- a/components/item-info.js
+++ b/components/item-info.js
@@ -179,7 +179,7 @@ export default function ItemInfo ({
         {item && item.mine && !item.noteId && !item.isJob && !item.parentId &&
           <CrosspostDropdownItem item={item} />}
         {/* for crosspost replies */}
-        {item && item.mine && !item.noteId && root.noteId && isParentCrossposted &&
+        {root && item && item.mine && !item.noteId && root.noteId && isParentCrossposted &&
           <CrosspostDropdownItem item={item} />}
         {me && !item.position &&
           !item.mine && !item.deletedAt &&
--- a/components/item-info.js
+++ b/components/item-info.js
@@ -175,9 +175,10 @@ export default function ItemInfo ({
             nostr note
           </Dropdown.Item>
         )}
+        {/* for crosspost top level items */}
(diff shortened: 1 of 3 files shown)


components/item-info.js
```diff
@@ -33,10 +35,19 @@ export default function ItemInfo ({
   const [canEdit, setCanEdit] =
     useState(item.mine && (Date.now() < editThreshold))
   const [hasNewComments, setHasNewComments] = useState(false)
+  const [isParentCrossposted, setIsParentCrossposted] = useState(false)
   const [meTotalSats, setMeTotalSats] = useState(0)
   const root = useRoot()
   const sub = item?.sub || root?.sub
 
+  const { data } = useQuery(ITEM_FULL, { variables: { id: item?.parentId } })
```
huumn, line 43: The performance implications of this are significant. For every comment, it queries the server for the entire parent. In a comment thread with 100 comments, this will make 100 requests to the server.

I think what we want to do instead is when rendering a `Comment` component from a parent item/comment, send it this info as props.

When a comment is in list of comments like on a profile, you'll want to use a different query fragment with a nested resolver for `parent` so that this data is gathered when all the other data is. An example of this is the nested resolver for `root`.

huumn (changes requested):
Awesome! I wasn't expecting this one.

I think the hard part of this is going to be getting all the data we need in a performant/consistent way.

components/item-info.js
AustinKelsay: Sweet thanks for the feedback! I'll get right on this

PR #1095, opened 2024-04-20
Fix no notification for @delete usage in post

### LINKED ISSUE #728: No notification for @delete usage in post
**Description**

Stackers don't get notifications if they use the built-in delete bot on posts. The post is still successfully deleted though.

**Steps to Reproduce**

Mention `@delete` in a post

**Expected behavior**

Receive notification that it worked.

SatsAllDay:
## Description
Closes #728 

This PR fixes the issue where using the at-delete directive in posts wasn't showing the success/confirmation toast like it's supposed to. This is accomplished by:

1. introducing `persistOnNavigate` option for toasts
2. pass `persistOnNavigate: true` when creating or editing a post, but not a comment

## Screenshots
N/A

## Additional Context
I suspect this was never an issue in local dev, but was an issue in prod, due to differences in how route changes are handled when running next.js in dev vs prod mode. Given that submitting a new post, or editing an existing post both result in a navigation to a new URL, my theory is that in dev mode, there is something that causes the `routeChangeStart` event _not_ to trigger, which is what's removing toasts on page navigation in production. Or, it may trigger, but the timing is different than in prod (relative to when the toast is added), causing this feature to work in dev but not in prod.

I placed breakpoints in prod and shoe-horned in a change representative of this PR and it seemed to do the trick, so that's how I verified it.
 
## Checklist
- [X] Are your changes backwards compatible?

## Summary by CodeRabbit

- **New Features**
	- Added persistence control for toasts on navigation, allowing specific notifications to remain visible during page transitions.

### DIFF
--- a/components/toast.js
+++ b/components/toast.js
@@ -110,7 +110,7 @@ export const ToastProvider = ({ children }) => {
   // Only clear toasts with no cancel function on page navigation
   // since navigation should not interfere with being able to cancel an action.
   useEffect(() => {
-    const handleRouteChangeStart = () => setToasts(toasts => toasts.length > 0 ? toasts.filter(({ onCancel, onUndo }) => onCancel || onUndo) : toasts)
+    const handleRouteChangeStart = () => setToasts(toasts => toasts.length > 0 ? toasts.filter(({ onCancel, onUndo, persistOnNavigate }) => onCancel || onUndo || persistOnNavigate) : toasts)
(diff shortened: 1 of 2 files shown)


ekzyis (approved):
Ahh, this makes sense. Very thankful for the detailed explanation in the additional context. I hate when I don't understand why something doesn't work, especially when it does work locally! This gives me peace.

Issue #519, opened 2023-09-25
Hidden message reply


Darth-Coin:
**A reply feature that will hide the text for other readers**
Will be nice and useful to have a way to reply to somebody, but only that user could see the text.
Other stackers MUST pay a fee to see it.
The fee could be defined by the OP when is replying.

This could be very useful when stackers are talking about some more private information but still want to share it with others.
Could have many use cases on SN.

Please consider it.

huumn:
I like the private reply idea! I'm not sure about people paying to see it, but I definitely like private messages!

ekzyis:
> but I definitely like private messages!

you mean in a different way than just #56 ?

Darth-Coin:
#56 is something different.
Hidden message is a public message with the option to be hidden for the rest but also option of pay to view.
The fee could go to SN pool just fine.

ekzyis:
Yes, I understood, was just wondering in what way this here would be different from #56 without the paywall (which I am also not sure about).

> This could be very useful when stackers are talking about some more private information but still want to share it with others.

Can you share some examples? This sound like a oxymoron for me. If something is truly private, I don't want other people to see it even if they pay for it

huumn:
> If something is truly private, I don't want other people to see it even if they pay for it

Same.

> you mean in a different way than just https://github.com/stackernews/stacker.news/issues/56 ?

Yes, because a private reply inherits the context of the thread. In a way, it's more natural and fits within existing ergonomics. A DM is like it's own thing.

It'd be cool if we can design a DM/private reply system that is somehow a hybrid of these ideas rather than independent things, but I'm not sure.

huumn:
Somewhat, but still not the same. I updated my comment. 

These are 3 different problems and 3 different solutions. It'd be nice to solve them all with 1 solution but it might not be possible.

Darth-Coin:
Maybe I want to send a PR-13 meme in a reply.
Like k00b said to me that is not nice to post them.
So I could hide them behind a paywall for example.

Maybe I want to send a reply with some text that is dedicated and related to previous comments, but want to be shown first only to the OP. If others really want to see it, must pay to SN pool. Is more like a game let's say.

huumn:
Added blocked label because we want private replies to be E2EE.

Darth-Coin:
Private replies / messages are something else.
What I wanted to say is like a spoiler alert, the comment in a public post is blurred and only those that pay to view will be able to see the spoiler.

[...]

Issue #56, opened 2021-10-21
Direct messages


huumn:
Kind of a nice to have but it'd help users connect more and could be interesting economically (i.e. users can configure what it costs to dm them).

ekzyis:
NIP-44 will most likely get merged end of November. The only thing that's left is a security audit. [And one is scheduled for in the second half of November.](https://github.com/nostr-protocol/nips/pull/746#issuecomment-1752656642)

huumn:
I've been fantasizing about DMs being rendered like comment trees. Figured I should share the fantasy.

PR #1104, opened 2024-04-24
add lnc attached wallet

huumn:
This 
- adds a `litd` container to `sndev` for testing 
- implements `lnc` as an attached wallet

This works but still has a few issues:
- [x] wasm loading is wonky ... in `strictMode`, `lnc.connect()` will be called twice which causes an infinite loop in wasm seemingly
     - I need to investigate how to better handle this and only conditionally load the wasm
- [x] need to check perms/budget as discussed [here](https://github.com/lightninglabs/lightning-terminal/discussions/749#discussioncomment-9222027)
     - we don't check budgets yet because we need more perms
- [x] `lnc` encrypts the keys with a stacker provided password before storing in the browser
     - alby [hardcodes a password](https://github.com/getAlby/bitcoin-connect/blob/86263c45a7cf31573a6ab2ed2683bb2657148b5a/src/connectors/LNCConnector.ts#L39) to deal with this but we probably give stackers the option to encrypt the storage
    - so for stackers that want encryption, I'll need a UI for prompting for the password to spend or to display the pairing phrase in a form
- [x] we have enough webln providers and shared code that it's time for a refactor/improvement to the abstraction .. I'll take stab at it possibly

We'll probably leave this as a TODO
- [x] for a given pairing phrase, only one session/tab/device can have access to it at a time
    - ~~I'm not sure if this is a mailbox limitation (in which case we can run our own ... or a litd limitation)~~
    - this is an unchangeable cryptographic thing according to [this dicussion](https://github.com/lightninglabs/lightning-terminal/discussions/751#discussioncomment-9237320) ... so we'll need to use a web worker or to connect/disconnect for every action
    - I "solved" this by disconnecting after every action, and reconnecting before it. ~~It's fragile as disconnections are not very reliable in my testing but I'll be curious how it flies in prod~~ found the problem

## Summary by CodeRabbit

- **New Features**
    - Introduced functionality for managing Lightning Network Client (LNC) connections in the app.
    - Added features for managing Lightning Node Connect for LND payments in the wallet settings.

- **Enhancements**
    - Improved provider management with the new `LNCProvider`.
    - Standardized provider statuses with the introduction of a `Status` object.
    - Enhanced validation rules for `pairingPhrase` and `password` fields.
    - Improved description extraction logic using `bolt11Tags`.
    - Unified status handling across wallet components with a new `status` prop.

- **Bug Fixes**
    - Ensured robust status verification and configuration checks with fixed conditional checks.

- **Refactor**
    - Replaced `enabled` properties with `status` for standardized state handling.

- **Style**
    - Added styling for warning indicators in wallet components.

- **Documentation**
    - No visible changes to end-users.

- **Chores**
    - Internal code clean-up and optimizations with no direct impact on end-user functionalities.

- **Revert**
    - Removed the link to "wallet logs" from the wallet page for a cleaner interface.


ekzyis:
Regarding passwords:

I envisioned that we stay close to bitcoin wallet UI/UX so we wouldn't prompt for a password but simply generate a BIP39 mnemonic for them that they should backup. I would prefer to avoid user-generated passwords. I see them as security holes since humans are bad at being random. Since it should be a familiar experience for most stackers, I think it would also meet expectations and thus be good UX. That's also not only what bitcoin wallets do, that's also what other websites with E2EE do (e.g. Matrix, RocketChat).

We could let stackers generate the optional BIP39 passphrase.

We could also use this passphrase for everything, including E2EE stuff (#56, #519) in the future.

> so for stackers that want encryption, I'll need a UI for prompting for the password to spend or to display the pairing phrase in a form

Is there no such concept as unlocking a wallet? This sounds like we need to prompt for the password for every payment. Is that true?

I envision unlocking in a browser context to work like loading decrypted content from localStorage into memory. There are security implications to consider: global variable vs only available inside a closure vs [web worker](https://thenewstack.io/leveraging-web-workers-to-safely-store-access-tokens/) etc.

huumn:
> This sounds like we need to prompt for the password for every payment. Is that true?

No, just once for each session.

components/item-act.js
```diff
@@ -271,7 +271,7 @@ export function useZap () {
     const item = cache.readFragment({
       id: `Item:${id}`,
       fragment: gql`
-        fragment ItemMeSats on Item {
+        fragment ItemMeSatsZap on Item {
           meSats
```
huumn, line 275: these were giving me duplicate fragment name warnings ... the fragment names are really used so I gave them slightly different names

components/form.js
```diff
@@ -1047,7 +1047,7 @@ function Client (Component) {
     const [,, helpers] = useField(props)
 
     useEffect(() => {
-      helpers.setValue(initialValue)
+      initialValue && helpers.setValue(initialValue)
```
huumn, line 1050: Without this, I was getting "controlling uncontrolled prop" errors

components/webln/index.js
```diff
@@ -3,29 +3,39 @@ import { LNbitsProvider, useLNbits } from './lnbits'
...
 
 const storageKey = 'webln:providers'
 
+export const Status = {
+  Initialized: 'Initialized',
+  Enabled: 'Enabled',
+  Locked: 'Locked',
+  Error: 'Error'
+}
+
```
huumn, line 33: I replaced the use of a state pair `(initialized, enabled)` with this enum like state for a wallet's status.

components/webln/index.js
```diff
@@ -107,9 +115,8 @@ function RawWebLNProvider ({ children }) {
     })
   }, [setEnabledProviders])
 
-  const value = { provider: { ...provider, sendPayment: sendPaymentWithToast }, enabledProviders, setProvider }
   return (
-    <WebLNContext.Provider value={value}>
+    <WebLNContext.Provider value={{ provider: provider ? { sendPayment: sendPaymentWithToast } : null, enabledProviders, setProvider }}>
```
huumn, line 119: I wanted a way for consumers of this context to not need to understand `provider` to use it ... so if it can be used, it's set. if it can't be used, it's null

pages/settings/wallets/index.js
```diff
@@ -27,12 +29,18 @@ export default function Wallet ({ ssrData }) {
       <div className='py-5 w-100'>
         <h2 className='mb-2 text-center'>attach wallets</h2>
         <h6 className='text-muted text-center'>attach wallets to supplement your SN wallet</h6>
+        <div className='text-center'>
+          <Link href='/wallet/logs' className='text-muted fw-bold text-underline'>
+            wallet logs
+          </Link>
+        </div>
```
huumn, line 36: I moved the attached wallet logs link to the attached wallet page because it didn't feel like it belonged on the `/wallet` page.

As a thought experiment, we could put every button on the site on the homepage and everything is "one click" away. That's obviously silly, so imho "one click" is not something we should optimize for singularly.

components/wallet-card.js
```diff
@@ -5,12 +5,33 @@ import Gear from '@/svgs/settings-5-fill.svg'
...
+import { Status } from './webln'
+
+export const isConfigured = status => [Status.Enabled, Status.Locked, Status.Error, true].includes(status)
+
+export function WalletCard ({ title, badges, provider, status }) {
+  const configured = isConfigured(status)
+  let indicator = styles.disabled
+  switch (status) {
+    case Status.Enabled:
+    case true:
```
huumn, line 17: this handles boolean status for serverside wallets

components/wallet-card.js
```diff
@@ -5,12 +5,33 @@ import Gear from '@/svgs/settings-5-fill.svg'
...
+      indicator = styles.success
+      break
+    case Status.Locked:
+      indicator = styles.warning
+      break
+    case Status.Error:
+      indicator = styles.error
```
huumn, line 24: indicator light is red only when we have errors

pages/settings/wallets/lnc.js
```diff
@@ -0,0 +1,119 @@
...
+        />
+        <PasswordInput
+          label={<>password <small className='text-muted ms-2'>optional</small></>}
+          name='password'
+          initialValue={defaultPassword ? '' : config?.password}
+          newPass={config?.password === undefined || defaultPassword}
+          hint='encrypts your pairing phrase when stored locally'
+        />
+        <ClientCheckbox
+          disabled={status !== Status.Enabled || isDefault || enabledProviders?.length === 1}
```
ekzyis, line 84: Shouldn't this be `!configured` instead of `status !== Status.Enabled` like for the other wallets?

huumn: I'm not sure. The checkbox rules were a bit unclear to me in examples.

ekzyis: > I'm not sure

In that case, I am sure that it should be `!configured` since it doesn't make sense to change the checkbox value (to make it a default) if it isn't even configured.

> The checkbox rules were a bit unclear to me in examples.

It tries to handle these three cases to allow changing the default value:

1) if wallet isn't configured, it's not a default and you can't make it a default
2) if it's already a default, you can't remove the default value. unsetting the default only works by making another provider the default. it's weird UX but it solves the issue that some provider has to be the default and I think it would be even weirder UX if making another provider the default is implicit.
3) if it's the only enabled provider, it must stay the default

_mhh, 3) might be included in 2)_

docker-compose.yml
```diff
@@ -367,6 +369,37 @@ services:
...
+      - '--uipassword=password'
+      - '--lnd-mode=remote'
+      - '--network=regtest'
+      - '--remote.lit-debuglevel=debug'
+      - '--remote.lnd.rpcserver=stacker_lnd:10009'
+      - '--remote.lnd.macaroonpath=/lnd/data/chain/bitcoin/regtest/admin.macaroon'
+      - '--remote.lnd.tlscertpath=/lnd/tls.cert'
+      - '--autopilot.disable'
+      - '--pool.auctionserver=test.pool.lightning.finance:12010'
+      - '--loop.server.host=test.swap.lightning.today:11010'
```
ekzyis, line 402: Weird, Github must have swallowed my comment here yesterday.

I tried to setup LNC by running the commands provided in the help but couldn't get it to work:

```
f72080b352ec:/# litcli accounts create --balance 100000
[litcli] unable to read macaroon path : open /root/.lit/mainnet/lit.macaroon: no such file or directory
f72080b352ec:/# litcli --macaroonpath /lnd/data/chain/bitcoin/regtest/admin.macaroon accounts create --balance 100000
[litcli] rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
f72080b352ec:/# litcli --macaroonpath /lnd/data/chain/bitcoin/regtest/admin.macaroon --tlscertpath /lnd/tls.cert accounts create --balance 100000
[litcli] rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
```

Am I doing something wrong?

huumn: `litd` is passed connection args on start. To run `litcli` and have it communicate with `lnd` you need to tell it where it is. Otherwise it assumes it's built-in.

ekzyis: Ahh, you mean built-in as in [integrated mode](https://docs.lightning.engineering/lightning-network-tools/lightning-terminal/run-litd#integrated-mode)? I see

huumn: yep

huumn: So the issue is these command need `--rpcserver localhost:8444` because the default port doesn't have tls.

I'm going to push an `sndev stacker_litcli` command that does this in a minute.

PR #1111, opened 2024-04-26
Store hashed and salted email addresses

### LINKED ISSUE #622: Store hash of email addresses rather than the plain text?
This is something that I've been background thinking about for awhile. 

For our purposes, we don't need to store email addresses that aren't subscribed to the newsletter.

I've been wondering if it might make sense to hash them with some salt

SatsAllDay:
## Description
Closes #622 
This PR updates the internals of SN to no longer persist email addresses used to login in plaintext, but rather stores a salted and hashed version of them instead.

* The salt is provided via `EMAIL_SALT` env var. ⚠️ new env var ⚠️ 
* The salted email is then hashed using `SHA256`, and then persisted in hex format in the db at `users.emailHash`.
* ~`users.email` column is removed from the DB~ The column is no longer removed, but all values are nulled-out via a migration job.
* The user settings page no longer shows your current linked email address (if you have one linked).
* New users signing up via email are still automatically enrolled in the newsletter, so no functional change there.
* New users signing up via email get an obfuscated nym by default - the first 10 chars of the salted hash of their email address. Similar to how we populate nyms for users signing up using key-based auth like LNAuth.
* For fun, I also added `./sndev email` to launch mailhog's inbox in your browser, for easy access

## Screenshots
User settings page allowing you to unlink email:
image

## Additional Context
The trickiest part of this PR is the DB migration. Since we are salting the emails before hashing them, using a secret salt that is not to be checked in to the codebase, but rather loaded via env var, it becomes challenging to perform the migration of existing email addresses to their salted/hashed versions in a reasonable way.

The approach I landed on is as follows:
1. DB migration adds the new `emailHash` column with uniqueness constraint.
2. It then submits a `saltAndHashEmails` job to `pgboss`, to be executed in 10 minutes in the future. This is slightly concerning because it leaves a window of time where the schema will match the code, but emails will not be hashed and therefore anyone trying to login via email during that time will not be able to do so. Maybe there's a better way to _quickly_ run this second part of the migration?
3. The DB migration also defines a function, which accepts a salt as a parameter, and then salts and hashes emails in the users table, then deletes the plaintext emails by ~removing the column, and removes the index~ assigning all rows to `NULL`. This function will be called by the worker.
4. The worker process picks up the job and invokes the function defined in step 3 above, providing the salt as an argument, loaded from the env var.

This feels super hacky, but I couldn't think of a better way without:
1. Compromising the integrity of the salt, or
2. Having a multitude of DB calls by using the prisma ORM to do this, since updating many rows with different values doesn't seem supported in a single query

As you an see, the approach was previously removing the `user.email` column altogether. However, that caused prisma to get upset, because the column was not removed via the migration script directly, but rather via an async job triggered through the worker. So for now, I kept `user.email`, and just nulled the values instead. It should be simple to remove the column in a future migration, after this one (and the async job) has executed, but it shouldn't be part of this PR.

## Checklist
- [X] Are your changes backwards compatible?
No, these changes replace `users.email` with `users.emailHash` in the database.
Yes, QA'd heavily.
- [X] For frontend changes: Tested on mobile?
N/A
- [X] Did you introduce any new environment variables? If so, call them out explicitly in the PR description.
Yes, `EMAIL_SALT`. I added it to `.env.development`, but not `.env.production`, since it's sensitive. This will need to be added wherever secrets are managed.

## A note on testing
This is a pretty critical piece of functionality, for users that auth via email. It's best to thoroughly test this from many angles. The following are the test cases/flows that I've considered and tested locally. Hopefully this gives some level of confidence in the changes. It also provides a good baseline for regression coverage after any subsequent commits are pushed to the PR.

1. Sign up new email
    1. Sign up as a new user via email address.
    2. Receive the magic link (verify that the email template implies creating a new user, not logging in to an existing account)
    3. Follow the magic link
    4. Land back in SN, authenticated as the new user
2. Log in with existing email
    1. In a fresh browser session, log in using the email provided in test 1 above
    2. Receive the magic link (verify that the email template implies logging in to an existing user, not creating a new account)
    3. Follow the magic link
    4. Land back in SN, authenticated as the pre-existing user
3. Add an unused, unverified email to an existing account
    1. Login to SN using some mechanism other than email. In my case, I linked to a GitHub account via OAuth
    2. Go to settings and supply an email address to link to the account
    3. Receive the magic link (assert that the link implies logging in to an existing account, not a new account)
    4. Follow the magic link
    5. Land back on SN settings page as the same original user, but now verify that the account has an email linked
4. Unlink an email address from an existing account
    1. login to an account that has auth besides email associated (e.g. GitHub)
    2. unlink the email address via settings page
    3. confirm that logging in via that email would create a new user
4. Add an occupied email to an existing account
    1. Login to SN using some mechanism other than email. In my case, I linked to a GitHub account via OAuth
    4. Go to settings and supply an email address to link to the account, BUT use an email address that is already associated with another account
    5. Verify that this operation fails, indicating that the supplied email is already in use by another account 
6. Login using users created via `./sndev login`
    1. Run `./sndev login user1`
    2. Launch the provided link to login
    3. Successfully authenticate and land in SN as user1
7. Login via email/magic link using users created from `./sndev login` 
    1. Using the user created in test 5 above, login to SN by entering the email address e.g. `user1@sndev.team`
    2. Receive the magic link (verify that the email implies an existing user)
    3. Follow the link
    6. Land in SN, authed as user1


pages/api/graphql.js
```diff
@@ -56,6 +56,7 @@ export default startServerAndCreateNextHandler(apolloServer, {
     const apiKey = req.headers['x-api-key']
     let session
     if (apiKey) {
+      // TODO replace email below? remove it?
       const [user] = await models.$queryRaw`
       SELECT id, name, email, "apiKeyEnabled"
```
SatsAllDay, line 61: @ekzyis can you explain why we select `id`, `name`, and `email` here? I see that we use those fields to populate the `sessionFields` object a few lines below, but I guess I'm wondering the significance of them? Are all of them necessary? I am asking because with this PR, I am removing `email` and replacing it with `emailHash`, so obviously we can't keep `email` in this query. I can replace it with `emailHash`, if that's valuable, but I'm not sure if it is? Is this just the list of fields we make available via `me` in various parts of the server-side code? I don't see `me.email` used anywhere, so presumably this change would be safe to make.

Thoughts?

ekzyis: I just selected these fields because that's how NextJS JWTs looked like by default. But only `id` and `name` are necessary.

>  Is this just the list of fields we make available via `me` in various parts of the server-side code? I don't see `me.email` used anywhere, so presumably this change would be safe to make.

Yes and I share your conclusion

SatsAllDay:
@huumn when you get a chance, can you take a look at the approach outlined in the PR description here? I still want to review the code for any cleanup or gaps before moving it out of draft state, but before I polish, I wanted a 👍 / 👎 on the approach and results/implications

ekzyis (commented):
> This is slightly concerning because it leaves a window of time where the schema will match the code, but emails will not be hashed and therefore anyone trying to login via email during that time will not be able to do so.

You mean that during login, the magic link will call the new `getUserByEmail` where the email is hashed for the lookup and thus will not find the user if the migration didn't finish yet? 

I don't know how many people login using email on average every 10min but I would guess it's a low number so I think this is okay.

But we could use `emailHash = hash(email) OR email = email` for the lookup if we think that's important enough, no?

Only looked at the code in Github, but lgtm

prisma/migrations/20240426144110_email_hash/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+CREATE UNIQUE INDEX "users.email_hash_unique" ON "users"("emailHash");
+
+-- hack ... prisma doesn't know about our other schemas (e.g. pgboss)
+-- and this is only really a problem on their "shadow database"
+-- so we catch the exception it throws and ignore it
+CREATE OR REPLACE FUNCTION submit_migrate_existing_user_emails_job() RETURNS void AS $$
+    BEGIN
+        -- Submit a job to salt and hash emails after the updated worker has spun-up
+        INSERT INTO pgboss.job (name, data, priority, startafter, expirein)
+        SELECT 'saltAndHashEmails', jsonb_build_object(), -100, now() + interval '10 minutes', interval '1 day';
```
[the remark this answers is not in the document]
SatsAllDay: Copy/paste from a different job insertion.

Conceptually I understand the purpose of `priority` but in practice, I'm not sure of the implications, especially in this case. I suppose this job might be higher priority than something like re-indexing items, so I could see an argument to set it to something higher than `-100`, but am definitely open to input.

SatsAllDay:
> You mean that during login, the magic link will call the new `getUserByEmail` where the email is hashed for the lookup and thus will not find the user if the migration didn't finish yet?

Precisely.

> I don't know how many people login using email on average every 10min but I would guess it's a low number so I think this is okay.

I didn't think it would be a huge deal, but I didn't like the idea of breaking login even temporarily.

 
> But we could use `emailHash = hash(email) OR email = email` for the lookup if we think that's important enough, no?

Yes! This should work to solve the issue, if we feel it's important enough to avoid the brief outage.

> Only looked at the code in Github, but lgtm

Cool, thank you! I will wait for k00b's input on approach before officially moving it out of draft state, but I think this is pretty much good to go, if he's on board with the approach.

prisma/migrations/20240426144110_email_hash/migration.sql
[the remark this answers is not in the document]
ekzyis: > I'm not sure of the implications, especially in this case

Ah ok, me neither haha, I just expected to not see an explicit priority in that case since you don't have to set it yourself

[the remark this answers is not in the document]
SatsAllDay: Yea, I copied it from the indexing bookmarked items migration job.

This is probably a good candidate to run at default priority.

huumn:
I think overall the approach sounds well worked out!

> This is slightly concerning because it leaves a window of time where the schema will match the code

This window isn't a huge deal. We only get a couple email logins per day I'd guess so odds are low this a problem.

In the past, when I migrated jwts, I kept the original jwt code in as a fallback should someone come along with an old jwt.

We could do something similar here just in case: If there's no matching `emailHash`, check `email`, if neither match, fail.

SatsAllDay:
> I think overall the approach sounds well worked out!

Awesome!

> This window isn't a huge deal. We only get a couple email logins per day I'd guess so odds are low this a problem.
> 
> In the past, when I migrated jwts, I kept the original jwt code in as a fallback should someone come along with an old jwt.
> 
> We could do something similar here just in case: If there's no matching `emailHash`, check `email`, if neither match, fail.

OK, that seems reasonable. I'll make that change, clean up the PR if there's anything left to tidy, then move it out of draft state. Thanks!

SatsAllDay:
Quick update, I think I have these final changes working, but as I am testing with users created from `./sndev login <nym>`, it is not working, so I am trying to figure out how fix that.

SatsAllDay:
OK, ready for review!

pages/api/auth/[...nextauth].js
```diff
@@ -217,7 +200,49 @@ const providers = [
...
+      }
+      // HACK! This is required to satisfy next-auth's check here:
+      // https://github.com/nextauthjs/next-auth/blob/5b647e1ac040250ad055e331ba97f8fa461b63cc/packages/next-auth/src/core/routes/callback.ts#L227
+      // since we are nulling `email`, but it expects it to be truthy there.
+      // Since we have the email from the input request, we can copy it here and pretend like we store user emails, even though we don't.
+      if (user) {
+        user.email = email
+      }
+      return user
+    }
```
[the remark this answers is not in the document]
Comment (SatsAllDay) [reply pages/api/auth/[...nextauth].js:244]: No, it does accept an object.

lib/crypto.js
```diff
@@ -0,0 +1,9 @@
+import { createHash } from 'node:crypto'
+
+export function hashEmail ({
+  email,
+  salt = process.env.EMAIL_SALT
+}) {
+  const saltedEmail = `${email.toLowerCase()}${salt}`
+  return createHash('sha256').update(saltedEmail).digest('hex')
+}
```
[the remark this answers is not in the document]
SatsAllDay: @huumn  thoughts on this one?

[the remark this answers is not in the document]
huumn: I think we are good as is.

.env.development
```diff
@@ -36,6 +36,10 @@ LNWITH_URL=
 LOGIN_EMAIL_SERVER=smtp://mailhog:1025
 LOGIN_EMAIL_FROM=sndev@mailhog.dev
 
+# email salt
+# openssl rand -hex 32
+EMAIL_SALT=202c90943c313b829e65e3f29164fb5dd7ea3370d7262c4159691c2f6493bb8b
```
SatsAllDay, line 41: This will need to be added to prod environments (with a different value, of course), however that works. But since it's a secret, I didn't add it to `.env.production`.

api/resolvers/user.js
```diff
@@ -44,7 +45,7 @@ async function authMethods (user, args, { models, me }) {
 
   return {
     lightning: !!user.pubkey,
-    email: user.emailVerified && user.email,
+    email: !!(user.emailVerified && user.emailHash),
```
SatsAllDay, line 48: Convert email to a bool instead of showing the email value, which we no longer have.

lib/crypto.js
```diff
@@ -0,0 +1,9 @@
+import { createHash } from 'node:crypto'
+
+export function hashEmail ({
+  email,
+  salt = process.env.EMAIL_SALT
```
SatsAllDay, line 5: Default to the salt from env, but in theory we can accept others, if that ever becomes useful

pages/api/graphql.js
```diff
@@ -57,7 +57,7 @@ export default startServerAndCreateNextHandler(apolloServer, {
     let session
     if (apiKey) {
       const [user] = await models.$queryRaw`
-      SELECT id, name, email, "apiKeyEnabled"
+      SELECT id, name, "apiKeyEnabled"
```
SatsAllDay, line 60: Remove this since we don't use email in server-side sessions.

pages/settings/index.js
```diff
@@ -714,15 +714,8 @@ function AuthMethods ({ methods, apiKeyEnabled }) {
           return methods.email
             ? (
               <div key={provider} className='mt-2 d-flex align-items-center'>
-                <Input
```
SatsAllDay, line 717: Remove the input field, just show a button to unlink email, if applicable

prisma/schema.prisma
```diff
@@ -162,10 +163,10 @@ model Wallet {
 }
 
 model WalletLog {
-  id        Int @id @default(autoincrement())
+  id        Int      @id @default(autoincrement())
```
SatsAllDay, line 166: Lint whitespace changes.

sndev
```diff
@@ -409,14 +428,16 @@ sndev__login() {
   # "SNDEV-TOKEN3_0W_PhDRZVanbeJsZZGIEljexkKoGbL6qGIqSwTjjI"
   # next-auth concats the token with the secret from env and then sha256's it
   token="d5fce54babffcb070c39f78d947761fd9ec37647fafcecb9734a3085a78e5c5e"
+  salt="202c90943c313b829e65e3f29164fb5dd7ea3370d7262c4159691c2f6493bb8b"
   # upsert user with nym and nym@sndev.team
+  email="$1@sndev.team"
   docker__exec db psql -U sn -d stackernews -q <<EOF
     INSERT INTO users (name) VALUES ('$1') ON CONFLICT DO NOTHING;
-    UPDATE users SET email = '$1@sndev.team' WHERE name = '$1';
+    UPDATE users SET email = '$email', "emailHash" = encode(digest(LOWER('$email')||'$salt', 'sha256'), 'hex') WHERE name = '$1';
```
SatsAllDay, line 436: Store the hashed email in the DB for `./sndev login`

SatsAllDay (commented):
Did a self-review, but I also have a lot of comments in the code that explains things, so hopefully all is clear

huumn:
This looks good at a glance. I'll try to get it review by tomorrow.

huumn:
Works perfectly. You're a surgeon @SatsAllDay!

SatsAllDay:
I've often compared making code changes to surgery, but I don't think I've ever outwardly verbalized it. You go in to this "living" thing, and make very specific changes to alter certain behavior, while (hopefully) leaving the rest of it as-is, unaffected. Sounds like surgery to me!

I've never performed actual surgery on a human, but I wonder if surgeons would be offended by the comparison :P

PR #1127, opened 2024-04-29
Fix unintended sharing of wallets and logs

### LINKED ISSUE #927: localstorage is shared in a shared browser
Got a nostr DM from someone who logged into another account in the same browser and their new account had access to their other account's NWC.

Given we store NWC in localstorage this makes sense. Can we create authenticated localstorage? If not, we should at least clear localstorage when someone logs out.

I don't expect this to be a common problem, but it's worth not letting this happen.

ekzyis:
## Description

Close #927

TODO:
- [x] suffix localStorage key for NWC, LNbits
- [x] suffix localStorage key for LNC
- [x] suffix IndexedDB database for wallet logs
- [x] detach attached wallets on logout
- [x] delete wallet logs on logout (done in #1101) 

## Checklist

Yes, stackers will still be able to access their current attached wallets without attaching them again. If a wallet wasn't found with the new key, we will fallback to the old key and migrate it to the new key.

Yes, but I couldn't test LNC since I don't know how to connect it. See https://github.com/stackernews/stacker.news/pull/1104#discussion_r1582239915.

### DIFF
--- a/components/logger.js
+++ b/components/logger.js
@@ -126,14 +126,14 @@ export function useServiceWorkerLogger () {
 const WalletLoggerContext = createContext()
 const WalletLogsContext = createContext()
 
-const initIndexedDB = async (storeName) => {
+const initIndexedDB = async (dbName, storeName) => {
   return new Promise((resolve, reject) => {
     if (!window.indexedDB) {
       return reject(new Error('IndexedDB not supported'))
     }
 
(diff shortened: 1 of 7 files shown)


huumn (changes requested):
looks good. Just need to resolve merge conflicts

ekzyis:
> looks good. Just need to resolve merge conflicts

Done

PR #1128, opened 2024-04-29
use random text for uploadMarker

### LINKED ISSUE #1096: Image alt attributes leak data

When users upload an image, we use the provided filename in the markdown image syntax for the image alt attribute. We use it as a marker so we know which text we can replace when upload is finished:

https://github.com/stackernews/stacker.news/blob/d7ecbbae3a86ac62d3ad30f5affa8d240c722ce0/components/form.js#L316-L343

However, this can leak information like when the image was taken as in this example:

``

Overwrite filename with random string or simply delete string for the image alt attribute. 

n/a

**Additional context**

When to overwrite filenames or delete them is not clear. If we do this on upload, then the user might be confused which image was uploaded since the filename no longer matches what they uploaded. The code also needs unique identifiers so simply deleting it on upload would break the `onSuccess` callback. We could do this after `onSuccess` though.

We could do this on submit on the client or server but then we need to parse for image alt attributes and then it's not clear which was provided by the user and which should be replaced. We could limit this replacement to images from `m.stacker.news` though.

[_sats for this ticket go to @Darth-Coin even though he probably does not want them_](https://stacker.news/items/512387)

itsrealfake:
in order to prevent leaking metadata in filenames, use a random string as the filename

## Description

closes #1096 

## Checklist

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -32,7 +32,6 @@ import Thumb from '@/svgs/thumb-up-fill.svg'
 import Eye from '@/svgs/eye-fill.svg'
 import EyeClose from '@/svgs/eye-close-line.svg'
 import Info from './info'
-import crypto from 'crypto'
 
 export function SubmitButton ({
   children, variant, value, onClick, disabled, nonDisabledText, ...props
@@ -317,8 +316,7 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, onKe
               ref={imageUploadRef}
               className='d-flex align-items-center me-1'
               onUpload={file => {
-                const altText = crypto.randomUUID()
-                const uploadMarker = `![Uploading ${altText}…]()`
+                const uploadMarker = `![Uploading ${file.name}…]()`
                 const text = innerRef.current.value
                 const cursorPosition = innerRef.current.selectionStart || text.length
                 let preMarker = text.slice(0, cursorPosition)
--- a/components/form.js
+++ b/components/form.js
@@ -32,6 +32,7 @@ import Thumb from '@/svgs/thumb-up-fill.svg'
 import Eye from '@/svgs/eye-fill.svg'
 import EyeClose from '@/svgs/eye-close-line.svg'
 import Info from './info'
+import crypto from 'crypto'
 
 export function SubmitButton ({
   children, variant, value, onClick, disabled, nonDisabledText, ...props
@@ -316,7 +317,8 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, onKe
               ref={imageUploadRef}
               className='d-flex align-items-center me-1'
               onUpload={file => {
-                const uploadMarker = `![Uploading ${file.name}…]()`
+                const altText = crypto.randomUUID()
+                const uploadMarker = `![Uploading ${altText}…]()`
                 const text = innerRef.current.value
                 const cursorPosition = innerRef.current.selectionStart || text.length
                 let preMarker = text.slice(0, cursorPosition)
--- a/components/image.js
+++ b/components/image.js
@@ -189,6 +189,7 @@ export const ImageUpload = forwardRef(({ children, className, onSelect, onUpload
 
   const s3Upload = useCallback(async file => {
     const img = new window.Image()
+    const originalFileName = file.name // store original file name before removing exif data & renaming file
     file = await removeExifData(file)
     return new Promise((resolve, reject) => {
       img.onload = async () => {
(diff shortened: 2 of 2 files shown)


itsrealfake:
Note, i have, as yet, been unable to get the repo to run.. so i'm shooting in the dark for a "fix".

> Prevent data leakage via image alt attributes (#1096) 	✅ 	

yay

> Ensure uniqueness of identifiers in `onSuccess` callback (#1096) 	❓ 	It's unclear how the new UUIDs interact with the `onSuccess` callback mechanism.

the new UUID doesn't interact with onSuccess. UUID will interpolate as a string, replacing the filename. i have no idea if there's a different UUID library already included in the repo that we could/should use, but certainly a different mechanism for generating a random string is an option worth considering.


> Avoid user confusion with filename changes (#1096) 	❌ 	The change does not address potential user confusion regarding mismatch between uploaded filenames and displayed identifiers.

i think it's straightforward enough to change the displayed name on upload. not opposed to doing something more complicated. seems challenging to avoid user confusion if we're changing the names.

> Limit alt attribute replacement to `m.stacker.news` images (#1096) 	❌ 	The change applies universally and does not restrict the replacement to images from `m.stacker.news`.

this was not a requirement of the ticket.

itsrealfake:
Converting to draft, i think i misunderestimated how straightfoward the change will be. 

seems like the `onUpload` & `onSuccess` functions here `./components/image.js` could be adjusted to support the fix.

ekzyis (changes requested):
Hi, welcome to contributing to SN and thanks for picking up #1128!

I pulled your code and tried  to upload an image but hit this error:

> TypeError: crypto__WEBPACK_IMPORTED_MODULE_24___default(...).randomUUID is not a function

Was there a problem with testing the code? For example, because [`./sndev start` doesn't work for you yet](https://github.com/stackernews/stacker.news/discussions/1059#discussioncomment-9248716)?

Anyway, you can't use `node:crypto` in the browser. It works when I remove the import and rely on the native [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID).

However, this breaks the `onSuccess` callback below since it doesn't find the marker anymore. It relied on the filename. So we need to find something we can use in both callbacks.

Maybe hashing the filename would work since you can generate the same hash in `onUpload`.

You'd also want to keep the file extension.

---

edit: Oh, guess I reviewed too soon, didn't see that you commented and realized the same already :eyes:

itsrealfake:
thanks for the quick review @ekzyis

correct: i'm still not able to get `./sndev start` working. else, i would have quickly realized i was trying to use a node library in the browser (whoops!)

hash of the filename is a fine idea and likely less cumbersome than crypto.someUUIDgeneratingThing() ... though we may run into an issue when two users upload `image.jpg` unless some randomness is added.

i'm thinking it could be easier to make a filename change inside of or near the [`removeExifData` function](https://github.com/stackernews/stacker.news/blob/df631878e0e586b83e9fd98be1854f2ec59b96d7/components/image.js#L192) in `componenets/image.js` and thinking about how best to replace the text in `onSuccess`... 

(fwiw, the image upload here on github just returns a UUID.)

ekzyis:
> though we may run into an issue when two users upload image.jpg unless some randomness is added.

That wouldn't be an issue, the altname is only required for the upload process itself as markers. They aren't identifiers, we generate unique URLs from m.stacker.news for every upload.

> i'm thinking it could be easier to make a filename change inside of or near the removeExifData function

Oh, I think you're right! We can immediately override the filename and then you might not need to change anything else 👀

That sounds like the way to go

ekzyis:
Wait, I hope I am not confusing you, but I think I was thinking too complicated while writing the ticket and here. I think all we need is this patch:

```diff
diff --git a/components/form.js b/components/form.js
index 8480ba6c..95f9135a 100644
--- a/components/form.js
+++ b/components/form.js
@@ -331,7 +331,7 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, onKe
               }}
               onSuccess={({ url, name }) => {
                 let text = innerRef.current.value
-                text = text.replace(`![Uploading ${name}…]()`, `![${name}](${url})`)
+                text = text.replace(`![Uploading ${name}…]()`, `![](${url})`)
                 helpers.setValue(text)
                 const s3Keys = [...text.matchAll(AWS_S3_URL_REGEXP)].map(m => Number(m[1]))
                 updateImageFeesInfo({ variables: { s3Keys } })
```

This simply removes the filename after upload finished since we no longer need it. As mentioned in my previous comment, we only needed it as a marker.

itsrealfake:
dang... well, reasons happen. maybe we can have a beer over it later

anyway, new PR https://github.com/stackernews/stacker.news/pull/1130

PR #1129, opened 2024-04-30
Force post options open when dirty or on errors

### LINKED ISSUE #491: Hidden user feedback on errors in options
**Description**

I just tried to post something and it didn't work and I got not feedback why:

2023-09-12-081439_743x368_scrot

Turned out it was because of an error in the options which were collapsed:

2023-09-12-081633_749x605_scrot

**Steps to Reproduce**

- Create error in options
- Hide options
- Try to post

**Expected behavior**

I get feedback why I can't post even if error is in collapsed options

**Additional context**

The percentage error message is also not displayed correctly below the field

### LINKED ISSUE #1045: Hidden prepopulated section entry potentially missed for deletion if user does not unroll the hidden section
**Description**
Check [this post](https://stacker.news/items/499422) for instance. I had started posting something a while ago (the 3blue1brown video), but then decided against it. Today, I decided to post something else (in this case, the link to Calle's Ecash post). I noticed the title and link were still populated by the title and link I had inputted for the 3blue1brown video. I deleted them and replaced them with the link for the Ecash video link and title. I did not click the options dropdown menu, so didn't see that the context was still prepulated. Resulting in this mismatch between current link/title and context.

**Steps to Reproduce**
1. Start a link post. Fill in title, link and context.
2. Do not click post. Navigate back to the main page.
3. Start a new post link. Notice the title and link being prepopulated. Delete these entries and replace them with new entries.
4. Optionally notice that if you click on options, the context entry is still prepopulated. If you do not notice this because it is collapsed, you'll post without noticing.

**Expected behavior**
Two potential solutions:
- Delete prepopulated context if user deletes prepopulated title and link entres
- If anything is prepopulated, automatically unroll the options section so that the user notices the hidden prepopulated entries.

**Additional context**
I remember another recent post by another user who encountered this same issue. I'll update here if I find it.

benalleng:
## Description

Closes #1045, closes #491

This will force the options tab open on post render if the inputs are dirty or opens them on submit if the options contain errors

## Additional Context

I designed it such that it should only be forced open when the advpostform contains errors. It will also start opened if the form is dirty at all, errors or not

If the options starts open because it is dirty or fills in the form with some valid inputs and then closes it, it will not reopen

## Checklist


components/accordian-item.js
```diff
@@ -3,12 +3,41 @@ import AccordionContext from 'react-bootstrap/AccordionContext'
...
+      setShow(hasForwardError || hasBoostError || hasMaxBidError ? formik?.isSubmitting : undefined)
+    }
+  }, [formik?.isSubmitting])
+
+  useEffect(() => {
+    if (show !== undefined && activeEventKey !== eventKey) {
+      decoratedOnClick()
+    }
+  }, [show])
+
```
huumn, line 40: I think all of this logic should be in `adv-post-form.js` ... `AccordianItem` is used in many places throughout the code base, and we don't want this affecting unrelated code.

I could be missing something though.

huumn (changes requested):
I left my main feedback as a comment.

components/accordian-item.js
benalleng: I had initially moved it because of what i viewed as a substaintially level of added complexity in https://github.com/stackernews/stacker.news/pull/1129/commits/ed3a958180d88c37588678ddf0a65523b10cfe83 vs just using the `hasForm` prop.  

Perhaps I could make it more clear about the useEffect dependencies not changing since `formik?` is null without the `hasForm` param `const formik = hasForm ? useFormikContext() : null`.  

Happy to move it back up one level to `adv-post-form` if you'd prefer that commit

benalleng: moved the logic in https://github.com/stackernews/stacker.news/pull/1129/commits/e4d273c19994af69999adbb561f9caa9fee1b460

huumn:
This does its job and fixes the bug. I didn't love the use of a ternary `show` ... I thought it was confusing so I removed it.

Note to myself: This whole accordion and localstorage draft thing needs a zoomed out refactor because it's a big hack. We keep tacking stuff onto our existing forms but the design of those were unconscious of all the things we'd add.

components/adv-post-form.js
```diff
@@ -30,18 +30,33 @@ export default function AdvPostForm ({ children, item }) {
...
+      }
+    }
+  }, [formik?.values, storageKeyPrefix])
 
-    setShow(hasForwardError || hasBoostError ? formik?.isSubmitting : undefined)
+  useEffect(() => {
+    // force show the accordian if there is an error and the form is submitting
+    const hasError = !!formik?.errors?.boost || formik?.errors?.forward?.length > 0
+    // if it's open we don't want to collapse on submit
+    setShow(show => show || (hasError && formik?.isSubmitting))
```
benalleng, line 59: On testing this branch I am not able to get it to open options on error.  I think we need to cirlce back to this setState.  I agree that the ternary behavior was confusing but it was the only way I get the errors to work.

huumn: That's true after it's manually closed. I'll fix it.

huumn: 71213179 does the ternary thing but is more explicit. 

The only reason this is needed is because of the `useEffect` dependency which we need. There has to be a better way to do this but it'll have to be part of a future form cleanup.

PR #1130, opened 2024-04-30
remove leaky text in uploadMarker

### LINKED ISSUE #1096: Image alt attributes leak data

itsrealfake:
## Description

closes #1096

## Checklist

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -331,7 +331,7 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, onKe
               }}
               onSuccess={({ url, name }) => {
                 let text = innerRef.current.value
-                text = text.replace(`![Uploading ${name}…]()`, `![${name}](${url})`)
+                text = text.replace(`![Uploading ${name}…]()`, `![](${url})`)
                 helpers.setValue(text)
                 const s3Keys = [...text.matchAll(AWS_S3_URL_REGEXP)].map(m => Number(m[1]))
                 updateImageFeesInfo({ variables: { s3Keys } })


ekzyis (approved):
Works great!

itsrealfake:
> Works great!

Swag. How does this get merged?

huumn:
I merge it when I'm in a reviewing/merging/award data entry mood.

itsrealfake:
while writing a comment which took advantage of this, i had a thought about a feature

when the user uploads their first image, a UX like 

### 1. Logged  user uploads from clipboard w/ e.g. CMD+V


### 2. Popup to save preference option

> "we automatically strip EXIF data in the upload process. 
> what would you like stacker.news to do when you upload images?
>   - do nothing
>   - show you an option dialogue to create a link to source
> (you can change this later in preferences)

### 3. Show a dialogue to optionally leave a link to source w/ pasted image.

> Would you like to add a link to the image source?
> You can change this option in user preferences.

ekzyis:
What do you mean with "link" here?

> would you like stacker.news to leave a link when paste images into a comment box?

PR #1131, opened 2024-04-30
Fix migration name

ekzyis:
## Description

We didn't deploy the migration with the bad name yet so we can simply rename the directory.

@stackernews/foss, you'll have to run `sndev delete` after this is merged

## Checklist

No, this will break any environment which already applied this migration. Hence `sndev delete`.

Made sure migration is sorted at the bottom now


SatsAllDay (approved):

hkarani:
Cool

PR #1137, opened 2024-05-01
Add explicit word break for items

### LINKED ISSUE #1125: Notifications with post titles without word breaks overflow horizontally
**Description**
On mobile, when viewing the notifications page, if a notification references a post that has a very long single word in it, the word is not broken via css, and therefore the notification overflows in the x-direction.

**Steps to Reproduce**
1. Somehow get notified of a post that has a very long word in the title
2. View notifications page

**Expected behavior**
Post titles and links in notifications should probably have some CSS word break behavior so that long words in titles don't cause overflow

**Screenshots**
IMG_9970
IMG_9969
IMG_9968
IMG_9967

**Logs**
N/A

**Environment:**
Likely only an issue on mobile, not desktop

**Additional context**
N/A

benalleng:
## Description

Closes #1125 
adds an explicit word break to the main and other item components to handle the text of long titles. Firefox handles this break by default but it seems that chrome and safari do not

## Screenshots

chrome linebreak (note the `break-all` behavior in the url where the rest of the title has `break-word`) 
image

firefox linebreak (no change from prod)
image

## Additional Context

This bit of css seems to only be necessary for chrome or safari based browsers as firefox seems to already have some default line-breaks built for slash characters. however, since I cannot explicitly state to make a linebreak on a "/" I opted to add `word-break: break-word;` instead. as a result the word breaks on firefox <-> chrome look slightly different

## Checklist

Yaaasss

Yaaasss 

Tested in firefox and chrome, not on safari

### DIFF
--- a/components/item.module.css
+++ b/components/item.module.css
@@ -90,6 +90,7 @@ a.link:visited {
     color: var(--theme-grey);
     vertical-align: text-top;
     margin-bottom: .125rem;
+    word-break: break-word;
 }
 
 .other svg {
@@ -156,6 +157,7 @@ a.link:visited {
 .main {
     display: flex;
     align-items: baseline;
+    word-break: break-word;
 }
 
 .children {
--- a/styles/globals.scss
+++ b/styles/globals.scss
@@ -320,6 +320,10 @@ $zindex-sticky: 900;
   color: var(--theme-grey) !important;
 }
 
+.text-reset {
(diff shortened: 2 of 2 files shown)


SatsAllDay (approved):
I didn't try it but the code change seems good. Screenshots would be helpful, especially to differentiate behavior between browsers?

PR #1145, opened 2024-05-04
honor mutes when sending push notifications

### LINKED ISSUE #717: Respect mutes all for push notifications
We currently *only* respect mutes for push notifications wrt replies.

SatsAllDay:
## Description
Closes #717 

This PR updates the push notification code paths to ensure that user-configured mutes are checked before sending push notifications, if applicable. The following code paths were updated:
* territory subscriptions - you only receive a push notification for a territory to which you're subscribed if the user who submitted the post is not a user you've muted
* mentions - if a user that you've muted mentions you, you will not receive a push notification for it
* user subscriptions - if for some reason you've muted and subscribed to the same user, you'll no longer receive push notifications when that user posts or comments (whichever you've subscribed to)

This PR also updates the API validation to disallow users from muting and subscribing to the same user - only one or the other is allowed. No migration is performed, so if any user got themselves into this state, they can keep it until they change it. Then they can't go back.

In order to explain this behavior to users, I updated the UI code that toggles subscriptions/mutes to show the error message from the API request in the error toast, instead of a generic "Failed to <mute|subscribe>" message.

## Screenshots
N/A

## Additional Context
N/A

## Checklist

Yes. No database changes or migrations. The only notable change is that users will no longer be able to mute and subscribe to the same user. However, if they've already done that, it will still be honored. They just can't re-configure such a thing.

Yes.

N/A

No.


api/resolvers/item.js
```diff
@@ -1227,7 +1227,7 @@ export const createMentions = async (item, models) => {
 
         // only send if mention is new to avoid duplicates
         if (mention.createdAt.getTime() === mention.updatedAt.getTime()) {
-          notifyMention(user.id, item)
+          notifyMention({ models, userId: user.id, item })
```
SatsAllDay, line 1230: I considered altering the surrounding code in this context to only invoke this `notifyMention` function for users that didn't mute OP, but instead I kept that logic within `notifyMention` for consistency with the other `notifyXXX` functions. It's slightly dangerous to assume that the caller of `notifyXXX` is doing the mute filtering, IMO.

It would probably be more performant to do it here, instead of doing a DB call for each mentioned user, so I am open to refactoring if we think that's a major concern. But for consistency and all that, I did it this way.

api/resolvers/user.js
```diff
@@ -112,6 +112,19 @@ export function viewValueGroup () {
   ) vv`
 }
 
+const isMuted = async ({ models, me, mutedId }) => {
```
SatsAllDay, line 115: Extracted this to a reusable function for code de-dupe.

PR #1150, opened 2024-05-06
Fix wallet detach unavailable

ekzyis:
## Description

The detach button was missing for CLN and LN address. This fixes it by using the same code as for LND.

## Additional Context

https://stacker.news/items/530554

## Checklist

yes

I did not test this actually fixes the issue but pretty sure it will from simply reading the code


SatsAllDay:
Thank you! I'll verify this works once deployed.

PR #1160, opened 2024-05-08
lnproxy payments

huumn:
So far this just adds lnproxy to local dev for beginning to test/play around with it.

I've made a few modifications to lnproxy (just configuration/ops stuff) in this fork: https://github.com/huumn/lnproxy-relay


ekzyis:
Since we're no longer using lnproxy (see #1178), can this PR be closed?

PR #1178, opened 2024-05-16
not-custodial zap beta

huumn:
This is just the scaffolding so far - a bunch of unrun, untested code that lays out *roughly* how we'll orchestrate and organize this.

Notable things:
- implements an [lnproxy-like](https://github.com/lnproxy/lnproxy-relay) payment forwarding mechanism using carefully created and observed hold invoices
    - I didn't use lnproxy because it maintains state in-memory and requires manual recovery in the case of unexpected errors
- creates an `InvoiceForward` table to maintain the state of an attempt/ongoing/completed forward
   - an invoice forward has both an `Invoice` for the incoming payment and a `Withdrawl` for the outgoing payment (assuming the incoming payment is received)
- organizes the `createInvoice` code into its own dir `api/createInvoice` attempting to standardize the interface for receiving wallets
- extends existing invoice/withdrawal subscription code to have a special case should they have an `InvoiceForward` associated
    - these special cases in `worker/wallet` is where all the state transitions happen

TODO:
- handle zap side effects (not addressed at all yet) and *optimistically*
    - add a column to `ItemAct` to show its associate payment state, e.g. `PENDING` or `PAID`, that gets updated with the `InvoiceForward` state changes
    - show `PENDING` actions as completed to stacker who they belong to, hide from everyone else
    - show `FAILED` actions as notifications
- walk through all possible state transitions of `InvoiceForward` and edge cases
    - mindful of reentry errors and idempotentcy
    - mindful of unsuccessful state transitions and/or failed persistence of state transitions
- corresponding frontend code
- test (obvs)
- retest all autowithdrawal methods because I refactored the code


worker/wallet.js
```diff
@@ -91,10 +92,9 @@ function subscribeToHodlInvoice (args) {
     sub.on('invoice_updated', async (inv) => {
       logEvent('hodl_invoice_updated', inv)
       try {
-        // record the is_held transition
-        if (inv.is_held) {
-          await checkInvoice({ data: { hash: inv.id }, ...args })
-          // after that we can stop listening for updates
+        await checkInvoice({ data: { hash: inv.id }, ...args })
+        // after settle or confirm we can stop listening for updates
+        if (inv.is_confirmed || inv.is_canceled) {
```
huumn, line 97: hold invoices don't get called on confirmation NOR cancellation, so if we want to know what's going on with them we have to stay subscribed until the fat lady lightning sings.

components/item-act.js
```diff
@@ -67,7 +67,7 @@ export default function ItemAct ({ onClose, itemId, down, children }) {
       const existingAmount = Number(window.localStorage.getItem(storageKey) || '0')
       window.localStorage.setItem(storageKey, existingAmount + amount)
     }
-    await act({
+    console.log(await act({
```
huumn, line 70: any modifications to this file are just to allow me to test the backend portion

worker/invoiceForward.js
```diff
@@ -0,0 +1,357 @@
+import {
```
huumn, line 1: I'm leaving this file as a reference for now. I'll delete before pulling out of draft.

worker/paidAction.js
```diff
@@ -1,9 +1,11 @@
+import { getPaymentFailureStatus, hodlInvoiceCltvDetails } from '@/api/lnd'
 import { paidActions } from '@/api/paidAction'
 import { datePivot } from '@/lib/time'
 import { Prisma } from '@prisma/client'
-import { getInvoice, settleHodlInvoice } from 'ln-service'
+import { getInvoice, getPayment, payViaPaymentRequest, settleHodlInvoice } from 'ln-service'
+import { MIN_SETTLEMENT_CLTV_DELTA } from 'wallets/wrap'
 
-async function transitionInvoice (jobName, { invoiceId, fromState, toState, toData, onTransition }, { models, lnd, boss }) {
+async function transitionInvoice (jobName, { invoiceId, fromState, toState, transition }, { models, lnd, boss }) {
```
huumn, line 8: A lot of the changes in this file are the result of this simplification, where `transitionInvoice` callers only define `transition`.

docker-compose.yml
```diff
@@ -237,19 +238,40 @@ services:
...
+                bitcoin-cli -named sendtoaddress address=$$address amount=0.01 fee_rate=$$(( $$RANDOM % 25 + 1 ))
+                ((++i))
+              done
+              echo "generating block..."
+              bitcoin-cli generatetoaddress 1 "$${nodes[ $$RANDOM % $${#nodes[@]} ]}"
+              if [[ $$(bitcoin-cli estimatesmartfee 6) =~ "\\"feerate\\":" ]]; then
+                echo "fee estimation succeeded..."
+                break
+              fi
+            done
```
huumn, line 274: In order for probing fee estimation, we needed to put transactions into our automined blocks.

wallets/wrap.js
```diff
@@ -0,0 +1,154 @@
+import { createHodlInvoice, parsePaymentRequest } from 'ln-service'
+import { estimateRouteFee } from '../api/lnd'
+
+const MIN_OUTGOING_MSATS = BigInt(900) // the minimum msats we'll allow for the outgoing invoice
+const MAX_OUTGOING_MSATS = BigInt(9_000_000_000) // the maximum msats we'll allow for the outgoing  invoice
+const MAX_EXPIRATION_INCOMING_MSECS = 900_000 // the maximum expiration time we'll allow for the incoming invoice
+const INCOMING_EXPIRATION_BUFFER_MSECS = 300_000 // the buffer enforce for the incoming invoice expiration
+const MIN_INCOMING_CLTV_DELTA = 200 // the minimum cltv delta we'll allow for the incoming invoice
+const MAX_OUTGOING_CLTV_DELTA = 1500 // the maximum cltv delta we'll allow for the outgoing invoice
```
huumn, line 9: I don't think this is the value we want in prod (ideally <600), but for some reason in dev our estimation of this is way bigger than it should be. More research needed.

api/paidAction/zap.js
```diff
@@ -78,10 +91,14 @@ export async function onPaid ({ invoice, actIds }, { models, tx }) {
...
-      SELECT "userId", msats FROM forwardees
+      SELECT "userId", msats, msats AS "stackedMsats" FROM forwardees
       UNION
       SELECT ${itemAct.item.userId}::INTEGER as "userId",
-        ${itemAct.msats}::BIGINT - (SELECT msats FROM total_forwarded)::BIGINT as msats
+        CASE WHEN ${!!invoice?.invoiceForward}::BOOLEAN
+          THEN 0::BIGINT
+          ELSE ${itemAct.msats}::BIGINT - (SELECT msats FROM total_forwarded)::BIGINT
+        END as msats,
+        ${itemAct.msats}::BIGINT - (SELECT msats FROM total_forwarded)::BIGINT as "stackedMsats"
```
huumn, line 101: We want `stackedMsats` to go up, but we don't want their balance to increase if it's a p2p zap.

api/resolvers/wallet.js
```diff
@@ -482,7 +474,16 @@ const resolvers = {
...
+    satsFeePaid: w => msatsToSats(w.msatsFeePaid),
+    preimage: async (withdrawl, args, { lnd }) => {
+      try {
+        if (withdrawl.status === 'CONFIRMED') {
+          return (await getPayment({ id: withdrawl.hash, lnd })).payment.secret
+        }
+      } catch (err) {
+        console.error('error fetching payment from LND', err)
+      }
+    }
```
huumn, line 486: Before, when notification would load withdraw notifications, it would grab the preimage which isn't even queried for.

wallets/wrap.js
huumn: This is resolved in [009ae43](https://github.com/stackernews/stacker.news/pull/1178/commits/009ae4332d516afd3015aa82db8b71b6c8bb4993).

`ln-service` expects absolute cltv heights, so we need to factor in current block height. I had assumed they were deltas.

api/paidAction/index.js
```diff
@@ -213,10 +215,32 @@ export async function retryPaidAction (actionType, args, context) {
...
+  // wrap it, and return the wrapped invoice
+  const { cost, models, lnd } = context
+  const userId = await paidActions[actionType]?.invoiceablePeer?.(args, context)
+
+  if (userId) {
+    const description = await paidActions[actionType].describe(args, context)
+    const { invoice, wallet } = await createUserInvoice(userId, {
+      msats: cost * BigInt(9) / BigInt(10),
+      description,
+      expiry: 600
```
huumn, line 229: Should probably fix this magic number.

docker-compose.yml
```diff
@@ -1,4 +1,8 @@
-x-env_file: &env_file .env.development
+x-env_file: &env_file
+  - path: .env.development
+    required: true
+  - path: .env.local
+    required: false
```
huumn, line 5: This is in more recent versions of docker compose. It allows us to assume an `.env.local` exists.

wallets/wrap.js
```diff
@@ -0,0 +1,169 @@
...
+      id: string,
+      description: string,
+      description_hash: string,
+      expires_at: Date,
+      cltv_delta: number,
+      mtokens: string
+    },
+    outgoingMaxFeeMsat: number
+  }
+*/
```
huumn, line 32: This comment is out of date. We now return a hold invoice rather than `invoiceParams`.

worker/wallet.js
```diff
@@ -92,10 +92,9 @@ function subscribeToHodlInvoice (args) {
     sub.on('invoice_updated', async (inv) => {
       logEvent('hodl_invoice_updated', inv)
       try {
-        // record the is_held transition
-        if (inv.is_held) {
-          await checkInvoice({ data: { hash: inv.id }, ...args })
-          // after that we can stop listening for updates
+        await checkInvoice({ data: { hash: inv.id }, ...args })
```
huumn, line 95: We record any possible transition rather than just held ones.

wallets/wrap.js
```diff
@@ -0,0 +1,169 @@
...
+      mtokens: inv.mtokens,
+      request: invoice,
+      timeout: FEE_ESTIMATE_TIMEOUT_SECS
+    })
+
+    const { current_block_height: blockHeight } = await getHeight({ lnd })
+    // we want the incoming invoice to have MIN_SETTLEMENT_CLTV_DELTA higher cltv delta than
+    // the expected ctlv_delta of the outgoing invoice's route
+    wrapped.cltv_delta = Number(timeLockDelay) + inv.cltv_delta - blockHeight + MIN_SETTLEMENT_CLTV_DELTA
+    console.log('routingFeeMsat', routingFeeMsat, 'timeLockDelay', timeLockDelay, 'blockHeight', blockHeight)
```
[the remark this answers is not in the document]
huumn: Added the comment to the code locally.

lib/cln.js
```diff
@@ -16,6 +16,8 @@ export const createInvoice = async ({ socket, rune, cert, label, description, ms
     agent = protocol === 'https:'
       ? new HttpsProxyAgent({ ...proxyOptions, ...httpsAgentOptions, rejectUnauthorized: false })
       : new HttpProxyAgent(proxyOptions)
+  } else if (process.env.NODE_ENV === 'development' && !cert) {
+    protocol = 'http:'
```
huumn, line 20: I couldn't get the cert chain to validate in polar even after generating my own.

api/paidAction/zap.js
```diff
@@ -78,10 +95,14 @@ export async function onPaid ({ invoice, actIds }, { models, tx }) {
       SELECT COALESCE(SUM(msats), 0) as msats
       FROM forwardees
     ), recipients AS (
-      SELECT "userId", msats FROM forwardees
+      SELECT "userId", msats, msats AS "stackedMsats" FROM forwardees
```
ekzyis, line 98: `recipients."stackedMsats"` is never used

api/paidAction/zap.js
ekzyis: Makes sense but you added `recipients."stackedMsats"` but never use it since afaict, you don't need to.

ekzyis (commented):
Partial review. Only test I did was to zap myself as anon with an attached wallet and it showed up as a p2p zap:

2024-08-06-151628_747x138_scrot

Then I got an overview over the code to see if I spot anything obvious but didn't look at anything very closely yet. Left some comments.

api/paidAction/README.md
```diff
@@ -54,6 +54,26 @@ Internally, pessimistic flows use hold invoices. If the action doesn't succeed,
 | update posts      | x           |            | x           |          | x          |             | x            |
 | update comments   | x           |            | x           |          | x          |             | x            |
 
+## Not-custodial zaps (ie p2p wrapped payments)
+Zaps, and possibly other future actions, can be performed peer to peer and non-custodially. This means that the payment is made directly from the client to the recipient, without the server taking custody of the funds. Currently, in order to trigger this behavior, the recipient must have a receiving wallet attached and the sender must have insufficient funds in their custodial wallet to perform the requested zap.
+
+This works by requesting an invoice from the recipient's wallet, wrapping it in a hold invoice paid to SN (to collect the sybil fee) which we serve to the sender. When the sender pays the wrapped invoice, we forward our own money to the recipient, who then reveals the preimage to us, allowing us to settle the wrapped invoice and claim the sender's funds. This effectively does what a lightning node does when forwarding a payment but allows us to do it at the application layer.
```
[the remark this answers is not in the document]
ekzyis: I think that wrapping means simply reusing the payment hash should be mentioned at least once.

api/paidAction/index.js
```diff
@@ -180,15 +182,17 @@ export async function retryPaidAction (actionType, args, context) {
   context.optimistic = true
   context.me = await models.user.findUnique({ where: { id: me.id } })
 
-  const { msatsRequested } = await models.invoice.findUnique({ where: { id: invoiceId, actionState: 'FAILED' } })
+  const { msatsRequested, actionId } = await models.invoice.findUnique({ where: { id: invoiceId, actionState: 'FAILED' } })
+  // TODO: msatsRequested includes routing fees if this was originally a wrapped invoice
```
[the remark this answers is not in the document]
huumn: I think I fixed this.

api/paidAction/index.js
```diff
@@ -211,24 +213,49 @@ export async function retryPaidAction (actionType, args, context) {
   }, { isolationLevel: Prisma.TransactionIsolationLevel.ReadCommitted })
 }
 
-const OPTIMISTIC_INVOICE_EXPIRE = { minutes: 10 }
-const PESSIMISTIC_INVOICE_EXPIRE = { minutes: 10 }
+const INVOICE_EXPIRE_SECS = 600
```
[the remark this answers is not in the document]
huumn: I think keeping them where they're used, if they're only used in that location, makes more sense.

api/resolvers/wallet.js
```diff
@@ -206,17 +199,19 @@ const resolvers = {
       if (include.has('withdrawal')) {
         queries.push(
           `(SELECT
-              id, created_at as "createdAt",
+              "Withdrawl".id, "Withdrawl".created_at as "createdAt",
               COALESCE("msatsPaid", "msatsPaying") as msats,
-              'withdrawal' as type,
+              CASE WHEN bool_and("InvoiceForward".id IS NULL) THEN 'withdrawal' ELSE 'p2p' END as type,
```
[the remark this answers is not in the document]
huumn: `bool_and` is an aggregate

api/typeDefs/item.js
```diff
@@ -22,6 +22,7 @@ export default gql`
     sats: Int!
     path: String
     act: String!
+    bolt11: String
```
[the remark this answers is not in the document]
huumn: Good catch. Leftovers from the merge

api/typeDefs/wallet.js
```diff
@@ -117,6 +117,7 @@ const typeDefs = `
     satsFeePaid: Int
     status: String
     autoWithdraw: Boolean!
+    direct: Boolean!
```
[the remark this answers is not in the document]
huumn: yep

api/paidAction/zap.js
```diff
@@ -10,6 +10,23 @@ export async function getCost ({ sats }) {
...
+      user: {
+        include: {
+          wallets: true
+        }
+      }
+    }
+  })
+
+  // request peer invoice if they have an attached wallet and have not forwarded the item
+  return item.user.wallets.length > 0 && item.itemForwards.length === 0 ? item.userId : null
```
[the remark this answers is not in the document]
huumn: Good point. I'm just going to punt it for this pr though

api/paidAction/zap.js
huumn: That's a bug!

huumn:
> We used to have this but we removed it in https://github.com/stackernews/stacker.news/commit/10e58d41c7b3c8e6b6886a9d09cdb0e75391ea75 since it wasn't needed anyway thanks to https://github.com/stackernews/stacker.news/pull/1098. Has this changed?

I needed this so that `COMPOSE_PROFILES` can be picked up from `.env.local` iirc

wallets/wrap.js
```diff
@@ -0,0 +1,166 @@
...
+      the expected ctlv_delta of the outgoing invoice's entire route
+
+      timeLockDelay is the absolute height the outgoing route is estimated to expire in the worst case.
+      It excludes the final hop's cltv_delta, so we add it. We subtract the blockheight,
+      then add on how many blocks we want to reserve to settle the incoming payment,
+      assuming the outgoing payment settles at the worst case (ie largest) height.
+    */
+    wrapped.cltv_delta = toPositiveNumber(
+      toPositiveNumber(timeLockDelay) + toPositiveNumber(inv.cltv_delta) -
+      toPositiveNumber(blockHeight) + MIN_SETTLEMENT_CLTV_DELTA)
```
ekzyis, line 137: I see, this is one of the critical parts. It makes sense to me as follows:

`estimateRouteFee` returns `timeLockDelay` which is the worst case for how many blocks the payment (given destination and amount) may take before the HTLC expires. This means we should wait at least that long. Since it's in absolute block height, we subtract the current block height to get a relative value.

We add the CLTV delta of the final hop to the receiver (`inv.cltv_delta`) since it's not included in `timeLockDelay` (not sure why not though).

We add `MIN_SETTLEMENT_CLTV_DELTA` (= 80) to give ourselves enough time to claim funds on-chain.

There is only one case in which _our_ CLTV delta becomes relevant _for us_: if our peer on the outgoing route decides to settle the payment on-chain; revealing the preimage (HTLC-success tx). We then have 80 blocks to use this preimage to settle the incoming HTLC (on-chain or off-chain) because when this happens, our peer on the incoming route must at least wait 80 blocks per our CLTV delta before they can claim their funds back on-chain (HTLC-timeout tx). This makes sure we always get paid when we forwarded the payment.

In the case our peer on the outgoing route is unresponsive, we have to wait the CLTV delta proposed by that peer (which we accepted) before our HTLC-timeout tx is valid. This makes sure we can always claim back our own funds on-chain. This would be a forced channel close with the unresponsive peer.[^1]

[^1]: It's not clear to me if that force-closure propagates back but I think not since responsive peers can still update their channel with a commitment tx that dropped this HTLC via normal revocation.

_`min_final_cltv_expiry_delta` of invoices as mentioned in BOLT11 was simply renamed to `cltv_delta` by `ln-service`._

wallets/wrap.js
```diff
@@ -0,0 +1,166 @@
...
+    console.log('routingFeeMsat', routingFeeMsat, 'timeLockDelay', timeLockDelay, 'blockHeight', blockHeight)
+
+    // validate the cltv delta
+    if (wrapped.cltv_delta > MAX_OUTGOING_CLTV_DELTA) {
+      throw new Error('Estimated outgoing cltv delta is too high: ' + wrapped.cltv_delta)
```
ekzyis, line 142: This means we're willing to lock our funds for 500 blocks which is ~3.5 days. Sounds high but if every hop wants 80 blocks for themselves, this means locking funds for 500 blocks only gets us 6 hops which in turn sounds reasonable (if not even too short).

wallets/wrap.js
```diff
@@ -0,0 +1,166 @@
...
+    } else if (wrapped.cltv_delta < MIN_INCOMING_CLTV_DELTA) {
+      // enforce a minimum cltv delta for the incoming invoice
+      wrapped.cltv_delta = MIN_INCOMING_CLTV_DELTA
```
ekzyis, line 145: Why do we need this check? Isn't specifying `min_final_cltv_expiry_delta` via `wrapped.cltv_delta` enough?

Afaict, this basically says that for short routes to the receiver, we want more time for ourselves. If so, why?

wallets/wrap.js
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    // validate the fee budget
+    const minEstFees = toPositiveNumber(routingFeeMsat)
+    const outgoingMaxFeeMsat = Math.ceil(outgoingMsat * MAX_FEE_ESTIMATE_PERCENT)
```
ekzyis, line 150: This means we're willing to pay 2.5% of the incoming amount in routing fees for paying the receiver invoice. It also means we're willing to give 25% of the sybil fee (which is 10% of the incoming amount) to the network. Sounds good to me.

wallets/wrap.js
```diff
@@ -0,0 +1,166 @@
...
+    if (minEstFees > outgoingMaxFeeMsat) {
+      throw new Error('Estimated fees are too high')
+    }
+
+    // calculate the incoming invoice amount, without fees
+    wrapped.mtokens = String(Math.ceil(outgoingMsat * ZAP_SYBIL_FEE_MULT))
```
ekzyis, line 156: Make them pay! @Darth-Coin would be proud.

ekzyis (commented):
Another partial review. Want to publish my comments about the invoice wrapping asap.

api/paidAction/index.js
```diff
@@ -237,39 +263,60 @@ async function createLndInvoice (actionType, args, context) {
...
-      userId: me?.id ?? USER_ID.anon,
-      actionType,
-      actionState,
-      actionArgs: args,
-      expiresAt,
-      actionId
-    }
-  })
+  const servedInvoice = wrappedInvoice ?? unwrappedInvoice
+  const request = parsePaymentRequest({ request: servedInvoice.request })
```
ekzyis, line 277: Mhh, I tested zapping someone with an attached wallet and without since I noticed that `unwrappedInvoice` is just a bolt11 in the wrapped case and wondered how `parsePaymentRequest` is going to work with `servedInvoice.request` being undefined in the unwrapped case.

Turns out it's not undefined since `unwrappedInvoice` is the return value of a LND call and not just the bolt11 in that case. So this code does not contain a bug but it's unexpected that the type of a variable might be different depending on context.

A comment might be useful here since I was also confused about `bolt11: unwrappedInvoice` in the `db.invoiceForward.create` call.

```suggestion
  // if we wrapped the invoice, the unwrapped invoice is a bolt11 string else it's a full lightning invoice object
  const servedInvoice = wrappedInvoice ?? unwrappedInvoice
  const request = parsePaymentRequest({ request: servedInvoice.request })
```

api/resolvers/wallet.js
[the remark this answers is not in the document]
ekzyis: Yes, but `SELECT bool_and(<statement>)` is the same as `SELECT <statement>` if there is only one statement inside `bool_and`, no?

wallets/wrap.js
huumn: Good question

api/paidAction/index.js
huumn: I'll take a closer look at this. The fact that I'm not sure what these types are is a sign this should be changed.

wallets/wrap.js
ekzyis: Resolved. This sets a lower bound for `wrapped.cltv_delta`.

api/resolvers/wallet.js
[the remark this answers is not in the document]
huumn: There's a `group by` clause. One statement !== one row

wallets/lnd/server.js
```diff
@@ -1,40 +1,13 @@
...
-
-    // we wrap both calls in one try/catch since connection attempts happen on RPC calls
-    await addWalletLog({ wallet: { type: 'LND' }, level: 'SUCCESS', message: 'connected to LND' }, { me, models })
-
-    return inv
-  } catch (err) {
-    // LND errors are in this shape: [code, type, { err: { code, details, metadata } }]
-    const details = err[2]?.err?.details || err.message || err.toString?.()
-    throw new Error(details)
-  }
```
huumn, line 37: A lot of these were doing silly things like duplicating `createInvoice` logic.

api/resolvers/wallet.js
```diff
@@ -23,7 +23,13 @@ function injectResolvers (resolvers) {
     console.log(resolverName)
 
     resolvers.Mutation[resolverName] = async (parent, { settings, ...data }, { me, models }) => {
-      await walletValidate(w, { ...data, ...settings })
+      // allow transformation of the data on validation (this is optional ... won't do anything if not implemented)
+      const validData = await walletValidate(w, { ...data, ...settings })
+      if (validData) {
+        Object.keys(validData).filter(key => key in data).forEach(key => { data[key] = validData[key] })
+        Object.keys(validData).filter(key => key in settings).forEach(key => { settings[key] = validData[key] })
+      }
```
huumn, line 31: After #1243, wallet field transforms like `ensureB64` weren't getting run, which meant it needed to be done every time `createInvoice` was called. This PR reintroduces it by using Yup's `transform` on `hexOrBase64` validations.

api/resolvers/wallet.js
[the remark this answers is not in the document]
ekzyis: Ohhh, I see

PR #1179, opened 2024-05-17
Show Alert Message on Settings Page

### LINKED ISSUE #790: Add alert to settings page when only one auth method is linked
People keep losing access to their accounts, mostly because they only link lnauth and either the wallet changes the derivation of the linking key or they change their lightning wallet and forget they've used it for auth.

So, when a stacker only has one auth method linked, display a warning at the top of the settings page that they should link another.

tsmith123:
Fixes #790 

## Description
This PR renders an alert message on the settings page when only a single auth method is activated.

## Screenshots

**Desktop**
Screenshot 2024-05-17 at 15 05 57

**Mobile**
Screenshot 2024-05-17 at 15 06 04


huumn:
I've been waiting for someone to pick this up!

Two things:

1. I think this should be for any auth method
2. It should be at the top of the settings page

While lnurl auth problems are the most common, we've had people lose access to email address and twitter auth before.

tsmith123:
Ah, no problem. Was a bit unsure of the requirements. Changes pushed and I've updated the screenshots to show how the message at the top will look.

PR #1184, opened 2024-05-18
Replace toasts - be more optimistic - prepaid and postpaid

### LINKED ISSUE #848: Make actions dependent on external payments complete optimistically
Currently, when you pay for a post, comment, or zap with an external wallet, the UI enters a sort of pending state until the payment completes. Ideally the UI would update as if the payment completed instantly - much like our existing custodial payments.

A solution must address two main questions:

1. what to do when the payment fails?
2. where is the data for the action stored while it's pending?

Taken together, I think that providing the ideal UX for (1) forces the answer for (2) to be on the server. The ideal UX for (1) is a notification that contains the data for the action (e.g. a comment that failed to pay its fee) that can then be retried. This will require storage of the action data on the server. Storing the action on the server also has the advantage of stackers not needing to be online when the action settles.

One way this could work:
1. tables for actions (e.g. "ItemAct" or "Item") that have pending payments, get two new columns:
    - `invoiceId` for the inflight invoice
    - `status` which is one of `SETTLED`, `PENDING`, or `FAILED`
2. for the logged in stacker initiating the action, we display `PENDING` actions as if they are already `SETTLED` more or less
3. for everyone else, they only see such actions that are `SETTLED`

### LINKED ISSUE #849: External pending payment UX
**tldr Toasts are perfectly okay for now - and maybe the best thing to do while we're experimenting - but we'll want something better once we nail the UX otherwise**

Some good discussion here: https://stacker.news/items/423928/r/k00b?commentId=428876

In addition to #848, we'll want something more refined than toasts as time goes on to facilitate cancels/retries. Basically these actions will be fundamental parts of the non-custodial UX so we shouldn't offload them to generic toast messages in the longrun.

ekzyis:
### Using undocumented Apollo API for optimistic behavior for zaps, poll votes and bounty payments

Every action which does not need a database id in the client now updates the UI optimistically even if payment is required. This includes (and is limited to) zaps, poll votes and bounty payments. This leads to a not-custodial UX that is equal to the familiar speed of the custodial UX since all payment delays are moved to the background, out of view from the user. When the invoice was paid, the action is submitted to the server as usual with payment hash + hmac.

To support this, the form component now supports `optimisticUpdate`. When this property is passed a function, its return value will be passed to the following function which calls [`QueryManager.markOptimisticMutation`](https://github.com/apollographql/apollo-client/blob/9c5a8cee40900125fe5037d573e6daedd619075f/src/core/QueryManager.ts#L587-L630) and [`cache.removeOptimistic`](https://github.com/apollographql/apollo-client/blob/9c5a8cee40900125fe5037d573e6daedd619075f/src/cache/inmemory/inMemoryCache.ts#L406-L412) to manually trigger the optimistic cache update and its revert:

```js
export function optimisticUpdate ({ mutation, variables, optimisticResponse, update }) {
  const { cache, queryManager } = getApolloClient()

  const mutationId = String(queryManager.mutationIdCounter)
  queryManager.markMutationOptimistic(optimisticResponse, {
    mutationId,
    document: mutation,
    variables,
    update
  })

  return () => cache.removeOptimistic(mutationId)
}
```

We call this function before triggering a payment. On payment or submit error, we rollback the optimistic update. The user is only informed in this case via notifications that are stored in `window.localStorage`.

_During the changes, it became clear that the property `invoiceable` should actually be called `prepaid`. This makes it purpose more clear._

### Optimistic replies and posts via post-paid support

For mutations where the id is required to update the UI in a sane way, the form component now supports a new property: `postpaid`.

If this property is passed, the form component will check the return value of the submit handler for an invoice. If one was found, it will try to pay it.

This allows us to use the server response to update the UI as fast as if it was paid via the custodial wallet. In the case a external payment is required, the item is inserted as "pending" which makes them only visible to the author and the server response will include an invoice. Since anons are not logged in and thus don't support this optimistic behavior, they don't have access to pending items but need to pay first. Therefore, forms and mutations support prepaid and postpaid simultaneously.

When the invoice is paid, we update the item status to active which makes it visible to everyone.

Based on and thus includes #1071 closes #849 closes #848 supersedes #1161.

## TODO

**Prepaid**

- [x] optimistic zaps
  - [x] optimism via `optimisticUpdate`
  - [x] ~persistence of optimistic update while payment is pending via `window.localStorage`~
  _we'll do this after #1178 since we need to be aware of pending stuff on the server anyway if proxy is used_
- [x] optimistic bounty payments
  - [x] optimism via `optimisticUpdate`
  - [x] ~persistence of optimistic update while payment is pending via `window.localStorage`~
  _we'll do this after #1178 since we need to be aware of pending stuff on the server anyway if proxy is used_
- [x] poll votes
  - [x] optimism via `optimisticUpdate`
  - [x] ~persistence of optimistic update while payment is pending via `window.localStorage`~
  _we'll do this after #1178 since we need to be aware of pending stuff on the server anyway if proxy is used_
- [x] error notification if payment failed
  _related to persistence to some degree since payment should also be considered "failed" if payment + associated action was lost due to page refresh, but we'll skip this case for now as mentioned above_   
- [x] ~error handling if payment failed (reverting persistence)~
  _see comments above_
- [x] zap undos

**Postpaid**

- [x] Add `payment.send` next to `payment.request` from #1071 
- [x] Add `postpaid` to form which checks for returned invoices and tries to pay them
- [x] insert pending items and return invoice
- [x] check for payments on server and update item status
- [x] check for payment failures and update item status
- [x] update queries to only show pending items to author
- [x] error notifications
- [x] on error notification click, navigate to prefilled replies/forms for good retry UX 
- [x] ignore failed items in spam calculation
- [x] make sure items transition to `ACTIVE` even if transition query failed after invoice confirmation 
- [x] don't notify users etc. if pending items are created
- [x] decide what to do with dead links in error notifications
- [x] ~update `updatedAt` manually in raw queries for notification red square to work~ _for some reason, this wasn't necessary_
- [ ] prevent replying to or zapping failed items

## TODO after [review](https://github.com/stackernews/stacker.news/pull/1184#issuecomment-2127919566)

- [ ] handle action in same tx as `confirm_invoice` to fix race conditions if autowithdrawals are enabled
- [ ] test edits of pending items:

> Editting a pending comment made a modal popup telling me the payment failed. Is that expected?

- [x] fix missing `resetForm`: 669ad67d

> When a top level comment failed, I clicked on the notification, then submitted the form to retry to the comment, it succeeded but the comment field did not clear 

## Videos

TBD

## Additional Context

- the idea of using `mutationIdCounter` and not `mutationIdCounter++` as in the library source code was that the mutation will also call an optimistic update with the same id to make sure the UI updates and overwrite the previous layer.  If I recall correctly, I found out it actually creates its own layer with its own (automated) revert. Therefore, I considered it not only be redundant but also a possible source of bugs. However, as mentioned, not sure, would need to check this behavior again.

## Checklist

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1386,15 +1386,15 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
     })
 
     let invoice
-    const actionData = { cost: err.cost, credits: err.balance }
     await models.$transaction(
       async (tx) => {
         // insert pending item
         ([item] = await tx.$queryRawUnsafe(
           `${SELECT} FROM create_item($1::JSONB, $2::JSONB, $3::JSONB, '${spamInterval}'::INTERVAL, $4::INTEGER[]) AS "Item"`,
-          JSON.stringify({ ...item, status: 'PENDING' }), JSON.stringify(fwdUsers), JSON.stringify(options), uploadIds));
+          JSON.stringify({ ...item, status: 'PENDING' }), JSON.stringify(fwdUsers), JSON.stringify(options), uploadIds))
 
         // set required invoice data to update item on payment
+        const actionData = { cost: err.cost, credits: err.balance };
         ([invoice] = await tx.$queryRaw`SELECT * FROM create_invoice(${lndInv.id}, NULL, ${lndInv.request},
           ${expiresAt}::timestamp, ${mtokens}, ${item.userId}::INTEGER, ${description}, NULL, NULL,
           ${invLimit}::INTEGER, ${balanceLimit}, 'ITEM'::"ActionType", ${item.id}::INTEGER, ${JSON.stringify(actionData)}::JSONB)`)
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1,6 +1,6 @@
 import { GraphQLError } from 'graphql'
 import { ensureProtocol, removeTracking, stripTrailingSlash } from '@/lib/url'
-import serialize from './serial'
+import serialize, { InsufficientFundsError } from './serial'
 import { decodeCursor, LIMIT, nextCursorEncoded } from '@/lib/cursor'
 import { getMetadata, metadataRuleSets } from 'page-metadata-parser'
 import { ruleSet as publicationDateRuleSet } from '@/lib/timedate-scraper'
@@ -9,7 +9,7 @@ import {
   ITEM_SPAM_INTERVAL, ITEM_FILTER_THRESHOLD,
   COMMENT_DEPTH_LIMIT, COMMENT_TYPE_QUERY,
   ANON_USER_ID, ANON_ITEM_SPAM_INTERVAL, POLL_COST,
-  ITEM_ALLOW_EDITS, GLOBAL_SEED, ANON_FEE_MULTIPLIER, NOFOLLOW_LIMIT, UNKNOWN_LINK_REL
+  ITEM_ALLOW_EDITS, GLOBAL_SEED, ANON_FEE_MULTIPLIER, NOFOLLOW_LIMIT, UNKNOWN_LINK_REL, JIT_INVOICE_TIMEOUT_MS, ANON_INV_PENDING_LIMIT, ANON_BALANCE_LIMIT_MSATS, INV_PENDING_LIMIT, USER_IDS_BALANCE_NO_LIMIT, BALANCE_LIMIT_MSATS
 } from '@/lib/constants'
 import { msatsToSats } from '@/lib/format'
 import { parse } from 'tldts'
@@ -21,6 +21,9 @@ import { datePivot, whenRange } from '@/lib/time'
 import { imageFeesInfo, uploadIdsFromText } from './image'
 import assertGofacYourself from './ofac'
 import assertApiKeyNotPermitted from './apiKey'
+import { createInvoice } from 'ln-service'
+import { Prisma } from '@prisma/client'
+import { createHmac } from './wallet'
 
 function commentsOrderByClause (me, models, sort) {
   if (sort === 'recent') {
@@ -1353,23 +1356,60 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
       fee = sub.baseCost * ANON_FEE_MULTIPLIER + (item.boost || 0)
     }
   }
-  fee += imgFees;
+  fee += imgFees
 
-  ([item] = await serialize(
-    models.$queryRawUnsafe(
+  try {
+    ([item] = await serialize(
+      models.$queryRawUnsafe(
       `${SELECT} FROM create_item($1::JSONB, $2::JSONB, $3::JSONB, '${spamInterval}'::INTERVAL, $4::INTEGER[]) AS "Item"`,
       JSON.stringify(item), JSON.stringify(fwdUsers), JSON.stringify(options), uploadIds),
-    { models, lnd, me, hash, hmac, fee }
-  ))
+      { models, lnd, me, hash, hmac, fee }
+    ))
+  } catch (err) {
+    if (err instanceof InsufficientFundsError) {
+      // create invoice and insert as pending item
+      const invLimit = me ? INV_PENDING_LIMIT : ANON_INV_PENDING_LIMIT
+      const balanceLimit = USER_IDS_BALANCE_NO_LIMIT.includes(Number(me?.id)) ? 0 : me ? BALANCE_LIMIT_MSATS : ANON_BALANCE_LIMIT_MSATS
+      const description = 'Creating item on stacker.news'
+      const expiresAt = datePivot(new Date(), { seconds: me ? JIT_INVOICE_TIMEOUT_MS : 180 })
+      const mtokens = err.cost - err.balance
 
-  await createMentions(item, models)
+      const lndInv = await createInvoice({
+        description: me.hideInvoiceDesc ? undefined : description,
+        lnd,
+        mtokens,
+        expires_at: expiresAt
+      })
+
+      let invoice
+      const actionData = { cost: err.cost, credits: err.balance }
+      await models.$transaction(
+        async (tx) => {
+          // insert pending item
+          ([item] = await tx.$queryRawUnsafe(
+          `${SELECT} FROM create_item($1::JSONB, $2::JSONB, $3::JSONB, '${spamInterval}'::INTERVAL, $4::INTEGER[]) AS "Item"`,
+          JSON.stringify({ ...item, status: 'PENDING' }), JSON.stringify(fwdUsers), JSON.stringify(options), uploadIds));
+
+          // set required invoice data to update item on payment
+          ([invoice] = await tx.$queryRaw`SELECT * FROM create_invoice(${lndInv.id}, NULL, ${lndInv.request},
+          ${expiresAt}::timestamp, ${mtokens}, ${item.userId}::INTEGER, ${description}, NULL, NULL,
+          ${invLimit}::INTEGER, ${balanceLimit}, 'ITEM'::"ActionType", ${item.id}::INTEGER, ${JSON.stringify(actionData)}::JSONB)`)
+        }, { isolationLevel: Prisma.TransactionIsolationLevel.Serializable })
+
+      invoice.hmac = createHmac(invoice.hash)
+
+      item.invoice = invoice
+    } else {
+      throw err
+    }
+  }
 
   await enqueueDeletionJob(item, models)
 
+  // TODO: don't notify users on pending items
   await createReminderAndJob({ me, item, models })
-
+  await createMentions(item, models)
   notifyUserSubscribers({ models, item })
-
   notifyTerritorySubscribers({ models, item })
 
   item.comments = []
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1396,6 +1396,7 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
           ${invLimit}::INTEGER, ${balanceLimit}, 'ITEM'::"ActionType", ${item.id}::INTEGER, ${JSON.stringify(actionData)}::JSONB)`)
         }, { isolationLevel: Prisma.TransactionIsolationLevel.Serializable })
 
+      // hmac is not required to submit action again but to allow user to cancel payment
       invoice.hmac = createHmac(invoice.hash)
 
       item.invoice = invoice
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -9,7 +9,7 @@ import {
   ITEM_SPAM_INTERVAL, ITEM_FILTER_THRESHOLD,
   COMMENT_DEPTH_LIMIT, COMMENT_TYPE_QUERY,
   ANON_USER_ID, ANON_ITEM_SPAM_INTERVAL, POLL_COST,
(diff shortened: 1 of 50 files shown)


worker/wallet.js
```diff
@@ -349,3 +355,44 @@ export async function checkPendingWithdrawals (args) {
...
+      })
+      if (item.boost > 0) {
+        await tx.$executeRaw(`SELECT item_act(${item.id}::INTEGER, ${item.userId}::INTEGER, 'BOOST'::"ItemActType", ${item.boost}::INTEGER)`)
+      }
+      if (item.maxBid) {
+        await tx.$executeRaw(`SELECT run_auction(${item.id}::INTEGER);`)
+      }
+    }, { isolationLevel: Prisma.TransactionIsolationLevel.Serializable })
+  }
+}
```
huumn, line 398: Just a heads up (I know this is in draft and this might not be what you intend to ship): worst case, this is 5 roundtrips to the db (because it's interactive and not pipelined), and if during that time any of the modified rows are updated by another transaction, this and the other tx will fail. 

In general, the longer a serializable tx takes to run, the more likely the failure. Everywhere in the code so far, we only do one roundtrip for serializable txs (and we wrap them all in a retry in `serializable`)

prisma/migrations/20240518020747_pending_items/migration.sql
```diff
@@ -0,0 +1,219 @@
...
+    INSERT INTO pgboss.job (name, data, retrylimit, retrybackoff, startafter)
+    VALUES ('imgproxy', jsonb_build_object('id', item.id), 21, true, now() + interval '5 seconds');
+
+    -- if item is pending, user pays missing sats later.
+    -- all remaining queries will run when invoice was paid and we update the item status.
+    IF item."status" = 'PENDING'::"Status" THEN
+        -- here, we immediately deduct as many of the sats that are required for the payment
+        -- to effectively "lock" them for it. the remainder will be paid via invoice.
+        -- if the payment fails, we release the locked sats by adding them to the balance again.
+        UPDATE users SET msats = GREATEST(msats - cost_msats - item.boost, 0) WHERE id = item."userId";
```
huumn, line 196: This pr is a big lift even without this *partial payment* change, but you might have this working well enough it's okay to add. But if it is overwhelming getting this extra done, I'd recommend postponing the *partial payments* until the *primary* objective of this PR is pristine. 

It's usually a lot easier to add extras when the foundation is settled.

worker/wallet.js
ekzyis: > worst case, this is 5 roundtrips to the db (because it's interactive and not pipelined) [...] Everywhere in the code so far, we only do one roundtrip for serializable txs (and we wrap them all in a retry in serializable)

Oh, I didn't realize that the `serialize` calls with an array of queries weren't multiple round trips.

But I agree with you, roundtrips are bad, especially when it's in a serialized tx. I didn't think too much about it here for now, it was just too convenient to use Prisma's API. 

But I noticed now that the only reason I use an interactive tx was to fetch the item author. I can do this outside of the tx and then use `serialize`.

prisma/migrations/20240518020747_pending_items/migration.sql
ekzyis: Mhh yes, but so far, I think it's only a matter of adding them back in case of failure which I have to handle anyway. So let's see

ekzyis: Ok, since deducting user balance is not idempotent and even though LND subscriptions don't trigger on invoice expiration (so idempotency shouldn't be necessary), I removed partial payments in 2c1807d2 to be sure.

ekzyis:
Regarding persistence of optimistic updates for prepaid stuff: I noticed we really don't need it. In 8e6f7bac, I (re)implemented client notifications and realized that if something is still pending when we reload the page, we can immediately show it as failed since this means the payment flow was interrupted. So there is never the case of something being pending but the UI not showing this.

components/item-act.js
```diff
@@ -41,222 +43,150 @@ const addCustomTip = (amount) => {
...
-      />
-      <div>
-        <Tips setOValue={setOValue} />
-      </div>
-      {children}
-      <div className='d-flex mt-3'>
-        <SubmitButton variant={down ? 'danger' : 'success'} className='ms-auto mt-1 px-4' value='TIP'>{down && 'down'}zap</SubmitButton>
-      </div>
-    </Form>
+    <ClientNotifyProvider additionalProps={{ itemId: item.id }}>
```
ekzyis, line 143: I hope I'll still like this approach to pass something to `notify` that the form component can't down the road

Can't tell yet if this is simple clever or confusing clever

api/resolvers/user.js
```diff
@@ -487,6 +490,20 @@ export default {
...
+          status: 'FAILED',
+          updatedAt: {
+            gt: lastChecked
+          }
+        }
+      })
+      if (newFailedItems) {
+        foundNotes()
+        return true
+      }
```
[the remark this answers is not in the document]
ekzyis: This comment reminds me that I probably need to manually update `updatedAt` in raw queries since it's not based on a trigger.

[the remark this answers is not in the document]
ekzyis: Mhh, even though I don't see a trigger that does this, `updatedAt` gets updated after a raw query updates an item:

These are the existing item triggers:

```
Triggers:
    index_item AFTER INSERT OR UPDATE ON "Item" FOR EACH ROW EXECUTE FUNCTION index_item()
    ncomments_after_comment_trigger AFTER INSERT ON "Item" FOR EACH ROW EXECUTE FUNCTION ncomments_after_comment()
    path_tgr BEFORE INSERT OR UPDATE ON "Item" FOR EACH ROW EXECUTE FUNCTION update_item_path()
    timestamp_item_on_insert AFTER INSERT ON "Item" FOR EACH ROW EXECUTE FUNCTION timestamp_item_on_insert()
```

huumn:
Things noticed after a few minutes of QA:

1. I boosted a post and it's stuck as `PENDING`
2. When a top level comment failed, I clicked on the notification, then submitted the form to retry to the comment, it succeeded but the comment field did not clear
3. When I have an autowithdrawal method attached, paying for things races with autowithdraw (after I paid for the post in (1), it autowithdrew 51 sats)

(2 & 3) aren't really regressions or "big deals" but (1) is pretty nasty as it looks to the person that created it that it succeeded but it didn't and it's unclear what they should do now. I'd guess there's a problem in the failure code but I don't know. It might be easier to recreate if you add an autowithdraw method.

That's after just a few minutes. There's a lot going on that's changed in this code, so it's hard for me to know what is causing what. 

I'll rebase #1178 on this and begin mending them together in the meantime.

api/resolvers/item.js
```diff
@@ -1353,26 +1378,81 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
...
+    // need to use interactive tx here to set invoice.actionId = item.id
+    await models.$transaction(
+      async (tx) => {
+        // insert pending item
+        ([item] = await tx.$queryRawUnsafe(
+          `${SELECT} FROM create_item($1::JSONB, $2::JSONB, $3::JSONB, '${spamInterval}'::INTERVAL, $4::INTEGER[]) AS "Item"`,
+          JSON.stringify({ ...item, status: 'PENDING' }), JSON.stringify(fwdUsers), JSON.stringify(options), uploadIds))
+
+        // set required invoice data to update item on payment
+        const actionData = { cost: err.cost, credits: err.balance };
```
huumn, line 1422: If these fields need to be saved, might we do better storing them as columns with types and uniform naming?

huumn:
Thinking through my own changes in #1178 alongside this makes me think we are going to need a refactor to make our planned changes robust.

If we solve the problem of (1) in a robust way and we can reason through the robustness of all the states of payments<->actions in this pr, we can ship this before the refactor. Bugs like this are evidence the code has become too much to handle.

ekzyis:
> 2. When a top level comment failed, I clicked on the notification, then submitted the form to retry to the comment, it succeeded but the comment field did not clear

Sounds like a missing `resetForm` call, should be easy to fix

> 3. Editting a pending comment made a modal popup telling me the payment failed. Is that expected?

No :eyes: I actually didn't test this, since I didn't change how items are updated and edits should have no cost associated to them so it shouldn't trigger any payment code. But will test now

> 1. I boosted a post and it's stuck as PENDING

> 4. When I have an autowithdrawal method attached, paying for things races with autowithdraw (after I paid for the post in (1), it autowithdrew 51 sats)

> If we solve the problem of (1) in a robust way and we can reason through the robustness of all the states of payments<->actions in this pr, we can ship this before the refactor.

One thing I wanted to do (and I think you also mentioned this) was to run the action in the same tx as the invoice gets confirmed. Currently, it's not since the state transition is not idempotent due to item side effects so I used the return value of `confirm_invoice` to see if it's the first confirmation.

Running both actions in the same tx should fix the race condition with autowithdrawals. I could do this by creating a postgres function that runs the code in `handleAction` since then I can rely on serialized transactions to check if it's the first confirmation (just like `confirm_invoice`).

We don't run into this problem with prepaid because the hash+hmac approach does that: creating and thus paying for the item is in the same tx as confirming the invoice. 

> Edit: it looks like the payment in (1) was confirmed but the Item is still pending.

There is a worker job which runs `finalizeAction` after the invoice is expired. It's there to make sure all state transitions are guaranteed to happen eventually (when the invoice expires). However, if the state transition to `ACTIVE` also fails there, it will be pending forever. That's a bug, I should just mark the item as `FAILED` then.

**So I think the most important remaining TODO here is to create `handle_action` as a postgres function that we can run in the same tx as `confirm_invoice`. This should fix conflicts with autowithdrawals and thus 1) and 4) (and possibly other payment problems)**

huumn:
FYI any of the code here might end up getting completely rewritten in the refactor and we *need* to do the refactor soon. All of our payment stuff was complicated before these changes we are making, and now it's even more so (ie there are three ways to pay for something). The code on the frontend got a lot better, but the backend code got a lot worse.

Just wanted to caution against getting too attached to any particular bug fix when we really need a *systemic* fix. Still, happy to ship this once it's deterministic and bug-free.

For a hint of how complicated this is, here's the abstraction requirements table from my notes:

|                     | anonable | optimistic | qr payable | p2p wrapped | side effects | fee credits payable |
| ------------------- | -------- | ---------- | ---------- | ----------- | ------------ | ------------------- |
| posts               | x        | x          | x          |             | x            | x                   |
| comments            | x        | x          | x          |             | x            | x                   |
| zaps                | x        | x          | x          | x           | x            | x                   |
| downzaps            |          | x          | x          |             | x            | x                   |
| poll votes          |          | x          | x          |             |              | x                   |
| territory purchases |          |            | x          |             |              | x                   |
| donations           | x        | x          | x          |             |              | x                   |
| update posts        |          | x?         | x          |             | x            | x                   |
| update comments     |          | x?         | x          |             | x            | x                   |

ekzyis:
I realized that fixing edits of pending items would also require changes to `update_item` (a similar flow as with `create_item` essentially) that will be overwritten by the systemic fix we discussed.

However, I think even the systemic fix we discussed wouldn't handle edits properly. Edits of pending items means that we now potentially have two invoices that could be paid separately but they require payment in order since it doesn't make sense to pay for an edit when the item itself wasn't paid yet and thus could still fail. Additionally, edits are the only actions that we have that mutate data instead of creating a row that we could mark as pending.

I see a few ways to approach this problem:

1) edits of pending items aren't allowed. you must first pay for the item before you can edit it. an edit that wasn't paid yet marks the item as pending again.
2) items never fail but can always be paid later via a new invoice
3) edits cancel the existing invoice for the pending item and create a new invoice that pays for the item + edit

(or a combination of them; 2) is something we discussed already iirc)

However, as mentioned, I consider this to be better done in the systemic fix instead of me messing with the backend more in this PR to get pending items working without bugs.

So I decided to indeed create a PR with only the frontend changes. I think we should merge them even if we'll also move zaps etc, to the server as pending acts since they include important changes to cleanup frontend code (removal of toast flows). They also already improve UX even if there are only few stackers who use attached wallets like me.

ekzyis:
Superseded by #1194 and #1195

PR #1191, opened 2024-05-26
Embed Rumble Video

### LINKED ISSUE #134: Rumble video embed like what was added for Youtube
Rumble is a video hosting network that is gaining in popularity, not just because of its monetization feature  (i.e., payouts to creators) but  but also because it is the relatively common destination for those who suffered from having their Youtube account terminated (censorship).  

This request is to add support for a Post for a link that is a Rumble video to include either the javascript embed or the iframe embed so that the video can be played within the SN post, similarly to how there is now Youtube embed, and Twitter embed support in SN (see Issue https://github.com/stackernews/stacker.news/issues/9).

tsmith123:
Fixes #134 

## Description

This PR allows embedding of Rumble videos (using specific embed link). It also adds a util function that:
 - helps to clean up some existing code including hard to read regexes
 - uses the URL constructor for more reliable parsing of embed urls
 - makes code reusable so both text and item-full component can use it

## Testing Notes
Rumble embed links have been tested (https://rumble.com/embed/12345/?pub=4)
YouTube traditional links have been tested (https://www.youtube.com/watch?v=12345)
YouTube short links have been tested (https://youtu.be/12345)
YouTube links with a start time have been tested (https://youtu.be/12345?feature=shared&t=30

## Screenshots

**Editor View**
Screenshot 2024-05-26 at 10 03 49

Screenshot 2024-05-26 at 10 03 55

Screenshot 2024-05-26 at 10 04 22

## Limitations
Rumble videos can only be embedded using the embed url specifically and not the video's webpage url. This is something that has been raised a few times already in different discussions so I won't go in to detail again.

There was a suggestion to use domino or something similar to make a request from the backend to the get the details required to support using webpage url like we do for autofilling titles etc. However, this wouldn't work because in the autofill scenario the request is done once and the title can be stored in the database. We don't really have that option for embed links and would make the whole process very messy.

I don't see a way around this unless Rumble starts adding the video id into the webpage url which might never happen. Still think embedding videos using the official embed link is a win though.

## Questions
I noticed that Tweets (or should they be called Xs now 😄 ) are supported in the `item-full.js` component but not the `text.js` component. Is this intentional?

## Notes
I've not tackled X/Twitter embeds as part of this PR.


tsmith123:
Adding @benalleng

huumn:
Code looks great. I'll QA first thing tomorrow

PR #1194, opened 2024-05-27
Frontend payment UX cleanup

ekzyis:
## Description

This includes all frontend changes from #1184 (especially the changes from https://github.com/stackernews/stacker.news/pull/1071) and supersedes it.

https://github.com/stackernews/stacker.news/issues/849 is not closed by this but this closes https://github.com/stackernews/stacker.news/issues/848

## Additional Context

See https://github.com/stackernews/stacker.news/pull/1184#issuecomment-2132916938

## Checklist

Yes

Yes

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -899,7 +899,7 @@ export default {
              WHERE act IN ('TIP', 'FEE')
              AND "itemId" = ${Number(id)}::INTEGER
              AND "userId" = ${me.id}::INTEGER)::INTEGER)`,
-          { models }
+          { models, lnd, hash, hmac }
         )
       } else {
         await serialize(
--- a/components/bounty-form.js
+++ b/components/bounty-form.js
@@ -108,7 +108,7 @@ export function BountyForm ({
       }}
       schema={schema}
       requireSession
-      invoiceable
+      prepaid
       onSubmit={
         handleSubmit ||
         onSubmit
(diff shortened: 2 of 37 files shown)


huumn:
I think this is the right move. The systemic fix is a pretty dramatic change. I'll QA and hopefully merge this today. 

How this'll go:
1. merge this
2. implement systemic overhaul (I have a local design that's nearly flushed out)
3. build not-custodial zaps #1178 into the systemic overhaul

huumn:
QA went pretty well until I did a bunch of zaps and the invoices started expiring.

For every expired invoice (about 10-15 of them), this modal popped up, one after the other on an unrelated page:

Screenshot 2024-05-27 at 12 30 10 PM

I'll do a line by line now, but QA is good other than the above.

huumn:
Nothing else came up in QA and line by line looks fine. Let me know if you want to fix the above issue before merging.

huumn (approved):
Good to merge but one bug.

huumn:
I did the same thing and the QR didn't show.

However, now I have payments failing that don't appear in my notifications:

Screenshot 2024-05-28 at 12 14 25 PM

Screenshot 2024-05-28 at 12 16 59 PM

I suspect not all the invoices are being recorded. Anyway, I'm going to merge as it's probably something easy for you to fix when you're back online.

ekzyis:
Good catch! When there are multiple errors which will show an modal, new modals replace the previous ones since `showModal` encloses the same values in each payment "thread". The call of one `showModal` in one thread does not update `showModal` in the other threads so they overwrite each other instead of stacking. The state is already "locked in".

I'm trying to get around this with `useRef` to have mutable state across components.

PR #1195, opened 2024-05-27
wip backend optimism

huumn:
This is a very rough draft of "backend optimism." To achieve an ideal non-custodial UX, we are doing what we're calling 2-stage optimism: 

1. frontend optimism: instantly reports success on the frontend while sending the requested action to the backend
2. backend optimism: stores the requested action in a `PENDING` state, returning success and appearing as-if the payment is confirmed to the actor

We then transition the action state to `PAID` or `FAILED` depending on the payment state. When the payment succeeds, we run the action side effects and the action is visible to everyone. When the payment fails, we report the failure in the stacker's notifications where they can retry the payment.

----------------------

Doing this requires an overhaul to all paid actions. The requirements on paid actions are summarized in the following table:

|                     | anonable & pessimistic | optimistic | qr payable | p2p wrapped | side effects | fee credits payable |
| ------------------- | ---------------------- | ---------- | ---------- | ----------- | ------------ | ------------------- |
| posts               | x                      | x          | x          |             | x            | x                   |
| comments            | x                      | x          | x          |             | x            | x                   |
| zaps                | x                      | x          | x          | x           | x            | x                   |
| downzaps            |                        | x          | x          |             | x            | x                   |
| poll votes          |                        | x          | x          |             |              | x                   |
| territory purchases |                        |            | x          |             |              | x                   |
| donations           | x                      | x          | x          |             |              | x                   |
| update posts        |                        | x?         | x          |             | x            | x                   |
| update comments     |                        | x?         | x          |             | x            | x                   |

----------------------

This pr intends to implement backend optimism for all actions. We aim to take a relatively declarative approach to creating paid actions, where paid actions implement a standard interface so that paid actions aren't directly concerned with managing the payments. So far, what this looks like is paid actions are implemented in a separate file in the `paidAction` folder with an interface that defines the following functionality:

1. `getCost` - returns the cost of the action given the action's arguments so that an invoice can be created
2. `perform` - inserts the action's records leaving them in a `PENDING` state with reference to the corresponding `invoiceId`
3. `onPaid` - transitions the records from `perform` to a `PAID` state and does any side effects (e.g. denormalization)

-----------------------

A lot of care will need to be taken wrt atomicity of state transitions and the performance of making `PENDING` actions appear `PAID` selectively.


api/paidAction/createItem.js
```diff
@@ -0,0 +1,92 @@
...
+      },
+      itemUploads: {
+        connect: uploadIds.map(id => ({ uploadId: id }))
+      },
+      itemAct: {
+        createMany: itemActs
+      }
+    })
+  ]
+  // TODO: run_auction for job or just remove jobs?
```
ekzyis, line 59: with "remove jobs" you mean literally to remove the job post type?

huumn:
@SocketSecurity ignore npm/@prisma/client@5.14.0
@SocketSecurity ignore npm/@prisma/engines@5.14.0
@SocketSecurity ignore npm/@prisma/client@5.4.2
@SocketSecurity ignore npm/@prisma/engines@5.4.2

prisma/migrations/20240530174040_paid_actions/migration.sql
```diff
@@ -0,0 +1,312 @@
+-- CreateEnum
+CREATE TYPE "InvoiceActionType" AS ENUM ('BUY_CREDITS', 'ITEM_CREATE', 'ITEM_UPDATE', 'ZAP', 'DOWN_ZAP', 'DONATE', 'POLL_VOTE', 'TERRITORY_CREATE', 'TERRITORY_UPDATE', 'TERRITORY_BILLING', 'TERRITORY_UNARCHIVE');
```
ekzyis, line 2: `BOUNTY_PAYMENT` is included in `ZAP` since it's like a zap, right?

Just want to make sure this wasn't added here on purpose.

worker/paidAction.js
```diff
@@ -0,0 +1,150 @@
...
+      // optimistic concurrency control (aborts if invoice is not in PENDING state)
+      await tx.invoice.update({
+        where: { id: dbInv.id, actionState: 'PENDING' },
+        data: {
+          actionState: 'PAID',
+          confirmedAt: new Date(invoice.confirmed_at),
+          confirmedIndex: invoice.confirmed_index,
+          msatsReceived: BigInt(invoice.received_mtokens)
+        }
+      })
```
ekzyis, line 29: Mhh, so the question here to make sure that `onPaid` is called exactly once is: **Do other transactions see this update before the transactions commits?**

To answer this question, I read the [section about the Read Committed isolation level](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED), the default isolation level of Postgres:

> `UPDATE`, `DELETE`, `SELECT FOR UPDATE`, and `SELECT FOR SHARE` commands behave the same as `SELECT` in terms of searching for target rows: they will only find target rows that were committed as of the command start time. However, such a target row might have already been updated (or deleted or locked) by another concurrent transaction by the time it is found. **In this case, the would-be updater will wait for the first updating transaction to commit or roll back** (if it is still in progress).

This seems to be what we want here. **Other transactions wait until they can see this update.**

So only one transaction should find a row with the pending state, update it, continue and commit. Every other transaction should wait at this statement and then see that there is no longer such a row. Therefore, it should not continue due to [Prisma returning a `RecordNotFound` exception in this case](https://www.prisma.io/docs/orm/reference/prisma-client-reference#return-type-4).

Since this behavior depends on the isolation level, I think it makes sense to explicitly set this transaction level here even though it should be the default.

worker/paidAction.js
```diff
@@ -0,0 +1,150 @@
...
+
+      await paidActions[dbInv.actionType].onPaid?.({ invoice: dbInv }, { models, tx })
+      await tx.$executeRaw`INSERT INTO pgboss.job (name, data) VALUES ('checkStreak', jsonb_build_object('id', ${dbInv.userId}))`
+    })
```
ekzyis, line 33:
```suggestion
    }, { isolationLevel: Prisma.TransactionIsolationLevel.ReadCommitted })
```

components/reply.js
```diff
@@ -55,61 +56,76 @@ export default forwardRef(function Reply ({
...
+          id: `Item:${id}`,
+          fields: {
+            invoiceActionState () {
+              return 'PAID'
             }
-          })
+          }
+        })
+      },
+      onPayError: (e, cache, { data: { upsertComment: { result: { id } } } }) => {
```
ekzyis, line 117: This throws `Cannot read properties of null (reading 'id')` if I am closing the QR code invoice as anon because `result` is null in that case.

So I think you need to check `paymentMethod` first. The cache update only needs to run if it was optimistic.

worker/paidAction.js
```diff
@@ -0,0 +1,150 @@
...
+    await models.$transaction(async tx => {
+      // optimistic concurrency control (aborts if invoice is not in PENDING_HELD state)
+      await tx.invoice.update({
+        where: { id: dbInv.id, actionState: 'PENDING_HELD' },
+        data: {
+          actionState: 'HELD',
+          isHeld: true,
+          msatsReceived: BigInt(invoice.received_mtokens)
+        }
+      })
```
ekzyis, line 79:
```suggestion
      }, { isolationLevel: Prisma.TransactionIsolationLevel.ReadCommitted })
```

worker/paidAction.js
```diff
@@ -0,0 +1,150 @@
...
+      // optimistic concurrency control (aborts if invoice is not in PENDING state)
+      await tx.invoice.update({
+        where: { id: dbInv.id, actionState: 'PENDING' },
+        data: {
+          actionState: 'FAILED',
+          cancelled: true
+        }
+      })
+      await paidActions[dbInv.actionType].onFail?.({ invoice: dbInv }, { models, tx })
+    })
```
ekzyis, line 135:
```suggestion
    }, { isolationLevel: Prisma.TransactionIsolationLevel.ReadCommitted })
```

ekzyis (commented):
> Is it easy to understand how paid actions work and how to implement them? ie DX

I think so, yes. However, I noticed this in the PR description which initially confused me:

> perform - inserts the action's records leaving them in a PENDING state with reference to the corresponding invoiceId

This might be an outdated comment since at least now, `perform` _might_ create pending state. If it does, we need `onPaid` to transition out of it.

So that was confusing first but it makes sense now. `perform` does not create pending state if only pessimism is supported like for donations and that's also why `onPaid` (and `onFail`) is optional. It's not needed if there are no side effects _and_ no pending state.

If the action supports pessimism _and_ optimism, `perform` creates pending state even if it was already paid to support both. In the pessimistic case where we already paid, we simply transition out of pending immediately.

So my conclusion is that it's easy enough considering what we're trying to achieve here. If we write down once well how all interactions between the functions and the support flags work (which is consistent with the code), I think it's a great design!

---

> Is the UX of paid actions good, regardless of whether the action is optimistic, pessimistic, anon, paid with a qr, etc.

I tested commenting and zapping so far. 

Regarding failed payment UX, I think it would be good to give stackers the option to either a) dismiss a failed item so it no longer shows up or b) retry payment to resolve  it.

I also noticed that the spam fees stayed up even if my payment failed for a reply. I don't think that should be the case.

But I need to play with the UX more. So far I mostly read the code and tried to understand the flow and potential issues.

---

> Is concurrency handled appropriately

I think so. See comment in `settleAction`. Afaict, only in its file payment code is run inside the worker and thus might get run by multiple workers. From reading the docs, it looks good. However, would be nice if we could test the behavior though instead of relying solely on documentation.

huumn:
Awesome review! I'm working on retries now and hoping to have those done by the end of the weekend. I'll incorporate your advice/thoughts as continue to do more passes over the code.

ekzyis:
> However, would be nice if we could test the behavior though instead of relying solely on documentation.

I realized we can test this using two `psql` sessions:

https://github.com/stackernews/stacker.news/assets/27162016/879da578-d491-4f2b-a9c7-5ab7bb13822b

As you can see, the second transaction tries to update the same item but blocks. I included `WHERE msats = X` to simulate the where condition no longer matching after the first transaction committed its update. So when the first transaction commits, the second transaction no longer finds the row and returns `UPDATE 0`.

So I am more confident that indeed only one worker will be able to proceed after `UPDATE "Invoice" SET "actionState" = 'PAID' WHERE id = $1 AND "actionState" = 'PENDING'`, assuming Prisma _really_ does throw here as mentioned in their docs.

huumn:
Excellent! I was planning to run that test myself to extra confirm (I've also been running several workers locally and seeing them contend for the transition and only having one succeed).

> So I am more confident that indeed only one worker will be able to proceed after UPDATE "Invoice" SET "actionState" = 'PAID' WHERE id = $1 AND "actionState" = 'PENDING', assuming Prisma really does throw here as mentioned in their docs.

You're right. I think we'll want to avoid assuming this and explicitly check ourselves - just in case prisma decides to change that.

prisma/migrations/20240530174040_paid_actions/migration.sql
```diff
@@ -0,0 +1,312 @@
...
+            SET "bountyPaidTo" = array_append("bountyPaidTo", item_id)
+            WHERE id = root_id;
+        END IF;
+    END IF;
+
+    RETURN 0;
+END;
+$$;
+
+CREATE OR REPLACE FUNCTION referral_act(item_act_id INTEGER)
```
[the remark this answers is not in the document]
huumn: Yep. I still need to review these more too to make sure we aren't going to have issues without serialization; basically anything read-modify-write is a problem.

worker/paidAction.js
huumn: Read Commited is the lowest isolation level in postgres:

> PostgreSQL's Read Uncommitted mode behaves like Read Committed

Still, being explicit will help readability.

api/paidAction/buyCredits.js
```diff
@@ -0,0 +1,25 @@
+// XXX we don't use this yet ...
+// it's just showing that even buying credits
+// can eventually be a paid action
+
+import { USER_ID } from '@/lib/constants'
+
+export const anonable = false
+export const supportsPessimism = false
+export const supportsOptimism = true
```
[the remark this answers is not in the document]
huumn: Pessimism means "use a hold invoice" and only "do stuff" when it's paid. `BUY_CREDITS` is a weird one because it's optimistic and **only** has side effects, ie `onPaid`, ie incrementing the stacker's balance, because `perform` is just making an invoice which `paidAction/index.js` does for us.

components/item-act.js
```diff
@@ -120,177 +71,144 @@ export default function ItemAct ({ onClose, item, down, children, abortSignal, o
...
+            act: {
+              result: {
+                id: item.id, sats: Number(amount), act: down ? 'DONT_LIKE_THIS' : 'TIP', path: item.path
+              }
+            }
+          }
+        : undefined,
+      // don't close modal immediately because we want the QR modal to stack
+      onCompleted: () => {
+        onClose?.()
```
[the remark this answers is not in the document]
huumn: It does stack as a result of this. `onCompleted` means the invoice has been when using `usePaidMutation`, so there's nothing to stack if this gets called.

Before, I was calling after `await act(...)` and it was causing the race.

api/resolvers/item.js
```diff
@@ -859,35 +863,12 @@ export default {
         }
       }
 
-      if (me && idempotent) {
```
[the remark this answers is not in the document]
huumn: Yep, gone. I think my turbo zaps didn't work well  (eg it'd go from 1 to 100 to 10 to 100 when I'd triple zap) because of `serialize`.

api/paidAction/zap.js
```diff
@@ -0,0 +1,169 @@
...
+      FROM zapper, zapped
+    )
+    UPDATE "Item"
+    SET
+      "weightedVotes" = "weightedVotes" + zap.weighted_vote,
+      upvotes = upvotes + zap.first_vote,
+      msats = msats + ${msats},
+      "lastZapAt" = now()
+    FROM zap
+    WHERE id = ${itemAct.itemId}`
```
huumn, line 126: @ekzyis you might find this interesting. It's good to be aware of at least. There was a lot of conflicting info about CTEs and subqueries out there, but this kind of thing is unsafe in `read committed`.

api/paidAction/index.js
ekzyis, line 1: Was there a particular reason for that? Just curious because I am also used `await import` in my PR to stay consistent with this here

huumn: No reason I'm confident in. I just suspected it might make a difference to webpack and I didn't benefit from the dynamic imports in a material way.

api/paidAction/zap.js
huumn: Resolved in [77cd1a3](https://github.com/stackernews/stacker.news/pull/1195/commits/77cd1a398cb60a05f8365a64777da16492ac1373) by creating a table for storing user<->item aggregates that can receive row level locks.

api/paidAction/index.js
ekzyis: Ok, I just realized that the dynamic imports caused extremely slow page loads if the file that was dynamically imported itself imported something from `@/components/form` (ab5cd268).

Also noticed another slowness related to `nodata` which I first fixed by removing the server query for now in deb476b3 but then realized that the dynamic import was also the root cause for this and thus reverted that commit in 264adf43.

So I see dynamic imports as footguns now :eyes:

huumn: I didn't even think about dependencies. Good to know.

api/paidAction/downZap.js
```diff
@@ -0,0 +1,44 @@
...
+    invoiceData = { invoiceId, invoiceActionState: 'PENDING' }
+  }
+
+  const itemAct = await tx.itemAct.create({
+    data: { msats: cost, itemId, userId: me.id, act: 'DONT_LIKE_THIS', ...invoiceData }
+  })
+
+  const [{ path }] = await tx.$queryRaw`SELECT ltree2text(path) as path FROM "Item" WHERE id = ${itemId}`
+  return { id: itemId, sats, act: 'DONT_LIKE_THIS', path, actId: itemAct.id }
+}
```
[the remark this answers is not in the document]
ekzyis: It's probably not a big deal performance-wise since it's just an index lookup but what I meant was that in components/item-act.js, the `modifyActCache` function does not update any ancestors for downzaps:

```javascript
  if (act === 'TIP') {
    // update all ancestors
    path.split('.').forEach(aId => {
      if (Number(aId) === Number(id)) return
      cache.modify({
        id: `Item:${aId}`,
        fields: {
          commentSats (existingCommentSats = 0) {
            return existingCommentSats + sats
          }
        }
      })
    })
  }
```

So this `path` query seems unnecessary since this action is only for downzaps.

Just want to make sure we're on the same page and I didn't miss something

api/paidAction/zap.js
ekzyis: I have read https://www.cybertec-postgresql.com/en/transaction-anomalies-with-select-for-update/ now and it makes sense.

I assume you have decided to go for a table that aggregates data instead of using `REPEATABLE READ` transaction isolation level since that would mean we would need to do retries which is bad for performance? Might also be less safe since we'd have to make sure we always use this isolation level where we could run into such problems. So having a dedicated table for such queries without lock footguns makes sense.

api/paidAction/downZap.js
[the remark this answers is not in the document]
huumn: Ah I see what you mean! I like the symmetry between item acts all returning the same stuff, but you're right.

components/payment.js
```diff
@@ -105,7 +123,7 @@ const useInvoice = () => {
     return inv
   }, [cancelInvoice])
 
-  return { create, isPaid, waitUntilPaid, cancel }
+  return { create, isPaid, waitUntilPaid: waitController.wait, stopWaiting: waitController.stop, cancel }
```
ekzyis: Maybe it makes sense to rename it to `pollUntilPaid` and `stopPolling`.

components/payment.js
```diff
@@ -132,6 +150,8 @@ export const useWebLnPayment = () => {
     } catch (err) {
       console.error('WebLN payment failed:', err)
       throw err
+    } finally {
+      invoice.stopWaiting()
```
ekzyis, line 152: Oh, maybe you could have just called `invoice.cancel` here instead of writing new code with `AbortController`?

Did you try that?

components/payment.js
huumn: hmmm I like your original `wait` names though because poll doesn't imply blocking.

huumn:
@ekzyis this is ready for final review now. I'm going to review it again myself tonight, then I'll merge it tomorrow.

api/paidAction/index.js
```diff
@@ -0,0 +1,309 @@
...
+      result: await action.retry({ invoiceId, newInvoiceId: invoice.id }, context),
+      invoice,
+      paymentMethod: 'OPTIMISTIC'
+    }
+  }, { isolationLevel: Prisma.TransactionIsolationLevel.ReadCommitted })
+}
+
+// TODO: switch these back before going to production
+const OPTIMISTIC_INVOICE_EXPIRE = { minutes: 10 } // { hours: 1 }
+const PESSIMISTIC_INVOICE_EXPIRE = { minutes: 10 } // { minutes: 10 }
```
huumn, line 232: Actually, I think we want 10 minutes for these. We aren't able to determine if they're failed until they expire, so anything longer could lead to items or zaps being much later the person intended. At least when they're short, we can prompt them to retry in a reasonable time frame.

api/paidAction/itemCreate.js
```diff
@@ -0,0 +1,242 @@
...
+    item = await tx.item.findFirst({
+      where: { invoiceId: invoice.id },
+      include: {
+        mentions: true,
+        itemReferrers: { include: { refereeItem: true } },
+        user: true
+      }
+    })
+    await tx.itemAct.updateMany({ where: { invoiceId: invoice.id }, data: { invoiceActionState: 'PAID' } })
+    await tx.item.updateMany({ where: { invoiceId: invoice.id }, data: { invoiceActionState: 'PAID', invoicePaidAt: new Date() } })
```
huumn, line 169: We have to use `updateMany` because it's not unique on item.

api/paidAction/itemUpdate.js
```diff
@@ -0,0 +1,155 @@
...
+      msats: boostMsats, act: 'BOOST', userId: me?.id || USER_ID.anon
+    })
+  }
+
+  // createMany is the set difference of the new - old
+  // deleteMany is the set difference of the old - new
+  // updateMany is the intersection of the old and new
+  const difference = (a = [], b = [], key = 'userId') => a.filter(x => !b.find(y => y[key] === x[key]))
+  const intersectionMerge = (a = [], b = [], key) => a.filter(x => b.find(y => y.userId === x.userId))
+    .map(x => ({ [key]: x[key], ...b.find(y => y.userId === x.userId) }))
```
huumn, line 46: We can't control the order of nested create, updates, and deletes in prisma. Besides, this is more accurate.

api/paidAction/pollVote.js
```diff
@@ -0,0 +1,65 @@
...
+  return { id: pollOptionId }
+}
+
+export async function onPaid ({ invoice }, { tx }) {
+  if (!invoice) return
+
+  await tx.itemAct.updateMany({ where: { invoiceId: invoice.id }, data: { invoiceActionState: 'PAID' } })
+  await tx.pollBlindVote.updateMany({ where: { invoiceId: invoice.id }, data: { invoiceActionState: 'PAID' } })
+  // anonymize the vote
+  await tx.pollVote.updateMany({ where: { invoiceId: invoice.id }, data: { invoiceId: null, invoiceActionState: null } })
```
huumn, line 54: It occurred to me while doing this, that `pollVote` `createdAt` and `updatedAt` times can also be used to deanonymize votes.

api/paidAction/territoryBilling.js
```diff
@@ -0,0 +1,69 @@
...
+
+  return await tx.sub.update({
+    // optimistic concurrency control
+    // make sure the sub hasn't changed since we fetched it
+    where: {
+      ...sub,
+      postTypes: {
+        equals: sub.postTypes
+      }
+    },
```
huumn, line 50: This is a bit heavy handed, but it's actually pretty convenient. We are able to make sure the territory hasn't been updated ***at all***.

api/paidAction/territoryUnarchive.js
```diff
@@ -0,0 +1,61 @@
...
+    // optimistic concurrency control
+    // make sure none of the relevant fields have changed since we fetched the sub
+    where: {
+      ...sub,
+      postTypes: {
+        equals: sub.postTypes
+      }
+    }
+  })
+}
```
huumn, line 57: A lot of this code is identical to `TERRITORY_UPDATE`, but I didn't want to labor over finding the differences.

api/paidAction/zap.js
```diff
@@ -0,0 +1,151 @@
...
+  const acts = await tx.itemAct.createManyAndReturn({
+    data: [
+      { msats: feeMsats, itemId, userId: me?.id || USER_ID.anon, act: 'FEE', ...invoiceData },
+      { msats: zapMsats, itemId, userId: me?.id || USER_ID.anon, act: 'TIP', ...invoiceData }
+    ]
+  })
+
+  const [{ path }] = await tx.$queryRaw`
+    SELECT ltree2text(path) as path FROM "Item" WHERE id = ${itemId}::INTEGER`
+  return { id: itemId, sats, act: 'TIP', path, actIds: acts.map(act => act.id) }
```
huumn, line 37: We need `actIds` to correlate `perform` and `onPaid` absent an `invoiceId`, e.g. paying with fee credits.

api/paidAction/zap.js
```diff
@@ -0,0 +1,151 @@
...
+      FROM "ItemUserAgg"
+      JOIN "Item" ON "Item".id = "ItemUserAgg"."itemId"
+      LEFT JOIN "Item" root ON root.id = "Item"."rootId"
+      WHERE "ItemUserAgg"."userId" = ${itemAct.userId}::INTEGER
+      AND "ItemUserAgg"."itemId" = ${itemAct.itemId}::INTEGER
+      AND root."userId" = ${itemAct.userId}::INTEGER
+      AND root.bounty IS NOT NULL
+    )
+    UPDATE "Item"
+    SET "bountyPaidTo" = array_remove(array_append(array_remove("bountyPaidTo", bounty.target), bounty.target), NULL)
```
huumn, line 127: We should've done this triple array call before. It basically makes `bountyPaidTo` into a set.

api/resolvers/image.js
```diff
@@ -10,8 +10,8 @@ export default {
 }
 
 export function uploadIdsFromText (text, { models }) {
-  if (!text) return null
-  return [...text.matchAll(AWS_S3_URL_REGEXP)].map(m => Number(m[1]))
+  if (!text) return []
+  return [...new Set([...text.matchAll(AWS_S3_URL_REGEXP)].map(m => Number(m[1])))]
```
huumn, line 14: This is useful for when there are multiple links to the same image in a post.

api/resolvers/item.js
```diff
@@ -63,7 +64,10 @@ export async function getItem (parent, { id }, { me, models }) {
     query: `
       ${SELECT}
       FROM "Item"
-      WHERE id = $1`
+      ${whereClause(
+        '"Item".id = $1',
+        activeOrMine(me)
```
huumn, line 69: Prevents 404ing on our own pending items.

api/resolvers/item.js
```diff
@@ -132,11 +136,14 @@ export async function itemQueryWithMeta ({ me, models, query, orderBy = '' }, ..
       LEFT JOIN "MuteSub" ON "Sub"."name" = "MuteSub"."subName" AND "MuteSub"."userId" = ${me.id}
       LEFT JOIN "SubSubscription" ON "Sub"."name" = "SubSubscription"."subName" AND "SubSubscription"."userId" = ${me.id}
       LEFT JOIN LATERAL (
-        SELECT "itemId", sum("ItemAct".msats) FILTER (WHERE act = 'FEE' OR act = 'TIP') AS "meMsats",
-          sum("ItemAct".msats) FILTER (WHERE act = 'DONT_LIKE_THIS') AS "meDontLikeMsats"
+        SELECT "itemId",
+          sum("ItemAct".msats) FILTER (WHERE "invoiceActionState" IS DISTINCT FROM 'FAILED' AND (act = 'FEE' OR act = 'TIP')) AS "meMsats",
+          sum("ItemAct".msats) FILTER (WHERE "invoiceActionState" IS NOT DISTINCT FROM 'PENDING' AND (act = 'FEE' OR act = 'TIP')) AS "mePendingMsats",
+          sum("ItemAct".msats) FILTER (WHERE "invoiceActionState" IS DISTINCT FROM 'FAILED' AND act = 'DONT_LIKE_THIS') AS "meDontLikeMsats"
         FROM "ItemAct"
```
huumn, line 143: Null comparisons in sql are weird.

api/resolvers/item.js
```diff
@@ -1078,8 +1073,8 @@ export default {
       return (msats && msatsToSats(msats)) || 0
     },
     meDontLikeSats: async (item, args, { me, models }) => {
-      if (!me) return false
-      if (typeof item.meMsats !== 'undefined') {
+      if (!me) return 0
+      if (typeof item.meDontLikeMsats !== 'undefined') {
```
huumn, line 1077: I think was an unexposed bug because `meMsats` is undefined when `me..` is.

api/resolvers/item.js
```diff
@@ -1318,169 +1239,61 @@ export const updateItem = async (parent, { sub: subName, forward, options, ...it
 
   // prevent update if it's not explicitly allowed, not their bio, not their job and older than 10 minutes
   const myBio = user.bioId === old.id
-  const timer = Date.now() < new Date(old.createdAt).getTime() + 10 * 60_000
+  const timer = Date.now() < new Date(old.invoicePaidAt ?? old.createdAt).getTime() + 10 * 60_000
 
-  if (!allowEdit && !myBio && !timer) {
+  if (!allowEdit && !myBio && !timer && !isJob(item)) {
```
huumn, line 1244: Jobs haven't been editable for a minute.

api/resolvers/user.js
```diff
@@ -320,6 +322,8 @@ export default {
             '"SubSubscription"."userId" = $1',
             '"Item".created_at > $2',
             '"Item"."parentId" IS NULL',
+            '"Item"."userId" <> $1',
+            activeOrMine(me),
```
huumn, line 326: `activeOrMine` was making it so we'd get notifications for our own items.

api/resolvers/wallet.js
```diff
@@ -380,18 +382,10 @@ export default {
...
-            cancelled: true
-          }
-        }),
-        { models }
-      )
-      return inv
+      await models.$executeRaw`
+        INSERT INTO pgboss.job (name, data, retrylimit, retrybackoff, priority)
+          VALUES ('checkInvoice', jsonb_build_object('hash', ${hash}), 21, true, 100)`
+      return await models.invoice.findFirst({ where: { hash } })
```
huumn, line 388: So that we can reuse the existing invoice checking code.

api/resolvers/wallet.js
```diff
@@ -545,7 +539,47 @@ export default {
...
+              SELECT id, act, "invoiceId", "invoiceActionState", msats
+              FROM "ItemAct"
+              WHERE "ItemAct"."invoiceId" = ${Number(invoice.id)}
+              AND "ItemAct"."userId" = ${me?.id}
+              AND act = ${action2act[invoice.actionType]}::"ItemActType"`
+          )?.[0]
+        default:
+          return null
+      }
+    }
```
huumn, line 582: We do this for the invoice failure notifications.

components/comment.js
```diff
@@ -137,129 +136,117 @@ export default function Comment ({
...
-  const meSats = item.meSats + pendingSats
-  const downSats = item.meDontLikeSats + pendingDownSats
-
-  return item.outlawed && !me?.privates?.wildWestMode
-    ? <Skull className={styles.dontLike} width={24} height={24} />
-    : downSats > meSats
-      ? <DownZap width={24} height={24} className={styles.dontLike} item={item} />
-      : pin ? <Pin width={22} height={22} className={styles.pin} /> : <UpVote item={item} className={styles.upvote} />
-}
-
```
huumn, line 262: I wanted to reuse these but didn't want to test the changes I needed to make 😄

components/fee-button.js
```diff
@@ -127,7 +127,12 @@ export default function FeeButton ({ ChildButton = SubmitButton, variant, text,
   return (
     <div className={styles.feeButton}>
       <ActionTooltip overlayText={!free && total === 1 ? '1 sat' : feeText}>
-        <ChildButton variant={variant} disabled={disabled} nonDisabledText={feeText}>{text}</ChildButton>
+        <ChildButton
+          variant={variant} disabled={disabled}
+          appendText={feeText}
+          submittingText={free || !feeText ? undefined : 'paying...'}
+        >{text}
+        </ChildButton>
```
huumn, line 135: Slightly better UX, so they have so info that we are working on submitting the payment.

components/invoice.js
```diff
@@ -1,37 +1,49 @@
...
+
+  useEffect(() => {
+    if (!invoice) {
+      return
+    }
+    if (waitFor && waitFor(invoice)) {
+      onPayment?.(invoice)
+    }
+    setExpired(new Date(invoice.expiredAt) <= new Date())
+  }, [invoice, onPayment, setExpired])
```
huumn, line 38: For pessimistic payments, we want to wait for the `actionState = 'PAID'`

components/invoice.js
```diff
@@ -1,37 +1,49 @@
...
 
   if (error) {
-    return <div>{error.toString()}</div>
+    return <div>{error.message}</div>
   }
 
-  // if webLn was not passed, use true by default
-  if (webLn === undefined) webLn = true
```
huumn, line 34: Set in props instead.

components/invoice.js
```diff
@@ -48,43 +60,31 @@ export default function Invoice ({ invoice, modal, onPayment, info, successVerb,
...
     webLn = false
+  } else if (invoice.expiresAt) {
+    variant = 'pending'
+    status = (
+      <CompactLongCountdown
+        date={invoice.expiresAt} onComplete={() => {
+          setExpired(true)
+        }}
+      />
+    )
```
huumn, line 71: Replace loading spinner + expiration countdown to be just the expiration countdown.

components/invoice.js
```diff
@@ -48,43 +60,31 @@ export default function Invoice ({ invoice, modal, onPayment, info, successVerb,
...
 
   return (
     <>
       {webLnError && !(webLnError instanceof WebLnNotEnabledError) &&
-        <div className='text-center text-danger mb-3'>
-          Payment from attached wallet failed:
-          <div>{webLnError.toString()}</div>
+        <div className='text-center fw-bold text-info mb-3' style={{ lineHeight: 1.25 }}>
+          Paying from attached wallet failed:
+          <code> {webLnError.message}</code>
```
huumn, line 81: I think we should begin using red text/backgrounds very sparingly. It should be an absolute emergency.

components/item-act.js
```diff
@@ -301,10 +242,10 @@ export class ActCanceledError extends Error {
 }
 
 export class ZapUndoController extends AbortController {
-  constructor () {
+  constructor ({ onStart = () => {}, onDone = () => {} }) {
     super()
-    this.signal.start = () => { this.started = true }
-    this.signal.done = () => { this.done = true }
+    this.signal.start = onStart
+    this.signal.done = onDone
```
huumn, line 248: This was coincidentally updating `UpVote`s state and causing it to rerender previously. Here, we make it explicitly update `UpVote`s state.

components/item-info.js
```diff
@@ -22,22 +22,23 @@ import { DropdownItemUpVote } from './upvote'
...
+import { useQrPayment } from './payment'
+import { useRetryCreateItem } from './use-item-submit'
 
 export default function ItemInfo ({
   item, full, commentsText = 'comments',
   commentTextSingular = 'comment', className, embellishUser, extraInfo, onEdit, editText,
   onQuoteReply, extraBadges, nested, pinnable, showActionDropdown = true, showUser = true
 }) {
-  const editThreshold = new Date(item.createdAt).getTime() + 10 * 60000
+  const editThreshold = new Date(item.invoicePaidAt ?? item.createdAt).getTime() + 10 * 60000
```
huumn, line 33: We start the ability to edit from the point were the invoice is paid.

components/item-info.js
```diff
@@ -58,7 +63,53 @@ export default function ItemInfo ({
...
+        <span> \ </span>
+        <span
+          className='text-reset pointer fw-bold'
+          onClick={onClick}
+        >
+          <Component />
+        </span>
+      </>
+    )
+  }
```
huumn, line 112: Admittedly a mess.

components/item-info.js
```diff
@@ -144,72 +195,66 @@ export default function ItemInfo ({
...
+                  <hr className='dropdown-divider' />
+                  <OutlawDropdownItem item={item} />
+                </>}
+              {item.mine && item.invoiceId &&
+                <>
+                  <hr className='dropdown-divider' />
+                  <Link href={`/invoices/${item.invoiceId}`} className='text-reset dropdown-item'>
+                    view invoice
+                  </Link>
+                </>}
```
huumn, line 235: Adds the ability to see the invoice that paid for an item.

components/modal.js
```diff
@@ -47,9 +47,17 @@ export default function useModal () {
...
+      const content = getCurrentContent()
+      const { persistOnNavigate } = content?.options || {}
+      if (!persistOnNavigate) {
+        onClose()
+      }
+    }
+
+    router.events.on('routeChangeStart', maybeOnClose)
+    return () => router.events.off('routeChangeStart', maybeOnClose)
+  }, [router.events, onClose, getCurrentContent])
```
huumn, line 60: Let us keep the modal open on navigation. This is useful for optimistic posting which navigates after the mutation completes.

components/notifications.js
```diff
@@ -404,16 +499,14 @@ function ItemMention ({ n }) {
...
-            </div>)}
-      </div>
+      {n.item?.title
+        ? <div className='ps-2'><Item item={n.item} /></div>
+        : (
+          <div className='pb-2'>
+            <RootProvider root={n.item.root}>
+              <Comment item={n.item} noReply includeParent rootText='replying on:' clickToContext />
+            </RootProvider>
+          </div>)}
```
huumn, line 509: The spacing of these notifications caused the bolt to have not padding off the left edge.

components/payment.js
```diff
@@ -60,53 +60,75 @@ const useInvoice = () => {
...
-    }
+    const { hash, cancelled } = data.invoice
+
     if (cancelled) {
       throw new InvoiceCanceledError(hash)
     }
-    return false
+
+    return that(data.invoice)
   }, [client])
```
huumn, line 75: More generic state checking for invoice.

components/qr.js
```diff
@@ -48,9 +47,6 @@ export function QrSkeleton ({ status, description, bolt11Info }) {
         <InputSkeleton />
       </div>
       <InvoiceStatus variant='default' status={status} />
-      <div className='text-muted text-center invisible'>
-        <SimpleCountdown date={Date.now()} />
-      </div>
```
huumn, line 53: Invoice status now displays the expiration countdown when it's pending.

components/territory-form.js
```diff
@@ -56,22 +36,28 @@ export default function TerritoryForm ({ sub }) {
...
+              fragment: gql`
+                fragment SubSubmitFragment on Sub {
+                  name
+                }`
+            })
+            if (existing.some(ref => readField('name', ref) === variables.name)) {
+              return existing
+            }
+            return [...existing, newSubRef]
           }
```
huumn, line 61: Takes care of an apollo warning.

components/webln/nwc.js
```diff
@@ -263,8 +263,7 @@ export function NWCProvider ({ children }) {
       logger.error('payment failed:', `payment_hash=${hash}`, err.message || err.toString?.())
       throw err
     } finally {
-      // For some reason, websocket is already in CLOSING or CLOSED state.
-      // relay?.close()
+      relay?.close()?.catch()
```
huumn, line 266: I let this call close any subscriptions from the try block.

fragments/notifications.js
```diff
@@ -1,12 +1,36 @@
...
+        ...ItemFullFields
+        ...PollFields
+      }
+      itemAct {
+        id
+        act
+        invoiceId
+        invoiceActionState
+      }
+    }
```
huumn, line 29: On retry, item.invoiceId and itemAt.invoiceId get replaced so we need them to have their own fields for the values.

lib/apollo.js
```diff
@@ -32,6 +32,37 @@ function getClient (uri) {
...
+          'ForwardedVotification',
+          'Revenue',
+          'SubStatus',
+          'TerritoryPost',
+          'TerritoryTransfer',
+          'Reminder',
+          'ItemMention',
+          'Invoicification'
+        ]
+      },
```
huumn, line 65: Apollo needs to be told explicitly about interface types.

lib/constants.js
```diff
@@ -12,14 +12,15 @@ export const UPLOAD_SIZE_MAX_AVATAR = 5 * 1024 * 1024
 export const IMAGE_PIXELS_MAX = 35000000
 // backwards compatibile with old media domain env var and precedence for docker url if set
 export const MEDIA_URL = process.env.MEDIA_URL_DOCKER || process.env.NEXT_PUBLIC_MEDIA_URL || `https://${process.env.NEXT_PUBLIC_MEDIA_DOMAIN}`
-export const AWS_S3_URL_REGEXP = new RegExp(`${MEDIA_URL}/([0-9]+)`, 'g')
+export const AWS_S3_URL_REGEXP = new RegExp(`${process.env.NEXT_PUBLIC_MEDIA_URL || `https://${process.env.NEXT_PUBLIC_MEDIA_DOMAIN}`}/([0-9]+)`, 'g')
```
huumn, line 15: This broke finding uploadIds in item.text in development.

package.json
```diff
@@ -67,7 +67,7 @@
     "page-metadata-parser": "^1.1.4",
     "pg-boss": "^9.0.3",
     "piexifjs": "^1.0.6",
-    "prisma": "^5.4.2",
+    "prisma": "^5.14.0",
```
huumn, line 70: Gives us `createManyAndReturn`

pages/invites/index.js
```diff
@@ -51,9 +51,7 @@ function InviteForm () {
             gift: Number(gift), limit: limit ? Number(limit) : limit
           }
         })
-        if (error) {
-          throw new Error({ message: error.String() })
-        }
+        if (error) throw error
```
huumn, line 54: I'm not sure why I ever did this weird error wrapping.

pages/invoices/[id].js
```diff
@@ -1,30 +1,18 @@
...
-        pollInterval: FAST_POLL_INTERVAL,
-        variables: { id: router.query.id },
-        nextFetchPolicy: 'cache-and-network'
-      })
 
   return (
     <CenterLayout>
-      {error && <div>{error.toString()}</div>}
-      {data ? <Invoice invoice={data.invoice} /> : <QrSkeleton description status='loading' bolt11Info />}
+      <Invoice id={router.query.id} query={INVOICE_FULL} poll description status='loading' bolt11Info webLn={false} />
```
Comment (huumn) [inline pages/invoices/[id].js:15]: did the todo

worker/paidAction.js
```diff
@@ -0,0 +1,133 @@
+import { paidActions } from '@/api/paidAction'
+import { datePivot } from '@/lib/time'
+import { Prisma } from '@prisma/client'
+import { getInvoice } from 'ln-service'
+
+async function transitionInvoice (jobName, { invoiceId, fromState, toState, toData, onTransition }, { models, lnd, boss }) {
+  console.group(`${jobName}: transitioning invoice ${invoiceId} from ${fromState} to ${toState}`)
```
huumn, line 7: We should start using `console.group` more! It'd be especially useful clientside.

worker/streak.js
```diff
@@ -13,12 +13,19 @@ export async function computeStreaks ({ models }) {
...
           FROM "Donation"
           WHERE (created_at AT TIME ZONE 'UTC' AT TIME ZONE 'America/Chicago')::date >= (now() AT TIME ZONE 'America/Chicago' - interval '1 day')::date
-      )) spending
+      )
+      UNION ALL
+      (SELECT "userId", floor(sum("SubAct".msats)/1000) as sats_spent
+        FROM "SubAct"
+        WHERE (created_at AT TIME ZONE 'UTC' AT TIME ZONE 'America/Chicago')::date >= (now() AT TIME ZONE 'America/Chicago' - interval '1 day')::date
+        AND "type" = 'BILLING'
+        GROUP BY "userId")) spending
```
huumn, line 28: we weren't including territory billing in cowboy streaks!

api/paidAction/index.js
```diff
@@ -0,0 +1,310 @@
...
+      actionState: 'HELD'
+    }
+  })
+
+  if (invoice.msatsReceived < cost) {
+    throw new Error('invoice amount too low')
+  }
+
+  return invoice
+}
```
ekzyis, line 310: This function looks suspiciously simple compared to all the checks in the previous version :monocle_face: 

But I guess making sure it's in `HELD` makes all the other checks indeed obsolete.

api/resolvers/user.js
```diff
@@ -424,7 +435,8 @@ export default {
             confirmedAt: {
               gt: lastChecked
             },
-            isHeld: null
+            isHeld: null,
+            actionType: null
```
ekzyis, line 439: In the future, `actionType` will be `BUY_CREDITS`?

api/resolvers/wallet.js
ekzyis: Why not import `checkInvoice` and run it ourselves?

Like this, this means that the invoice isn't actually cancelled yet when the mutation returns.

Even the response would indicate that. So a hypothetical user interface where we have a cancel button next to the invoice state, the invoice state wouldn't update on cancel (since it's not cancelled yet).

We don't have this anywhere yet afaict but I considered being able to cancel deposits manually (instead of letting them expire) in the past.

api/resolvers/wallet.js
```diff
@@ -545,7 +539,47 @@ export default {
...
+        case 'ZAP':
+        case 'DOWN_ZAP':
+        case 'POLL_VOTE':
+          return (await models.$queryRaw`
+              SELECT id, act, "invoiceId", "invoiceActionState", msats
+              FROM "ItemAct"
+              WHERE "ItemAct"."invoiceId" = ${Number(invoice.id)}
+              AND "ItemAct"."userId" = ${me?.id}
+              AND act = ${action2act[invoice.actionType]}::"ItemActType"`
+          )?.[0]
```
ekzyis, line 578:
```suggestion
          return (await models.$queryRaw`
              SELECT id, act, "invoiceId", "invoiceActionState", msats
              FROM "ItemAct"
              WHERE "ItemAct"."invoiceId" = ${Number(invoice.id)}::INTEGER
              AND "ItemAct"."userId" = ${me?.id}::INTEGER
              AND act = ${action2act[invoice.actionType]}::"ItemActType"`
          )?.[0]
```

Why `$queryRaw` though? Wouldn't `models.itemAct` work?

ekzyis (commented):
This is a partial review. I read every single new line in the backend now and made sure I understand what it's doing. Left comments where I had questions.

Only components/, fragments/ and pages/ is left to review.

Looks good so far!

pages/invites/index.js
ekzyis: Mhh, the only thing that I can tell this does it to lose where the error actually happened, lol.

api/paidAction/buyCredits.js
```diff
@@ -0,0 +1,26 @@
...
+
+export async function getCost ({ amount }) {
+  return satsToMsats(amount)
+}
+
+export async function onPaid ({ invoice }, { tx }) {
+  return await tx.users.update({
+    where: { id: invoice.userId },
+    data: { balance: { increment: invoice.msatsReceived } }
```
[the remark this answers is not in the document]
huumn: This might've been a copilot autofill thing. I think we'll do mcredits for the fee credit balance

api/resolvers/notifications.js
```diff
@@ -330,6 +332,22 @@ export default {
...
+        AND (
+          "Invoice"."actionType" = 'ITEM_CREATE' OR
+          "Invoice"."actionType" = 'ZAP' OR
+          "Invoice"."actionType" = 'DOWN_ZAP' OR
+          "Invoice"."actionType" = 'POLL_VOTE'
+        )
+        ORDER BY "sortTime" DESC
+        LIMIT ${LIMIT})`
+      )
+
```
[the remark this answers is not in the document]
huumn: Without someone explicitly hitting a `cancel` button it feels weird to assume they meant to cancel when they closed a modal. The binary success/failure is a lot easier to think through too. 

The approach I took was "they tried to pay and the flow didn't complete" = failed.

----------

One improvement on retries that I'd like to make is doing "implicit retries" where if you repeat a failed action like you do in your video (without hitting the `retry` button), the backend recognizes it's repeating something that failed and does a retry rather than creating a new action.

api/paidAction/README.md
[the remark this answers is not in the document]
huumn: The way I approached writing mine was "what would I write a future version of myself when I have to modify this code in the future."

api/paidAction/README.md
```diff
@@ -0,0 +1,204 @@
...
+| ----------------- | ----------- | ---------- | ----------- | -------- | ---------- | ----------- | ------------ |
+| zaps              | x           | x          | x           | x        | x          | x           | x            |
+| posts             | x           | x          | x           | x        | x          |             | x            |
+| comments          | x           | x          | x           | x        | x          |             | x            |
+| downzaps          | x           | x          |             |          | x          |             | x            |
+| poll votes        | x           | x          |             |          | x          |             |              |
+| territory actions | x           |            | x           |          | x          |             |              |
+| donations         | x           |            | x           | x        | x          |             |              |
+| update posts      | x           |            | x           |          | x          |             | x            |
+| update comments   | x           |            | x           |          | x          |             | x            |
```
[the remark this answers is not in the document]
huumn: That'd be cool

api/paidAction/zap.js
huumn: Mutating input parameters is an anti-pattern because it's a side effect and side effects are hidden from the calling context. Introducing side effects to functions mean that when we change the calling context, we would have to know about all the potential side effects of the functions it calls, wherever they are, because they can't be seen from the calling context.

api/resolvers/user.js
huumn: Yes, good point. All invoices should have an `actionType` and run through the state machine in the hopefully near future.

api/resolvers/wallet.js
huumn: Good rec!

api/resolvers/wallet.js
huumn: Probably!

components/item-act.js
```diff
@@ -50,35 +49,110 @@ const setItemMeAnonSats = ({ id, amount }) => {
...
+      onClose?.()
+      try {
+        await abortSignal.pause({ me, amount })
+      } catch (error) {
+        if (error instanceof ActCanceledError) {
+          return
+        }
+      }
+    }
+    strike()
```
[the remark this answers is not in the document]
huumn: oh hmmm I'll give this a shot

api/resolvers/item.js
```diff
@@ -132,11 +136,14 @@ export async function itemQueryWithMeta ({ me, models, query, orderBy = '' }, ..
...
-        SELECT "itemId", sum("ItemAct".msats) FILTER (WHERE act = 'FEE' OR act = 'TIP') AS "meMsats",
-          sum("ItemAct".msats) FILTER (WHERE act = 'DONT_LIKE_THIS') AS "meDontLikeMsats"
+        SELECT "itemId",
+          sum("ItemAct".msats) FILTER (WHERE "invoiceActionState" IS DISTINCT FROM 'FAILED' AND (act = 'FEE' OR act = 'TIP')) AS "meMsats",
+          sum("ItemAct".msats) FILTER (WHERE "invoiceActionState" IS NOT DISTINCT FROM 'PENDING' AND (act = 'FEE' OR act = 'TIP')) AS "mePendingMsats",
+          sum("ItemAct".msats) FILTER (WHERE "invoiceActionState" IS DISTINCT FROM 'FAILED' AND act = 'DONT_LIKE_THIS') AS "meDontLikeMsats"
         FROM "ItemAct"
         WHERE "ItemAct"."userId" = ${me.id}
         AND "ItemAct"."itemId" = "Item".id
+        AND "Item"."userId" <> ${me.id}
```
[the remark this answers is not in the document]
huumn: I think I did this so the post `FEE` was not picked up, but that might not be the right thing to do. I'll double check.

api/resolvers/item.js
```diff
@@ -724,10 +736,6 @@ export default {
       if (old.bio) {
         throw new GraphQLError('cannot delete bio', { extensions: { code: 'BAD_INPUT' } })
       }
-      // clean up any pending reminders, if triggered on this item and haven't been executed
-      if (hasReminderCommand(old.text)) {
-        await deleteReminderAndJob({ me, item: old, models })
-      }
```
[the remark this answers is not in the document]
huumn: It's because I removed `deleteReminderAndJob`, but I can add it back

components/item-act.js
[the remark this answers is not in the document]
huumn: Good rec

huumn:
Excellent review! You caught a lot of good stuff

api/resolvers/item.js
```diff
@@ -802,7 +809,6 @@ export default {
         return await updateItem(parent, { id, ...item }, { me, models })
```
ekzyis, line 809: To test edits that cost something, I updated the image fees function such that an image always costs 100 sats (like for anons). I noticed that `lnd` was undefined inside `createDbInvoice` due to this:

```suggestion
        return await updateItem(parent, { id, ...item }, { me, models, lnd })
```

Same for line 800, here is a full patch since I can't comment on line 800:

```diff
diff --git a/api/resolvers/item.js b/api/resolvers/item.js
index 95e76820..f83382d0 100644
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -797,7 +797,7 @@ export default {
       item.maxBid ??= 0

       if (id) {
-        return await updateItem(parent, { id, ...item }, { me, models })
+        return await updateItem(parent, { id, ...item }, { me, models, lnd })
       } else {
         return await createItem(parent, item, { me, models, lnd, hash, hmac })
       }
@@ -806,7 +806,7 @@ export default {
       await ssValidate(commentSchema, item)

       if (id) {
-        return await updateItem(parent, { id, ...item }, { me, models })
+        return await updateItem(parent, { id, ...item }, { me, models, lnd })
       } else {
         item = await createItem(parent, item, { me, models, lnd, hash, hmac })
         return item
```

ekzyis (changes requested):
Another review pass. Noticed a bug with edits. See comment.

components/use-paid-mutation.js
```diff
@@ -0,0 +1,165 @@
...
+export function usePaidMutation (mutation,
+  { onCompleted, ...options } = {}) {
+  options.optimisticResponse = addOptimisticResponseExtras(options.optimisticResponse)
+  const [mutate, result] = useMutation(mutation, options)
+  const waitForWebLnPayment = useWebLnPayment()
+  const waitForQrPayment = useQrPayment()
+  const invoiceWaiter = useInvoice()
+  const client = useApolloClient()
+  // innerResult is used to store/control the result of the mutation when innerMutate runs
+  const [innerResult, setInnerResult] = useState(result)
```
ekzyis, line 26: Is this needed anywhere? I searched for `usePaidMutation` and I only found code which only used the first returned element.

ekzyis (changes requested):
Final review

Requesting changes due to the `lnd` bug mentioned in the previous review.

components/use-paid-mutation.js
huumn: I'm not sure what you mean exactly. Do you mean is innerResult needed? We return it from the function.

The idea with `usePaidMutation` to mostly mirror `useMutation` so that if you're familiar with calling `useMutation`, it should be easy to use `usePaidMutation`.

components/use-paid-mutation.js
```diff
@@ -0,0 +1,165 @@
...
+  }, [waitForWebLnPayment, waitForQrPayment])
+
+  const innerMutate = useCallback(async ({
+    onCompleted: innerOnCompleted, ...innerOptions
+  } = {}) => {
+    innerOptions.optimisticResponse = addOptimisticResponseExtras(innerOptions.optimisticResponse)
+    let { data, ...rest } = await mutate(innerOptions)
+
+    // use the most inner callbacks/options if they exist
+    const { onPaid, onPayError, forceWaitForPayment, persistOnNavigate } = { ...options, ...innerOptions }
```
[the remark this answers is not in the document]
huumn: No, I don't think so. It's another solution to the problem `persistOnNavigate` solves.

components/use-paid-mutation.js
```diff
@@ -0,0 +1,165 @@
...
+            variables: {
+              ...options.variables,
+              ...innerOptions.variables,
+              hmac: invoice.hmac,
+              hash: invoice.hash
+            }
+          }))
+          // block until the invoice to be marked as paid
+          // for pessimisitic actions, they won't show up on navigation until they are marked as paid
+          await invoiceWaiter.waitUntilPaid(invoice, inv => inv?.actionState === 'PAID')
```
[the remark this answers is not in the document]
huumn: LND doesn't call it paid. It calls it settled and held. We call it paid and held.

If we zoom out, afaict the confusing thing is the name of this function here which suggests that paid has multiple meanings. I didn't want to rename this function initially, but I can rename it.

components/use-paid-mutation.js
ekzyis: > We return it from the function.

Yes but I don't see it used anywhere and I wondered if that was just old code or something.

> The idea with `usePaidMutation` to mostly mirror `useMutation` so that if you're familiar with calling `useMutation`, it should be easy to use `usePaidMutation`.

Ah, right, that makes sense.

api/resolvers/item.js
huumn: these also need hash and hmac. Nice catch

PR #1197, opened 2024-05-27
Make Polls Anonymous

### LINKED ISSUE #1192: Make polls anonymous
We current record a person's poll choice. For bitcoin-related polls this limits participation.

We should record that a person voted in a poll and prevent voting again, but we don't need to know their poll choice.

It'd be nice to also apply this retroactively.

SatsAllDay:
## Description
Closes #1192 

Introduce a `PollBlindVote` DB table that tracks when a user votes in a poll, but does not track which choice they made.

Alter the `PollVote` DB table to remove the `userId` column, meaning `PollVote` now tracks poll votes anonymously - it captures votes per poll option, but does not track which user submitted the vote.

Update the `poll_vote` DB function to work with both tables now.

Update the `item.poll` resolver to calculate `meVoted` based on the `PollBlindVote` table instead of `PollVote`.

## Checklist


huumn:
Try as I might I couldn't fault it :)

PR #1200, opened 2024-05-29
Use refs for modals

ekzyis:
## Description

This fixes concurrent `showModal` calls with stale values which lead to modals getting replaced instead of stacked.

fixes https://github.com/stackernews/stacker.news/pull/1194#issuecomment-2135756183

## Checklist

Yes

Searched for `showModal` calls and thus tested modals for QR codes, custom zaps, zoomable images, avatar upload, delete, anon info, "you will lose your work"-modal. There are more but looks good to me

### DIFF
--- a/components/modal.js
+++ b/components/modal.js
@@ -1,4 +1,4 @@
-import { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react'
+import { createContext, useCallback, useContext, useEffect, useMemo, useReducer, useRef } from 'react'
 import Modal from 'react-bootstrap/Modal'
 import BackArrow from '@/svgs/arrow-left-line.svg'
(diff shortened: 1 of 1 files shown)


huumn:
Curious. How does this fix the issue described in the comment? That issue was unrelated to modals afaict

ekzyis:
> Curious. How does this fix the issue described in the comment? That issue was unrelated to modals afaict

It's related to modals because the payments weren't lost but their errors opened up a modal but you never saw all of them because only the modal for the most recent error was shown. The previous modals were simply replaced without running the `onClose` of the previous ones. So it only looked like they were lost, but they were still pending.[^1]

[^1]: But they should get marked as failed with "payment was interrupted" as the reason

Issue #690, opened 2023-12-15
Only view posts that have at least X amount of sats invested in them


huumn:
@benthecarman requested a "cowboy hats only" mode as a way to filter out people only sharing content to farm sats or "assmilk" as DarthCoin puts it. I think having a sat threshold is more generic and useful.

This should probably only be applied to `recent` to start which suffers from this the worst.

It should be available in settings and default to `0` ... this would replace `freebie` mode as `freebie` mode basically means `investment=0`

What we mean by X amount of sats invested: 
- the item cost >=X sats to post
- or the item has >=X sats zapped to it
- or the item has >=X sats of boost

ekzyis:
Ideas for the setting name:

- saloon mode

_(I assume this is going to be just another setting like greeter and wild west mode)_

PR #1205, opened 2024-05-29
Allow API key access by default

ekzyis:
## Description

This PR updates every user to allow them API key access by default. The info was also updated to mention that this access can be disabled at any time.

Motivation for this is to allow anyone to build on SN without having to ask for permission in ~meta. API key access was disabled by default in #915 to limit our exposure to this new auth method. With #980 and #989 in place and after having gauged initial interest, I think it's safe to remove this requirement now.

Additionally, any item created with API keys will now be marked as `bot`. 

## Screenshots

2024-05-29-025646_594x85_scrot

## Checklist

Yes

Yes

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -1341,6 +1341,9 @@ export const createItem = async (parent, { forward, options, ...item }, { me, mo
     item.url = removeTracking(item.url)
   }
 
+  // mark item as created with API key
+  item.apiKey = me?.apiKey
+
   const uploadIds = uploadIdsFromText(item.text, { models })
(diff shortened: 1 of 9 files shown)


ekzyis:
Closing because we currently have different priorities and not the resources to deal with any potential downfall of this.

For the same reasons, we're not going to allow any more API keys for the moment. 

We discussed that having additional filters like #690 would give us more confidence in allowing general API access.

ekzyis:
@k00b, I think we can cherry-pick 71db5e87819fa20180a0b1bf8e10247ba7196391 and 6459295e49f6de2fac014dd3bbee37f79f1b4a73 from this branch. What do you think?

PR #1207, opened 2024-05-29
Close related section

### LINKED ISSUE #837: Related section not collapsed after first reply
https://stacker.news/items/426061

We want to keep this functionality but it should collapse related posts on the first comment submission.

tsmith123:
Fixes #837 

## Description
This PR ensures that the Related Posts section is closed once a comment is made.

This issue took me a while to find but eventually found the problem within Accordion component.
- Made use of `activeKey` and `onSelect` props
- Implemented `onSelect` handler to update `activeKey` value
- Set `activeKey` value on load
- Use `null` instead of `undefined` (Accordion doesn't seem to behave when using `undefined`)

## Video
https://github.com/stackernews/stacker.news/assets/7639854/cf6cb2a3-29fc-4428-848b-1889ceead087

## Testing
Please QA this if possible as I may have overlooked something.

### DIFF
--- a/components/item-full.js
+++ b/components/item-full.js
@@ -176,7 +176,14 @@ function TopLevelItem ({ item, noReply, ...props }) {
       </article>
       {!noReply &&
         <>
-          <Reply item={item} replyOpen placeholder={item.ncomments > 3 ? 'fractions of a penny for your thoughts?' : 'early comments get more zaps'} onCancelQuote={cancelQuote} onQuoteReply={quoteReply} quote={quote} />
+          <Reply
+            item={item}
+            replyOpen
+            placeholder={item.ncomments > 3 ? 'fractions of a penny for your thoughts?' : 'early comments get more zaps'}
+            onCancelQuote={cancelQuote}
+            onQuoteReply={quoteReply}
+            quote={quote}
+          />
           {
           // Don't show related items for Saloon items (position is set but no subName)
           (!item.position && item.subName) &&
--- a/components/item-full.js
+++ b/components/item-full.js
@@ -195,7 +195,7 @@ function TopLevelItem ({ item, noReply, ...props }) {
           !item.deletedAt &&
           // Don't show related items for items with bounties, show past bounties instead
           !(item.bounty > 0) &&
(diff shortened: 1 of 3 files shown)


components/item-full.js
```diff
@@ -188,7 +195,7 @@ function TopLevelItem ({ item, noReply, ...props }) {
           !item.deletedAt &&
           // Don't show related items for items with bounties, show past bounties instead
           !(item.bounty > 0) &&
-            <Related title={item.title} itemId={item.id} show={item.ncomments === 0} />
+            <Related title={item.title} itemId={item.id} show={item.ncomments % 2 === 0} />
```
ekzyis, line 198: Was this only for testing? This makes the related section show for posts with even comment count

ekzyis (commented):
Looks good except one thing, see comment

components/reply.js
```diff
@@ -31,7 +31,15 @@ export function ReplyOnAnotherPage ({ item }) {
...
-export default forwardRef(function Reply ({ item, onSuccess, replyOpen, children, placeholder, onQuoteReply, onCancelQuote, quote }, ref) {
+export default forwardRef(function Reply ({
+  item,
+  replyOpen,
+  children,
+  placeholder,
+  onQuoteReply,
+  onCancelQuote,
+  quote
+}, ref) {
```
ekzyis, line 42: I see that `onSuccess` was removed here. Accident or on purpose?

tsmith123: It wasn't being used so I removed it

components/item-full.js
tsmith123: Oh bugger. Yes that was for testing. I'll put that back. It saved me from making new posts each time and waiting 10 mins

tsmith123:
Thanks @ekzyis - good spot on the conditional. I've put that back. My testing was done with the proper conditional btw.

components/item-full.js
tsmith123: Done

components/reply.js
ekzyis: I see, makes sense

ekzyis (approved):

Issue #667, opened 2023-12-06
In SN post, show links from other SN posts to it


kristapsk:
Back in the day, some 15-20 years ago, in blogging there was that [trackback](https://en.wikipedia.org/wiki/Trackback) thing when some blog post contained link to some other blog post it notified other site about that and it could be shown like a comment below the post. Ok, that thing is probably dead today. But it would be useful internally for SN to show references to SN post from other SN posts. For example, I mentioned [Heroes of Might and Magic III](https://stacker.news/items/340167) post in my post about [King's Bounty](https://stacker.news/items/341376). Would be cool that in HoMMIII post visitor could see that it is mentioned in KB post.

huumn:
Notify the author: https://stacker.news/items/438451

abhiShandy:
Would you like to add a cost to cite other posts since it is going to create a comment (and comments are not free)?

ekzyis:
No, no cost, we want to incentivize linking to other content for evergreenness

I also don't think we want to have it as replies, maybe references could be in the "related" section with some special indicator or inside an own section? But these sections don't show for comments (even if they are the root item of the page), so mh.

huumn:
Agreed, it shouldn't create a comment. 

Without a better idea, I think it should just be a link in the overflow menu, ie the `...` on the item perhaps shown as `X references`.

When clicked it takes you to a page like the related page: https://stacker.news/items/522161/related

ekzyis:
We could also show it in the item info `100 sats \ 3 comments \ 5 references` if references > 0. That would work for comments, too.

But maybe that's too prominent

huumn:
It's mostly the screen real estate I'm concerned about.

Maybe: `100 sats \ 3 comments \ 5 refs`

Still, a lot to look at. Alternatively:

<img width="918" alt="Screenshot 2024-04-28 at 5 06 58 PM" src="https://github.com/stackernews/stacker.news/assets/34140557/4bc868d6-0745-438b-acea-48c9f32ab156">

dillon-co:
what about a second dropdown next to or under `related posts` labeled `referenced by` ?

PR #1208, opened 2024-05-29
Item mention notifications

ekzyis:
## Description

This adds notifications if one of your posts or comments were mentioned.

I also added parsing of references as `#<itemId>` to links so internal references are now linked both ways. This means that `#1234` will now be parsed as a link to https://stacker.news/items/1234 (see d449d8bf).

Related to https://github.com/stackernews/stacker.news/issues/667 but doesn't close it since this PR is only about notifications

## Screenshots

2024-05-31-102231_514x83_scrot

## Additional Context

I initially wanted to include the type of item that was referred in the title. So if someone referred to a post of yours, it would show `one of your posts was mentioned` instead of `one of your comments was mentioned`. However, that made the code more complicated (see d9740250).

I now use `items` to mean both but I think that's not a good name since it might be confusing for stackers what we mean by that. I thought about using "submissions" instead but that's a completely new term so maybe even more confusing.

## Checklist

Yes

Yes

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -15,7 +15,7 @@ import { msatsToSats } from '@/lib/format'
 import { parse } from 'tldts'
 import uu from 'url-unshort'
 import { actSchema, advSchema, bountySchema, commentSchema, discussionSchema, jobSchema, linkSchema, pollSchema, ssValidate } from '@/lib/validate'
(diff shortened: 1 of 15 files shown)


huumn:
I get a ton of false positives from `#<num>` being an item link now. I had five false positive item mentions this morning from people sharing github release notes. 

I recall talking about this in the past and that we were going to wait until `#` autocompleted like `@` ... it'd allow us to distinguish between someone saying "I'm #1" and someone want to share `#1`.

Issue #1218, opened 2024-06-03
Comments non-deterministically missing after submission


huumn:
It happened to @SatsAllDay in response to a post here: https://stacker.news/items/560966/r/k00b?commentId=560967

It happened to me in a response to a comment here: https://stacker.news/items/560297/r/k00b?commentId=561044

ekzyis:
I think this is the cause:

![2024-06-03-181343_1053x370_scrot](https://github.com/stackernews/stacker.news/assets/27162016/e06d8a2a-f8a3-4918-81f1-1c51f39dd1d3)

I encountered this error while working on https://github.com/stackernews/stacker.news/pull/1184 it but I thought it was part of my postpaid changes which weren't included in https://github.com/stackernews/stacker.news/pull/1194

ekzyis:
I can reliably reproduce the comment not showing up when I zap the parent first. But the cache warning does not show up. So maybe it's unrelated or another issue. I can't reproduce the cache warning.

huumn:
If I uncomment the optimistic mutation code, it appears to fix it.

ekzyis:
This also seems to fix it (at least for successful zaps):

```diff
diff --git a/components/item-act.js b/components/item-act.js
index 29babfd4..f4e360e1 100644
--- a/components/item-act.js
+++ b/components/item-act.js
@@ -277,7 +277,7 @@ export function useZap () {

       let hash, hmac;
       [{ hash, hmac }, cancel] = await payment.request(satsDelta)
-      await zap({ variables: { ...variables, hash, hmac } })
+      await zap({ variables: { ...variables, hash, hmac }, optimisticResponse, update })
     } catch (error) {
       revert?.()

diff --git a/lib/apollo.js b/lib/apollo.js
index d950ed68..17296bc1 100644
--- a/lib/apollo.js
+++ b/lib/apollo.js
@@ -259,7 +259,7 @@ function getClient (uri) {
 export function optimisticUpdate ({ mutation, variables, optimisticResponse, update }) {
   const { cache, queryManager } = getApolloClient()

-  const mutationId = String(queryManager.mutationIdCounter++)
+  const mutationId = String(queryManager.mutationIdCounter)
   queryManager.markMutationOptimistic(optimisticResponse, {
     mutationId,
     document: mutation,
```

But only removing `++` does not.

huumn:
All mutations share the queryManager. So not incrementing it means all mutations would share that mutationId.

[...]

PR #1220, opened 2024-06-04
Avoid manual optimistic updates for now

ekzyis:
## Description

Related to #1218 

Manual optimistic updates break cache updates for comments (probably updates for all following mutations). This PR comments out the code that breaks them for now.

TODO:
- [x] ~also comment out~ delete manual optimistic update code for bounty and poll votes

## Checklist

Yes

Yes, tested zaps and custom zap with commenting afterwards, bounty payments, polls


ekzyis:
I kept the broken optimistic code for zaps as comments so I can easier study it. But it's probably better to completely focus on #1195 instead of trying to fix code that #1195 will make obsolete anyway.

PR #1223, opened 2024-06-05
Handle Peertube Embeds

### LINKED ISSUE #107: Peertube embed like what was added for Youtube
Peertube is a decentralised video hosting network, based on free/libre software.   The bitcoin ecosystem embraces free, open source software (FOSS) and thus numerous web shows and podcasts either originate as a Peertube video, or are cross-distributed (syndicated ?) so that the show or podcast is available as a Peertube video.

This request is to add support for a Post for a link that is a Peertube video to include the iframe embed so that the video can be played within the SN post, similarly to how there is now Youtube embed, and Twitter embed support in SN (see Issue #9).

tsmith123:
Fixes #107 (with a caveat, see notes below)

## Description
This PR allows users to embed a video hosted on PeerTube.

## Screenshots

image

image

## Testing

The following urls were used to test in New Post and Comment.

https://peertube.satoshishop.de/w/d2ecb738-bee7-48db-8f7f-0009382cb24d
https://peertube.zoz-serv.org/w/aFWFxWPKYxmLFteHNJeBKG?start=0s&subtitle=en

**Note the different domains in the urls**

## Notes
As PeerTube allows videos to be hosted on different domains we can't specify a frame-src domain in the csp frame-src settings. The only way to permit video embeds from PeerTube is to use the allow all (*) value instead. **I will leave it you to decide whether this is too big a security risk**.


middleware.js
```diff
@@ -42,7 +42,9 @@ export function middleware (request) {
     // unsafe-inline for styles is not ideal but okay if script-src is using nonces
     "style-src 'self' a.stacker.news 'unsafe-inline'",
     "manifest-src 'self'",
-    'frame-src www.youtube.com platform.twitter.com rumble.com',
+    // Using wildcard (*) is the only way to get peertube videos working
+    // as the domain for each is different
+    'frame-src *',
```
tsmith123, line 47: If this is deemed too much of a security risk then we will probably need to cancel this PR (unless there's a way to dynamically create the frame-src value. I did try `peertube.*` and `peertube.*be` etc but these don't seem to be valid.

components/item-full.js
```diff
@@ -114,6 +114,20 @@ function ItemEmbed ({ item }) {
...
   }
 
+  if (provider === 'peertube') {
+    return (
+      <div className={styles.videoWrapper}>
+        <div className={styles.videoContainer}>
+          <iframe
+            title='PeerTube Video'
+            allowFullScreen=''
+            src={meta?.href}
```
ekzyis, line 124:
```suggestion
            src={meta?.href}
            sandbox='allow-scripts'
```

See comment regarding `frame-src`.

middleware.js
ekzyis: Makes sense. [I just learned that one can sandbox iframes](https://web.dev/articles/sandboxed-iframes). For PeerTube embeds, only `allow-scripts` seems to be required (see suggestion). With these restrictions, I think it's okay to use a wildcard on `frame-src`.

I also think `frame-src` isn't actually that important since we don't allow arbitrary iframes anyway, we only use iframes for the origins that our code parses which we therefore already inherently trust. So it might have been just a nice layer of defense. Additionally, according to my research, iframes are bad if _other_ sites frame your content in a malicious way which we don't allow via the `X-Frame-Options: DENY` header.

components/item-full.js
```diff
@@ -114,6 +114,20 @@ function ItemEmbed ({ item }) {
     )
   }
 
+  if (provider === 'peertube') {
+    return (
+      <div className={styles.videoWrapper}>
+        <div className={styles.videoContainer}>
+          <iframe
+            title='PeerTube Video'
+            allowFullScreen=''
```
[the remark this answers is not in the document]
tsmith123: Yeah, it was a copy and paste. I'll update both to allow full screen.

components/item-full.js
tsmith123: Nice 👍

lib/url.js
```diff
@@ -110,6 +110,16 @@ export function parseEmbedUrl (href) {
         }
       }
     }
+
+    if (hostname.includes('peertube.') && href.includes('/w/')) {
```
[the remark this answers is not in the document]
tsmith123: Yeah this was an awkward one. Hopefully my changes will match the majority of them.

middleware.js
tsmith123: Sounds like a "go" then 😄

ekzyis (approved):

tsmith123:
Can this be merged?

ekzyis (changes requested):
> Can this be merged?

Sorry to leave you hanging. @huumn and me discussed this PR yesterday and the conclusion was that we will not merge this as-is since it increases the attack surface significantly for something which isn't used much (not many share PeerTube links).

During my review, I missed that it's easy to create a domain which would trigger the peertube embed code and to therefore execute arbitrary javascript code against the page on which this potentially malicious embed is included.[^1]

So we'd rather see a whitelist approach for commonly linked PeerTube sites like [bitcointv.com](https://bitcointv.com/).[^2]

However, this PR is still eligible for awards since you essentially helped us to refine #107.

[^1]: I am not sure if this is as severe as an XSS since it runs the code inside an iframe but it might be.
[^2]: I don't know which other sites should be added to this whitelist. I only know of bitcointv.com because it's mentioned in [this SN item](https://stacker.news/items/11327) which was mentioned in [this Github comment](https://github.com/stackernews/stacker.news/issues/107#issuecomment-1037799988)

huumn:
Yes I think whitelisting trusted peertube sources will be our preferred approach to doing this, absent some other method for trusting what is being embedded.

tsmith123:
Cool, so I've restored the original frame-src values and added `bitcointv.com` and `peertube.tv` to the list.

tsmith123:
Think this needs another review guys 👍

tsmith123:
@huumn @ekzyis sorry to ask again guys but is there anything else I can do with this PR to get it over the line? It's been here for a while now.

huumn:
@tsmith123 it's us, not you. I've been super heads down on https://github.com/stackernews/stacker.news/pull/1195 ... which is taking every bit of working memory I have.

I'll try to give this a review today or tomorrow if @ekzyis doesn't get to it. @kravhen I'll try to get you some feedback too.

lib/url.js
```diff
@@ -120,6 +120,16 @@ export function parseEmbedUrl (href) {
         }
       }
     }
+
+    if (hostname.includes('peertube.tv') || hostname.includes('bitcointv.com')) {
```
ekzyis, line 124: `hostname.includes` is still prone to matching malicious domains since any domain could contain a subdomain that matches this.

You needs to use `endsWith` or an exact match.

ekzyis (changes requested):
Sorry for the slow review, time flies. Found an issue with the domain matching code.

ekzyis (approved):

PR #1229, opened 2024-06-06
wip: Optimistic updates via pending sats in item context

ekzyis:
## Description

Since we removed optimistic updates in #1220 because the implementation broke replies, there is no visual feedback if you zap from a non-custodial wallet. This is bad UX since it looks like zaps don't work since nothing happens. However, the payment is just pending.

This PR fixes this by implementing optimistic updates via a new context `ItemContext` (maybe not the best name). This context keeps track of pending sats. This is therefore completely separate of the Apollo cache and thus does not interfere with any mutation. All we need to do is to add pending sats to the item context, remove them after the mutation (if it failed or not) and hook into the context in the components that need to be aware of pending sats.

## Checklist


ekzyis:
> As I review, and I'm just curious because it would've been my first approach, did you try using `read/writeFragment` for this?

Mhh, I think it was one of my first approaches when I started implementing optimistic UX (so not in this PR but like weeks ago) but I must have had some problems with (undoing) it

huumn:
This works well. The only thing I'm noticing in QA is that for a brief moment, it appears as if I zapped double.

e.g. If I zap 100 sats, it'll briefly show 200 sats, then 100.

I'll see if it's easy to fix, but it isn't a huge deal. I still need to QA attached wallets too.

huumn:
Ah, I think this happens because `useZap` both does your optimistic context AND the `optimisticResponse`. I don't think it needs both anymore.

ekzyis:
> Ah, I think this happens because `useZap` both does your optimistic context AND the `optimisticResponse`. I don't think it needs both anymore. In prod, where there's more latency, this would probably be more noticeable and pretty unsettling (and it affects not just attached wallet zaps but all of them).

Ah yes, the optimistic context is undone in the `finally` branch. We could remove the optimistic response and only run `revert` during errors.

huumn:
I committed my changes to remove the doubling. Please QA it again yourself. Then I can deploy tomorrow if that works for you.

ekzyis:
lgtm

PR #1233, opened 2024-06-12
Donate sats forwarded to anon

### LINKED ISSUE #1232: Sats forwarded to anon don't go to rewards
**Description**

If sats are forwarded to [@anon](https://stacker.news/anon), the sats don't go to rewards like anon zaps but increase the balance of anon.

**Steps to Reproduce**

1. Create item and forward 100% sats to anon
2. Zap item with 100 sats item (can be anon)
3. Check /rewards
4. Rewards increased by 5 sats via fees and rewards via anon's stack didn't change at all[^1]
5. Running `SELECT id, msats, "stackedMsats" FROM users WHERE id = 27` reveals that 90 sats went to anon's balance

[^1]: 5 sats go to territory founder as revenue since all subs in seed.sql have "rewardsPct" set to 50

**Expected behavior**

Rewards increased by 5 sats via fees and by 90 sats via anon's stack

**Additional context**

Reported by [@0xbitcoiner](https://stacker.news/0xbitcoiner) [here](https://stacker.news/items/570337?commentId=570345).

ekzyis:
## Description

Fix #1232 by donating any sats forwarded to anon.

Diff of `item_act` function:

```diff
@@ -61,6 +61,11 @@ BEGIN
             UPDATE users
             SET msats = msats + fwd_msats, "stackedMsats" = "stackedMsats" + fwd_msats
             WHERE id = fwd_entry."userId";
+
+            IF fwd_entry."userId" = 27 THEN
+                -- anon forwards go to rewards as donations
+                PERFORM donate((fwd_msats / 1000)::INTEGER, "fwd_entry"."userId");
+            END IF;
         END LOOP;

         -- Give OP any remaining msats after forwards have been applied
```

## Additional Context

I initially wanted to include anon forwards in anon's stack in rewards. However, after looking in the code and considering the required changes, I think it's better to simply show them as donations.

I considered following approaches:

**1. no changes to rewards query**

In `item_act`, pretend that anon was zapped instead of adding forwarded sats to anon's balance. This would mean adding a row to `ItemAct`. But this would then mean that for anon forwards, three `ItemAct` rows are inserted (2x `TIP` + `FEE`) with their msats sum being higher than the zap itself. This might break some assumptions in queries but decreasing the amount of msats for the original `TIP` might as well.

So this approach seems to complicated for this requirement.

**2. changes to rewards query**

We could add a column  `forwarded` to the `Donations` table. Donations as part of anon forwards are marked with forwarded whereas normal donations are not. We then query the rows with forwarded set separately as part of anon's stack in the rewards query. This is a lot simpler but adds a column to `Donations` for something very specific.

## Checklist

Yes

tACK 3ca6fe1e

- Tested zaps and custom zaps from custodial balance, from attached wallet and as anon
- Tested replying as stacker and anon, editing as stacker
- Tested posting as stacker and anon

No

### DIFF
--- a/prisma/migrations/20240612085341_fix_anon_fwds/migration.sql
+++ b/prisma/migrations/20240612085341_fix_anon_fwds/migration.sql
@@ -1,0 +1,93 @@
+-- fix anon forwards: don't add sats forwarded to anon to anon's balance
+CREATE OR REPLACE FUNCTION item_act(item_id INTEGER, user_id INTEGER, act "ItemActType", act_sats INTEGER)
+RETURNS INTEGER
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    user_msats BIGINT;
+    act_msats BIGINT;
+    fee_msats BIGINT;
+    item_act_id INTEGER;
+    fwd_entry record; -- for loop iterator variable to iterate across forward recipients
+    fwd_msats BIGINT; -- for loop variable calculating how many msats to give each forward recipient
+    total_fwd_msats BIGINT := 0; -- accumulator to see how many msats have been forwarded for the act
+BEGIN
+    PERFORM ASSERT_SERIALIZED();
+
+    IF act_sats <= 0 THEN
+        RETURN 0;
+    END IF;
+
+    act_msats := act_sats * 1000;
+    SELECT msats INTO user_msats FROM users WHERE id = user_id;
+    IF act_msats > user_msats THEN
+        RAISE EXCEPTION 'SN_INSUFFICIENT_FUNDS';
+    END IF;
+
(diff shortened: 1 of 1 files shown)


ekzyis:
Only 5016 sats were forwarded to anon:

```sql
SELECT SUM(msats * pct / 100) FROM "ItemForward"
JOIN "Item" ON "Item"."id" = "ItemForward"."itemId"
WHERE "ItemForward"."userId" = 27;

-- returns 5,015,980 as of today
```

Since anon zaps were included in rewards, rewards only missed these 5k sats.

ekzyis:
Closing because this conflicts with #1195 so we will wait for #1195

PR #1235, opened 2024-06-12
Fix anon payment verification

ekzyis:
tACK 63280987

* Tested zaps and custom zaps from custodial balance, from attached wallet and as anon
* Tested replying as stacker and anon, editing as stacker
* Tested posting as stacker and anon


ekzyis:
Oh, I missed some `serialize` calls

PR #1243, opened 2024-06-20
Wallet definitions with uniform interface

ekzyis:
## Description

This replaces all attached wallets with definitions that use a uniform interface.

The wallet definitions export following properties:

- `name`: the wallet name. will be used as a wallet identifier across the code (used with `useWallet` etc.)
- `fields`: this defines which fields are required to configure this wallet
- `card`: properties for use with `WalletCard`
- `validate`: will be called to validate configuration before saving it
- `schema`: schema for formik validation
- `sendPayment`: for wallets that support sending
- `server`: for wallets that support receiving and are saved on the server
  - `walletType`: the wallet type as defined in the db schema (must match a value in enum `WalletType`)
  - `walletField`: the wallet field name as defined in the db schema (must match a wallet column of the `Wallet` table)
  - `resolverField`: name of GraphQL mutation. Mutation and resolver will be generated using this field. Example: `upsertWalletLND`
  - `testConnect`: function that will be called with config and context during save to create a test invoice
  - `createInvoice`: function that will be called with config and context during autowithdrawal. Must return bolt11.

**Other changes**

- replaced confusing UX around "default payment/autowithdrawal method" with a checkbox to enable/disable a wallet + drag and drop for priority
- replaced occurrences of `WebLN` with `Wallet`: `useWebLN` -> `useWallet`
- reversed wallet logs order and removed follow functionality

## TODO

- [x] sending wallets (client)
  - [x] lnbits
  - [x] nwc
  - [x] lnc
    - [x] ~fix lnc + allow "locked" as a status~ _(I couldn't set a password for LNC. It fails at [this line](https://github.com/lightninglabs/lnc-web/blob/f65cf92b50b3d580072d76a40eff763a97d7a67b/lib/util/credentialStore.ts#L81) even though I am trying to **set** a password not unlocking/decrypting something. Mnemonics also seem to work only once.)_
- [x] receiving wallets (server)
  - [x] ln address
  - [x] lnd
  - [x] cln
- [x] fix import error
- [x] use drag and drop to update wallet priorities
- [x] move all/most/some (?) wallet-specific backend code into wallet definition
- [x] test if autowithdrawal order is consistent with card order
- [ ] [optional] replace `nostr-tools` with own code
- [x] break config, payment request, and context into independent parameters for `sendPayment`
- [x] fix separate errors per invalid bip39 word
- [x] make dnd work on mobile _(using touches instead of dnd on mobile now, see 2051dd0e)_
- [x] fix `draggable` not set on first page load
- [x] add example wallet definition as documentation
- [x] remove `React` dependency in wallet defs, render markdown instead
- [x] remove `Yup` dependency in wallet defs, generate schema from fields instead
- [x] fix LND and test other wallets since bbcfc2fa
- [x] ~replace `testConnect` with `validate` so we don't have two functions for validation~
- [x] ~add `formikValidate` and rename `schema` to `yupSchema`, `validate` to `validateOnClient`, `testConnect` to `validateOnServer`~ _we now export `fieldValidation`, `testConnectClient` and `testConnectServer`, see [this comment](https://github.com/stackernews/stacker.news/pull/1243#discussion_r1684506466)_
- [x] write new updated documentation

## Checklist

Yes, only added boolean column `enabled` to `Wallet` table

Yes, tested every wallet functionality and order in which autowithdrawals are attempted. Same for picking which enabled wallet to use for payments.

As mentioned above, I was not able to test LNC passwords. Configuring a password always failed with `The password provided is not valid`.

Tested touch code as a replacement for DnD to change priority of wallets via wallet cards

No


components/logger.js
```diff
@@ -4,6 +4,8 @@ import fancyNames from '@/lib/fancy-names.json'
 import { gql, useMutation, useQuery } from '@apollo/client'
 import { WALLET_LOGS } from '@/fragments/wallet'
 import { getWalletBy } from '@/lib/constants'
+// TODO: why can't I import this without errors?
+// import { getWalletByName } from './wallet'
```
ekzyis, line 8: This import breaks the build with this error:

```
Module build failed: UnhandledSchemeError: Reading from "node:crypto" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
 ⨯ node:crypto
```

This is somehow related to the import of `@/components/wallets/lnbits` in `@/components/wallet`. If I remove the `await import`, it works:

```diff
diff --git a/components/wallet/index.js b/components/wallet/index.js
index fbddb07f..76ac8e18 100644
--- a/components/wallet/index.js
+++ b/components/wallet/index.js
@@ -6,7 +6,7 @@ import { SSR } from '@/lib/constants'

 // wallet definitions
 export const WALLET_DEFS = [
-  await import('@/components/wallet/lnbits')
+  // await import('@/components/wallet/lnbits')
 ]

 export const Status = {
```

But it also doesn't work if I use a "normal import":

```diff
diff --git a/components/wallet/index.js b/components/wallet/index.js
index fbddb07f..9fcb6e21 100644
--- a/components/wallet/index.js
+++ b/components/wallet/index.js
@@ -3,10 +3,11 @@ import { useMe } from '@/components/me'
 import useLocalState from '@/components/use-local-state'
 import { useWalletLogger } from '@/components/logger'
 import { SSR } from '@/lib/constants'
+import lnbits from '@/components/wallet/lnbits'

 // wallet definitions
 export const WALLET_DEFS = [
-  await import('@/components/wallet/lnbits')
+  lnbits
 ]

 export const Status = {
```

fails with same error.

So I realized that `import { bolt11Tags } from '@/lib/bolt11'` in components/wallets/lnbits.js import is causing this. When I remove it, it works.

I can reproduce the error if I import `bolt11` in components/logger.js.

But why? Why does `bolt11` break the build if imported in components/logger.js?

ekzyis: I don't know what problem JS had with logger.js, but moving the code that required `getWalletByName` into wallet-logger.js fixed it. See 4a4d6698.

huumn:
For this:

>  [ ] receiving wallets (server)

I already began doing this in #1178 in [api/createInvoice](https://github.com/stackernews/stacker.news/tree/11b0a7382c81989b1965cdce807082a2b4b482f7/api/createInvoice)

components/wallet/lnbits.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    const preimage = checkResponse.preimage
+    logger.ok('payment successful:', `payment_hash=${hash}`, `preimage=${preimage}`)
+    return { preimage }
+  } catch (err) {
+    logger.error('payment failed:', `payment_hash=${hash}`, err.message || err.toString?.())
+    throw err
+  }
+}
```
huumn, line 65: I know this is super WIP, but just in case, one of the benefits of unifying the interface should be that all the logging and generic error handling can go in `wallet/index.js`

ekzyis: Ah yes, now only `validate` logs messages since I think they should be custom per wallet and not generic. Every other logging is handled in `wallet/index.js`.

components/wallet-logger.js
```diff
@@ -0,0 +1,321 @@
...
+      const isBottom = dY >= -1 && dY <= 1
+      if (isBottom) {
+        setFollow(true)
+      }
+
+      scrollY.current = y
+    }
+    tableRef.current?.addEventListener('scroll', onScroll)
+    return () => tableRef.current?.removeEventListener('scroll', onScroll)
+  }, [])
```
huumn, line 63: I know it's not the subject of this PR, but I've been wondering if we might avoid the scroll by sorting in descending order - that is, most recent entries first. Our `recent` sort is works this way as do most UI's on the internet with the exception of chat apps.

ekzyis: Mhh, I remember considering this. I think I thought there would be an expectation for logs to work like printing line by line and thus scroll downwards. But I think we can indeed avoid this follow functionality by simply reversing the order.

ekzyis: Done in 2ee983d0

components/wallet/nwc.js
```diff
@@ -0,0 +1,161 @@
...
+            const msg = 'EOSE received without info event'
+            logger.error(msg)
+            reject(new Error(msg))
+          }
+          sub?.close()
+        }
+      })
+    })
+  } finally {
+    // For some reason, websocket is already in CLOSING or CLOSED state.
```
huumn, line 95: probably because it's garbage collected once the `try` block is finished

components/wallet/lnc.js
```diff
@@ -0,0 +1,190 @@
...
+          </div>
+        </Form>
+      )
+    }
+    )
+  })
+}
+
+// FIXME: pass me, status, showModal in useWallet hook
+export async function sendPayment ({ bolt11, pairingPhrase, password: configuredPassword, me, status, showModal, logger }) {
```
huumn, line 120: consider breaking the config, payment request, and context into independent parameters. Maybe something like:

`sendPayment(bolt11: string, config: Object, context: Object)`

components/wallet/nwc.js
ekzyis: Mhh, that `sub?.close()` should only close the subscription. But this comment is about the relay connection itself. Just because a subscription was closed, the relay connection shouldn't close? Mhh, but maybe it's closing the relay because it's the last subscription that was closed? I need to check the library code. 

This is also a comment from when I implemented NWC.

components/wallet/lnc.js
```diff
@@ -0,0 +1,190 @@
...
+
+export const schema = object({
+  pairingPhrase: array()
+    .transform(function (value, originalValue) {
+      if (this.isType(value) && value !== null) {
+        return value
+      }
+      return originalValue ? originalValue.split(/[\s]+/) : []
+    })
+    .of(string().trim().oneOf(bip39Words, ({ value }) => `'${value}' is not a valid pairing phrase word`))
```
ekzyis, line 69: This creates a separate error for every invalid word which looks bad

2024-06-25-155030_725x113_scrot

components/wallet/nwc.js
huumn: I'd guess it doesn't keep the connection open without an active subscription.

components/payment.js
```diff
@@ -124,19 +124,19 @@ const useWebLnPayment = () => {
           .catch(reject)
       })
```
huumn, line 125: Heads up: `waitUntilPaid`'s interval leaks. It polls even after our attempt to pay it fails and will keep polling until it expires.

I'm messing around with how to fix it, but it's tricky.

huumn: I ended up using an `AbortController` in [a41c710](https://github.com/stackernews/stacker.news/pull/1195/commits/a41c710751c55647aadcc0da4dafe73c4c8a0d4a)

ekzyis: > It polls even after our attempt to pay it fails and will keep polling until it expires.

Mhh, this means the invoice wasn't properly cancelled. Aren't we always cancelling it on failure? Any payment error or closing the modal should automatically cancel the invoice.

But I see the value in calling `stopWaiting` in the `finally` block to make sure. I just think our code should also cancel any invoices that we know are not going to get paid.

api/resolvers/wallet.js
```diff
@@ -318,9 +318,10 @@ export default {
         where: {
           userId: me.id
         },
-        orderBy: {
-          createdAt: 'asc'
-        }
+        orderBy: [
+          { createdAt: 'desc' },
+          { id: 'desc' }
+        ]
```
ekzyis, line 324: Logs have same timestamp within the same tx so I use `id` as a tie breaker for the real log insertion order with new logs first.

components/wallet-buttonbar.js
ekzyis, line 1: had to put this into its own file to dynamically import it to fix hydration issues

worker/wallet.js
```diff
@@ -253,7 +253,7 @@ async function checkWithdrawal ({ data: { hash }, boss, models, lnd }) {
       if (dbWdrwl.wallet) {
         // this was an autowithdrawal
         const message = `autowithdrawal of ${numWithUnits(msatsToSats(paid), { abbreviate: false })} with ${numWithUnits(msatsToSats(fee), { abbreviate: false })} as fee`
-        await addWalletLog({ wallet: dbWdrwl.wallet.type, level: 'SUCCESS', message }, { models, me: { id: dbWdrwl.userId } })
+        await addWalletLog({ wallet: dbWdrwl.wallet, level: 'SUCCESS', message }, { models, me: { id: dbWdrwl.userId } })
```
ekzyis, line 256: Fixed missing autowithdraw success logs

pages/settings/wallets/index.js
```diff
@@ -1,29 +1,54 @@
...
-  const cln = wallets.find(w => w.type === W.CLN.type)
+    const [source] = newOrder.splice(sourceIndex, 1)
+    const newTargetIndex = sourceIndex < targetIndex ? targetIndex - 1 : targetIndex
+    const append = sourceIndex < targetIndex
+
+    newOrder.splice(newTargetIndex + (append ? 1 : 0), 0, source)
+
+    await Promise.all(
+      newOrder.map((w, i) =>
+        w.setPriority(i).catch(console.error)
```
ekzyis, line 48: This will trigger a mutation for every server wallet but only if the priority actually changed.

I forgot to add some error handling here. There is no feedback on errors so the order might get out of sync with the actual order in the server. Will add some.

pages/settings/wallets/index.js
```diff
@@ -1,29 +1,54 @@
...
+const WalletCard = dynamic(() => import('@/components/wallet-card'), { ssr: false })
+
+export const getServerSideProps = getGetServerSideProps({ authRequired: true })
 
 export default function Wallet ({ ssrData }) {
-  const { data } = useQuery(WALLETS)
+  const { wallets } = useWallets()
+
+  const [sourceIndex, setSourceIndex] = useState()
+  const [targetIndex, setTargetIndex] = useState()
```
ekzyis, line 17: I hope this drag and drop feature is still in line with our wallets plans but I think it's cool.

Might need to add some explanation what drag & drop does. It configures the priority per wallet so if a wallet fails, the next wallet will be tried.

It's only used during autowithdrawals yet though. The `useWallet` hook on the client does not support multiple enabled wallets yet. It will only try the first enabled.

components/wallet/lnc.js
ekzyis: Done in 9af8e633

pages/settings/wallets/index.js
ekzyis: Add error toasts in f947a450. It will print separate errors for each call which looks bad but I think errors here should be rare:

2024-07-08-131800_1920x1080_scrot

huumn (changes requested):
This is only a partial review. This is excellent at a high-level. I only gave NWC a spin, but it works well and the drag and drop is slick. Lots of code reuse. 

I still haven't fixed LNC or done a full line-by-line, so we'll need to do that before merging, but at least the NWC uncaught connection exceptions are also blocking merging. 

I also think we'll want to move `wallet` out of `components` into a top-level directory called `wallets`.

Otherwise, this can probably be merged, but we might want to spend a bit more time refining the interface design first. We can make that decision together in a call ultimately, but I wanted to share some of my early thoughts and ask a few questions before proposing refinements.

The sacrifices made to put clientside and serverside code in the same file appear to complicate things more than expected. The point of the single file suggestion was encapsulation and it appears we can't encapsulate the serverside portion for a number of reasons and many of them are outside of our control. I'm not sure it'll ever be fully encapsulated given we need a DB table for every backend wallet type, and a typed resolver for upserting them via the api, but also because they tend to have dependencies that can only run on the serverside.

Still, it might be worth pursuing a *mostly* one file design if we can decomplicate the import issues and dynamic schema stuff. As is this aspect is pretty difficult to maintain (for us) and reason about (for a wallet contributor). You've probably tried some of these alternatives but I wanted to suggest them to get an idea of what else we might try:

1. Did you try named imports rather than `import * as LND` and still experience clientside/worker import errors?
2. One alternative to a one file design is a one folder design. For each wallet we can have a directory containing a `client.js` and a `server.js` file, e.g. `wallets/lnd/client.js` and `wallets/lnd/server.js` and maybe a `wallets/lnd/index.js` for universal things.
3. Did you try dynamic imports for serverside only deps inside of `testConnect` and `createInvoice`? I imagine that would work and make the interface cleaner even if it isn't a very good idea.
4. Another thing we can try as a last resort to simplify the serverside interface is explicitly telling webpack using the `ignorePlugin` to ignore those imports on the clientside.

I think the one folder idea is probably cleanest, but idk.

-------------

Something small that'll be easy to refine later, if not immediately, is I think we should remove any assumption of understanding `Yup` or `React` or `graphql` in these wallet definition files. The ideal is probably not forcing them to import anything, allowing them to import some package for their specific wallet, and just let them write js to define a wallet.

- I'm pretty sure we can easily remove `React` by supporting markdown for `card` fields
- For removing `Yup`, I think allowing a `validate` function to every object in `fields` would allow the wallet definer to use regex or whatever to validate.

components/wallet/lnbits.js
```diff
@@ -0,0 +1,109 @@
...
+]
+
+export const card = {
+  title: 'LNbits',
+  subtitle: React.createElement(
+    React.Fragment,
+    {},
+    'use ',
+    React.createElement('a', { href: 'https://lnbits.com/', target: '_blank', rel: 'noreferrer' }, 'LNbits'),
+    ' for payments'),
```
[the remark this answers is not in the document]
ekzyis: Good idea!

components/wallet/lnd.js
```diff
@@ -0,0 +1,110 @@
...
+  subtitle: 'autowithdraw to your Lightning Labs node',
+  badges: ['receive only', 'non-custodial']
+}
+
+export const schema = LNDAutowithdrawSchema
+
+export const server = {
+  walletType: 'LND',
+  walletField: 'walletLND',
+  resolverName: 'upsertWalletLND',
```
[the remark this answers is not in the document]
ekzyis: Mhh, no, I think you are right. When I wrote this code, I wasn't sure if injecting GraphQL typedefs and resolver was already too much magic so I at least kept the names of what gets injected explicit, lol

components/wallet/nwc.js
```diff
@@ -0,0 +1,133 @@
...
+    // relay?.close()
+    if (relay) logger.info(`closed connection to ${relayUrl}`)
+  }
+}
+
+export async function sendPayment (bolt11, { nwcUrl }, { logger }) {
```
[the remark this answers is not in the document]
ekzyis: You're right, I didn't see that the relay connection is still open. I trusted my previous code (that was running in prod for a while) and the error message that said the connection is already closed when calling `relay.close()` but it indeed isn't:

2024-07-15-133057_451x53_scrot
*state after payment*

Another reason why we should write our own relay code.

> Locally, it throws an uncaught exception sometime after payment that hits the error barrier. I'm surprised you didn't run into that.

I got some `Error: sending on closed connection` in the past locally (not sure if while working on this PR though) but I always shrugged them off as something specific to the dev build. They also seemed to happen randomly.  :sweat_smile: I also couldn't reproduce the error here. How long did you wait for it after payment?

Anyway, I'm going to use `relay?.close()?.catch()` as you did in https://github.com/stackernews/stacker.news/commit/ca11ac9fb85547096d61f398443f7d0b649a5e1f.

ekzyis:
> I also think we'll want to move `wallet` out of components into a top-level directory called `wallets`.

Makes sense!

> 1. Did you try named imports rather than import * as LND and still experience clientside/worker import errors?

I did not! Do you mean this with named imports (not sure, looks weird):

<details>
<summary>named import patch</summary>

```diff
diff --git a/components/wallet/index.js b/components/wallet/index.js
index 3e747c78..64be19b1 100644
--- a/components/wallet/index.js
+++ b/components/wallet/index.js
@@ -8,7 +8,7 @@ import { bolt11Tags } from '@/lib/bolt11'
 import * as lnbits from '@/components/wallet/lnbits'
 import * as nwc from '@/components/wallet/nwc'
 import * as lnc from '@/components/wallet/lnc'
-import * as lnd from '@/components/wallet/lnd'
+import { name as lndName, fields as lndFields, card as lndCard, schema as lndSchema, server as lndServer } from '@/components/wallet/lnd'
 import * as lnAddr from '@/components/wallet/lightning-address'
 import * as cln from '@/components/wallet/cln'
 import { gql, useApolloClient, useQuery } from '@apollo/client'
@@ -18,7 +18,7 @@ import { useShowModal } from '@/components/modal'
 import { useToast } from '../toast'

 // wallet definitions
-export const WALLET_DEFS = [lnbits, nwc, lnc, lnd, lnAddr, cln]
+export const WALLET_DEFS = [lnbits, nwc, lnc, { name: lndName, fields: lndFields, card: lndCard, schema: lndSchema, server: lndServer }, lnAddr, cln]

 export const Status = {
   Initialized: 'Initialized',
diff --git a/components/wallet/lnd.js b/components/wallet/lnd.js
index 0e0e1731..a60522b8 100644
--- a/components/wallet/lnd.js
+++ b/components/wallet/lnd.js
@@ -4,6 +4,8 @@ import { ensureB64 } from '@/lib/format'
 import { datePivot } from '@/lib/time'
 import { LNDAutowithdrawSchema } from '@/lib/validate'

+import { authenticatedLndGrpc } from 'ln-service'
+
 export const name = 'lnd'

 export const fields = [
```

</details>

If I do this and then import a nodejs dependency like `import { authenticatedLndGrpc } from 'ln-service'` in the file that I import in the client the build still breaks with this error:

```
 ⨯ ./node_modules/@grpc/grpc-js/build/src/channel-credentials.js:20:0
Module not found: Can't resolve 'tls'

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./node_modules/@grpc/grpc-js/build/src/index.js
./node_modules/lightning/lnd_grpc/unauthenticated_lnd_grpc.js
./node_modules/lightning/lnd_grpc/index.js
./node_modules/lightning/index.js
./node_modules/ln-service/index.js
./components/wallet/lnd.js
./components/wallet/index.js
./components/wallet-logger.js
./pages/_app.js
```

I also tried this and same error:

<details>
<summary>default import patch</summary>

```diff
diff --git a/components/wallet/index.js b/components/wallet/index.js
index 3e747c78..bdabd9f3 100644
--- a/components/wallet/index.js
+++ b/components/wallet/index.js
@@ -8,7 +8,7 @@ import { bolt11Tags } from '@/lib/bolt11'
 import * as lnbits from '@/components/wallet/lnbits'
 import * as nwc from '@/components/wallet/nwc'
 import * as lnc from '@/components/wallet/lnc'
-import * as lnd from '@/components/wallet/lnd'
+import lnd from '@/components/wallet/lnd'
 import * as lnAddr from '@/components/wallet/lightning-address'
 import * as cln from '@/components/wallet/cln'
 import { gql, useApolloClient, useQuery } from '@apollo/client'
diff --git a/components/wallet/lnd.js b/components/wallet/lnd.js
index 0e0e1731..4e10e6fc 100644
--- a/components/wallet/lnd.js
+++ b/components/wallet/lnd.js
@@ -4,6 +4,8 @@ import { ensureB64 } from '@/lib/format'
 import { datePivot } from '@/lib/time'
 import { LNDAutowithdrawSchema } from '@/lib/validate'

+import { authenticatedLndGrpc } from 'ln-service'
+
 export const name = 'lnd'

 export const fields = [
@@ -108,3 +110,11 @@ export const server = {
     return invoice.request
   }
 }
+
+export default {
+  name,
+  fields,
+  card,
+  schema,
+  server
+}
```

</details>

> 3. Did you try dynamic imports for serverside only deps inside of testConnect and createInvoice? I imagine that would work and make the interface cleaner even if it isn't a very good idea.

Also no and I also thought it should work but it doesn't for some reason:

<details>
<summary>dynamic import patch</summary>

```diff
diff --git a/components/wallet/lnd.js b/components/wallet/lnd.js
index 0e0e1731..3a1b7bdc 100644
--- a/components/wallet/lnd.js
+++ b/components/wallet/lnd.js
@@ -58,8 +58,9 @@ export const server = {
   resolverName: 'upsertWalletLND',
   testConnect: async (
     { cert, macaroon, socket },
-    { me, models, addWalletLog, lnService: { authenticatedLndGrpc, createInvoice } }
+    { me, models, addWalletLog }
   ) => {
+    const { authenticatedLndGrpc, createInvoice } = await import('ln-service')
     try {
       cert = ensureB64(cert)
       macaroon = ensureB64(macaroon)
```

</details>

throws same error as above.

> 4. Another thing we can try as a last resort to simplify the serverside interface is explicitly telling webpack using the ignorePlugin to ignore those imports on the clientside.

Mhh, would be interesting to look into because I want to be less of a webpack noob but wouldn't that mean we're simply moving the leaky abstraction from "dependency injection" to "webpack configuration"?

I imagine this to work by telling webpack to ignore `ln-service`, `lib/cln` etc. (all the imports that break the client) on the client so if a contributor wants to use a new dependency that only works on the server, they would need to specify this dependency in webpack now instead of where they currently get injected.

So to me it sounds like we're not gaining anything if we're solving this issue in another way that also isn't encapsulated.

> I think the one folder idea is probably cleanest, but idk.

Yes, sounds like the simplest solution for everyone and can't get much simpler than that. I'll look into this.

> I'm pretty sure we can easily remove `React` by supporting markdown for `card` fields

Mhh yes, good idea!

> For removing `Yup`, I think allowing a `validate` function to every object in `fields` would allow the wallet definer to use regex or whatever to validate.

Mhh, so basically constructing the Yup schema from the fields like I am constructing the mutation or the input props, right? Makes sense

components/wallet/lightning-address.js
```diff
@@ -0,0 +1,52 @@
...
+
+export const name = 'lightning-address'
+export const shortName = 'lnAddr'
+
+export const fields = [
+  {
+    name: 'address',
+    label: 'lightning address',
+    type: 'text',
+    hint: 'tor or clearnet',
```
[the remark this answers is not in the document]
ekzyis: Yes, will remove this!

But I think in theory and in the future, we could support Tor for LN addresses using the same proxy agents we use for CLN.

huumn:
> If I do this and then import a nodejs dependency like import { authenticatedLndGrpc } from 'ln-service' in the file that I import in the client the build still breaks with this error

Is this true when you don't do `import * as lnd` in `wallet/index.js`? Intuition would suggest tree shaking would remove any packages never referenced in clientside code, but I doubt it works if we are implicitly importing things like `module.server` on the client.

Regardless, folders might make better sense anyway.

huumn:
> `import { name as lndName, fields as lndFields, card as lndCard, schema as lndSchema, server as lndServer } from '@/components/wallet/lnd'`

How about:

`import { name, fields, card, schema, server } as lnd from '@/components/wallet/lnd'`

components/wallet/lnd.js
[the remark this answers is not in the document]
ekzyis: Mhh, I actually think `walletType` and `walletField` are good reminders that database changes are required for receiving wallets. `walletType` must match a `WalletType` enum value and `walletField`  must match a column in the `Wallet` table (as documented in _example.js). I could assume that they will always be called in this way but I think it will lead to "oh, you can't name your column / enum value that way". So it's better to be explicit here imo.

The value for `resolverName` however can be set to anything. So I will add code that always uses `upsertWallet${module.name}` (removing any characters that might be illegal in this context).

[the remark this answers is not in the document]
ekzyis: I decided to use `walletField` in f01ce79a since it's semantically closer to what the resolver is about

ekzyis:
I did the split in 5b2e835 now. We can now import stuff like `ln-service` in wallets/[wallet]/server.js.

The app still builds but for some reason, this line doesn't doesn't work as expected while building the worker:

```
import walletDefs from 'wallets/server'
```

It doesn't break on import but when trying to iterate over the default export in `injectResolvers`. It then throws `TypeError: import_server.default is not iterable`.

When I print `walletDefs`, it just prints `undefined`.

But when I fixed this, I think this is indeed a better interface design. I like this more than dependency injection!

components/payment.js
```diff
@@ -131,7 +131,7 @@ export const useWalletPayment = () => {
   const wallet = useWallet()
 
   const waitForWalletPayment = useCallback(async ({ id, bolt11 }, waitFor) => {
-    if (!wallet) {
+    if (!wallet?.enabled) {
```
huumn, line 134: If we didn't have a wallet (e.g. anon), this function would throw

pages/settings/wallets/[wallet].js
```diff
@@ -44,7 +44,7 @@ export default function WalletSettings () {
       <Form
         initial={initial}
         schema={schema}
-        onSubmit={async (values) => {
+        onSubmit={async ({ amount, ...values }) => {
```
Comment (huumn) [inline pages/settings/wallets/[wallet].js:47]: I'm not sure if we need `amount` anymore but this was getting saved in the configs for wallets

pages/settings/wallets/[wallet].js
```diff
@@ -99,37 +99,42 @@ export default function WalletSettings () {
...
+        autoFocus: i === 0
+      }
+      if (type === 'text') {
+        return <ClientInput key={i} {...rawProps} />
+      }
+      if (type === 'password') {
+        return <PasswordInput key={i} {...rawProps} newPass />
+      }
+      return null
+    })
```
Comment (huumn) [inline pages/settings/wallets/[wallet].js:139]: Most of what I do here is add a `readOnly` prop and if `props.hidden` don't show the field.

For lnc, it needs to store other fields that don't concern the user, so this lets us keep those without butchering the wallet code or keeping 2 copies of the config.

wallets/index.js
```diff
@@ -73,8 +73,8 @@ export function useWallet (name) {
       // validate should log custom INFO and OK message
       // validate is optional since validation might happen during save on server
       // TODO: add timeout
-      await wallet.validate?.(newConfig, { me, logger })
-      await saveConfig(newConfig)
+      const validConfig = await wallet.validate?.(newConfig, { me, logger })
+      await saveConfig(validConfig ?? newConfig)
```
huumn, line 77: This is the most invasive change. It allows `validate` to return any modifications to the config before storing. `lnc` needs to store diffie helman keys, so when validate generates them, we return them in the config.

wallets/lnc/client.js
```diff
@@ -141,3 +99,63 @@ function validateNarrowPerms (lnc) {
   // TODO: need to check for more narrow permissions
   // blocked by https://github.com/lightninglabs/lnc-web/issues/112
 }
+
+// default credential store can go fuck itself
+class LncCredentialStore {
```
huumn, line 104: The default credential store had so many side effects it was incredibly difficult to troubleshoot. I started forking `lnc-web` (which we'll probably want to do eventually anyway), but simply removing the side effects was most of what we needed to make this work.

This also remove the password. As we discussed, we'll want to do one password to rule them all eventually, so there's no reason to try to support the buggy library one in the interim.

wallets/nwc/client.js
```diff
@@ -18,7 +18,7 @@ export async function validate ({ nwcUrl }, { logger }) {
   logger.ok(`connected to ${relayUrl}`)
 
   try {
-    return await new Promise((resolve, reject) => {
+    await new Promise((resolve, reject) => {
```
huumn, line 21: did this because  `validate` now expects `undefined` or a new config returned. This didn't need to be returned anyway.

lib/wallet.js
```diff
@@ -0,0 +1,106 @@
...
+import { autowithdrawSchemaMembers, hexOrBase64Validator, lightningAddressValidator } from '@/lib/validate'
+import { TOR_REGEXP } from '@/lib/url'
+import { B64_URL_REGEX } from '@/lib/format'
+
+export function generateResolverName (walletField) {
+  const capitalized = walletField[0].toUpperCase() + walletField.slice(1)
+  return `upsertWallet${capitalized}`
+}
+
+export function generateSchema (wallet) {
```
huumn, line 11: It looks like doing this will require us creating our own language *on top* of Yup. This won't block deployment, but I don't think we want to do this because we'll have to document it and Yup/Formik is always going to have better docs.

Formik supports [two methods of validation](https://formik.org/docs/guides/validation). So I think the move is going to be letting wallet implementers write one or the other:
1. some function that we validate with (where they can do regex or whatever) and throw errors in Formik's expected shape
2. provide their own Yup schema and we can just link yup's docs in our own docs

huumn (approved):
I still need to QA receiving wallets (LNC ate so much of my time ...), but I should be able to do that tomorrow and if that looks good we can merge.

lib/wallet.js
ekzyis: Ahh, that's what you meant. I should have asked for clarification since I also wondered if this is still too depended on Yup. Didn't think about native Formik form-level validation (or field-level). Sorry for the miscommunication!

I'll revert my changes about `generateSchema` and do the following: wallet definitions can contain either `formikValidate` or `yupSchema` for validation before submit, `validateOnSubmit` which is the function that we call ourselves during submit on the client via `wallet?.validate` and `validateOnServer` which runs on the server in `upsertWallet` and thus can actually create test invoices. This is therefore also related to this TODO:

> - [ ] replace `testConnect` with `validate` so we don't have two functions for validation

So I will not replace `testConnect` with `validate` but rename them: `testConnect` to `validateOnServer` and `validate` to `validateOnSubmit`. You mentioned to only have one `validate` since both do basically the same but I think it's better to keep them separate since when we have wallets that do both, we might want to be able to run both. Also, the new names make it clear where they run whereas with only one function, it would depend on the wallet context: is it a wallet stored on the client or server?[^1] But as mentioned, we will lose this context when wallets are stored on both.[^2]

[^1]: assuming I didn't get you wrong again
[^2]: until we store encrypted spending permissions on the server but even then, we might want to run some validation during submit before the config hits the server which we don't want to run for each field change via `formikValidate` or `yupSchema`.

pages/settings/wallets/[wallet].js
Comment (ekzyis) [reply pages/settings/wallets/[wallet].js:47]: Ah no, thanks for fixing it. I also noticed that but always when I was already doing something else and then I forgot.

components/payment.js
ekzyis: Thanks for the fix but I think [1b6de0b](https://github.com/stackernews/stacker.news/pull/1243/commits/1b6de0bb96a0909280eaec926f6473fc646afbb1) is a better fix.

My intention with `useWallet()` (with no name as argument) was supposed to always return an enabled wallet if one exists else null. So we shouldn't need to check if it's enabled if we use `useWallet()`.

lib/wallet.js
huumn: `validateOnClient` might make it more consistent

huumn: Also, it might be worth using `testConnectClient` and `testConnectServer` instead because we have both connection validation AND field validation.

> formikValidate or yupSchema for validation before submit

It might be worth giving it one name `fieldValidation` and then checking for the type returned, e.g. `typeof fieldValidation === 'function'` or `typeof fieldValidation === 'YupSchema'` or something like that.

ekzyis: Mhh, as implemented in 86a630f2, we support both at the same time so it's actually not an XOR even though in practice, it's probably not useful to validate a wallet using both. I thought `Formik` might not accept both params (`validate` and `validationSchema`) at the same time but it does.

I tested the code using this patch:

```diff
diff --git a/wallets/lightning-address/index.js b/wallets/lightning-address/index.js
index 51dd665f..9ce4fc91 100644
--- a/wallets/lightning-address/index.js
+++ b/wallets/lightning-address/index.js
@@ -18,6 +18,10 @@ export const card = {
   badges: ['receive only', 'non-custodialish']
 }

+export const formikValidate = ({ address }) => {
+  return { address: 'test error' }
+}
+
 export const yupSchema = lnAddrAutowithdrawSchema

 export const walletType = 'LIGHTNING_ADDRESS'
```

and to test the validation call on the server in `upsertWallet`, I commented out `Formik.validate`:

```diff
diff --git a/pages/settings/wallets/[wallet].js b/pages/settings/wallets/[wallet].js
index c9fd8f91..442a74ff 100644
--- a/pages/settings/wallets/[wallet].js
+++ b/pages/settings/wallets/[wallet].js
@@ -40,7 +40,7 @@ export default function WalletSettings () {
       {!wallet.walletType && <WalletSecurityBanner />}
       <Form
         initial={initial}
-        validate={wallet.formikValidate}
+        // validate={wallet.formikValidate}
         schema={wallet.yupSchema}
         onSubmit={async ({ amount, ...values }) => {
           try {
```

But I'll look into using `typeof` so we can use same exported property for both for better naming.[^1]

[^1]: [`Formik` does support functions for `validationSchema`](https://formik.org/docs/api/formik#validationschema-schema----schema), too but I think we don't need to support that for wallet definitions. I don't think we need to support functions that return yup schemas there.

huumn:
Autowithdraw works great. Priority works too. This has come together really nicely.

 I'll do a line by line of the serverside after SNL then merge.

ekzyis:
> Autowithdraw works great. Priority works too. This has come together really nicely.
> 
> I'll do a line by line of the serverside after SNL then merge.

Okay, I just pushed the last changes regarding validation feedback in 52dded22 and b7541352.

I'll write up some markdown documentation for adding a new wallet but you can go ahead and merge.

huumn:
I have more to review serverside still. I'll finish up tomorrow.

ekzyis:
This is how the documentation in [1048c89](https://github.com/stackernews/stacker.news/pull/1243/commits/1048c89a622050fbbd8a55f37d1694fdb3201e39) renders on Github:

# Wallets

Every wallet that you can see at [/settings/wallets](https://stacker.news/settings/wallets) is implemented as a plugin in this directory.

This README explains how you can add another wallet for use with Stacker News.

> [!NOTE]
> Plugin means here that you only have to implement a common interface in this directory to add a wallet.

## Plugin interface

Every wallet is defined inside its own directory. Every directory must contain an _index.js_ and a _client.js_ file.

An index.js file exports properties that can be shared by the client and server.

Wallets that have spending permissions / can pay invoices export the payment interface in client.js. These permissions are stored on the client.[^1]

[^1]: unencrypted in local storage until we have implemented encrypted local storage.

A _server.js_ file is only required for wallets that support receiving by exposing the corresponding interface in that file. These wallets are stored on the server because payments are coordinated on the server so the server needs to generate these invoices for receiving. Additionally, permissions to receive a payment are not as sensitive as permissions to send a payment (highly sensitive!).

> [!NOTE]
> Every wallet must have a client.js file (even if it does not support paying invoices) because every wallet is imported on the client. This is not the case on the server. On the client, wallets are imported via
>
> ```js
> import wallet from 'wallets/<name>/client'
> ```
>
> vs
>
> ```js
> import wallet from 'wallets/<name>/server'
> ```
>
> on the server.
>
> To have access to the properties that can be shared between client and server, server.js and client.js always reexport everything in index.js with a line like this:
>
> ```js
> export * from 'wallets/<name>'
> ```
>
> If a wallet does not support paying invoices, this is all that client.js of this wallet does. The reason for this structure is to make sure the client does not import dependencies that can only be imported on the server and would thus break the build.

> [!WARNING]
> Wallets that support spending **AND** receiving have not been tested yet. For now, only implement either the interface for spending **OR** receiving until this warning is removed.

> [!TIP]
> Don't hesitate to use the implementation of existing wallets as a reference.

### index.js

An index.js file exports the following properties that are shared by imports of this wallet on the server and wallet:

- `name: string`

This acts as an ID for this wallet on the client. It therefore must be unique across all wallets and is used throughout the code to reference this wallet. This name is also shown in the [wallet logs](https://stacker.news/wallet/logs).

- `shortName?: string`

Since `name` will also be used in [wallet logs](https://stacker.news/wallet/logs), you can specify a shorter name here which will be used in logs instead.

- `fields: WalletField[]`

Wallet fields define what this wallet requires for configuration and thus are used to construct the forms like the one you can see at [/settings/wallets/lnbits](https://stacker.news/settings/walletslnbits).

- `card: WalletCard`

Wallet cards are the components you can see at [/settings/wallets](https://stacker.news/settings/wallets). This property customizes this card for this wallet.

- `fieldValidation: (config) => { [key: string]: string } | Yup.ObjectSchema`

This property defines how Formik should perform form-level validation. As mentioned in the [documentation](https://formik.org/docs/guides/validation#form-level-validation), Formik supports two ways to perform such validation.

If a function is used for `fieldValidation`, the built-in form-level validation is used via the [`validate`](https://formik.org/docs/guides/validation#validate) property of the Formik form component.

If a [Yup object schema](https://github.com/jquense/yup?tab=readme-ov-file#object) is set, [`validationSchema`](https://formik.org/docs/guides/validation#validationschema) will be used instead.

This validation is triggered on every submit and on every change after the first submit attempt.

Refer to the [Formik documentation](https://formik.org/docs/guides/validation) for more details.

- `walletType?: string`

This field is only required if this wallet supports receiving payments. It must match a value of the enum `WalletType` in the database.

- `walletField?: string`

Just like `walletType`, this field is only required if this wallet supports receiving payments. It must match a column in the `Wallet` table.

> [!NOTE]
> This is the only exception where you have to write code outside this directory for a wallet that supports receiving: you need to write a database migration to add a new enum value to `WalletType` and column to `Wallet`. See the top-level [README](../README.md#database-migrations) for how to do this.

#### WalletField

A wallet field is an object with the following properties:

- `name: string`

The configuration key. This is used by [Formik](https://formik.org/docs/overview) to map values to the correct input. This key is also what is used to save values in local storage or the database. For wallets that are stored on the server, this must therefore match a column in the corresponding table for wallets of this type.

- `label: string`

The label of the configuration key. Will be shown to the user in the form.

- `type: 'text' | 'password'`

The input type that should be used for this value. For example, if the type is `password`, the input value will be hidden by default using a component for passwords.

- `optional?: boolean | string = false`

This property can be used to mark a wallet field as optional. If it is not set, we will assume this field is required else 'optional' will be shown to the user next to the label. You can use Markdown to customize this text.

- `help?: string | { label: string, text: string }`

If this property is set, a help icon will be shown to the user. On click, the specified text in Markdown is shown. If you additionally want to customize the icon label, you can use the object syntax.

- `editable?: boolean = true`

If this property is set to `false`, you can only configure this value once. Afterwards, it's read-only. To configure it again, you have to detach the wallet first.

- `placeholder?: string = ''`

Placeholder text to show an example value to the user before they click into the input.

- `hint?: string = ''`

If a hint is set, it will be shown below the input.

- `clear?: boolean = false`

If a button to clear the input after it has been set should be shown, set this property to `true`.

- `autoComplete?: HTMLAttribute<'autocomplete'>`

This property controls the HTML `autocomplete` attribute. See [the documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) for possible values. Not setting it usually means that the user agent can use autocompletion. This property has no effect for passwords. Autocompletion is always turned off for passwords to prevent passwords getting saved for security reasons.

#### WalletCard

- `title: string`

The card title.

- `subtitle: string`

The subtitle that is shown below the title if you enter the configuration form of a wallet.

- `badges: string[]`

The badges that are shown inside the card.

### client.js

A wallet that supports paying invoices must export the following properties in client.js which are only available if this wallet is imported on the client:

- `testConnectClient: async (config, context) => Promise<void>`

`testConnectClient` will be called during submit on the client to validate the configuration (that is passed as the first argument) more thoroughly than the initial validation by `fieldValidation`. It contains validation code that should only be called during submits instead of possibly on every change like `fieldValidation`.

How this validation is implemented depends heavily on the wallet. For example, for NWC, this function attempts to fetch the info event from the relay specified in the connection string whereas for LNbits, it makes an HTTP request to /api/v1/wallet using the given URL and API key.

This function must throw an error if the configuration was found to be invalid.

The `context` argument is an object. It makes the wallet logger for this wallet as returned by `useWalletLogger` available under `context.logger`. See [components/wallet-logger.js](../components/wallet-logger.js).

- `sendPayment: async (bolt11: string, config, context) => Promise<{ preimage: string }>`

`sendPayment` will be called if a payment is required. Therefore, this function should implement the code to pay invoices from this wallet.

The first argument is the [BOLT11 payment request](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md). The `config` argument is the current configuration of this wallet (that was validated before). The `context` argument is the same as for `testConnectClient`.

> [!IMPORTANT]
> As mentioned above, this file must exist for every wallet and at least reexport everything in index.js so make sure that the following line is included:
>
> ```js
> export * from 'wallets/<name>'
> ```
>
> where `<name>` is the wallet directory name.

> [!IMPORTANT]
> After you're done implementing the interface, you need to import this wallet in _wallets/client.js_ and add it to the array that is the default export of that file to make this wallet available across the code:
>
> ```diff
> // wallets/client.js
> import * as nwc from 'wallets/nwc/client'
> import * as lnbits from 'wallets/lnbits/client'
> import * as lnc from 'wallets/lnc/client'
> import * as lnAddr from 'wallets/lightning-address/client'
> import * as cln from 'wallets/cln/client'
> import * as lnd from 'wallets/lnd/client'
> + import * as newWallet from 'wallets/<name>/client'
>
> - export default [nwc, lnbits, lnc, lnAddr, cln, lnd]
> + export default [nwc, lnbits, lnc, lnAddr, cln, lnd, newWallet]
> ```

### server.js

A wallet that supports receiving file must export the following properties in server.js which are only available if this wallet is imported on the server:

- `testConnectServer: async (config, context) => Promise<void>`

`testConnectServer` is called on the server during submit and can thus use server dependencies like [`ln-service`](https://github.com/alexbosworth/ln-service).

It should attempt to create a test invoice to make sure that this wallet can later create invoices for receiving.

Again, like `testConnectClient`, the first argument is the wallet configuration that we should validate. However, unlike `testConnectClient`, the `context` argument here contains `me` (the user object) and `models` (the Prisma client).

- `createInvoice: async (amount: int, config, context) => Promise<bolt11: string>`

`createInvoice` will be called whenever this wallet should receive a payment. The first argument `amount` specifies the amount in satoshis. The second argument `config` is the current configuration of this wallet. The third argument `context` is the same as in `testConnectServer` except it also includes `lnd` which is the return value of [`authenticatedLndGrpc`](https://github.com/alexbosworth/ln-service?tab=readme-ov-file#authenticatedlndgrpc) using the SN node credentials.


> [!IMPORTANT]
> Don't forget to include the following line:
>
> ```js
> export * from 'wallets/<name>'
> ```
>
> where `<name>` is the wallet directory name.

> [!IMPORTANT]
> After you're done implementing the interface, you need to import this wallet in _wallets/server.js_ and add it to the array that is the default export of that file to make this wallet available across the code:
>
> ```diff
> // wallets/server.js
> import * as lnd from 'wallets/lnd/server'
> import * as cln from 'wallets/cln/server'
> import * as lnAddr from 'wallets/lightning-address/server'
> + import * as newWallet from 'wallets/<name>/client'
>
> - export default [lnd, cln, lnAddr]
> + export default [lnd, cln, lnAddr, newWallet]
> ```

PR #1249, opened 2024-07-01
Fix push notifications

ekzyis:
* [x] fix wrong author in reply push notification
* [x] fix duplicate mention push notifications on edits
* [x] fix missing item mention push notifications on create
* [x] zaps push notifications use stale value because we didn't commit the transaction yet

I noticed that we can't fully eliminate duplicates if we delete existing mentions. For example, if I mention someone, then edit my comment to remove the mention and then edit again mentioning them again, they will receive another notification.

I think we didn't delete them before.


huumn:
THANK YOU!

PR #1256, opened 2024-07-07
Fix missing stackedMsats update + wrong sybil fee

ekzyis:
## Description

A stacker reported to me that even though [this stacker](https://stacker.news/vostrnad) stacked 20k+ sats on [their post](https://stacker.news/items/600187), `stacked` in their profile still showed 0 sats.

I noticed that the new zap code introduced in #1195 is missing the update to the `users.stackedMsats` column:

https://github.com/stackernews/stacker.news/blob/371084016740fc29751e5964e17c687420c6ea57/api/paidAction/zap.js#L86-L88

After I fixed that, I noticed that 99 sats were added to `stackedMsats` instead of the expected 90 sats for a 100 sat zap.

Then I also noticed that the sybil fee is calculated wrong. It divides by 100 which is equal to 1% instead of dividing by 10 for 10% since `x * 0.1 = x * 1/10`.

## Checklist

yes

yes

### DIFF
--- a/api/paidAction/zap.js
+++ b/api/paidAction/zap.js
@@ -84,7 +84,9 @@ export async function onPaid ({ invoice, actIds }, { models, tx }) {
       WHERE users.id = forwardees."userId"
     )
     UPDATE users
-    SET msats = msats + ${itemAct.msats}::BIGINT - (SELECT msats FROM total_forwarded)::BIGINT
+    SET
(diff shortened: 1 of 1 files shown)


ekzyis:
Mhh, I think this means `stackedMsats` is wrong for everyone now and we have to recalculate it.

PR #1259, opened 2024-07-08
Ignore canceled invoices in item submit error handling

### LINKED ISSUE #1257: Text Overflowing Out Of Bounds For "invoice canceled" Notification
**Description**
When displaying a notification for a canceled invoice, the text overflows out of the bounds of the toast box, causing it to be partially hidden and unreadable.

**Steps to Reproduce**
Trigger a toast notification for a canceled invoice.

0. Be unauthenticated to force a payment.
1. Go to any post, write anything within the text area to leave a comment then press "reply" for 100 sats.
2. A window will pop up with the QR code/address to send the 100 sats, close this by pressing the "X" from the top-right side.
3. In the bottom-right side a notification will pop up with the message "invoice canceled" and a hash.
4. Observe:
    4.1 The hash string is overflowing out of bounds.
    4.2 There is no "X" top-right side of the red box notification to close it out.
    4.3 The notification will not go away after a few seconds, similar in the case of a green notification.
    4.4 As a result of 4.2 and 4.3, you can stack multiple notifications until you refresh the page.

**Expected behavior**
A) Text from the hash string is either cut to fit the notification box, or the notification box should be dynamic to fit the the text.
Other option as a quick fix can be so that the hash string can wrap on a 3rd line instead of overflowing or have it cut off.
B) User can manually close the notification.
C) Notification goes away by itself without the need for a page refresh.

**Screen
Captureshots**

**Environment:**
Same behavior on Win10 (Firefox & Chrome) and Debian (Firefox).

ekzyis:
## Description

Fix #1257 

## Checklist

Yes

Yes

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -33,6 +33,7 @@ import EyeClose from '@/svgs/eye-close-line.svg'
 import Info from './info'
 import { useMe } from './me'
 import classNames from 'classnames'
+import { InvoiceCanceledError } from './payment'
 
 export class SessionRequiredError extends Error {
   constructor () {
@@ -844,7 +845,9 @@ export function Form ({
       }
     } catch (err) {
       console.log(err.message, err)
-      toaster.danger(err.message ?? err.toString?.())
+      if (!(err instanceof InvoiceCanceledError)) {
+        toaster.danger(err.message ?? err.toString?.())
+      }
       return
     }
 


huumn (changes requested):
This change shouldn't be in `Form`. `Form` shouldn't need to know about payments. I believe this belongs in `use-item-submit`.

Also, if the issue is overflow, should the solution be to remove the error?

If the issue is "we don't need this error," then removing it makes sense.

ekzyis:
> Also, if the issue is overflow, should the solution be to remove the error?

Is it really useful to tell the user that they canceled the invoice? Also, since it's a red error toast, it looks like something went wrong even though nothing happened that should concern them: the invoice was successfully canceled. You mentioned we want to show less red stuff to the user which makes sense to me.

It also only happens when anon tries to post something so it's inconsistent.

huumn:
> Mhh, we need to throw the onSubmit in Form because it should not be considered a successful submit. Else we will clear local storage. So I think we unfortunately have to handle it at this level since it's related to control flow at this level.

Okay, I'll try to figure an alternative out.

huumn:
Done in 6cf16d3da7790e4d36fe42295242506e9a3cc9fd ... Basically I made it so that paid actions only throw when invoice isn't cancelled ... lots of paid action mutations didn't use `Form` anyhow, so this will also fix this issue for those paid actions ... in the parent commit I fixed the overflow for future stuff

PR #1263, opened 2024-07-13
random zap amounts

### LINKED ISSUE #1112: Random zap setting
Add a setting that when enabled allows someone to zap a random amount within a specified range.

This will require adding new user settings and modifying zaps to be random when enabled.

SatsAllDay:
## Description
Closes #1112

Adds an option to enable random zap amounts as a user setting.

Users can opt in to random zapping in their settings. If they enable the setting, they must pick a minimum and maximum random zap in their user settings before being able to save this option.

Random zaps take precedence over a default zap value.

Random zap config is validated to ensure that a min and max is specified, and that the min is less than the max, if the feature is enabled.

## Additional Context
N/A

## Checklist

Yes

Yes

I made the desktop window smaller but that's about it.

No.


api/typeDefs/user.js
```diff
@@ -98,6 +98,9 @@ export default gql`
     noteItemMentions: Boolean!
     nsfwMode: Boolean!
     tipDefault: Int!
+    tipRandom: Boolean!
```
SatsAllDay, line 101: I named everything `tipRandom...`, following the existing pattern from `tipDefault` and `turboTipping`. Would you prefer me to use `zap` instead? e.g. `zapRandom`, `zapRandomMin`, etc.? @huumn

huumn: tip is fine. We've been so lazy with the names

.github/workflows/lint.yml
```diff
@@ -1,4 +1,4 @@
-name: Eslint Check
+name: Lint Check
```
SatsAllDay, line 1: Just a nit because we don't actually use eslint.

components/item-act.js
```diff
@@ -101,7 +101,7 @@ export default function ItemAct ({ onClose, item, down, children, abortSignal })
   return (
     <Form
       initial={{
-        amount: me?.privates?.tipDefault || defaultTips[0],
+        amount: defaultTipIncludingRandom(me?.privates) || defaultTips[0],
```
SatsAllDay, line 104: Honor random zap settings here, too.

components/upvote.js
```diff
@@ -67,11 +67,20 @@ export function DropdownItemUpVote ({ item }) {
   )
 }
 
-export const nextTip = (meSats, { tipDefault, turboTipping }) => {
+export const defaultTipIncludingRandom = ({ tipDefault, tipRandom, tipRandomMin, tipRandomMax }) =>
```
SatsAllDay, line 70: A reusable function for determining the next zap value that honors `tipRandom` and `tipDefault`

lib/validate.js
```diff
@@ -540,8 +540,31 @@ export const actSchema = object({
   act: string().required('required').oneOf(['TIP', 'DONT_LIKE_THIS'])
 })
 
-export const settingsSchema = object({
+export const settingsSchema = object().shape({
```
SatsAllDay, line 543: Had to do this for the cyclic dependency issue noted below

SatsAllDay:
I think this is ready for review. Let me know of any feedback!

prisma/schema.prisma
```diff
@@ -30,6 +30,9 @@ model User {
   apiKeyHash                String?              @unique(map: "users.apikeyhash_unique") @db.Char(64)
   apiKeyEnabled             Boolean              @default(false)
   tipDefault                Int                  @default(100)
+  tipRandom                 Boolean              @default(false)
+  tipRandomMin              Int?
+  tipRandomMax              Int?
```
huumn, line 35: I don't think we need the boolean if we can just check for null on min and max. I'd suggest we use a postgres [range type](https://www.postgresql.org/docs/current/rangetypes.html) too but [prisma doesn't support it](https://github.com/prisma/prisma/issues/3287).

huumn (changes requested):
Sorry for the slow review. The only thing blocking is the sort of redundant boolean. Not a huge deal, but we already have so many columns.

(I'll want to clean up the UI of the zap accordion too, but given that involves zap undos too, I'll do that after merging.)

SatsAllDay:
> Sorry for the slow review. The only thing blocking is the sort of redundant boolean. Not a huge deal, but we already have so many columns.
> 
> 
> 
> (I'll want to clean up the UI of the zap accordion too, but given that involves zap undos too, I'll do that after merging.)

Ok, thank you for the feedback! FWIW, I went with the redundant Boolean so users could explicitly enable/disable the feature which I felt was a more clear pattern, and also so that users could disable the feature without wiping out their configured values.

In any case, I can make the adjustments, hopefully in the next day or two :)

huumn:
I think the boolean in the UI does make sense. For the db, I'd just prefer the values. Toggling off an on wouldn't persist old values but I don't suspect that'd happen a lot.

SatsAllDay:
What are your thoughts on keeping `tipRandom: Boolean!` as part of the `UserPrivates` in the GraphQL query API, and just having a manual resolver in `api/resolvers/user` which does the null check on `tipRandomMax` and `tipRandomMin` to resolve it, but still removing it from `SettingsInput` and the db? Basically leave it as a read-only field, computed in the API. I find it nice to have `me.privates.tipRandom` in the API response to consume in various spots in the UI.

I'm ok either way, but this idea occurred to me as I am working on the above feedback and wanted to get your take.

SatsAllDay:
OK, I think we're ready for another review pass!

components/upvote.js
```diff
@@ -67,11 +67,20 @@ export function DropdownItemUpVote ({ item }) {
...
   // what should our next tip be?
-  if (!turboTipping) return (tipDefault || 1)
+  const calculatedDefault = defaultTipIncludingRandom({ tipDefault, tipRandom, tipRandomMin, tipRandomMax })
+
+  if (!turboTipping) {
+    return calculatedDefault
+  }
 
-  let sats = tipDefault || 1
+  let sats = calculatedDefault
```
huumn, line 83: Ohhh cool compatible with turbo zapping!

huumn (approved):
Still need to QA but no objections otherwise.

PR #1264, opened 2024-07-15
feat: added NGN currency

### LINKED ISSUE #1226: Addition of Naira(NGN) to fiat currencies
Not really a problem, more like a reference

Addition of the Naira(NGN) as part of the currencies as they're more African users from that location

Screenshot 2024-06-05 140430

**Additional context**
1ngn == 1sat, so this shouldn't be so hard to implement

aniskhalfallah:
## Description

Closes #1226 
Added Naira currency.

## Screenshots

proof_ngn-stacker news

## Checklist

Yes, by deleting the line that adds the Naira Currency

Yes

N/A

N/A

### DIFF
--- a/lib/currency.js
+++ b/lib/currency.js
@@ -5,6 +5,7 @@ export const CURRENCY_SYMBOLS = {
   GBP: '£',
   USD: '$',
   NZD: '$',
+  NGN: '₦',
   ZAR: 'R '
 }
 


ekzyis (approved):
Sorry for the slow review, works well!

A nitpick (which was already the case prior to this PR) is that we show the new currency symbol with the old value until the new data price data is fetched. This can take up to 30s.

I briefly looked into refetching it immediately using `useEffect` in `PriceProvider` but since we do the price lookup asynchronously in the backend, it requires more changes than I expected for such a small detail.

aniskhalfallah:
Hello @ekzyis Thanks for the review ! 
I noticed the issue after you mentioned it, I looked into it by using `useEffect` to refetch after changing the currency: 

```
useEffect(() => {
    if (fiatCurrency) {
      refetch({ fiatCurrency });
    }
  }, [fiatCurrency, refetch]);
```

I tested it, and in my requests I see a query 1 second after I change the currency

useEffect
useEffect2
useEffect3

I am not sure whether this fixes the issue or not since you said it requires more changes, Let me know what you think.

PR #1274, opened 2024-07-23
Add WebLN for sending payments

ekzyis:
## Description

Since WebLN requires no configuration, I had to add some additional code to handle this edge case:

* `useWallet` now exposes `hasConfig` which is consumed by `WalletButtonBar` to only show `detach` if a wallet actually has config we can detach and to only show the security banner if there is config we store
* added new exportable property `available` that overrides default behavior of when `enabled` checkbox is clickable (see README)
* making sure `config.enabled` is always false if `available` is false

## Checklist

Yes

Yes


huumn:
I'm going to check this out to play with it. But I never liked that `webln` is on by default enabled if it's present. I'd like stackers to be able to opt-in/out like other wallets.

huumn:
Cool, it requires explicit enabling. 

attach/detach/enable UX is a bit inconsistent but it's not super clear how it should be.

PR #1276, opened 2024-07-24
fix: fetch price immediately after currency changes

aniskhalfallah:
## Description

as mentioned by @ekzyis [here](https://github.com/stackernews/stacker.news/pull/1264#pullrequestreview-2183230416), when we change the currency in settings, we don't immediately query the price, we wait for 30 seconds before doing so, which leaves us with a new currency symbol but with the old value. this PR aims to fix that.

## Screenshots

Screenshot 2024-07-24 150633

## Checklist

Yes
Yes

No

### DIFF
--- a/api/resolvers/price.js
+++ b/api/resolvers/price.js
@@ -19,12 +19,16 @@ async function getPrice (fiat) {
   if (cache.has(fiat)) {
     const { price, createdAt } = cache.get(fiat)
     const expired = createdAt + expiresIn < Date.now()
-    if (expired) fetchPrice(fiat).catch(console.error) // update cache
-    return price // serve stale price (this on the SSR critical path)
+    if (expired) {
+      const newPrice = await fetchPrice(fiat);
+      return newPrice;
+    } else {
+      return price;
+    }
   } else {
-    fetchPrice(fiat).catch(console.error)
+    const newPrice = await fetchPrice(fiat);
+    return newPrice;
   }
-  return null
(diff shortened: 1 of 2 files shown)


ekzyis (changes requested):
We want to only await the promise in the case where the currency was updated. As the comment mentioned, this is else on the SSR critical path so this would stall requests until the Coinbase API (over which we have no control) responds.

So I think you need a flag that you set in the `useEffect` query to await the price in this special case here.

aniskhalfallah:
OK thank you for the feedback, I have included a new flag `changedCurrency` so we only await the response when the currency changes. I think we are ready for a review now :)

ekzyis (commented):
You forgot to update the GraphQL type definition and query so `changedCurrency` never arrived at the server. I fixed that in 4fb250c0.

I hope you don't mind but I also refactored the code in 5ef94ade since there was some duplicate `if (changedCurrency) ...` code. I also renamed `changedCurrency` to `fromCache` to reflect better what this variable is used for.

**However, I noticed that `refetch` not only refetches the data with new variables, but also sets these new variables for following polls.**

So this mean if we set `fromCache: false` in the refetch, the price polls will from now on also use `fromCache: false` which is not intended...

Not sure what's the best way to fix this unintended interaction between refetching and polling.

Tbh, I suggest you work on something else since this seems to indeed be more complicated than fixing this bug is relevant.

PR #1278, opened 2024-07-30
Support receiving via LNbits 

ekzyis:
## Description

This PR implements receiving via LNbits. This will be the first wallet that supports both.

**Design**

- Wallet fields are marked with `clientOnly` or `serverOnly`.
If both aren't set, we assume it's a field shared by client and server.
- `generateMutation` and `injectTypeDefs` are aware which fields should be sent to the server
- `fieldValidation` checks if config is valid for spending OR receiving using [`schema.when`](https://github.com/jquense/yup?tab=readme-ov-file#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema)
- `saveConfig` extracts client/local/spending and server/receiving config and only saves them if they are valid

**Restrictions**

- enabled and priority only exists per wallet, not per send/recv
This means if a wallet is configured for both and you want to use it for receiving, the priority for spending is the same. I think this actually matches user expectations since send from and receive to same wallet is convenient, so I don't think this will be a huge problem.

TODO:

- [x] update validation to check if it's spending or receiving config
- [x] save + fetch server config
- [x] implement `lnbits.testConnectServer`
- [x] implement `lnbits.createInvoice`
- [x] rename "local config" to "client config" to be consistent with "server config"
- [x] make sure priority and enabled are synced between client and server (or other solution to potential mismatch of data that lives at client and server) _(`setPriority` calls `saveConfig` which sets same priority locally and on server. However, initial priority is 0 on server and undefined locally but `walletPrioritySort` should handle this gracefully in a way that makes sure that wallets are ordered in the way they will be used.)_ 
- [x] ~fix cannot unset admin key since local config is only stored if it's valid: include a dropdown menu to `detach` button to only detach config for spending or receiving (?)~ _mhh, will ignore for now, maybe not that big of an issue_
- [x] fix logger on client disabled if wallet supports receiving
- [ ] separate status indicators for payments and receivals instead of green dot if any direction is enabled (maybe do this in another PR)
- [x] fix undefined stale `walletId` used in `clearConfig` if we refresh config page and then detach _(fixed in f65a9e86)_ 

**Additional changes**

* created `walletValidate` function to reuse logic which checks type of `wallet.fieldValidation` to run validation
* wallet logger now logs payments and receivals separately:

2024-08-03-173143_254x121_scrot

This was required since the previous wallet logger implementation assumed that only server or client (but not both) are logging so there were duplicate `wallet updated` logs.

## Additional Context

* adding a server wallet is quite tedious. Need to mention the new wallet as a fragment in multiple queries and took me some time to fix a bug caused by forgetting to create a trigger for the new LNbits wallet to update the `Wallet` table 
* [this comment](https://github.com/stackernews/stacker.news/pull/1278/files#diff-fe57f778594fcebe6b8d96b7e6e0355bec7f8ca042a8397dc6a7fb508674344fR177-R183):

```
// NOTE:
//   verifying the local/server configuration before saving it
//   prevents unsetting just one configuration if both are set.
//   This means there is no way of unsetting just one configuration
//   since 'detach' detaches both.
//   Not optimal UX but the trade-off is saving invalid configurations
//   and maybe it's not that big of an issue.
```

* I don't really like the name "receivals"

## Checklist

Yes, only added a table `WalletLNbits` and a new value `LNBITS` to type `WalletType`

Yes, tested payments and autowithdrawals from/to LNbits + detaching & updates + LND autowithdrawals + order in which wallets are attempted for payments and autowithdrawals

No


wallets/lnbits/server.js
```diff
@@ -0,0 +1,30 @@
...
+  const payment = await res.json()
+  return payment.payment_request
+}
+
+export async function testConnectServer ({ url, invoiceKey }) {
+  return await _createInvoice({ url, invoiceKey, amount: 1, expiry: 1 })
+}
+
+export async function createInvoice ({ amount, maxFee }, { url, invoiceKey }) {
+  return await _createInvoice({ url, invoiceKey, amount, expiry: 360 })
```
ekzyis, line 29: TODO: Check if `expiry` is really given in seconds

ekzyis:
LNbits can now send and receive in f64bbe7. Will stop force-pushing now.

Updated remaining TODOs in PR description + all kind of testing left.

wallets/lnbits/server.js
ekzyis: Yes, verified by creating an invoice with the LNbits API using `expiry: 1` and then pasting the invoice into https://lightningdecoder.com/ and seeing that time expire date is one second after timestamp.

huumn:
> I don't really like the name "receivals"

how about "receives"

ekzyis:
> how about "receives"

Mhh, yes, slightly better. Did that in c9313cf9.

I also considered "withdrawals" which imo makes sense until #1178 is merged but would thus only move this naming problem to later. Or "paying invoices" instead of "payments" and "creating invoices" instead of "receives" ...

Other naming issue:

I also wondered if `testServerConnect` should be called `testCreateInvoice` since we always create test invoices in it.

Thinking this further, `testClientConnect` could also be called `testSendPayment`. This would require changing the current sanity checks in `testClientConnect` to actually attempt payments. The surrounding code could generate a HODL invoice and let `testSendPayment` pay it. The surrounding code then checks if the payment was held in which case validation succeeds and the invoice is canceled to return the funds.

This would be stronger validation which makes sense for LNbits and NWC since it's possible to enter a configuration that passes their validation but that can't actually pay invoices: you can also fetch a LNbits wallet with `invoiceKey` instead of `adminKey` or fetch the NWC info event with a wrong secret.

I can create a ticket for this validation with a HODL invoice if this makes sense to you.

huumn:
> I can create a ticket for this validation with a HODL invoice if this makes sense to you.

I think this makes sense. Your sense is as good as my sense. :)

ekzyis:
This is ready for review now.

As mentioned in the description, detach UX can be improved since it only allows detaching payments + receives.

I think we can ship without being able to detach only one direction.

Separate status indicators can also be added later imo.

I tested complete user flows for attaching LNbits for payments and/or autowithdrawals and also tested LND autowithdrawals to make sure other wallets didn't break for some reason. I hope I didn't miss anything but it seems to work well enough now to open it up for review.

wallets/index.js
```diff
@@ -106,46 +110,114 @@ export function useWallet (name) {
...
+    }
+    if (hasServerConfig) {
+      const newServerConfig = extractServerConfig(wallet.fields, newConfig)
+      await walletValidate(wallet, newServerConfig)
+        .then(() => true)
+        .catch(() => false)
+        .then(async (valid) => {
+          if (valid) return await setServerConfig(newServerConfig)
+        })
+    }
```
ekzyis, line 210: I created `walletValidate` since I needed to reuse the validation logic here to make sure only valid configurations are saved.

Since `fieldValidation` makes sure that at least one configuration (for payments or for receives)  is valid, we don't know if both are valid so we extract the corresponding configuration and validate it independently to see if it's valid before saving it.

api/resolvers/wallet.js
```diff
@@ -353,7 +347,7 @@ const resolvers = {
   },
   WalletDetails: {
     __resolveType (wallet) {
-      return wallet.address ? 'WalletLNAddr' : wallet.macaroon ? 'WalletLND' : 'WalletCLN'
+      return wallet.address ? 'WalletLNAddr' : wallet.macaroon ? 'WalletLND' : wallet.rune ? 'WalletCLN' : 'WalletLNbits'
```
ekzyis, line 350: there is probably a way to generate this code

api/typeDefs/wallet.js
```diff
@@ -2,19 +2,22 @@ import { gql } from 'graphql-tag'
...
       let args = 'id: ID, '
-      args += w.fields.map(f => {
-        let arg = `${f.name}: String`
-        if (!f.optional) {
-          arg += '!'
-        }
-        return arg
-      }).join(', ')
+      args += w.fields
+        .filter(isServerField)
```
ekzyis, line 13: this prevents sending something like `adminKey` to the server

api/typeDefs/wallet.js
```diff
@@ -74,7 +77,12 @@ const typeDefs = `
     cert: String
   }
 
-  union WalletDetails = WalletLNAddr | WalletLND | WalletCLN
+  type WalletLNbits {
+    url: String!
+    invoiceKey: String!
+  }
+
+  union WalletDetails = WalletLNAddr | WalletLND | WalletCLN | WalletLNbits
```
ekzyis, line 85: can probably also be generated

components/autowithdraw-shared.js
```diff
@@ -25,10 +25,15 @@ export function AutowithdrawSettings ({ wallet }) {
...
+  const [mounted, setMounted] = useState(false)
+  useEffect(() => {
+    setMounted(true)
+  }, [])
+
   return (
     <>
       <Checkbox
-        disabled={!wallet.isConfigured}
+        disabled={mounted && !wallet.isConfigured}
```
ekzyis, line 36: Required to enable the checkbox if a wallet supports both but only payments are configured and we reload the configuration page.

In that case, we render this checkbox on the server as disabled so we need to render it again on the client where the wallet will load the configuration from local storage.

pages/settings/wallets/[wallet].js
```diff
@@ -57,15 +57,11 @@ export default function WalletSettings () {
 
             await wallet.save(values)
 
-            if (values.enabled) wallet.enable()
-            else wallet.disable()
-
```
Comment (ekzyis) [inline pages/settings/wallets/[wallet].js:62]: Moved into `saveConfig` of `useConfig`

pages/settings/wallets/[wallet].js
```diff
@@ -106,7 +102,7 @@ export default function WalletSettings () {
 
 function WalletFields ({ wallet: { config, fields, isConfigured } }) {
   return fields
-    .map(({ name, label, type, help, optional, editable, ...props }, i) => {
+    .map(({ name, label, type, help, optional, editable, clientOnly, serverOnly, ...props }, i) => {
```
Comment (ekzyis) [inline pages/settings/wallets/[wallet].js:105]: Destructuring unused props to avoid DOM warnings

wallets/index.js
```diff
@@ -32,6 +33,22 @@ export function useWallet (name) {
...
+  const disablePayments = useCallback(() => {
+    disableWallet(name, me)
+    logger.info('payments disabled')
+  }, [name, me, logger])
+
+  if (wallet) {
+    wallet.isConfigured = _isConfigured
+    wallet.enablePayments = enablePayments
+    wallet.disablePayments = disablePayments
+  }
```
ekzyis, line 50: Mhh, maybe not the best code style but this makes `wallet.enablePayments()` etc. available in `useConfig` even though `wallet` is passed before this code runs :thinking:

wallets/index.js
```diff
@@ -49,53 +66,40 @@ export function useWallet (name) {
...
-      logger.ok(_isConfigured ? 'wallet updated' : 'wallet attached')
+      validConfig = await wallet.testConnectClient?.(newConfig, { me, logger })
     } catch (err) {
-      const message = err.message || err.toString?.()
-      logger.error('failed to attach: ' + message)
+      logger.error(err.message)
       throw err
     }
-  }, [_isConfigured, saveConfig, me, logger])
+    await saveConfig(validConfig ?? newConfig, { logger })
```
ekzyis, line 90: Only `testConnectClient` is now wrapped in `try/catch` because only its errors need to be logged.  `saveConfig` already handles logging within.

api/resolvers/wallet.js
huumn: The 3rd parameter passed to [`__resolveType`](https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces/#resolving-a-union) is this [info object](https://github.com/graphql/graphql-js/blob/a24a9f35b876bdd0d5050eca34d3020bd0db9a29/src/type/definition.ts#L891-L902). It might have the grandparent in it to determine the type.

huumn (changes requested):
When I try to attach it just for sending or receiving on their own, it logs

```
39s | [lnbits] | ok | wallet found
39s | [lnbits] | info | trying to fetch wallet
```

Then the form finishes submitting, but it doesn't store the config, or light the indicator.

---------------

When I add both the admin and invoice key, it submits and lights the indicator, but only stores the invoice key.

```
31s | [lnbits] | ok | receives enabled
31s | [lnbits] | ok | wallet attached for receives
31s | [lnbits] | ok | wallet found
31s | [lnbits] | info | trying to fetch wallet
```

ekzyis:
Sounds like I accidentally broke something, looking into it

huumn:
Could you also add an `ATTACH.md` doc when you get a chance. In order to test both sides simultaneously both the docker network and the host need to reach the lnbits container at the same hostname, and I assume you already figured out some way to do that.

ekzyis:
> Sounds like I accidentally broke something, looking into it

Yes, I broke something during my refactor in 0b7691a9 that I thought was foolproof just by reading it. Fixed in bcd62f1e.

> I assume you already figured out some way to do that.

No, I simply patched the `testServerConnect` function to always use `lnbits:5000` for url to test send+recv together. Need to figure out the best general solution.[^1]

[^1]: Just realized I accidentally checked that hardcoded url in again in 652f4436. Removed in d0c47a72.

What are the `CONNECT` labels for?

huumn:
> Regarding https://github.com/stackernews/stacker.news/commit/d2b7ac5a2d01ff05130e93fa2156c2841089cf28: What are the CONNECT labels for?

They appear in `sndev status` to show you what url to find the GUI or service at. And, when you run `sndev open SERVICE`, e.g.  `sndev open lnbits`, it'll open the GUI in the browser.

huumn (commented):
Once this wallet is added, I can't unset either the invoice key or the admin key. Is that expected? If it is, we should probably make those fields `readOnly`.

Otherwise, this works.

ekzyis:
It's bad UX but yes. It's related to validation: you should be able to save if spending or receiving works but we only save the part of the config that works. That means you can't remove a value and save since that will break spending or receiving. You can can only detach both using the detach button.

That's consistent with how it worked before except that now the submit goes through while before the validation before submit prevented the save.

I briefly considered adding separate detach buttons or tabs but abandoned that idea for the first iteration on wallets with send+recv.

wallets/index.js
```diff
@@ -106,46 +110,120 @@ export function useWallet (name) {
...
+      let valid = true
+      try {
+        await walletValidate(wallet, newServerConfig)
+      } catch {
+        valid = false
+      }
+
+      if (valid) await setServerConfig(newServerConfig)
+    }
+  }, [setClientConfig, setServerConfig, wallet])
```
huumn, line 217: This is missing `has...Config` deps.

huumn: Also, couldn't this set or unset local/server config if the fields are blank? Or am I missing something?

wallets/index.js
```diff
@@ -106,46 +110,120 @@ export function useWallet (name) {
...
+
+  const clearConfig = useCallback(async ({ logger }) => {
+    if (hasClientConfig) {
+      clearClientConfig()
+      wallet.disablePayments()
+      logger.ok('wallet detached for payments')
+    }
     if (hasServerConfig) await clearServerConfig()
-  }, [wallet])
+  }, [clearClientConfig, clearServerConfig, wallet])
```
huumn, line 226: also missing `has...Config`

wallets/index.js
ekzyis: > This is missing `has...Config` deps.

Done in 32c9f8c0.

> Also, couldn't this set or unset local/server config if the fields are blank? Or am I missing something?

If the fields are blank, `valid` will be false and `setLocal/ServerConfig` will not be called.

I also wanted to avoid saving invalid configurations so we can continue to assume that configurations are valid if they were saved.

wallets/index.js
ekzyis: Done in 32c9f8c0

ekzyis:
> If it is, we should probably make those fields `readOnly`.

Good idea! Done in 9210de99 + d7721ed2.

PR #1283, opened 2024-08-03
Fix wallet.createInvoice signature

ekzyis:
Originally, only `amount` was intended to be first argument of `createInvoice` (see README).

However, `createInvoice` for LN addresses required `maxFee` to be passed in so the schema validation succeeds.

It's dumb that `maxFee` has to be part of the signature for every wallet just because of this.

`createInvoice` for LN addresses now simply passes in `maxFee: 0` to pass schema validation.

This is still dumb but less and works.

### DIFF
--- a/wallets/cln/server.js
+++ b/wallets/cln/server.js
@@ -22,7 +22,7 @@ export const testConnectServer = async (
 }
 
 export const createInvoice = async (
-  { amount },
+  amount,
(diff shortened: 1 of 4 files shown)


huumn:
This is obsolete once #1178 is merged. It doesn't overload the withdrawal lightning address function. Let's prioritize that.

PR #1291, opened 2024-08-09
replace greeter mode with investment filter

### LINKED ISSUE #690: Only view posts that have at least X amount of sats invested in them

huumn:
closes #690

This basically turns greeter mode from `yes, see freebies` or `no, don't see freebies` -to-> only see *posts* with at least X sats "invested." For comments, if `investmentFilter` > 0, freebies are collapsed unless they've been upvoted.

This is also compatible with wild west mode.

By default `investmentFilter = 10` and also applies to lurkers, giving them better spam filtering. We can increase this default for lurkers and new stackers as needed.

### DIFF
--- a/pages/settings/index.js
+++ b/pages/settings/index.js
@@ -140,7 +140,7 @@ export default function Settings ({ ssrData }) {
             hideTwitter: settings?.hideTwitter,
             imgproxyOnly: settings?.imgproxyOnly,
             wildWestMode: settings?.wildWestMode,
-            investmentFilter: settings?.investmentFilter,
+            satsFilter: settings?.satsFilter,
             nsfwMode: settings?.nsfwMode,
             nostrPubkey: settings?.nostrPubkey ? bech32encode(settings.nostrPubkey) : '',
             nostrCrossposting: settings?.nostrCrossposting,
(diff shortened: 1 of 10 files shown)


pages/settings/index.js
```diff
@@ -467,34 +472,39 @@ export default function Settings ({ ssrData }) {
             label={<>don't create referral links on copy</>}
             name='noReferralLinks'
           />
-          <div className='form-label'>content</div>
-          <Checkbox
+          <h4>content</h4>
+          <Input
             label={
-              <div className='d-flex align-items-center'>wild west mode
+              <div className='d-flex align-items-center'>filter posts with fewer than this many sats invested
```
ekzyis, line 478: Mhh, that's a long label. Would "investment filter" be too confusing even though we have the info icon? Or "sats filter" since I think "investment filter" is also too long and not descriptive enough?

If the label is renamed to "sats filter", I would also rename every use of "investmentFilter" to "satsFilter".

ekzyis (approved):
Looks good. Tested with an anon post (`cost = 100`) and updating settings + manually updating item cost.

Some notes:

* `Item.freebie` can be removed from `Item` table and only exist as a field resolver that returns `item.cost === 0` (similar to `freedFreebie`)
* the freebie filter used weighted votes which was more sybil resistant but imo it's worth trading sybil resistance for a simple filter like this. Also, there is still the outlaw filter which uses weighted votes.

huumn:
Good feedback. Will fix.

huumn:
I think removing the freebie column makes this backwards incompatible ... crap.

Issue #1142, opened 2024-05-03
Blink wallet sending attachment


huumn:
Create a blink sending attachment clientside.

See [Alby's code for an example](https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/connectors/ConnectGaloy/index.tsx) and [blink's docs](https://dev.blink.sv/api/btc-ln-send).

SatsAllDay:
Perhaps a noob question: what is meant by "attachment" in "blink sending attachment"? I tried to grok it via the provided links but I'm not sure it has registered.

huumn:
You'd be adding blink to the wallets in `attach wallets`.

SatsAllDay:
Wow, duh lol thanks!

riccardobl:
Is anyone working on this? If not i think i can take care of it 🫡

ekzyis:
No, nobody is working on this. I assigned this ticket to you now. If you get this done, you would be the first contributor who added a wallet! 

[This documentation](https://github.com/stackernews/stacker.news/blob/master/wallets/README.md) should help you. Let us know if you have any questions or feedback how to improve it!

PR #1293, opened 2024-08-10
Blink wallet sending attachment

riccardobl:
## Description

Adds blink sending attachment. Solves https://github.com/stackernews/stacker.news/issues/1142 .

## Screenshots

image

## Additional Context

This attachment currently works only with the  blink's staging environment due to an issue in cors configuration, should stay on hold until the fix is pushed to production.

## Checklist

Yes

Yes

Yes

No


ekzyis:
> This attachment currently works only with the blink's staging environment due to an issue in cors configuration, the pr should stay on hold until the fix is pushed to production.

Is there some discussion in https://github.com/GaloyMoney/blink-mobile or elsewhere that I can subscribe for updates regarding this? I searched for "CORS" but found nothing.

riccardobl:
> Is there some discussion in https://github.com/GaloyMoney/blink-mobile or elsewhere that I can subscribe to for updates regarding this? I searched for "CORS" but found nothing.

The conversation happened in the galoy dev chat, but there are these two related prs: 
https://github.com/GaloyMoney/charts/pull/6885
https://github.com/GaloyMoney/charts/pull/6888

Also, it seems it went to production 1 hour ago,  so, later today i will rebase this pr and do a final test and then it should be ready to go :+1: 


> And do you have any feedback regarding the [wallet documentation](https://github.com/stackernews/stacker.news/blob/master/wallets/README.md) or ease of implementation? Was it straight-forward or did you have to read a lot of the existing code to get an idea how to add a wallet (additionally to the documentation)? You are the first contributor who added a wallet so we're very curious how easy it was and if there is anything lacking that we could improve.

It was pretty straight-forward actually, the only doubt i had was if i had to put these variable in index.js or client.js https://github.com/stackernews/stacker.news/blob/b7b4e4688d4fef9387f7f273e83a29325380b376/wallets/blink/index.js#L3-L5
the documentation seems to imply that index.js should export only the specified properties , but i ended up putting them there anyway since they are variables that server.js could use at some point, not sure if this respect the design

wallets/blink/client.js
```diff
@@ -0,0 +1,181 @@
...
+async function request (authToken, query, variables = {}) {
+  const options = {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'X-API-KEY': authToken
+    },
+    body: JSON.stringify({ query, variables })
+  }
+  return fetch(galoyBlinkUrl, options).then(response => response.json())
```
ekzyis, line 180: If I enter an invalid blink API key, I get this error:

```
0s	[blink]	ERROR	Unexpected token '<', "<html> <h"... is not valid JSON
4s	[blink]	INFO	trying to fetch wallet
```

This is not a user-friendly error message; it looks like a bug. Fortunately, it does respond with `401 Unauthorized` (and not `200 OK` like most GraphQL APIs even on errors) so we can use it:

```suggestion
  const res = await fetch(galoyBlinkUrl, options)
  if (res.status === 401) {
    throw new Error('API responded with 401 Unauthorized')
  }
  return await res.json()
```

lib/validate.js
```diff
@@ -712,6 +712,32 @@ export const nwcSchema = object({
...
+    .test(async (currency, context) => {
+      try {
+        const supportedCurrencies = ['USD', 'BTC']
+        await string().required('required').oneOf(supportedCurrencies, 'must be one of ' + supportedCurrencies.join(', ')).validate(currency)
+      } catch (err) {
+        return context.createError({ message: err.message })
+      }
+      return true
+    })
+})
```
ekzyis, line 739: You don't need `.test` if you don't do custom validation. For NWC, I had to use it to parse the NWC url and individually validate sections of it + control validation order. You don't need that here.

Therefore, this should suffice:

```suggestion
export const blinkSchema = object({
  apiKey: string()
    .required('required')
    .matches(/^blink_[A-Za-z0-9]+$/, { message: 'must match pattern blink_A-Za-z0-9' }),
  currency: string()
    .transform(value => value.toUpperCase())
    .required('required')
    .default('BTC')
    .oneOf(['USD', 'BTC'], 'must be BTC or USD')
})
```

ekzyis (changes requested):
> It was pretty straight-forward actually

Cool!

> the documentation seems to imply that index.js should export only the specified properties , but i ended up putting them there anyway since they are variables that server.js could use at some point, not sure if this respect the design

That's fine, you're good.

Regarding the code: I unfortunately couldn't test the payment code since Blink apparently isn't available in the US.[^1] But looking at the [API docs](https://dev.blink.sv/api/btc-ln-send), it looks good. Left some suggestions regarding other stuff that I'd like to see resolved before merging.

Thank you for the PR!

[^1]: I used an American phone number during signup and later it said my IP address is not allowed to signup.

wallets/blink/client.js
riccardobl: done, i've changed it to catch other errors as well

lib/validate.js
riccardobl: done

riccardobl:
The current master has an issue that affect this pr, it seems something broke validation for optional fields in commit ae73b0c19fd277e1dfd2bac289a869b1dc980cfb .
The 'currency' field in the blink connector is defaulted to "BTC" by client.js, if unspecified, however since ae73b0c19fd277e1dfd2bac289a869b1dc980cfb if the field is left empty in the attach screen, the wallet does not attach and redirects back to settings/wallets without error. 
I've tried to rebase this pr on the commit before ae73b0c19fd277e1dfd2bac289a869b1dc980cfb and it works as expected there.

ekzyis (approved):
> The current master branch has something that affects this pr, it seems something broke validation for optional fields in commit https://github.com/stackernews/stacker.news/commit/ae73b0c19fd277e1dfd2bac289a869b1dc980cfb .
The 'currency' field in the blink connector is defaulted to "BTC" by client.js, if unspecified, however since https://github.com/stackernews/stacker.news/commit/ae73b0c19fd277e1dfd2bac289a869b1dc980cfb if the field is left empty in the attach screen, the wallet does not attach and redirects back to settings/wallets without error.
I've tried to rebase this pr on the commit before and it works as expected there.

You're right, something changed. We now run validation twice since ae73b0c: [once via Formik props](https://github.com/stackernews/stacker.news/blob/ec2383b99889f47f651860f9e41f38fbcd1940a2/pages/settings/wallets/%5Bwallet%5D.js#L36-L39) and [once just before saving](https://github.com/stackernews/stacker.news/blob/ec2383b99889f47f651860f9e41f38fbcd1940a2/wallets/index.js#L188-L204).

This is the case since the validation Formik runs just makes sure receiving OR sending works while the validation before save makes sure the configuration we're about to save is valid (spending config is saved on the client, while receiving config is saved on the server).

The save no longer works since `currency` is set to an empty string at that point which means the default doesn't apply so the `oneOf` check fails. It fails silently since it's assumed at least one save will work if Formik validation worked.

We can workaround this by returning the default using `transform` if `currency` is the empty string:

```diff
diff --git a/lib/validate.js b/lib/validate.js
index 776cacdd..39ba6e63 100644
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -717,8 +717,7 @@ export const blinkSchema = object({
     .required('required')
     .matches(/^blink_[A-Za-z0-9]+$/, { message: 'must match pattern blink_A-Za-z0-9' }),
   currency: string()
-    .transform(value => value.toUpperCase())
-    .default('BTC')
+    .transform(value => value ? value.toUpperCase() : 'BTC')
     .oneOf(['USD', 'BTC'], 'must be BTC or USD')
 })
```

I went ahead and pushed these changes (and other); I hope you don't mind.

PR #1296, opened 2024-08-13
Remove unnecessary me from addWalletLog params

ekzyis:
Since `wallet` contains `userId`, we don't need to pass `me` to `addWalletLog`.

Tested all `addWalletLog` calls. Works.

_related to [this comment](https://github.com/stackernews/stacker.news/pull/1178/files#r1700885838) in #1178 and based on and merging into #1178 to avoid merge conflicts later_


api/resolvers/wallet.js
```diff
@@ -577,8 +577,9 @@ async function upsertWallet (
     } catch (err) {
       console.error(err)
       const message = 'failed to create test invoice: ' + (err.message || err.toString?.())
-      await addWalletLog({ wallet, level: 'ERROR', message }, { me, models })
-      await addWalletLog({ wallet, level: 'INFO', message: 'receives disabled' }, { me, models })
+      wallet = { ...wallet, userId: me.id }
```
ekzyis, line 580: `wallet` might not exist yet here so it doesn't contain `userId`

PR #1302, opened 2024-08-15
Use custom relay API

ekzyis:
## Description

In preparation for NWC receives, I replaced  the relay code for NWC payments with own code.

As discussed before, `Relay` from `nostr-tools` was cumbersome to use. It was also buggy since `relay.close()` throws `websocket already in CLOSING state` unexpectedly.

The custom abstraction over `window.WebSocket` in this PR makes our interactions with nostr relays easier.

For example, the relay API from `nostr-tools` is focused on streaming events. However, we usually simply want to retrieve all existing events for a filter or publish an event and wait for a response. Before, we had to wrap `relay.subscribe` calls with promises to achieve the behavior we want. Now, `relay.fetch` handles promise resolution (and rejection). It simply returns all events it has seen until EOSE or the first event after EOSE if EOSE was the first received message.

Callers can now be concerned with less as can be seen by the reduction of relay code in `testConnectClient` and `sendPayment` of NWC. 

TODO:
- [x] replace NWC code that uses `nostr-tools`
- [x] replace crossposting code that uses `nostr-tools` 
- [x] replace nip57 code that uses `nostr`

## Additional context

- [x] use [`isomorphic-ws`](https://github.com/heineiuo/isomorphic-ws) to use same code on client and server?

## Checklist

Yes

Yes, tested NWC, crossposting and nip57 zap receipts

No


ekzyis:

lib/nostr.js
```diff
@@ -15,6 +17,149 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
+      ws.removeEventListener('message', listener)
+    }
+  }
+
+  async fetch (filter, { timeout } = {}) {
+    const ws = this.ws
+
+    let listener
+    const ackPromise = new Promise((resolve, reject) => {
+      const id = crypto.randomUUID()
```
ekzyis, line 115: Mhh, NWC autowithdrawals (#1310) fail because of this error:

```
worker        | ReferenceError: crypto is not defined
worker        |     at Object.encrypt (file:///app/node_modules/nostr-tools/lib/esm/index.js:1282:19)
worker        |     at nwcCall (file:///app/wallets/nwc/index.js:47:35)
worker        |     at async createInvoice (file:///app/wallets/nwc/server.js:13:18)
worker        |     at async createInvoice (file:///app/wallets/server.js:48:23)
worker        |     at async autoWithdraw (file:///app/worker/autowithdraw.js:41:31)
worker        |     at async file:///app/worker/index.js:80:9
worker        |     at resolveWithinSeconds (/app/node_modules/pg-boss/src/manager.js:35:14)
worker        |     at <anonymous> (/app/node_modules/p-map/index.js:57:22)
```

`import crypto from 'crypto'` did not help

ekzyis: Fixed in 5515c63c

huumn:
Eventually I'd like to use one package or the other, or just yank both of them out and take only what we use.

```js
import { signId, calculateId, getPublicKey } from 'nostr'
import { finalizeEvent, nip04 } from 'nostr-tools'
import { nip19 } from 'nostr-tools'
```

PR #1306, opened 2024-08-16
Add Random link and basic query

tsmith123:
## Checklist


api/resolvers/item.js
```diff
@@ -80,6 +80,8 @@ const orderByClause = (by, me, models, type) => {
       return 'ORDER BY "Item".msats DESC'
     case 'zaprank':
       return topOrderByWeightedSats(me, models)
+    case 'random':
+      return 'ORDER BY "Item".msats DESC'
```
tsmith123, line 84: This ORDER BY is just a placeholder - can't get RAND() or NEWID() to work here. Get a 404 if the query doesn't work.

api/resolvers/item.js
```diff
@@ -398,6 +400,27 @@ export default {
...
+        case 'random':
+          items = await itemQueryWithMeta({
+            me,
+            models,
+            query: `
+              ${selectClause(type)}
+              ${relationClause(type)}
+              ${whereClause(
+                '"Item"."deletedAt" IS NULL',
+                '"Item"."weightedVotes" > 10',
```
tsmith123, line 412: This probably shouldn't go here?

Also, 10 is just a random number as I don't know what it should be. It seems that no results are return if I set it to about 25.

api/resolvers/item.js
```diff
@@ -398,6 +400,27 @@ export default {
...
+                type === 'posts' && '"Item"."subName" IS NOT NULL',
+                subClause(sub, 5, subClauseTable(type), me, showNsfw),
+                typeClause(type),
+                await filterClause(me, models, type),
```
tsmith123, line 416: Stripped out the `whenClause`

api/resolvers/item.js
```diff
@@ -398,6 +400,27 @@ export default {
             orderBy: orderByClause(by || 'zaprank', me, models, type)
           }, ...whenRange(when, from, to || decodedCursor.time), decodedCursor.offset, limit, ...subArr)
           break
+        case 'random':
```
tsmith123, line 403: Copied `top` query and tweaked...

api/resolvers/item.js
```diff
@@ -398,6 +400,27 @@ export default {
...
+                subClause(sub, 5, subClauseTable(type), me, showNsfw),
+                typeClause(type),
+                await filterClause(me, models, type),
+                muteClause(me))}
+              ${orderByClause(by || 'random', me, models, type)}
+              OFFSET $3
+              LIMIT $4`,
+            orderBy: orderByClause(by || 'random', me, models, type)
+          }, ...whenRange(when, from, to || decodedCursor.time), decodedCursor.offset, limit, ...subArr)
+          break
```
tsmith123, line 423: **Questions:**
Why is there a `whenClause` and then a `whenRange`?
Why are there two `orderBy`s?

I can't find what $1, $2 etc so left them as they are from `top` query

tsmith123: So how come $3 and $4 can mean different things depending on the query?

huumn: These are parameterized queries and the numbers correspond to parameter position: https://minervadb.xyz/how-parameterized-queries-and-bind-parameters-are-implemented-in-postgresql/

api/resolvers/item.js
ekzyis: `RAND()` is a MySQL thing, PostgreSQL uses `RANDOM()`, see https://www.postgresql.org/docs/13/functions-math.html

api/resolvers/item.js
tsmith123: Thanks for that. I actually am aware of names params but my question is more specific to the context of the app. I can't see how the params are being passed so that limit can correspond to $3 for one query and $4 for another one.

api/resolvers/item.js
tsmith123: Ah right. I'll try this.

api/resolvers/item.js
huumn: It depends on where you put `limit` in the parameter list.

```js
}, ...whenRange(when, from, to || decodedCursor.time), decodedCursor.offset, limit, ...subArr)
```

Above, limit is the `$4`th parameter because `whenRange` produces two paramters.

api/resolvers/item.js
tsmith123: Yeah this works 👍

PR #1307, opened 2024-08-16
Msats to sats floor

### LINKED ISSUE #1303: Lack of msats support from LNURL providers
**Description**

My SN wallet is above the limit I have set. Went to the attach wallet config page for it and I see many log messages like this.

`13m	[lnAddr]	error	creating invoice for SN: autowithdrawal failed: Error: invoice has incorrect amount`

My wallet is a lightning address type of wallet. This seems to have started about 19 hours ago.

**Steps to Reproduce**

Set up an Alby wallet with lightning address and set your wallet limit. Now receive sats above the limit.

**Expected behavior**

huumn:
fixes #1303

Mostly this floors any msats passed to lnaddr's `createInvoice`. This change also required modify `wrapInvoice` to take the served invoice's msats as a parameter (as it can't be induced anymore). For p2p zaps, the excess msats effectively get paid to SN.

### DIFF
--- a/wallets/cln/server.js
+++ b/wallets/cln/server.js
@@ -3,7 +3,7 @@ import { createInvoice as clnCreateInvoice } from '@/lib/cln'
 export * from 'wallets/cln'
 
 export const testConnectServer = async ({ socket, rune, cert }) => {
-  return await createInvoice({ msats: 1, expiry: 1, description: '' }, { socket, rune, cert })
+  return await createInvoice({ msats: 1000, expiry: 1, description: '' }, { socket, rune, cert })
 }
 
 export const createInvoice = async (
--- a/wallets/server.js
+++ b/wallets/server.js
@@ -53,7 +53,21 @@ export async function createInvoice (userId, { msats, description, descriptionHa
 
       const bolt11 = await parsePaymentRequest({ request: invoice })
       if (BigInt(bolt11.mtokens) !== BigInt(msats)) {
-        throw new Error(`invoice has incorrect amount ${bolt11.mtokens} !== ${msats}`)
+        if (BigInt(bolt11.mtokens) > BigInt(msats)) {
+          throw new Error(`invoice is for an amount greater than requested ${bolt11.mtokens} > ${msats}`)
+        }
+        if (BigInt(bolt11.mtokens) === 0n) {
+          throw new Error('invoice is for 0 msats')
+        }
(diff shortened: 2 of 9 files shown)


wallets/cln/server.js
```diff
@@ -3,7 +3,7 @@ import { createInvoice as clnCreateInvoice } from '@/lib/cln'
 export * from 'wallets/cln'
 
 export const testConnectServer = async ({ socket, rune, cert }) => {
-  return await createInvoice({ msats: 1, expiry: 1, description: '' }, { socket, rune, cert })
+  return await createInvoice({ msats: 1000, expiry: 1, description: '' }, { socket, rune, cert })
```
huumn, line 6: I also made the test invoice less likely to be rejected for msats reasons.

wallets/wrap.js
```diff
@@ -145,13 +155,13 @@ export default async function wrapInvoice (bolt11, { description, descriptionHas
 
     // validate the fee budget
     const minEstFees = toPositiveNumber(routingFeeMsat)
-    const outgoingMaxFeeMsat = Math.ceil(outgoingMsat * MAX_FEE_ESTIMATE_PERCENT)
+    const outgoingMaxFeeMsat = Math.ceil(msats * MAX_FEE_ESTIMATE_PERCENT)
```
huumn, line 158: This was always intended to be 2.5% of the incoming amount, so now it is.

wallets/server.js
```diff
@@ -53,7 +53,21 @@ export async function createInvoice (userId, { msats, description, descriptionHa
...
+        }
+        if (BigInt(msats) - BigInt(bolt11.mtokens) >= 1000n) {
+          throw new Error(`invoice has a different satoshi amount ${bolt11.mtokens} !== ${msats}`)
+        }
+
+        await addWalletLog({
+          wallet,
+          level: 'INFO',
+          message: `wallet does not support msats so we floored ${msats} msats to nearest sat ${BigInt(bolt11.mtokens)} msats`
+        }, { models })
```
ekzyis, line 70: I like this but I noticed we only have wallet logs if the user invoice creation fails but not if the autowithdrawal itself fails. So then only this log will show up in the logs which looks confusing.

But that's unrelated to this PR, that was already the case since #1178. I'll fix this in another PR.

ekzyis (approved):
Looks good, only left two comments but not blocking

PR #1309, opened 2024-08-17
Generate more server wallet code

ekzyis:
## Description

This PR makes it easier to add server wallets by generating more boilerplate plumbing code:

* generate `type WalletLND { ... }` etc.
* generate `union WalletDetails = WalletLND | ...`
* I could unfortunately not see a way to generate `__resolveType` with its fourth argument as mentioned [here](https://github.com/stackernews/stacker.news/pull/1278/files#r1707709821) so I left a comment there

<details>
<summary>info object</summary>

```
GraphQLUnionType {
  name: 'WalletDetails',
  description: undefined,
  resolveType: [Function: __resolveType],
  extensions: [Object: null prototype] {},
  astNode: {
    kind: 'UnionTypeDefinition',
    description: undefined,
    name: { kind: 'Name', value: 'WalletDetails' },
    directives: [],
    types: [ [Object], [Object], [Object], [Object] ]
  },
  extensionASTNodes: [],
  _types: [
    GraphQLObjectType {
      name: 'WalletLND',
      description: undefined,
      isTypeOf: undefined,
      extensions: [Object: null prototype] {},
      astNode: [Object],
      extensionASTNodes: [],
      _fields: [Object: null prototype],
      _interfaces: []
    },
    GraphQLObjectType {
      name: 'WalletCLN',
      description: undefined,
      isTypeOf: undefined,
      extensions: [Object: null prototype] {},
      astNode: [Object],
      extensionASTNodes: [],
      _fields: [Object: null prototype],
      _interfaces: []
    },
    GraphQLObjectType {
      name: 'WalletLightningAddress',
      description: undefined,
      isTypeOf: undefined,
      extensions: [Object: null prototype] {},
      astNode: [Object],
      extensionASTNodes: [],
      _fields: [Object: null prototype],
      _interfaces: []
    },
    GraphQLObjectType {
      name: 'WalletLNbits',
      description: undefined,
      isTypeOf: undefined,
      extensions: [Object: null prototype] {},
      astNode: [Object],
      extensionASTNodes: [],
      _fields: [Object: null prototype],
      _interfaces: []
    }
  ]
}

```

</details>

* I was able to generate wallet fragments (`... on WalletLND { ... }` etc.) but that broke client imports because `fragments/wallet` is imported on the client and I imported `wallets` from `wallets/server` in there so I also only left a comment

## Additional Context

[...]

PR #1310, opened 2024-08-17
Support receiving with NWC

ekzyis:
## Description

This PR adds receiving to NWC.

Based on #1302, #1309

## Checklist


ekzyis:

wallets/nwc/index.js
```diff
@@ -1,12 +1,26 @@
+import { Relay } from '@/lib/nostr'
+import { parseNwcUrl } from '@/lib/url'
 import { nwcSchema } from '@/lib/validate'
+import { finalizeEvent, nip04 } from 'nostr-tools'
 
 export const name = 'nwc'
 
 export const fields = [
   {
     name: 'nwcUrl',
```
ekzyis, line 10: Kept this the same name to avoid having to migrate from the old name for existing wallets

huumn:
It'd be nice if these permission strings, not just nwc but all of them, allowed us to remove permissions from them somehow. Like the permissions could be aggregated and disaggregated. I'm not sure cryptography is there yet though.

ekzyis:
> It'd be nice if these permission strings, not just nwc but all of them, allowed us to remove permissions from them somehow. Like the permissions could be aggregated and disaggregated. I'm not sure cryptography is there yet though.

So you mean like [macaroons](https://github.com/lightningnetwork/lnd/blob/master/docs/macaroons.md) except the receiver can also add caveats? I don't see a way without the wallet cooperating.

I think the ideal UX would be that we tell the wallet which permissions we want and then the user can confirm, so basically what [NWA](https://github.com/nostr-protocol/nips/pull/851) is about.

huumn:
Yep, I meant I'd like to do something that's likely infeasible. I agree, NWA is probably the best thing we've got for this.

ekzyis:
This is ready for review now. I've tested it by disabling following checks:

```diff
diff --git a/lib/validate.js b/lib/validate.js
index c2b85c5e..0fc9d094 100644
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -717,14 +717,14 @@ export const nwcSchema = object().shape({
   nwcUrl: string().nwcUrl().when(['nwcUrlRecv'], ([nwcUrlRecv], schema) => {
     if (!nwcUrlRecv) return schema.required('required if connection for receiving not set')
     return schema.test({
-      test: nwcUrl => nwcUrl !== nwcUrlRecv,
+      test: nwcUrl => true, // nwcUrl !== nwcUrlRecv,
       message: 'connection for sending cannot be the same as for receiving'
     })
   }),
   nwcUrlRecv: string().nwcUrl().when(['nwcUrl'], ([nwcUrl], schema) => {
     if (!nwcUrl) return schema.required('required if connection for sending not set')
     return schema.test({
-      test: nwcUrlRecv => nwcUrlRecv !== nwcUrl,
+      test: nwcUrlRecv => true, // nwcUrlRecv !== nwcUrl,
       message: 'connection for receiving cannot be the same as for sending'
     })
   })
diff --git a/wallets/nwc/server.js b/wallets/nwc/server.js
index 80393b2f..f9a49192 100644
--- a/wallets/nwc/server.js
+++ b/wallets/nwc/server.js
@@ -10,10 +10,10 @@ export async function testCreateInvoice ({ nwcUrlRecv }) {
     throw new Error('make_invoice not supported')
   }

-  const unrestricted = await withTimeout(hasMethod(nwcUrlRecv, 'pay_invoice'), timeout)
-  if (unrestricted) {
-    throw new Error('pay_invoice must not be supported')
-  }
+  // const unrestricted = await withTimeout(hasMethod(nwcUrlRecv, 'pay_invoice'), timeout)
+  // if (unrestricted) {
+  //   throw new Error('pay_invoice must not be supported')
+  // }

   return await withTimeout(createInvoice({ msats: 1000, expiry: 1 }, { nwcUrlRecv }), timeout)
 }
```

This allows us to use the same connection string from the NWC container twice and use a connection for receiving even though it supports `pay_invoice`.

wallets/index.js
```diff
@@ -77,17 +77,7 @@ export function useWallet (name) {
...
-    // testConnectClient is optional since validation might happen during save on server
-    // TODO: add timeout
-    let validConfig
-    try {
-      validConfig = await wallet.testConnectClient?.(newConfig, { me, logger })
-    } catch (err) {
-      logger.error(err.message)
-      throw err
-    }
-    await saveConfig(validConfig ?? newConfig, { logger })
```
ekzyis, line 90: This ran always if `testConnectClient` existed even if no client config was specified.

Moving it into the if block where the client config actually gets saved fixes that.

wallets/index.js
```diff
@@ -233,10 +230,16 @@ function isConfigured ({ fields, config }) {
...
+  let val = fields.every(f => {
+    return f.optional ? true : !!config?.[f.name]
   })
 
+  // however, a wallet is not configured if all fields are optional and none are set
+  // since that usually means that one of them is required
+  if (val && fields.length > 0) {
+    val = !(fields.every(f => f.optional) && fields.every(f => !config?.[f.name]))
+  }
+
```
ekzyis, line 242: After all fields were marked as optional, NWC was marked as configured and thus a detach button was shown. This code prevents that.

huumn (changes requested):
I can't nwc to attach for receiving. It keeps timing out regardless of the timeout I set. Alby hub says it connected though sometime after timeout.

ekzyis:
I was able to reproduce the timeout issue with Alby Hub. There seems to be something wrong with how the responses are relayed (relay or Hub error) since we sometimes receive the response and sometimes not. I reached out to Alby for support.

Recordings: [video 1](https://files.ekzy.is/public/sn/2024-08-20%2018-06-47.mp4) [video 2](https://files.ekzy.is/public/sn/2024-08-20%2018-21-43.mp4)

wallets/index.js
```diff
@@ -197,6 +187,13 @@ function useConfig (wallet) {
       }
 
       if (valid) {
+        try {
+          await wallet.testSendPayment?.(newConfig, { me, logger })
```
huumn, line 191: For LNC, `testSendPayment` returns diffie hellman keys that need to be saved. You deleted this logic on line 80.

huumn (changes requested):

wallets/index.js
huumn: I went ahead and fixed it.

Issue #864, opened 2024-02-20
Login with Nostr NIP-46


nostrband:
**Is your feature request related to a problem? Please describe.**
I don't like using an extension to store Nostr keys, and extension doesn't work on mobile. 

**Describe the solution you'd like**
Adding support for Nostr NIP-46 (remote key access protocol) could solve the problem.

**Describe alternatives you've considered**
Implementing the NIP46 protocol could be an option. Simpler solution is to use a nostr library - nostr-tools and ndk both have nip46 support. Even simpler one is to use a nip-46 client implementation - nostr-login and window.nostr.js both provide a proxy object that converts nip07 extension calls to nip46 calls.

lungfulmeme:
implementation of nsec.app is really sleek and simple. would enjoy using this for SN

ekzyis:
This look like what I always wanted for nostr login. Thanks for the ticket!

brugeman:
> This might also solve #727 for nostr auth in a PWA:

It should, I use a nostr pwa app with nip46 pretty nicely.

riccardobl:
Hey, is anyone working on this? If not, i can do it :+1:

huumn:
Nope, I'll assign you!

PR #1311, opened 2024-08-18
NIP-46 auth

riccardobl:
## Description

Solves https://github.com/stackernews/stacker.news/issues/864 .
This PR updates the behavior of the "Login with Nostr" button by adding an intermediary page that presents the following login options:
- **Login with extension**:  the current behavior using the browser extension eg. alby
- **Login by pasting a bunker:// token**
- **Login by using a nip-05 address**
- **Login by copying or scanning the nostrconnect:/// url with the signer**

## Screenshots
Screenshot from 2024-08-18 11-34-53

Screenshot from 2024-08-18 11-35-01

image

image

image

## Additional Context

- I've updated the nostr-tools dependency to the latest version

Still wip

## Checklist

yes

wip

no


huumn:

lib/nostr.js
```diff
@@ -16,24 +20,52 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
+        }, 5000)
+      }
+    }
+
+    ws.onopen = () => {
+      // Resubscribe to all subscriptions on reconnect
+      for (const sub of this.subs) {
+        ws.send(JSON.stringify(['REQ', sub.id, ...sub.filters]))
+      }
+    }
```
ekzyis, line 68: Mhh, if you add these listeners conditionally, I think we don't need store `this.reconnect` as state. This is the only place you use it so you could consume it immediately when it's passed into `static connect` without any assignment to `this`. For example like this:

```diff
diff --git a/lib/nostr.js b/lib/nostr.js
index 7af160cf..c4b5bbe4 100644
--- a/lib/nostr.js
+++ b/lib/nostr.js
@@ -68,9 +68,15 @@ export class Relay {
     }
   }

-  static async connect (url, { timeout } = {}) {
+  static async connect (url, { reconnect, timeout } = {}) {
     const relay = new Relay(url)
     await relay.waitUntilConnected({ timeout })
+
+    if (reconnect) {
+      // add reconnect logic to websocket
+      // ...
+    }
+
     return relay
   }
```

This might sound petty but I just wrote this code so I'd like to keep it as stateless as possible, lol

lib/nostr.js
```diff
@@ -16,24 +20,52 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
+    this.url = relayUrl
+    this.subs = []
+    this.reconnectOnClose = reconnectOnClose
+    this._closed = false
+    this.error = undefined
+    this.reconnect()
+  }
 
-    ws.onmessage = function (msg) {
+  reconnect () {
```
ekzyis, line 38: This function does more than just handle `reconnect`. The listeners for notices or errors should be added in the constructor.

lib/nostr.js
```diff
@@ -16,24 +20,52 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
       console.error('websocket error: ' + err)
       this.error = err
     }
 
-    this.ws = ws
+    ws.onclose = () => {
+      if (this.reconnectOnClose && !this._closed) {
+        setTimeout(() => {
+          this.reconnect()
+        }, 5000)
```
ekzyis, line 59: Mhh, this only attempts to reconnect once. Is this intended?

I also initially thought there might be a listener leak here because `reconnect` adds listeners again but  I guess since you assign to `onclose`, `onopen` etc. that's not the case. You overwrite them.

lib/nostr.js
```diff
@@ -16,24 +20,52 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
+      }
+    }
+
+    ws.onopen = () => {
+      // Resubscribe to all subscriptions on reconnect
+      for (const sub of this.subs) {
+        ws.send(JSON.stringify(['REQ', sub.id, ...sub.filters]))
+      }
```
ekzyis, line 67: Should this run even if reconnecting was not enabled?

lib/nostr.js
```diff
@@ -76,6 +112,12 @@ export class Relay {
     if (state !== WebSocket.CLOSING && state !== WebSocket.CLOSED) {
       this.ws.close()
     }
+    for (const sub of this.subs) {
+      if (sub.closed) continue
+      sub.reject(new Error('relay closed'))
+    }
+    this.subs = []
+    this._closed = true
```
ekzyis, line 120: I believe you only need this to not run the reconnect logic on close? You could also simply remove the listener if you don't want it to get run, no need for `this._closed`.

_I also wonder if this line should be before `this.ws.close()` :thinking:_

ekzyis (commented):
Hi, thanks for the PR and integrating it with my relay code!

I left some suggestions regarding the reconnect logic.

Will look at `Relay.subscribe` and the rest tomorrow.

lib/nostr.js
riccardobl: afaik you cannot reconnect an 'isomorphic-ws' , so this function is recreating the socket on reconnection, and for this reason it needs to add the listener to the new websocket there.

lib/nostr.js
riccardobl: this reconnects everytime the socket is closed,  on reconnection a new socket is always created, if it is closed again a new socket will be created again an so forth

lib/nostr.js
riccardobl: i could add an if there if you want, but unless there is a reconnection this.subs will be empty so the for loop will exit immediately

lib/nostr.js
riccardobl: yes, you are right, it should go before this.ws.close().
I've also added a check inside the reconnection timeout, just in case it is set right before some code calls close().

this._closed is used in `get closed ` and `get connected` to make the relay look closed and disconnected even if the socket is still closing underneath.

lib/nostr.js
riccardobl: the ws field should be considered private imo since the relay class is an abstraction over the use of websockets, so it wouldn't be a good practice to 'leak' it outside of the class like so

lib/nostr.js
```diff
@@ -16,24 +20,52 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
 ]
+export const DEFAULT_NIP46_RELAYS = [
+  'wss://relay.damus.io/',
+  'wss://relay.snort.social/'
+]
 
 export class Relay {
-  constructor (relayUrl) {
-    const ws = new WebSocket(relayUrl)
+  constructor (relayUrl, reconnectOnClose = false) {
```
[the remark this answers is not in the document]
riccardobl: > Can you make this an object parameter with empty object as default? I would also name it simply reconnect.

done

lib/nostr.js
riccardobl: i made a different change : https://github.com/stackernews/stacker.news/pull/1311/commits/a8d572ba452ac01b310fae9e660dfce5cbf9d9c4 to not set this.reconnectOnClose in the constructor, maybe this can be a compromise?

lib/nostr.js
ekzyis: Ah, I see. Makes sense then!

components/nostr-auth.js
```diff
@@ -1,32 +1,291 @@
...
+              handleError(e)
+            }
+          }}
+        >
+          {text || 'Login'} with extension
+        </Button>
+      </Row>
+      <Row className='w-100 mt-4 text-muted small'>
+        <AccordianItem
+          header='Which NIP-46 signed can I use?'
```
ekzyis, line 257:
```suggestion
          header='Which NIP-46 signers can I use?'
```

ekzyis (commented):
Hey @riccardobl, sorry for the delayed review.

I tested your code today with [nsec.app](https://nsec.app/) and pasted the `bunker://` string into SN and it works well! I haven't tested `nostrconnect://` though. I expected clicking on the QR code would open nsec.app but it didn't. Guess it doesn't register for this scheme?

Anyway, looking through the code, I have two questions:

1. Have you looked into the NIP-46 implementation from NDK[^1]? Maybe I missed something or forgot the reason, but I wonder why we don't simply use NDK's implementation. That should work just as well with less lines of code we have to maintain ourselves, no?

2. I am currently confused why you need long-living connections with a subscription pool (`this.subs`) for your implementation. Was it not possible to implement NIP-46 with the existing implementation of `relay.fetch`? I wrote it for NWC and afaict, NIP-46 has a similar request-response flow so I wonder why it needed changes. Is it because we need to handle connections to multiple relays at once? That's something my relay code is lacking.

[^1]: https://github.com/nostr-dev-kit/ndk/tree/master/ndk/src/signers/nip46

lib/nostr.js
ekzyis: Ahh, makes sense

components/nostr-auth.js
```diff
@@ -1,32 +1,291 @@
...
+      if (signer) {
+        signer.close()
+      }
+    }
+  }, [signer])
+
+  // initialize signer
+  useEffect(() => {
+    if (!data) return
+    (async () => {
```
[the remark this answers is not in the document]
riccardobl: true, it is a left over from a previous version of the code. Removed.

riccardobl:
> I expected clicking on the QR code would open nsec.app but it didn't. Guess it doesn't register for this scheme?

Clicking or scanning the QR is for the [nip-46 nostrconnect:// flow](https://github.com/nostr-protocol/nips/blob/master/46.md#direct-connection-initiated-by-the-client) , currently this is not supported by many signers. The only one that i've found that is production looking is [nostrum](https://github.com/nostr-connect/nostrum/) , but it seems it diverges from the spec ( https://github.com/nostr-connect/nostrum/issues/22 ) so it doesn't work properly atm. 
I made sure to implement the full nip-46 despite the current status of the various signers.

> Have you looked into the NIP-46 implementation from NDK[1](https://github.com/stackernews/stacker.news/pull/1311#user-content-fn-1-4c7a1391edf1a30ea3d8399142d8f36d)? Maybe I missed something or forgot the reason, but I wonder why we don't simply use NDK's implementation. That should work just as well with less lines of code we have to maintain ourselves, no?

Mainly because it is not a dependency that is currently available in stacker news and it didn't feel right to add a new dependency just for this, also ndk doesn't support nostrconnect:// flow from what i can tell

> I am currently confused why you need long-living connections with a subscription pool (this.subs) for your implementation. Was it not possible to implement NIP-46 with the existing implementation of relay.fetch? I wrote it for NWC and afaict, NIP-46 has a similar request-response flow so I wonder why it needed changes. Is it because we need to handle connections to multiple relays at once? That's something my relay code is lacking.

Because we would need to re-subscribe at least 3 times to the same filter to use the fetch logic, it would be slower and provide no benefit imo, i think it makes sense to use fetch where you don't have a back and forth communication

ekzyis (approved):
Hi @riccardobl, thanks again for your PR!

We've decided to merge it into an intermediate branch[^1] since we're looking into using only [NDK](https://github.com/nostr-dev-kit/ndk) for all our nostr code instead of the library mess of [`nostr-tools`](https://github.com/nbd-wtf/nostr-tools) and [`nostr`](https://github.com/jb55/nostr-js) that we currently have. We will do this ourselves in this branch before merging it into `master` since it's out of the scope of this PR. Your changes look good and you couldn't have known that we will be looking into NDK (we didn't even know that ourselves before I reviewed this PR) so it's okay that you didn't use NDK.

I will approve this PR and @huumn will send you your sats for #864. Looking forward to more contributions from you!

[^1]: I will merge the bugfixes you provided for existing relay code separately into `master`

PR #1320, opened 2024-08-20
disable freebies setting

huumn:
Introduces a setting for disabling freebies. By default, this is null. When the first sending wallet is attached, `disableFreebies` is set to true IFF `disableFreebies` is null, otherwise the setting is completely controlled by the stacker.

### DIFF
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -628,10 +628,10 @@ export default {
       }
 
       // disable freebies if it hasn't been set yet
-      models.user.update({
+      await models.user.update({
         where: { id: me.id, disableFreebies: null },
         data: { disableFreebies: true }
-      }).catch()
+      })
 
       return true
     },
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -622,6 +622,19 @@ export default {
   },
 
   Mutation: {
+    disableFreebies: async (parent, args, { me, models }) => {
+      if (!me) {
+        throw new GraphQLError('you must be logged in', { extensions: { code: 'UNAUTHENTICATED' } })
+      }
+
+      // disable freebies if it hasn't been set yet
+      models.user.update({
+        where: { id: me.id, disableFreebies: null },
+        data: { disableFreebies: true }
+      }).catch()
+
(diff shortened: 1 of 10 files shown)


api/resolvers/user.js
```diff
@@ -622,6 +622,19 @@ export default {
...
+    disableFreebies: async (parent, args, { me, models }) => {
+      if (!me) {
+        throw new GraphQLError('you must be logged in', { extensions: { code: 'UNAUTHENTICATED' } })
+      }
+
+      // disable freebies if it hasn't been set yet
+      models.user.update({
+        where: { id: me.id, disableFreebies: null },
+        data: { disableFreebies: true }
+      }).catch()
```
ekzyis, line 634: maybe better to await this since we don't await on the client so the client could print the error?

pages/settings/index.js
```diff
@@ -252,6 +253,20 @@ export default function Settings ({ ssrData }) {
...
           />
+          <Checkbox
+            label={
+              <div className='d-flex align-items-center'>disable freebies
+                <Info>
+                  <p>Some posts and comments can be created without paying. However, that content has limited visibility.</p>
+
+                  <p>If you disable freebies, you will always be prompted to pay for your posts and comments.</p>
+
+                  <p>If you attach a sending wallet, we disable freebies for you unless you have unchecked this box.</p>
```
ekzyis, line 264: Mhh, the box is already unchecked by default (`disableFreebies = NULL`). So wouldn't "interacted" make more sense?

> If you attach a sending wallet, we disable freebies for you unless you interacted with this box and saved.

ekzyis (approved):
Tested and works well. Left some suggestions but can be ignored.

api/resolvers/user.js
huumn: I call this in `enablePayments` though. I thought erroring about a side effect might be weird.

ekzyis: Yes but you don't await it there, you only log errors so it wouldn't error on error

huumn: Good point

PR #1321, opened 2024-08-21
Rename LNbits primary key column from 'int' to 'id'

ekzyis:
## Description

This didn't cause any bugs because we use the id of the `Wallet` table.

## Checklist

Yes, this shouldn't cause any downtime

Yes

No

### DIFF
--- a/prisma/migrations/20240821072645_rename_lnbits_pk/migration.sql
+++ b/prisma/migrations/20240821072645_rename_lnbits_pk/migration.sql
@@ -9,7 +9,6 @@
 ALTER TABLE "WalletLNbits" RENAME COLUMN "int" TO "id";
 
 UPDATE "Wallet"
-SET wallet = to_jsonb(NEW)
-WHERE id IN (
-  SELECT "walletId" FROM "WalletLNbits"
-)
+SET wallet = to_jsonb("WalletLNbits")
+FROM "WalletLNbits"
+WHERE "Wallet".id = "WalletLNbits"."walletId";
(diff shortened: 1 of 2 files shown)


prisma/migrations/20240821072645_rename_lnbits_pk/migration.sql
```diff
@@ -0,0 +1,15 @@
...
+
+*/
+-- AlterTable
+ALTER TABLE "WalletLNbits" RENAME COLUMN "int" TO "id";
+
+UPDATE "Wallet"
+SET wallet = to_jsonb(NEW)
+WHERE id IN (
+  SELECT "walletId" FROM "WalletLNbits"
+)
```
huumn, line 15: Is this valid sql? What does NEW refer to in an update statement?

huumn: It wasn't. I fixed it.

huumn:
> Did you QA this? Could we deploy this straight to production? Please answer below:
>
> Yes

👀 It's important we're able to trust this answer.

---------------

I've thought about changing this to "On a scale of 1-10 how well have your QA'd this change and any features it might affect? How did you QA this?"

prisma/migrations/20240821072645_rename_lnbits_pk/migration.sql
ekzyis: Whoops, I remember I ran an update query manually but must have forgotten to check what I committed

ekzyis:
> I've thought about changing this to "On a scale of 1-10 how well have you QA'd this change and any features it might affect? How did you QA this?"

Good idea!

PR #1325, opened 2024-08-22
Show item details via context menu

ekzyis:
## Description

This PR fixes the accessibility issue that `item.meSats`, `item.upvotes` or the exact timestamp is only visible on hover which isn't available on mobile. It introduces `details` as a new entry in the item context menu. On click, it opens this modal:

2024-08-21-191009_519x272_scrot

I also added `item.cost` for everyone to see even though it was only visible to the author before since I don't consider it sensitive. 

## Screenshots

Mobile:

localhost_3000_(iPhone SE)

## Checklist

Yes, I only additionally resolve `item.cost` in item queries now which isn't a breaking change

Yes, tested logged in and logged out.

Yes, see screenshot

No

### DIFF
--- a/components/item-info.js
+++ b/components/item-info.js
@@ -294,7 +294,7 @@ function InfoDropdownItem ({ item }) {
               <div>{item.meSats}</div>
             </>
           )}
-          <div>upvotes</div>
+          <div>zappers</div>
           <div>{item.upvotes}</div>
         </div>
       )
--- a/components/item-info.js
+++ b/components/item-info.js
@@ -25,6 +25,7 @@ import UserPopover from './user-popover'
 import { useQrPayment } from './payment'
 import { useRetryCreateItem } from './use-item-submit'
 import { useToast } from './toast'
+import { useShowModal } from './modal'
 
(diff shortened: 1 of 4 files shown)


components/item-info.js
```diff
@@ -269,3 +271,39 @@ export default function ItemInfo ({
...
+          <div>{item.cost}</div>
+          <div>sats</div>
+          <div>{item.sats}</div>
+          {me && (
+            <>
+              <div>sats from me</div>
+              <div>{item.meSats}</div>
+            </>
+          )}
+          <div>upvotes</div>
```
huumn, line 297:
```suggestion
          <div>zappers</div>
```

components/item.module.css
```diff
@@ -154,6 +154,12 @@ a.link:visited {
     grid-template-columns: auto minmax(0, 1fr);
 }
 
+.details {
+    display: grid;
```
huumn, line 158: nice! grids are da best

huumn (approved):
Just want to not use the term `upvotes`

PR #1327, opened 2024-08-23
Skip wallets tests on priority update

ekzyis:
## Description

This fixes that priority updates reuse the wallet save logic which means that the wallet gets unnecessarily tested again.

## Additional context

I didn't create a new mutation because `saveConfig` contains the logic to determine if client and/or server needs to be updated. I therefore only added a flag `priorityOnly` that then skips the wallet tests (`testSendPayment`, `testCreateInvoice`). It does not skip validation since validation is used to check if we should save the configuration on the client or server. Since all saved configurations are already valid this shouldn't ever be a problem when only priority is changed.

## Checklist

Yes, very backwards compatible, much wow

9\. I tested changing priority between three wallets: LNC only sending, WebLN no config, NWC send+recv.

I also read through the code and the changes are quite simple: I basically only skip `testSendPayment` and `testCreateInovice`. Everything else stayed the same. 

### DIFF
--- a/wallets/index.js
+++ b/wallets/index.js
@@ -71,7 +71,7 @@ export function useWallet (name) {
   const setPriority = useCallback(async (priority) => {
     if (_isConfigured && priority !== config.priority) {
       try {
-        await saveConfig({ ...config, priority }, { logger })
+        await saveConfig({ ...config, priority }, { logger, priorityOnly: true })
       } catch (err) {
         toaster.danger(`failed to change priority of ${wallet.name} wallet: ${err.message}`)
       }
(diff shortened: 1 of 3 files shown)


wallets/index.js
```diff
@@ -161,13 +161,16 @@ function useConfig (wallet) {
...
+    const { enabled, priority } = config || {}
     config = {
       ...config,
       ...serverConfig
     }
     // wallet is enabled if enabled is set in client or server config
     config.enabled ||= enabled
+    // priority might only be set on client or server
+    // ie. if send+recv is available but only one is configured
+    config.priority ||= priority
```
ekzyis, line 173: if a wallet had send+recv available but only spend was configured, `priority` from the server config was undefined and overrode the priority set on the client.

PR #1333, opened 2024-08-26
Improve freebies

huumn:
## Description
- removes freebie posts (and removed from territory form)
- ranks unfreed freebie and deleted comments at bottom regardless of sort (e.g. recent)

## Additional Context

I left the `allowFreebies` territory setting for backwards compat.

## Checklist

Should be.

`6`: I didn't test unarchiving territories or anon posts/comments.

n/a

no


ekzyis (approved):
Looks good, I tested creating territories, creating freebie and non-freebie comments logged in, anon posts and comments and sorting.

PR #1356, opened 2024-09-03
improve dupes

### LINKED ISSUE #1344: Convert item `dupes` resolver for links to use regex instead of `SIMILAR TO`
Regex plays better with postgresql indices and we should be able to express our query in regex anyway.

huumn:
closes #1344

## Description

This makes dupe check use regex and adds an index on `url`.

Additionally:
- improves dupe check loading ui/ux
- debounces dupe check calls

## Checklist

8. I tested an article url and both youtube and hn urls and variations.

### DIFF
--- a/components/link-form.js
+++ b/components/link-form.js
@@ -16,6 +16,7 @@ import { useMe } from './me'
 import { ItemButtonBar } from './post'
 import { UPSERT_LINK } from '@/fragments/paidAction'
 import useItemSubmit from './use-item-submit'
+import useDebounceCallback from './use-debounce-callback'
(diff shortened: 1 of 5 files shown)


components/link-form.js
```diff
@@ -39,9 +42,7 @@ export function LinkForm ({ item, sub, editThreshold, children }) {
       dupes(url: $url) {
         ...ItemFields
       }
-    }`, {
-    onCompleted: () => setPostDisabled(false)
-  })
```
huumn, line 44: This didn't do anything

PR #1358, opened 2024-09-04
Optout of display of images and video (show them as links)

### LINKED ISSUE #610: Setting to click to reveal images
Images can make the reading experience on sn noisy. Offer a setting that prevents displaying images unless the image link is clicked on.

Details: when enabled, show a normal link to any images included in a post/comment that when clicked open up the image overlay.

huumn:
closes #610 

## Description
This introduces a new setting to render images/video as links instead of images/video. They open like a link when clicked - in a new tab.

By default and for anon, images and videos are shown.

`6` Tested image and video links anon/logged in with all combos of imgproxy only and show images and videos.

I did not test proxied images.

### DIFF
--- a/components/image.js
+++ b/components/image.js
@@ -177,7 +177,7 @@ export default function ZoomableImage ({ src, srcSet, ...props }) {
   if (!src) return null
 
   if (trustedDomain) {
-    if (me?.privates?.showImagesAndVideos !== false) {
+    if (props.tab === 'preview' || !me || me.privates.showImagesAndVideos) {
       return (
         <TrustedImage
           src={src} srcSet={srcSet}
--- a/components/image.js
+++ b/components/image.js
@@ -19,13 +19,25 @@ export function decodeOriginalUrl (imgproxyUrl) {
   return originalUrl
 }
 
-function ImageOriginal ({ src, topLevel, rel, tab, children, onClick, ...props }) {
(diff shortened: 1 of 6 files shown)


components/image.js
```diff
@@ -171,12 +177,16 @@ export default function ZoomableImage ({ src, srcSet, ...props }) {
   if (!src) return null
 
   if (trustedDomain) {
-    return (
-      <TrustedImage
-        src={src} srcSet={srcSet}
-        onClick={handleClick} onError={handleError} {...props}
-      />
-    )
+    if (me?.privates?.showImagesAndVideos !== false) {
```
ekzyis, line 180: At this point it might be easier to read `!me || me.privates.showImagesAndVideos`.

Additionally, since I've seen above that you want to show images in the preview regardless of the setting, you also need to use it here else trusted images will not render in preview if the setting is enabled.

```suggestion
    if (props.tab === 'preview' || !me || me.privates.showImagesAndVideos) {
```

_afaik, we can assume `privates` exist_

ekzyis (changes requested):
Found an inconsistency with how images are handled in preview, see suggestion.

> I did not test proxied images.

I did by letting an image render and then copy that image address which starts with `localhost:3001`. It's picked up as a trusted domain so will be handled like any other trusted domain.

Issue #1304, opened 2024-08-15
Encrypted wallet credentials device sync


ekzyis:
**Is your feature request related to a problem? Please describe.**

Spending wallets are currently stored unencrypted on client devices as this disclaimer shows:

![2024-08-15-174318_685x177_scrot](https://github.com/user-attachments/assets/38916082-d9c6-423a-b43c-3c2bcc03c64c)

We don't store them on the server because we don't want to be able to spend users' funds.

As the disclaimer mentions, a user has to reenter their credentials on other devices. This is bad UX. 

**Describe the solution you'd like**

Users can enable device sync with a checkbox, for example on /settings/wallets.

When this checkbox is enabled, we will ask them for a password or generate cryptographically secure mnemonics for them. A key pair will be derived from this secret using [`window.crypto.subtle.deriveKey()`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveKey). The secret is stored on the device (and only on the device) so it can be looked up again by the user; for example to enter it into a different device to generate the same key pair. This exchange could be done manually or with the help of QR codes.

If a client device with a key pair saves a wallet, it will encrypt with AES-GCM and attempt to save the credentials on the server. The save will fail if there are existing encrypted credentials for this wallet on the server to prevent overwrites. In this case, the credentials are only saved locally. If the save is successful, existing local (unencrypted) credentials are deleted.

If a client device with a key pair fetches a wallet, it will also fetch any existing encrypted credentials for this wallet from the server and try to decrypt them. Since AES-GCM uses authenticated encryption, decryption will graciously fail if the configuration was not encrypted with our public key. If decryption succeeds, it will use these instead of any existing local unencrypted wallet credentials. They are deleted instead.

If a client device with a key pair deletes a wallet, it will also submit the encrypted credentials (or a hash of them) to be deleted. This is done so we can compare them with existing encrypted wallet credentials on the server. If they match, they are also deleted on the server, not only the configuration for receiving.

So essentially, each wallet on the server will additionally have a single column to share encrypted data between devices with keys shared out-of-band.

This scheme not only enhances UX but also security since credentials are no longer stored unencrypted on client devices. They still store a key which could be accessed with a XSS vuln for example[^1] but an attacker would also need access to the encrypted credentials by taking over a user session (more severe than XSS) or getting access to the data on the server (also more severe than XSS).

[...]

PR #1373, opened 2024-09-07
Encrypted device sync

riccardobl:
## Description

This PR introduces a general purpose approach to end to end device sync and implements it in wallet attachments to solve https://github.com/stackernews/stacker.news/issues/1304 .

## Screenshots
image
image
image
image
image
image
image
image

## Additional Context
It is developed to be a drop-in replacement to `components/use-local-state.js`, usage is as follows:

```js
import useVaultStorageState from '@/components/use-user-vault-state'

const [value, setValue, clearValue] = useVaultStorageState('storage-key')
// ...
console.log(value)
// setValue(123)
// clearValue()

```

the useVaultStorageState hook handles everything internally and detects if the user has device sync enabled and fallbacks to localStorage if not, so there is no other code required to use this feature other than the one presented above.

useVaultStorageState needs to rely on localStorage to store the passphrase and the stored values if the user doesn't have device sync enabled. To prevent possible name conflicts it prefixs every value with "vault:" and suffixes with the userid.

## Checklist

**yes it is backward compatible and everything that will be set using `user-vault-state` and everything that was set before using wallet attachments will automatically migrate to the encrypted device sync when the user enables it**

in details:
- `user-vault-state` is backward compatible: if a key is not available with the new vault:key:userId syntax, it will try to read it  normally from the localStorage
- configured wallet attachments are automatically migrated when the user enables device sync and confirms the migration dialog
- if an user doesn't have device sync activated yet and does some configurations, all the entries set using `user-vault-state`  will be automatically migrated to device sync when the users enables it (and confirms the dialog).

8

yes

no


ekzyis (commented):
First of all, thank you for your work, this is AWESOME! #1304 is actually the most important wallet feature right now but we realized that we didn't set `priority:high` on it. We corrected this now so you actually get 2x more sats for #1304.

Regarding the code: I've played around with it and looked through the code and I like how you interfaced it nicely with the existing wallet code (just as you mentioned), nice work! I also tested syncing of wallets using Brave and Firefox and it works well on the happy path.

However, I noticed that it's possible to get the devices out of sync by resetting one. The other device shows disconnected now but is not actually reset (key still exists on device):

https://github.com/user-attachments/assets/736f680b-235d-4919-84ef-014ea8a44a8d

If you now have a wallet configured that will get synced across the devices, decryption errors are thrown on the device that wasn't reset and is now "connected" with a wrong passphrase:

https://github.com/user-attachments/assets/52c4fa6c-bab8-4587-8fcc-9ac16bfa05ca
 
Another thing I noticed is that the UI/UX could be improved. Going through the setup, I think we can simplify it. I think it's currently a bit intimidating to use since it's in a separate tab which makes it stand out and it might involve more than one step to setup because of the possible prompt for migration which sounds scary. After setup, there are three buttons in different colors.

For example, do we need this "scary" migration prompt? Does it make sense to enable this setting but _not_ migrate? And can we express the state without three separate buttons that can be overwhelming and confusing?

I think we might be able to express everything a user might want to do with a single button on the settings page in a familiar layout and colors. The flow could be similar to how API keys are implemented. I'll give it a shot.

I've left some other suggestions but they are mostly about minor things.

riccardobl:
Thank you for the review.

> However, I noticed that it's possible to get the devices out of sync by resetting one. The other device shows disconnected now but is not actually reset (key still exists on device):

This should be fixed now, i've added some code to clear the stored key if the local and remote hashes mismatch

> For example, do we need this "scary" migration prompt? Does it make sense to enable this setting but not migrate?

The only situation i could think of, where this makes sense, is when you have different conflicting settings on different devices and you want to chose which one is migrated. I think another solution is making the migration not overwrite keys, so that the first device that connects gets the precedence. I've changed the code to do that and got rid of the popup.


Also changed the ux to have a single button under general, what do you think? :

image

image

image


image

image

image

image

image



There was also some code in `wallets/index.js` that used localStorage to find the highest priority wallet and to check if it was "disabled locally" (used only in webln when the browser doesn't support it), so i had to change things around a bit.

ekzyis:
> The only situation i could think of, where this makes sense, is when you have different conflicting settings on different devices and you want to chose which one is migrated. I think another solution is making the migration not overwrite keys, so that the first device that connects gets the precedence. I've changed the code to do that and got rid of the popup.

Cool, works great. Now it's just two clicks without much to consider to enable it, that should be easy enough.

> Also changed the ux to have a single button under general, what do you think? :

Looks great, so much better!

> There was also some code in wallets/index.js that used localStorage to find the highest priority wallet and to check if it was "disabled locally" (used only in webln when the browser doesn't support it), so i had to change things around a bit.

Mhh, I was a bit worried about conditionally skipping all other hooks since that's frowned upon but seems to work fine here.

---

I've changed some code in the `useVault...` hooks in my last commits, especially in d03fda22, see commit message.

Changes worth pointing out:

* simplified `localStorage` I/O functions since they were using a lot of args that were then additionally used in the opposite order to construct the storage key. That was very confusing. See d03fda22.
* changed migration since I saw no need for the `vault:` prefix to ever exist in local storage (except for the key). See d03fda22.
* the encryption/decryption methods now always assume JSON as input and return JSON. See afad3e44.
* I realized that the WebLN wallet should not be synced across devices. I added `:local-only` as a suffix (which is then suffixed by the user id) to detect such wallets that should always be stored locally. See bc0fde51.
* renamed a bunch of stuff from `UserVault` to `Vault`. It got tiring to read `User...` all the time. Imo, it's pretty fast pretty clear that it's a vault for users, this information doesn't have to be repeated all over the place. I like short names when possible :) See 3015e8ee.

What's left to do is to fix the following bug and test some more. Maybe we can also improve UX by making it clearer that this is for the wallets (the only current use case).

TODO:

- [ ] fix priority updates via drag & drop

You can drag the wallets around to change which wallet will be used for payments. However, it seems to immediately overwrite the new value with `priority` with the old value without `priority` for some reason.

- [ ] test everything really hard (again) until no more new bugs are found (like 7cf0466e), test different combinations of usage with multiple devices (enable, connect, reset, disconnect etc.)

- [ ] make it clearer that this is to sync wallets

The wallets are never mentioned but that's currently the only use case for this vault. Maybe we should also show the button on /settings/wallets.

riccardobl:
> simplified localStorage I/O functions since they were using a lot of args that were then additionally used in the opposite order to construct the storage key. That was very confusing. See https://github.com/stackernews/stacker.news/commit/d03fda226581e15c6f9a3245b4b80c8b56d8462d.
changed migration since I saw no need for the vault: prefix to ever exist in local storage (except for the key). See https://github.com/stackernews/stacker.news/commit/d03fda226581e15c6f9a3245b4b80c8b56d8462d.

I suggest to revert this change, the prefixes and suffixes are used to know which entries in the localStorage belong to the vault, and are used to migrate them and to avoid conflicts with other localStorage calls. With this change there is the risk to overwrite user vault entries by mistake, also to make the migration work without prefixes, you locked it to be wallet only, while the original code was developed to be general purpose and backward compatible with wallets.
If you were to use the code with this change on another thing, you would have to invent your own prefix and then add it to line 79 of `use-user-vault-state.js`. While you didn't have to worry about it before.

riccardobl:
>  fix priority updates via drag & drop
> You can drag the wallets around to change which wallet will be used for payments. However, it seems to immediately overwrite the new value with priority with the old value without priority for some reason.

This was introduced in, still trying to figure out the cause
[Refactor local<>vault storage interface d03fda2](https://github.com/stackernews/stacker.news/pull/1373/commits/d03fda226581e15c6f9a3245b4b80c8b56d8462d)

wallets/index.js
```diff
@@ -48,17 +48,17 @@ export function useWallet (name) {
   const _isConfigured = isConfigured({ ...wallet, config })
 
   const enablePayments = useCallback(() => {
-    enableWalletLocally(name, me)
+    saveConfig(config => ({ ...config, enabled: true }), { skipTests: true })
```
riccardobl, line 51: My understanding was that enable/disablePayments are supposed to disable or enable a payment method for the local machine, while the state in the vault should change only by unchecking `enabled` in the wallet settings

wallets/index.js
```diff
@@ -251,9 +255,9 @@ function useConfig (wallet) {
         valid = false
       }
 
-      if (valid) await setServerConfig(newServerConfig, { priorityOnly })
+      if (valid) await setServerConfig(newServerConfig, { priorityOnly: skipTests })
     }
-  }, [hasClientConfig, hasServerConfig, setClientConfig, setServerConfig, wallet])
+  }, [config, hasClientConfig, hasServerConfig, setClientConfig, setServerConfig, wallet])
```
riccardobl, line 265: having `config` as dependency in saveConfig causes the callback to be triggered twice.
Not sure why, could be a race condition because `saveConfig` side effect is updating `config`, or could be because some code upstream is depending on the `saveConfig` state.

Removing config as dependency solves the priority issue

ekzyis:
> I suggest to revert this change, the prefixes and suffixes are used to know which entries in the localStorage belong to the vault, and are used to migrate them [...] also to make the migration work without prefixes, you locked it to be wallet only, while the original code was developed to be general purpose and backward compatible with wallets.
If you were to use the code with this change on another thing, you would have to invent your own prefix and then add it to line 79 of use-user-vault-state.js. While you didn't have to worry about it before.

Ah, I think I see where my confusion about the `vault:` prefix comes from. I think you're more concerned about _forward_ compatibility here when you say "general purpose" since new code that uses the vault will need to be aware that migration needs an update and you can't simply use `useVault` and be done. That's fair.

I didn't see this point because I thought the `vault:` entries are for backwards compatibility. I didn't understand this because for me, "migration" was simply the process to send whatever is already stored locally to the server when device sync is enabled. It's obvious you need to know what you want to send and what not so it wasn't a concern of backwards/forwards compatibility for me. But I think you saw "migration" as the process to move old data to a new schema that is then forwards compatible (`vault:` prefixes), so as a backwards compatibility thing.

I like how detached your code is from the wallets / it being general purpose. It makes sure that one does not have to keep in mind how the wallets are implemented to understand the i/o code (and vice versa) which helps a lot with maintenance, but I do believe that the complexity of the migration and i/o code nullified that maintenance benefit. I believe that

<details>

<summary>the old migration code</summary>

```js
// use to migrate the local storage to vault (do not overwrite existing vault entries)
export function useLocalStorageToVaultMigration () {
  const { me } = useMe()
  const [setVaultValue] = useMutation(SET_ENTRY)

  // count how many local storage keys can be migrated
  const count = SSR ? 0 : getMigrableStorageKeys(me?.id).length

  // migrate local storage to vault
  const migrate = async () => {
    if (SSR) return
    const vaultKey = getLocalStorage(me?.id, 'vault', { prefix: 'key' })
    if (!vaultKey) throw new Error('vault key not found')
    const keys = getMigrableStorageKeys(me?.id)
    let migratedCount = 0
    for (const key of keys) {
      const localStorageValue = window.localStorage.getItem(key.localKey)
      if (localStorageValue) {
        const encryptedValue = await encryptStorageData(vaultKey.key, localStorageValue)
        const res = await setVaultValue({ variables: { key: key.vaultKey, value: encryptedValue, skipIfSet: true } })
        if (res?.data?.setVaultEntry) {
          console.info(localStorageValue, 'migrated to vault', vaultKey.key, res)
          window.localStorage.removeItem(key.localKey)
          migratedCount++
        } else {
          console.warn(localStorageValue, 'could be migrated to vault', vaultKey.key)
        }
      }
    }
    return migratedCount
  }

  return [count, migrate]
}

function getMigrableStorageKeys (userId) {
  // get all the local storage keys that can be migrated
  if (!userId) return []
  const out = []
  const vaultPrefix = 'vault:'
  const vaultSuffix = ':' + userId
  for (const key of Object.keys(window.localStorage)) {
    if (key.startsWith(vaultPrefix) && key.endsWith(vaultSuffix)) {
      // everything that is set by the local storage vault (ie. if the user doesn't have device sync enabled yet)
      out.push({
        vaultKey: key.substring(vaultPrefix.length, key.length - vaultSuffix.length),
        localKey: key
      })
    } else if (key.startsWith('wallet:') && key.endsWith(':' + userId)) {
      // every old setting related to the wallet attachments
      out.push({
        vaultKey: key,
        localKey: key
      })
    }
    // check here for more keys that can be migrated if needed
  }
  return out
}
```

</details>

which required a) to keep in mind that the storage key that was passed into `useVault` is manipulated to always have the required `vault:` prefix and

<details>
<summary>b) use these functions for local storage</summary>

```js
function setLocalStorage (userId, key, value, prefix = 'vault') {
  window.localStorage.setItem(prefix + ':' + key + ':' + userId, JSON.stringify(value))
}

function getLocalStorage (userId, key, { prefix = 'vault', backwardCompatible = false }) {
  const v = window.localStorage.getItem(prefix + ':' + key + ':' + userId)
  if (v) {
    return JSON.parse(v)
  }
  if (backwardCompatible) {
    const vb = window.localStorage.getItem(key)
    return vb ? JSON.parse(vb) : null
  }
  return null
}

function unsetLocalStorage (userId, key, prefix = 'vault') {
  window.localStorage.removeItem(prefix + ':' + key + ':' + userId)
}
```

</details>

that were used

<details>
<summary>like this</summary>

```js
// can you tell where these are reading/writing from/to local storage
// without actively having to keep the internals in mind?
// their function signature also confused me a lot (important for maintenance).
// they looked pretty complicated for what should be an easy task:
// "read or write something from/to local storage"

getLocalStorage(me?.id, storageKey, { backwardCompatible: true })

getLocalStorage(me?.id, 'vault', { prefix: 'key' })

unsetLocalStorage(me?.id, storageKey)

setLocalStorage(me?.id, 'vault', { passphrase, key, hash }, 'key')
```

</details>

is a lot harder to understand than

<details>
<summary>the new migration code</summary>

```js
export function useVaultMigration () {
  const { me } = useMe()
  const [setVaultEntry] = useMutation(SET_ENTRY)

  // migrate local storage to vault
  const migrate = useCallback(async () => {
    const vaultKey = getLocalKey(me.id)
    if (!vaultKey) throw new Error('vault key not found')

    let migratedCount = 0

    for (const storageKey of Object.keys(window.localStorage)) {
      if (!storageKey.startsWith('wallet:')) continue
      if (storageKey.includes(':local-only:')) continue

      const value = getLocalStorage(storageKey)
      if (!value) continue

      const encrypted = await encryptJSON(vaultKey.key, value)

      const { data } = await setVaultEntry({ variables: { key: storageKey, value: encrypted, skipIfSet: true } })
      if (data?.setVaultEntry) {
        unsetLocalStorage(storageKey)
        window.localStorage.removeItem(storageKey.localKey)
        migratedCount++
        console.log('stored encrypted value in vault:', storageKey, value, encrypted)
      } else {
        console.error('failed to store encrypted value in vault:', storageKey)
      }
    }

    return migratedCount
  }, [me?.id])

  return migrate
}
```
</details>

with these

<details>
<summary>local storage functions</summary>

```js
function setLocalStorage (key, value) {
  window.localStorage.setItem(key, JSON.stringify(value))
}

function setLocalKey (userId, key) {
  return setLocalStorage(`vault:key:${userId}`, key)
}

function getLocalStorage (key) {
  const v = window.localStorage.getItem(key)
  return v ? JSON.parse(v) : null
}

function getLocalKey (userId) {
  return getLocalStorage(`vault:key:${userId}`)
}

function unsetLocalStorage (key) {
  window.localStorage.removeItem(key)
}

function unsetLocalKey (userId) {
  return unsetLocalStorage(`vault:key:${userId}`)
}
```

</details>

And If I understood correctly, all this complexity is there so new code does not need changes to the migration code.

I think we can keep this simple _and_ solve this problem: we could store the storage keys in a global context for which we tried to read from using `useVault`. Since that hook always runs on first render, we won't miss any during migration. So we should be able to do something like this:

```diff
diff --git a/components/use-vault.js b/components/use-vault.js
index 8c0d944f..981ae11c 100644
--- a/components/use-vault.js
+++ b/components/use-vault.js
@@ -65,6 +65,7 @@ export function useVaultConfigurator () {
 export function useVaultMigration () {
   const { me } = useMe()
   const [setVaultEntry] = useMutation(SET_ENTRY)
+  const { storageKeys } = useVaultConfigurator()

   // migrate local storage to vault
   const migrate = useCallback(async () => {
@@ -73,10 +74,7 @@ export function useVaultMigration () {

     let migratedCount = 0

-    for (const storageKey of Object.keys(window.localStorage)) {
-      if (!storageKey.startsWith('wallet:')) continue
-      if (storageKey.includes(':local-only:')) continue
-
+    for (const storageKey of storageKeys) {
       const value = getLocalStorage(storageKey)
       if (!value) continue

@@ -102,11 +100,15 @@ export function useVaultMigration () {
 // used to get and set values in the vault
 export default function useVault (unscopedStorageKey, defaultValue) {
   const { me } = useMe()
+  const { rememberStorageKey } = useVaultConfigurator()

   // scope all storage to user
   const storageKey = `${unscopedStorageKey}:${me?.id}`

   const localOnly = storageKey.includes(':local-only:')
+  if (!localOnly) {
+    rememberStorageKey(storageKey)
+  }

   const [setVaultValue] = useMutation(SET_ENTRY)
   const [value, innerSetValue] = useState(undefined)
```

What do you think?

> and to avoid conflicts with other localStorage calls. With this change there is the risk to overwrite user vault entries by mistake

How?

wallets/index.js
ekzyis: Ah, I see. No, `enabled` (and `priority`) should always be part included in the wallet configuration on any device.

wallets/index.js
ekzyis: Ah, cool, thanks, that would make sense. But removing `config` as a dependency would break this:

```js
if (typeof newConfig === 'function') {
  newConfig = newConfig(config)
}
```

which I added to no longer access local storage directly in `useWallet` and use `saveConfig` instead but I think I could have done simply this:

```diff
   const enablePayments = useCallback(() => {
-    saveConfig(config => ({ ...config, enabled: true }), { skipTests: true })
+    saveConfig(({ ...config, enabled: true }), { skipTests: true })
     logger.ok('payments enabled')
     disableFreebies().catch(console.error)
-  }, [saveConfig, logger])
+  }, [config, logger])

   const disablePayments = useCallback(() => {
-    saveConfig(config => ({ ...config, enabled: false }), { skipTests: true })
+    saveConfig(({ ...config, enabled: false }), { skipTests: true })
     logger.info('payments disabled')
-  }, [name, me, logger])
+  }, [config, name, me, logger])
```

I will test this out later.

riccardobl:
>  forward compatibility

yes, the intent was to build a component that can be standalone. To adapt old code we would need to manually filter the localStorage entries (eg. for wallet: ), but someone using the vault for new code should not worry about its internals.

> which required a) to keep in mind that the storage key that was passed into useVault is manipulated to always have the required vault: prefix and

This should not be required, because when you use the vault you should never try to access its content with localStorage, even if the user doesn't have device sync enabled. What is set in the vault should be retrieved from the vault interface, so you won't have to worry about the prefix that is added and removed internally.

> is a lot harder to understand than

i agree, with this point

>> and to avoid conflicts with other localStorage calls. With this change there is the risk to overwrite user vault entries by mistake

> How?

Let's say i want to store a user scoped variable to the local storage, but not on the vault
```
localStorage.setItem("myvar:userId",2)

```
this should not appear to be accessible from the vault

with:
```
const [value,setValue]  = useVault("myvar")
```
if the user has device sync disable.
Otherwise you'd need to track the usage of localStorage and vault and ensure that they do not use the same keys.
By having a "vault" prefix we make this highly unlikely (unless someone is really asking for troubles), so we can keep the two things logically separated.


I don't like the react context solution very much (what the useVault key is conditional?) but i think i can rewrite the old migration code on top of the new code and make it simpler

components/use-vault.js
```diff
@@ -0,0 +1,313 @@
...
+  // migrate local storage to vault
+  const migrate = useCallback(async () => {
+    const vaultKey = getLocalKey(me.id)
+    if (!vaultKey) throw new Error('vault key not found')
+
+    let migratedCount = 0
+
+    for (const storageKey of Object.keys(window.localStorage)) {
+      if (!storageKey.startsWith('wallet:')) continue
+      if (storageKey.includes(':local-only:')) continue
```
riccardobl, line 78: the new code doesn't check if the key belongs to the current user id

components/use-vault.js
```diff
@@ -0,0 +1,313 @@
...
+  }, [vaultData, me?.privates?.vaultKeyHash])
+
+  const setValue = useCallback(async (newValue) => {
+    const vaultKey = getLocalKey(me.id)
+
+    const useVault = vaultKey && vaultKey.key && vaultKey.hash === me.privates.vaultKeyHash
+
+    if (useVault && !localOnly) {
+      const encryptedValue = await encryptJSON(vaultKey.key, newValue)
+      await setVaultValue({ variables: { key: storageKey, value: encryptedValue } })
```
riccardobl, line 169: it shouldn't use the scoped key here, the unique constraint is userId+key in the database

ekzyis:
> This should not be required, because when you use the vault you should never try to access its content with localStorage, even if the user doesn't have device sync enabled. What is set in the vault should be retrieved from the vault interface, so you won't have to worry about the prefix that is added and removed internally.

Yes, it should not be required but I as a dev have to read that code over and over again so it is required for me to understand that this is happening.

I agree, what is set in the vault should be retrieved from the vault interface. That's why I no longer use the local storage interface in the wallet code, only `useVault` (it was still used before for enable/disable). You don't need to enforce this contract so hard, we can assume some competence that another dev will not retrieve something they didn't even create themselves or check if the key already exists before they write something into local storage. Also, it's still possible if someone uses `vault:...`. So afaict, this is as safe as simply using `wallet:` without all that confusing code.

> By having a "vault" prefix we make this highly unlikely (unless someone is really asking for troubles), so we can keep the two things logically separated.

Therefore, my point is that the vault prefix is redundant since we already prefix our stuff to avoid the same possible conflicts you are talking about.

> By having a "vault" prefix we make this highly unlikely 

It's already highly unlikely imo so the complexity is not worth it

> I don't like the react context solution very much

Why not?

> (what the useVault key is conditional?) 

Why would that be the case and why would that be a problem?

components/use-vault.js
ekzyis: It doesn't need to check because it can assume whoever calls `useVault` knows what they are doing

components/use-vault.js
ekzyis: But is it a problem?

riccardobl:
> Why would that be the case and why would that be a problem?

Imagine this scenario: 

- you interact with stackernews with device sync off
- some conditional code uses useVault on some keys
- now you have a localStorage entry for the vault
- you close the browser and after some time you come back and decide to enable device sync
- the code that calls useVault doesn't trigger
- you configure device sync
- the entry from the previous session is not migrated

> It's already highly unlikely imo so the complexity is not worth it

can you please check the last commit? I agree that the previous version might have been hard to read, but the new version now is much simpler, i believe.

ekzyis:
> some conditional code uses useVault on some keys
> ...
> the code that calls useVault doesn't trigger

I think that would never be the case. Hooks should not be called conditionally. That will cause other bugs anyway:

> **Don’t call Hooks inside loops, conditions, nested functions, or try/catch/finally blocks.**

-- https://react.dev/reference/rules/rules-of-hooks

components/use-vault.js
riccardobl: It could be because if you have to change the way the key is scoped in future, you will have to migrate every key in the vault. Since we don't need this additional information in the key, it is better to not add it, imo.

ekzyis: You reminded me that I've actually been thinking about hashing the keys so we don't even know what the users are storing. The names leak metadata.

ekzyis (approved):
@riccardobl, I made some more changes. Brief summary:

* wrong passphrases now show up as a validation error and not as a toast
* added QR code and scanner for easy import/export of passphrase with copy on click
* I noticed that one could set very weak passphrases like just a single letter. The resulting encryption would be trivial to crack. I decided to not allow custom passphrases to ensure strong encryption.

---

@huumn, this is ready for review now.

Summary of changes:

* `useLocalState` which was used previously to store spending permissions in local storage was replaced by `useVault`
* `useVault` reads/writes data from/to local storage if vault was not configured yet
* else it reads/writes encrypted data from/to the server
* it can be setup in the settings with only a passphrase that is automatically generated to avoid insecure passphrases

* backwards compatibility: existing wallets without the `vault:` prefix are migrated to the server vault when device sync is enabled.
* wallets which use the new wallet field `perDevice` are never migrated to the server. This is set for WebLN since WebLN requires browser extensions which are not available on mobile.
* new wallets are stored using the `vault:` prefix so we can easily find them during migration to server vault
* if devices are connected, they share wallets + status (enabled or disabled) + priority as expected
* devices can be connected by scanning the QR code or copy+paste

huumn:
Just a heads up: I'll be reviewing this tomorrow. I suspect it'll be merged then too.

ekzyis:
Just realized that the wallet security disclaimer probably needs an update since what it says is no longer true _if_ you enabled device sync.

2024-09-30-152635_736x177_scrot

riccardobl:
> [Someone reminded me](https://stacker.news/items/705629?commentId=705953) that when this is enabled, we no longer need to delete the wallets on logout.

This reminds me that it is probably a good practice to delete the vault key on logout.

My last two commits:

- Fix the wallet reset on logout to affect only localStorage entries
- Remove the vault key from localStorage on logout
- Make use-vault work for anonymous users (fallbacks to localOnly)

huumn:
I'm impressed - code is clear and simple, and this problem isn't simple at all. Please lmk if you're looking for a job!

I was only able to get to this at the end of my day. The one thing we might really want to change before merging it is storing the keys as non-extractable in indexeddb rather than in localstorage. See:  https://github.com/denoland/deno/issues/11481#issuecomment-899623594. (I mentioned this to ek last week, but neither of us had done enough research to know this could be done.)

Basically, we'd generate the keys on device, store the `CryptoKey` object with `extractable: false` in indexeddb, then trash the passphrase so us or no one else can read it (or the derived key) again. This changes the device sync flow quite a bit sadly, but it seems worth it for the security advantage. 

Regardless, you've more than qualified for the full bounty and I'll send that to you the moment you send me a 2m sat invoice. If you'd like, ek or I can make the remaining changes. Up to you!

I made myself notes while reviewing that I'm going to put here for convenience (don't feel like you need address these if you don't want to):

- expected copy button for device sync on initial setup - required clicking "manage device sync" to copy it afterward
- connect form error shows serverside errors in form field hint area
- reset/disconnect usage is a bit confusing - I suspect we can get away with "disconnect" then only show "reset" when there isn't already a passphrase
    - connect button should be far from reset button
- indexeddb storage for non-extractable `CryptoKey` object storage?
    - from  [https://pomcor.com/2017/06/02/keys-in-browser/](https://pomcor.com/2017/06/02/keys-in-browser/) ... it should apply to symmetric keys too
        >  In the Web Cryptography API, generation of an RSA or ECDSA key pair produces two CryptoKey objects, one containing the private key, the other containing the public key. When the key pair is generated, the private key can be made non-extractable from its CryptoKey object. This means that it cannot later be extracted from the object by JavaScript code embedded in a Web page, even if that Web page has the same origin as the Web page containing the JavaScript code that invoked the key generation procedure. A CryptoKey object is not persistent by itself, and it is not an ordinary JavaScript object that could be encoded as a string for storage in localStorage, but it can be stored in a database accessed through the IndexedDB API. 
    - also see: [https://github.com/denoland/deno/issues/11481#issuecomment-899623594](https://github.com/denoland/deno/issues/11481#issuecomment-899623594)

riccardobl:
> I'm impressed - code is clear and simple, and this problem isn't simple at all.  Please lmk if you're looking for a job!

Thank you for the kind words! It is also thanks to @ekzyis improvements and suggestions. 
I'll reach out to you on DevChat 🫡

> I was only able to get to this at the end of my day. The one thing we might really want to change before merging it is storing the keys as non-extractable in indexeddb rather than in localstorage. 

I've read the post you linked, and i am pretty sure it can be implemented easily, but for our specific use case:

- i believe we are already protected against trivial non targeted attacks (eg. an app/tab using some exploit to  snoop into every localStorage looking for encrypted data and secrets), since we store only the passphrase, while all the encrypted data is queried from the server. So an attack needs to be somewhat targeted or sophisticated anyway.
- we don't need to worry about protecting the passphrase if it gets leaked, since  @ekzyis  changed it to be always randomly generated, so there is no risk that the user is reusing a password to some other service .
- we also need the user to be able to encrypt and decrypt, so both the non-extractable key and the passphrase give an attacker the same capabilities.

The only security improvement this would make, that i could think of is that, since we use bip39, an app (with enough permissions) could scan the filesystem for known words and should be able to find the passphrase in the format it is currently stored, while using the non extractable CryptoKey will provide some level of obfuscation, but: 
- it won't help if the attack is targeted or sophisticated anyway, since it could be looking for the key patterns
- if we are looking to obfuscate the passphrase pattern for non targeted filesystem scans, (or someone opening the dev tools or an exploit for scanning the localStorage), we could just encrypt it with the user id inside localStorage

My main concerns in switching the logic to indexeddb are:
- more lines of code, since it has a more complex async interface
- indexeddb used to not work in incognito/private mode for some browsers, so i would probably write a fallback to localStorage, unless there is updated data on this, making it even more convoluted

That said, the end result shouldn't be **too bad** code wise imo, so we could still move forward with this. Thoughts? 
Please let me know if i am missing something.


> If you'd like, ek or I can make the remaining changes. Up to you!

No worries, I'll look into fixing at least some of them 🫡

riccardobl:
> reset/disconnect usage is a bit confusing - I suspect we can get away with "disconnect" then only show "reset" when there isn't already a passphrase

could something like this work?
image
image

the reset button appears only if device sync is configured but the user is disconnected

> connect button should be far from reset button

I am not sure how to make this look good in the current layout (maybe an extra top margin on the reset button?) 

> expected copy button for device sync on initial setup - required clicking "manage device sync" to copy it afterward

added


> connect form error shows serverside errors in form field hint area

is something like this preferable: https://github.com/stackernews/stacker.news/pull/1373/commits/98089bdf77224a9d7744c8041daea616561a0c30 ?
or should it show in a toast?

huumn:
> we also need the user to be able to encrypt and decrypt, so both the non-extractable key and the passphrase  

The non-extractable key cannot be trivially serialized and sent off device afaik, so the attacker would have to decrypt the vault on the device and send it off. That seems a step or two more targeted. 

Beyond that, not having access to the key ourselves will prevent us from accidentally doing something that reveals it.

The CryptoKey API was specifically designed for this kind of thing, so I suspect it's the best tool we have for the job.

> give an attacker the same capabilities

Neither method is bulletproof, but I don't think they're the same. Most security stuff is a game of inches afaict.

huumn:
> > reset/disconnect usage is a bit confusing - I suspect we can get away with "disconnect" then only show "reset" when there isn't already a passphrase
> 
> could something like this work? image image
> 
> the reset button appears only if device sync is configured but the user is disconnected

Yep, I think that works.

> > connect button should be far from reset button
> 
> I am not sure how to make this look good in the current layout (maybe an extra top margin on the reset button?)

They can stay close in the current layout. We should probably create a reset prompt/barrier though if we don't already have one.

> > expected copy button for device sync on initial setup - required clicking "manage device sync" to copy it afterward
> 
> added image
> 

Cool! Lots of buttons, but I don't mind the "packaging" of v0 things being wip so long as the functionality is there.

> > connect form error shows serverside errors in form field hint area
> 
> is something like this preferable: [98089bd](https://github.com/stackernews/stacker.news/commit/98089bdf77224a9d7744c8041daea616561a0c30) ? or should it show in a toast?

Toast seems to be the pattern we are using these days.

riccardobl:
> The non-extractable key cannot be trivially serialized and sent off device afaik, so the attacker would have to decrypt the vault on the device and send it off. That seems a step or two more targeted.
>Beyond that, not having access to the key ourselves will prevent us from accidentally doing something that reveals it.

Makes sense,  I hadn’t considered it that way.
The last commits implement the indexeddb stored non extractable object, if for some reason indexeddb is unavailable it fallsback to storing the passphrase in localStorage.

I moved the serialization/deserialization code into the set/get localkey functions, so the rest of the code needs to deal only with CryptoKey objects without worrying if it comes from the indexedDB or localStorage backends.

huumn:
I ended up changing two things:

- I made the passphrase a textarea because it was unreadable on mobile
- I removed the QR because:
    - before device sync is enabled, it can't be used (no "connect" option yet on other devices)
    - after device sync is enabled, we can't read the passphrase

huumn:
I ended up reverting this because given the way wallets are loaded, every time `useWallet` or `useWallets` is called, which is every where, it was making a request to the server for every sending wallet we support (both on the server and the client).

It does this mostly because we don't know the wallet priority and need to grab all of them to see the wallet priority, and the `useWallet` code was written assuming everything is stored locally.

To support device sync, we'll likely need a polling context that grabs the vault periodically so that we have a predictable number of API calls to support it.

This shouldn't be too difficult but it's a bit more involved. I have some work locally making progress on this.

PR #1387, opened 2024-09-10
css to put adjacent images/video into vertical alignment

huumn:
## Description

I don't know if this should ship. It needs more testing but I was frustrated by only one image/video being on each line.

The goal to allow for responsive grid-like behavior with just css.

## Screenshots

Screenshot 2024-09-10 at 5 20 25 PM

`3`

no


ekzyis (approved):
Looks good to me, tested it out with images and videos. I think we can ship and see.

PR #1405, opened 2024-09-13
Fix different slugs in ToC vs text

ekzyis:
## Description

I've noticed that the slugs generated in the FAQ are not always consistent with the slugs that are used in the Table of Contents. This means that if you click on an entry in the ToC, you might not get scrolled anywhere. This seems to happen if you move from a post to the FAQ. I assume that's the case since that case triggers a lot of changes in the `<Text>` component; calling `GithubSlugger.slug` a lot of times with the same inputs which always generates unique slugs.

At first I thought this is caused by strict mode but strict mode isn't enabled in production afaik and it also still happens if it's disabled locally.

To reproduce:

1. Go to https://stacker.news/items/683598
2. Click link to FAQ in the footer
3. Notice how site loads slow because a lot of renders are triggered
4. Click on a heading and see it ends with a number
5. Click on ToC and see they don't end with a number

## Additional Context

This PR assumes that headings are unique since using the `slug` function instead of `GithubSlugger` does not generate unique slugs on repeated calls; fixing the inconsistency between the slugs in the text and ToC but might cause duplicate slugs getting generated for different headings with same text.

However, I believe this is rarely the case.

## Checklist

yes

`6`

n/a

no

### DIFF
--- a/components/table-of-contents.js
+++ b/components/table-of-contents.js
@@ -5,7 +5,7 @@ import TocIcon from '@/svgs/list-unordered.svg'
 import { fromMarkdown } from 'mdast-util-from-markdown'
 import { visit } from 'unist-util-visit'
 import { toString } from 'mdast-util-to-string'
-import GithubSlugger from 'github-slugger'
+import { slug } from 'github-slugger'
 import { useRouter } from 'next/router'
 
 export default function Toc ({ text }) {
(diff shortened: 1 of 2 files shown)


huumn:
nice, somehow I missed they had a global slugger!

PR #1408, opened 2024-09-14
rethinking boost

### LINKED ISSUE #194: Feature request: Add Boost capability to comment replies
This request is to permit a person to "boost" their own comment reply (similar to how a person can boost their own Post).

Discussed in an SN Daily post [here](https://stacker.news/items/70570).

### LINKED ISSUE #1351: We need a new model for job post promotion that doesn't assume custodial funds
Current job promotion logic assumes a custodial wallet. The costs should probably be paid at one time and count towards a month worth of promotion.

huumn:
closes #1351 
closes #194 

This is done. It can be improved but it's good enough imo.

Screenshot 2024-09-18 at 7 41 29 PM
Screenshot 2024-09-18 at 7 41 41 PM
Screenshot 2024-09-18 at 7 42 34 PM
Screenshot 2024-09-18 at 7 43 31 PM

The intent here is to:

1. make job promotion use boost like normal items
2. make item "promotion" possible using boost 
    - highest boost in territory gets pinned to top
    - when viewing "home" highest boost in all followed territories gets pinned at top
4. make adding boost not reliant on edits
    - if an item is our own, whether a comment or post, show the lightning bolt flipped horizontally which when pressed opens a modal to apply boost (this way they can work for comments too)

TODO

- [x] allow jobs to be deleted
- [x] expire boost after 30 days
- [x] allow items to be boosted beyond edit window
- [x] allow highest boosted item in territory to be pinned
- [x] unify zap/downzap/boost button components/styling
- [x] comment ranking for boost
- [x] make sure boost is excluded from top

EXTRA TODO

- [x] hint if boosted item is high enough to pin it or not
- [ ] review performance and see if there's something better that can be done
- [x] squash all migrations for this into one


components/adv-post-form.js
```diff
@@ -26,9 +28,116 @@ const FormStatus = {
...
+  const boostMessage = useMemo(() => {
+    if (data?.boostPosition?.home || data?.boostPosition?.sub) {
+      const boostPinning = []
+      if (data?.boostPosition?.home) {
+        boostPinning.push('homepage')
+      }
+      if (data?.boostPosition?.sub) {
+        boostPinning.push(`~${item?.subName || sub?.name}`)
+      }
+      return `pins to the top of ${boostPinning.join(' and ')}`
```
ekzyis, line 118: This shows `pins to the top of ~undefined` for comments because `boostPosition.sub` is true but `item.subName` and `sub.name` are undefined.

I think comment boosts are cool but this scoring shouldn't run for comments or we should include a item score.

api/resolvers/item.js
```diff
@@ -635,18 +664,42 @@ export default {
...
+        outlawed: false
+      }
+      if (id) {
+        where.id = { not: Number(id) }
+      }
+
+      return {
+        home: await models.item.count({ where }) === 0,
+        sub: await models.item.count({ where: { ...where, subName: sub } }) === 0
+      }
```
ekzyis, line 702: See other comment about comment boosts showing `~undefined`:

If we fix that `sub` arrives as the string `"undefined"` here for comments using this:

```diff
diff --git a/components/adv-post-form.js b/components/adv-post-form.js
index 660160fe..43773fcb 100644
--- a/components/adv-post-form.js
+++ b/components/adv-post-form.js
@@ -90,8 +90,8 @@ export function BoostItemInput ({ item, sub, act = false, ...props }) {
   const [boost, setBoost] = useState(Number(item?.boost) + (act ? BOOST_MULT : 0))

   const [getBoostPosition, { data }] = useLazyQuery(gql`
-    query BoostPosition($id: ID, $boost: Int) {
-      boostPosition(sub: "${item?.subName || sub?.name}", id: $id, boost: $boost) {
+    query BoostPosition($sub: String, $id: ID, $boost: Int) {
+      boostPosition(sub: $sub, id: $id, boost: $boost) {
         home
         sub
       }
@@ -102,7 +102,7 @@ export function BoostItemInput ({ item, sub, act = false, ...props }) {

   useEffect(() => {
     if (boost) {
-      getPositionDebounce({ variables: { boost: Number(boost), id: item?.id } })
+      getPositionDebounce({ variables: { sub: item?.subName || sub?.name, boost: Number(boost), id: item?.id } })
     }
   }, [boost, item?.id])

```

we can always return `false` for `sub` for comments here like this:

```diff
diff --git a/api/resolvers/item.js b/api/resolvers/item.js
index fc8a1b03..c92001d9 100644
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -698,7 +698,7 @@ export default {

       return {
         home: await models.item.count({ where }) === 0,
-        sub: await models.item.count({ where: { ...where, subName: sub } }) === 0
+        sub: sub ? await models.item.count({ where: { ...where, subName: sub } }) === 0 : false
       }
     }
   },
```

Additionally, I think we only want to run this over posts, not comments since they don't affect the ranking on the homepage or territories:

```diff
diff --git a/api/resolvers/item.js b/api/resolvers/item.js
index c92001d9..03bba2ca 100644
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -690,7 +690,8 @@ export default {
         boost: { gte: boost },
         status: 'ACTIVE',
         deletedAt: null,
-        outlawed: false
+        outlawed: false,
+        parentId: null
       }
       if (id) {
         where.id = { not: Number(id) }
```

ekzyis (commented):
Looks good, played around with it and found an issue with comment boosts.

I also think we can do better with the icon. Flipping it only helps to distinguish it from zaps. But wouldn't be an arrow up be a better symbol for boosting? For example using [`arrow-up-double-line`](https://remixicon.com/icon/arrow-up-double-line):

2024-09-19-214902_346x49_scrot

It looks a bit like a rank insignia which can be associated with boosting something.

components/adv-post-form.js
huumn: Fixed on master

huumn:
> I also think we can do better with the icon. Flipping it only helps to distinguish it from zaps. But wouldn't be an arrow up be a better symbol for boosting? 

If we assume people are looking to boost, I agree. Zooming out, if we assume people aren't looking to boost, my guess is it'd be kind of annoying to look at yet another icon imo.

ekzyis:
Mhh, interesting reasoning. This isn't considered to be a different icon since it's not different enough?

Btw, I also wondered if you wanted to use `upbolt.svg` since you named the import `UpBolt`

huumn:
Yeah, I thought it looked similar enough. Maybe not according to 0xBitcoiner

PR #1424, opened 2024-09-24
batch zap requests

### LINKED ISSUE #1389: Zap responses still race to update client cache
This continues to be an issue. I'll do 3 zaps in quick succession and responses will still sometimes arrive out of order and the UI will show something like 1, 10, 100, 10, 100.

huumn:
This uses a conditional batching apollo link if successive zaps are made so they can be serially/batch executed so updating the client cache doesn't race.

This batching link is only used if there's a wallet attached or using custodial sats so that QR popup isn't delayed by the debounce.

fixes #1389 

QA: `7`

### DIFF
--- a/components/item-act.js
+++ b/components/item-act.js
@@ -13,6 +13,7 @@ import { usePaidMutation } from './use-paid-mutation'
 import { ACT_MUTATION } from '@/fragments/paidAction'
 import { meAnonSats } from '@/lib/apollo'
 import { BoostItemInput } from './adv-post-form'
+import { useWallet } from '../wallets'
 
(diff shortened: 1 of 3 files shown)


ekzyis:
wow, this looks cool and very useful, will check out now

huumn:
if you await sleep(random) in the act resolver it should let you see the it treats the symptom

ekzyis (approved):

huumn:
> wow, this looks cool and very useful, will check out now

We are probably under-leveraging apollo links. Something to keep in mind next time we look at account switching.

PR #1425, opened 2024-09-24
Image carousel

### LINKED ISSUE #986: Image Carousel / Slideshow
Not exactly a problem, just an UX improvement

When zooming an image, allow the user to slide/swipe through the other images of that post or comment

Closing the zoomed in image and clicking/tapping the next one 

**Additional context**
It would be much easier to navigate through the images of [this](https://stacker.news/items/484362) post, for example

image

ekzyis:
## Description

Close #986

This PR uses a React ref to keep track of images. This ref is passed to every image so every image can render every other image in the best resolution. This is used to implement a carousel.

## Video

https://github.com/user-attachments/assets/36570da4-046c-4687-8255-14b3835f8c0a

## Checklist

yes

`6`. I am not entirely sure if the order will always be the same as in the text but so far this was always the case. I also only tested viewing a post with images.

yes, see video

no


huumn:
Cool! I wonder if we could do this for an entire comment thread (e.g. meme monday) by creating an optional images ref in `item-full`.

I'll want to mess with the styling some (I prefer full screen images to go edge to edge), but looks good on first glance.

ekzyis:
> I'll want to mess with the styling some (I prefer full screen images to go edge to edge), but looks good on first glance.

Me too but I had problems with getting `onclick` to trigger if the div that contains the svg goes below the image using `z-index: -1`. I then settled on reserving some space for them.

_Not sure if I tried `z-index: -1` for the svg only_ :thinking:

> Cool! I wonder if we could do this for an entire comment thread (e.g. meme monday) by creating an optional images ref in item-full.

We definitely could! I just didn't think it'd be that useful but didn't think of meme monday, that'd be great!

huumn:
> We definitely could! I just didn't think it'd be that useful or confusing UX but didn't think of meme monday, that'd be great! I'll add this.

It might be nice to scroll through the thread to comment as you click through the carousel too (but maybe not must-have).

huumn:
This is what twitter looks like for a rough reference.

Screenshot 2024-09-24 at 9 24 33 AM

ekzyis:
* carousel now goes over all images in a comment section
* images now use `max-width: 100vw` again

https://github.com/user-attachments/assets/f756b266-7920-473b-8a26-94fa4182d6b7

huumn:
You were right about the scroll! It's too weird.

- I refactored the carousel - the abstraction was pretty leaky (but fits right in with the crazy markdown component). 
    - It only touches a minimal amount of lines in relevant files now.
- I added keydown and swipe events too
- I fixed the modal's overflow to link to currently viewed image - it wasn't getting updated before

huumn:
The only remaining issue with this afaict is that the carousel order depends on rendering order. In most cases it should work as expected, but if an image is external and loads more slowly the order will not be as expected.

ekzyis:
> The only remaining issue with this afaict is that the carousel order depends on rendering order. In most cases it should work as expected, but if an image is external and loads more slowly the order will not be as expected.

Mhh, if we would know the order of items on a page this could be an easy fix. We could use `item.position` which is currently only used for pins and add a boolean to indicate if an item is pinned to that position or not.

huumn:
It's not just that the images are out of order in terms of item order, they can also get out of order for an item with multiple images.

Depending on render order is the problem. Solving this is probably going to be pretty involved unfortunately.

PR #1426, opened 2024-09-24
Include extension in S3 key

ekzyis:
## Description

It's annoying when downloading pictures from m.stacker.news that the extension is not included in the URL.

## Checklist

yes

`7`. Tested locally with one image.

n/a

no


huumn:
> 7. Tested locally with one image.

I didn't expect this to break anything either, but testing with one image should not qualify as a `7`.

We want these numbers to have meaning.

PR #1428, opened 2024-09-24
Prevent Disguised Links

### LINKED ISSUE #1397: It's still possible to disguise links as other links
**Description**

~This code exists to avoid that one can write [stacker.news](https://www.youtube.com/watch?v=dQw4w9WgXcQ) which looks like it's a link to stacker.news but it's actually not (this works on Github too though).~

https://github.com/stackernews/stacker.news/blob/36e9f3f16fd3ef0ea7edf32fe84ab252f7f53020/components/text.js#L181-L189

The above was wrong, it's actually the `isRawURL` check in this code that was written to avoid disguised links.

However, it only works for
`[https://stacker.news](https://www.youtube.com/watch?v=dQw4w9WgXcQ)]`
and not
`[stacker.news](https://www.youtube.com/watch?v=dQw4w9WgXcQ)`.

https://github.com/stackernews/stacker.news/blob/450c969dfc7535b41406b29d913404bdbc3295be/components/media-or-link.js#L13-L24

It doesn't even get hit for `stacker.news` as text.

**Steps to Reproduce**

1. Write `[stacker.news](https://www.youtube.com/watch?v=dQw4w9WgXcQ)`
2. Click preview
3. See that it gets rendered as a link to stacker.news but it's a link to youtube.com

**Expected behavior**

It should refuse to use the text chosen by the user if it looks like it's pretending to be another link

**Logs**
If applicable, add your browsers console logs.

**Environment:**

**Additional context**

* https://stacker.news/items/682259

* Maybe we shouldn't attempt to stop this since Github also doesn't? So mark as `wontfix`?

tsmith123:
## Description
Fixes #1397 

When adding links with a "text part" and an "href part" to the editor, if the text part contains a string that resembles a link then ensure that it's the same as the href part. If it isn't, then render the href part only.

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -12,7 +12,7 @@ import Thumb from '@/svgs/thumb-up-fill.svg'
 import { toString } from 'mdast-util-to-string'
 import copy from 'clipboard-copy'
 import MediaOrLink from './media-or-link'
-import { IMGPROXY_URL_REGEXP, parseInternalLinks, decodeProxyUrl } from '@/lib/url'
+import { IMGPROXY_URL_REGEXP, isMisleadingLink, parseInternalLinks, decodeProxyUrl } from '@/lib/url'
 import reactStringReplace from 'react-string-replace'
 import { rehypeInlineCodeProperty, rehypeStyler } from '@/lib/md'
 import { Button } from 'react-bootstrap'
@@ -176,17 +176,11 @@ export default memo(function Text ({ rel, imgproxyUrls, children, tab, itemId, o
         return href
       }
 
-      // If [text](url) was parsed as <a> and text is not empty and not a link itself,
-      // we don't render it as an image since it was probably a conscious choice to include text.
-      let text
+      let text = children[0]
 
-      // This prevents "stacker.news" string being used in text part of link
-      // to make it look like a stackernews link - text part will be stripped out
-      // used a regex to allow for easy improvements
-      if (children.length > 0 && children[0].match(/stacker.news/)) {
-        children[0] = href
-      } else {
-        text = children[0]
+      // Prevents text being used to appear like a link
+      if (isMisleadingLink(text, href)) {
+        text = href
       }
 
       let url
@@ -264,7 +258,7 @@ export default memo(function Text ({ rel, imgproxyUrls, children, tab, itemId, o
       }
 
       // assume the link is an image which will fallback to link if it's not
-      return <TextMediaOrLink src={href} rel={rel ?? UNKNOWN_LINK_REL} {...props}>{children}</TextMediaOrLink>
+      return <TextMediaOrLink src={href} rel={rel ?? UNKNOWN_LINK_REL} {...props}>{text}</TextMediaOrLink>
     },
     img: TextMediaOrLink
   }), [outlawed, rel, itemId, Code, P, Heading, Table, TextMediaOrLink])
--- a/components/text.js
+++ b/components/text.js
@@ -263,7 +263,6 @@ export default memo(function Text ({ rel, imgproxyUrls, children, tab, itemId, o
         // ignore errors like invalid URLs
       }
 
-      console.log('HIT 1')
       // assume the link is an image which will fallback to link if it's not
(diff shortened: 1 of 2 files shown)


huumn:
I think the goal here is to prevent impersonation of any link - not just SN links. It's pretty non-trivial though because we'd need a list tlds. Or maybe simply checking for `\[a-zA-Z0-9]\\.[a-zA-Z0-9]\` and seeing that it doesn't match the url would be enough.

@ekzyis made the issue maybe he has a better idea for what this should do.

tsmith123:
No probs. I'll wait for refinement on the requirements and then make the change.

ekzyis:
Yes, I intended to solve generally misleading links. I think we can use the regexp @huumn mentioned and URL parsing to check if it looks misleading like this:

```diff
diff --git a/components/media-or-link.js b/components/media-or-link.js
index 76ac52d7..5639d3f6 100644
--- a/components/media-or-link.js
+++ b/components/media-or-link.js
@@ -11,6 +11,7 @@ import YouTube from 'react-youtube'
 import useDarkMode from './dark-mode'
 
 function LinkRaw ({ href, children, src, rel }) {
+  // prevent misleading users via URLs from origin X disguised as from origin Y
   const isRawURL = /^https?:\/\//.test(children?.[0])
   return (
     // eslint-disable-next-line
diff --git a/components/text.js b/components/text.js
index ac5c2851..8c5c7758 100644
--- a/components/text.js
+++ b/components/text.js
@@ -176,9 +176,21 @@ export default memo(function Text ({ rel, imgproxyUrls, children, tab, itemId, o
         return href
       }
 
-      // If [text](url) was parsed as <a> and text is not empty and not a link itself,
-      // we don't render it as an image since it was probably a conscious choice to include text.
-      const text = children[0]
+      let text = children[0]
+      // we use the actual URL as the text if text in [text](url) is a misleading
+      if (/^\s*(\w+\.)+\w+/.test(text)) {
+        let misleading = false
+        try {
+          if (new URL(text).origin !== new URL(href).origin) misleading = true
+        } catch {}
+
+        try {
+          if (new URL('https://' + text).origin !== new URL(href).origin) misleading = true
+        } catch {}
+
+        if (misleading) text = href
+      }
+
       let url
       try {
         url = !href.startsWith('/') && new URL(href)
@@ -254,7 +266,7 @@ export default memo(function Text ({ rel, imgproxyUrls, children, tab, itemId, o
       }
 
       // assume the link is an image which will fallback to link if it's not
-      return <TextMediaOrLink src={href} rel={rel ?? UNKNOWN_LINK_REL} {...props}>{children}</TextMediaOrLink>
+      return <TextMediaOrLink src={href} rel={rel ?? UNKNOWN_LINK_REL} {...props}>{text}</TextMediaOrLink>
     },
     img: TextMediaOrLink
   }), [outlawed, rel, itemId, Code, P, Heading, Table, TextMediaOrLink])

```

This seemed to work during my shallow testing.

tsmith123:
Hey @ekzyis 
Thanks for the suggestion. However, I don't think it works correctly as it always seems to return `true` for misleading whether the text part is the same as the href part. For example:

- [stacker.news](http://stacker.news) -> misleading (this shouldn't be considered misleading!)
- [stacker.com](http://stacker.news) -> misleading

Therefore, I've modified the logic slightly and I've also moved it into it's own function called `isMisleadingLink` (feel free to suggest a new name).

My only concern it when using the example from the issue description (with the npub link) it renders an embed. This might be due to you chaning the `children` to `text` but I don't know SN well enough to know if this is a recession or desired behaviour so if you could check this that would be cool.

huumn:
I ended up fixing this in #1430 as I was refactoring the nasty `Text` component. I used your disguised link logic so I'll award you the sats you would've earned if this were merged.

tsmith123:
Cheers @huumn appreciate that 👍

tsmith123:
Did your refactor get merged?

huumn:
Yep, it's merged. I haven't the last few days of awards though. I'll try to do it by end of day.

PR #1430, opened 2024-09-24
fixes #1395

### LINKED ISSUE #1395: Don't let media/embeds appear inline in a text node
https://stacker.news/items/681960

https://stacker.news/items/681960/edit

huumn:
- This refactors all plugins (gfm exempted) into a single` rehype` plugin 
- It prevents making embeds out of links that has `text` sibling
- It prevents images/videos from inlining with its text siblings
- Begins moving some custom element component logic into non-rerendering rehype custom elements
- It collages adjacent images into a flex grid, whether 
    - the images are all in the same `p` tag
    - the images are in independent `p` tags

The one failing of this improvement is that images/videos that aren't markeddown as images, ``, won't collage. This can be fixed for proxied images by passing in `imgproxyURLs` to the plugin, but even without it seems like a fair trade.

QA: `6`

TODO: 

- [x] should probably check code blocks, headings, and lists to make sure they still work.
- [x] fix break of item/user hover popover
- [x] refactor embeds
- [x] refactor more of the component customization into the rehype plugin
- [x] refactor text component into less of a mess
- [ ] refactor as much of the media-to-link component into the rehype plugin if possible
- [x] refactor all sn remark and rehype plugins into a single rehype plugin
    - each tree `visit` or `visitParents` or `findAndReplace` walks every node in the tree


huumn:
The ideal solution to this might be less heavy handed to this and create rehype plugins for embeds/media instead of treating them all as `a`.

huumn:
I skipped the media-or-link refactor. The biggest mess of it all is the `imgproxyURL` handling. That should probably be moved to the server so it's typed and doesn't need to be unpacked in the view-related code.

PR #1444, opened 2024-10-01
Map localhost:<port> to lnbits:5000 on server

ekzyis:
This enables attaching LNbits locally by using the host network for `app` and `worker`.

I also setup a LNbits superuser and made it persistent.

Thanks to this persistent superuser, we now also have a persistent wallet which we can use for attaching with fixed keys.

However, that wallet is not funded yet but that's for another PR when deemed valuable enough.


huumn:
Did you check the `ATTACH` instructions for other containers? I suspect this impacts some of those too.

I'm a bit worried this might impact other services as well like our mock s3 and imgproxy.

ekzyis:
Oh no, I didn't. Will check `ATTACH` instructions and impact on other services. I thought that only app and worker load these env vars but I am no longer sure.

ekzyis:
Instead of using `network_mode: "host"`, I have decided to simply allow to enter multiple URLs separated by `;` during local dev. This avoids impact on other services caused by `network_mode: "host"`.

Issue #1437, opened 2024-09-26
Bio drafts are not saved in local storage


ekzyis:
**Description**

If I write a long bio and then move somewhere else before posting it, it is gone forever.

**Steps to Reproduce**

1. Write bio
2. Go to homepage
3. Go back
4. Bio draft is gone

**Expected behavior**

Bio draft still exists

huumn:
https://github.com/stackernews/stacker.news/pull/1455

PR #1454, opened 2024-10-02
cache bio drafts

toyota-corolla0:
## Description
Address #1437 

Current pr does not cache by user. `localStorageBioDraftKey` needs user id added in. I am not sure if there is a scheme used for determining local storage keys. 

## Checklist

### DIFF
--- a/pages/[name]/index.js
+++ b/pages/[name]/index.js
@@ -21,7 +21,7 @@ export const getServerSideProps = getGetServerSideProps({
   notFound: data => !data.user
 })
 
-export function BioForm ({ handleDone, bio }) {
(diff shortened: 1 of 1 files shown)


huumn:
We have this built into forms already. If you set the `storageKeyPrefix` to `me.id` it should do the trick.

toyota-corolla0:
@huumn I made new pr #1455  cause the git history on here was so messy. Thanks!

toyota-corolla0:
@huumn I believe I was incorrectly paid for this pr.

ekzyis:
Appreciate the honesty :)

PR #1460, opened 2024-10-05
Encrypted device sync and server side client wallets

riccardobl:
re: https://github.com/stackernews/stacker.news/pull/1373#issuecomment-2394499691

This PR modifies how client wallets are managed and introduces the "vault", a method for attaching client-side encrypted data to server entries.

Below is a quick overview :

# Vault and Vault Owners
The vault is essentially a local storage system that can be securely synced across user devices via client-side end-to-end encryption. 

It operates in two modes:
- **If device sync is disabled**: Data is read and written locally.
- **If device sync is enabled**: Data is read and written encrypted to the server database.

Vault entries are associated with polymorphic owners, meaning that entries (both in the database and in local storage) are linked to different types of owners. However, I chose not to enforce this relationship directly in the database, which might be a topic for further discussion. Instead, the relationship is inferred automatically from GraphQL types.

In practice, allowing a type to own vault entries requires making that type implement VaultOwner in the GraphQL schema.

Example:

```graphql
type Wallet implements VaultOwner {
  ...
}
```

That's all that's needed—no other schema or code changes are required.

You can query and set vault entries on the client using `use-vault.js`, which provides both a React hook (`useVault()`) and an asynchronous function (`openVault()`). Both functions require an owner parameter, which can either be the result of a query for a `VaultOwner` type (which has an `id` and `__typename` field) or a simple JavaScript object with `id` and `type` fields.

A typical React flow looks like this:

```js
const {data} = useQuery(WALLET)
const [value, setValue, clearValue, refreshData] = useVault(data.wallet, 'mysecret')
```

Using `useVault()` or `openVault()`, the code queries the GraphQL resolver, which verifies if the type implements `VaultOwner` and then either retrieves or sets the data.

# Local Storage
When device sync is disabled, the vault's backend is handled by a standalone local storage component, provided by `use-local-storage` and built on top of IndexedDB.

To open local storage, pass a `database` name and an optional `namespace` to `useLocalStorage()` or `openLocalStorage()`. The namespace logically separates different types of entries (in the vault’s case, different owners) to avoid key conflicts.

A namespace is an array, which is combined into a single string. For example, `['wallet', 1]` would be stored as `'wallet:1'`.

# Device Sync
Device sync functionality remains unchanged from the previous PR.

# Client Wallets
Previously, sending wallets were managed solely on the client side. Now, they are stored in the `Wallet` table along with receiving wallets (wallets that both send and receive use a single row). The `upsertWalletXXX` mutations are used to update both sending and receiving wallets, allowing you to set status, priority, and receiving properties. These mutations now accept `canSend` and `canReceive` flags, indicating if a wallet can send only, receive only, or do both.

To enable this, I've made all arguments optional in the `upsertWalletXXX` mutations, but the resolver will still check them checks if `canReceive=true`.

# Improved Wallets Query
The `wallets()` query can now filter by sending, receiving, and status, and it can also sort by priority.

# Polling Context for Wallets
Wallets are now fetched within a polling context, reducing API spam from frequent calls to `useWallet()`.

# Removed `perDevice` Wallets and Added `isAvailable()`
Previously, `perDevice` was added to exclude the WebLN wallet from device sync. However, this caused other issues and may not have been the right approach. For instance, if you have a desktop and laptop with WebLN enabled, but a mobile device without it, you might still want to prioritize WebLN on all devices where it's available.

To solve this, I’ve removed the `perDevice` field and added an optionally exportable `isAvailable()` function for wallets. This allows wallets to be disabled at runtime, and it's currently applied only to the WebLN wallet.

# New Fields for Sending Wallets
Sending wallets now need to export `walletType` and `walletField`, similar to receiving wallets. These fields are only used to build mutations, so there’s no need to define a wallet kind model in the schema.

Before this PR, we didn’t need to worry about marking fields as `clientOnly` for sending wallets, as they were entirely client-side. Now that the distinction between client and server is less clear, we need to mark fields as either `clientOnly` or `serverOnly`. If a wallet contains only client-side fields, you can mark all of them by exporting export `const clientOnly = true`.

# Importing Client Wallets on the Server and Avoiding Client-Side Code in `wallet/index.js`
Since the server must be aware of client-only wallets, they are now imported into `wallets/server.js`. To avoid importing browser-only code in `client.js`, the `index.js` file for each wallet will be imported instead:

```js
import * as blink from 'wallets/blink'
```

Do not use:
~~import * as blink from 'wallets/blink/client'~~

The way sending wallets are imported in `wallets/client.js` remains unchanged.

Make sure not to include browser-specific code (e.g., WebLN checks) in `index.js`, this should go into `client.js`.

# Avoiding Orphaned Vault Entries
Since the vault-owner relationship is "virtual" (as not enforced in the database), extra care is needed to avoid orphaned entries. For instance, if a wallet is deleted, its corresponding vault entries should also be cleaned up using the clean method from `useVault()`.

TODO:
- [x] clientside encryption
- [x] ownable vault entries
- [x] client wallets database entries
- [x] migrate old localStorage based wallets

Testing:
- [x] initialize vault
- [x] connect to vault
- [x] disconnect from vault
- [x] reset vault
- [x] sending wallets save
- [x] receiving wallets save
- [x] sending wallets update
- [x] receiving wallets update
- [x] sending wallets sync
- [x] receiving wallets sync
- [x] sending works
- [x] receiving works
- [x] priority works (visually)
- [x] priority works in sending
- [x] priority works in receiving
- [x] autowithdraw settings save
- [x] autowithdraw settings update
- [x] old wallets migration
- [x] webln should appear disabled in unsupported devices even when device sync is enabled
- [x] logout clears unsynched wallets
- [x] logout clears vault key
- [x] logout doesn't clear synched wallets

---
~~Uses a polling context to fetch the whole user vault with a single request.~~

~~Should work fine as long as we don't use the vault to store a lot of data (eg. dms), then we should probably adapt the code to reconstruct a local encrypted indexeddb cache and fetch only the diffs, should be easy enough and not require too many changes, but seems overkill for its current use case. Thoughts?~~


huumn:
> Should work fine as long as we don't use the vault to store a lot of data (eg. dms)

When we do dms (and other things like encrypted comments), we'll probably want to move away from the key-value nature of the vault and store enough metadata to do relational queries.

ekzyis:
Fyi, we usually use the draft setting in Github to indicate if something is still a draft

prisma/migrations/20241013200637_client_wallets_2/migration.sql
ekzyis, line 1: I think there is something wrong with the migrations. You delete here the tables you created in the previous one.

api/resolvers/vault.js
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Ensures the passed ownerType represent a valid type that extends VaultOwner in the graphql schema.
+ * Throws a GqlInputError otherwise
+ * @param {*} info the graphql resolve info
+ * @param {string} ownerType the ownerType to check
+ * @throws GqlInputError
+ */
+function checkOwner (info, ownerType) {
+  const gqltypeDef = info.schema.getType(ownerType)
+  const ownerInterfaces = gqltypeDef?.getInterfaces ? gqltypeDef.getInterfaces() : null
```
ekzyis, line 144: not tested but I believe you can write this like this:

```suggestion
  const ownerInterfaces = gqltypeDef?.getInterfaces?.()
```

ekzyis (commented):
took a quick look, didn't pull the code yet

prisma/migrations/20241013200637_client_wallets_2/migration.sql
riccardobl: yes, i've changed the code to not require wallet type tables for sending-only wallets, so the tables i made initially were not needed anymore, i forgot to merge the two migrations together.

components/device-sync.js
```diff
@@ -0,0 +1,269 @@
...
+  }, [])
+
+  const onConnect = useCallback(async (values, formik) => {
+    if (values.passphrase) {
+      try {
+        await setVaultKey(values.passphrase)
+        await migrate()
+        apollo.cache.evict({ fieldName: 'BestWallets' })
+        apollo.cache.gc()
+        await apollo.refetchQueries({ include: ['BestWallets'] })
```
huumn, line 112: I'm just browsing right now and haven't taken a close look, but I suspect we'll eventually want the vault to not be this tightly "coupled" with wallets.

huumn (commented):
I thinks this is going to require quite a few iterations of feedback to get right. Communication overhead might be high so if at any point you want me to implement the changes I'm suggesting, let me know.

I've only really tested the changes and looked at them briefly - mostly focusing on the new wallet stuff and that it simply works as expected before even creating the vault - but I wanted to get started synchronizing our understanding before recommending any specific changes.

First, I'm curious why the vault entries are associated with `users` rather than with a `Wallet` in the db? We discussed the latter so I suspect you had good reason to not go that route. The point of that recommendation was being able to treat both send and receive wallets mostly the same.

The goal was to create a query where we'd `getWallets` and the response would be like:

```
    getWallets () {
      id
      priority
      type
      updatedAt
      canSend
      canReceive
      enabled
      wallet {
        __typename
        ... on WalletLightningAddress {
          address
        }
        ... on WalletLnd {
          socket
          macaroon
          cert
        }
        ... on WalletCln {
          socket
          rune
          cert
        }
        ... on WalletLnbits {
          url
          invoiceKey
        }
        ... on WalletNwc {
          nwcUrlRecv
        }
        ... on WalletPhoenixd {
          url
          secondaryPassword
        } 
      }
      encryptedValues: {
           id
           iv
           key
           value
      } 
    }
```

Then on the client, we'd decrypt values using `{decrypt} = useVault()` as needed and when upserting a wallet encrypting the values as needed using `{encrypt} = useVault()`. (`useVault()` might also need to return `{ onPassphraseChange } = useVault()` so we can re-encrypt the wallet values).

After these changes, the only difference between send/receive wallet config (assuming device sync is used) would become:
- send wallets have these encrypted fields
- send wallets have fields that require additional clientside validation

Then we can get rid of a lot of weird code (or at least move it into its own part of the codebase for backwards compatibility).

--------------------

If a lot of this code is to support a migration, and we're confident we need to migrate independent of device sync, I think we should break out the migration into its own change so that we can create the device sync abstraction cleanly. 

If the user is using device sync, do we really need to store anything in indexeddb other than the encryption key? I think that might also be an opportunity to save code - we don't need to be storing these values in multiple places (for read perf, apollo's cache is already storing the values on the client in memory).

---------------------

As far as small things I've noticed so far: I think `iv` should get its own db column (and probably its own field in indexeddb, e.g. `{iv, value}`), so that it can be validated and we aren't storing stringified JSON on the server sent from the client.

----------------

This ticket is a big lift for even ek or I and you're new to the codebase. It makes sense we'd need to take a lot of swings at it first. Again, let me know if this isn't making sense or you think it's worth me making these changes instead.

riccardobl:
I think there are some misunderstanding in the way this code works:

> First, I'm curious why the vault entries are associated with `users` rather than with a `Wallet` in the db? We discussed the latter so I suspect you had good reason to not go that route. The point of that recommendation was being able to treat both send and receive wallets mostly the same.

They are associated to the user in the schema to have a relation to allow the user to clear its vault, eg. to remove data from the server or change password.
However they also have an ownerId and ownerType that in the case of wallet are Wallet and the wallet  id.
The only unorthodox thing about this implementation  is that it uses the graphql type as ownerType instead of remapping it to something that is known by the database.

> The goal was to create a query where we'd getWallets and the response would be like:

This is something that can be implemented into this design, on the wallet resolver we can select all the entries associated to the ownerType='Wallet'  and walletId and return them with the wallet fields. 
However i didn't do that because it makes the vault code a bit more complex and it didn't seem worth the 1 api call it saves.


> If a lot of this code is to support a migration, and we're confident we need to migrate independent of device sync, I think we should break out the migration into its own change so that we can create the device sync abstraction cleanly.

There are two kinds of migrations happening here: 
- from an unsynched vault to a synched vault
- from the old storage to the vault

i think what you are referring to is the second one, and it is happening in this code 

https://github.com/stackernews/stacker.news/blob/afac53a879a2dbc2e1e6483d92bf960b6c2a3476/wallets/index.js#L498-L527

so i'd say it is already out of the device sync abstraction


> If the user is using device sync, do we really need to store anything in indexeddb other than the encryption key? I think that might also be an opportunity to save code - we don't need to be storing these values in multiple places (for read perf, apollo's cache is already storing the values on the client in memory).

That's already the case when device sync is enabled

> I think iv should get its own db column (and probably its own field in indexeddb, e.g. {iv, value}), so that it can be validated and we aren't storing stringified JSON on the server sent from the client.

I am not sure, since these values mean something only to the client, and since we could change the encryption algorithm at a later time, i would suggest keeping them into a generic field. We could even turn it into a binary field since iv is of a known length.

> in addition to the iv it might be worth storing the vaultKeyHash with the vault entries so that we can detect if we're "locked out" of reading a value

This implementation stores the vaultKeyHash in the user table. It is not possible to be locked out from reading a value, because the system ensures the users is using only one vault key across all the devices by checking it against vaultKeyHash, if for any reason the key changes, the user is disconnected from the vault.
The problem of having  a (potentially) different vaultKeyHash associated to every encrypted vault entry, is that we open to the possibility of having an user using multiple different vaultkeys on different devices, this makes the migration from an unsynched to a synched vault a lot more complex, even just to figure out who gets the precedence.

huumn:
All of that's fair. I just wanted to share my thoughts before I had a chance to look deeper - mostly so we have more time to update the PR. I'm sorry if anything I said was off-base.

I'll continue reviewing the PR and let you know if anything else comes up.

PR #1479, opened 2024-10-16
Blink Receive

riccardobl:
## Description

Blink receive attachment

## Screenshots
image

## Additional Context

Like in the sending attachment, there is a currency field but it is enforced to BTC, since currently blink doesn't support invoices below 1 cent in their USD wallet, making it unable to receive small p2p zaps.

## Checklist

yes

8

yes

no


wallets/blink/common.js
ekzyis, line 1: I put stuff like this in the index.js file of the wallet in the past but I let you decide

wallets/blink/server.js
```diff
@@ -0,0 +1,62 @@
...
+      amount: msatsToSats(msats),
+      expiresIn: Math.floor(expiry / 60) || 1,
+      memo: description,
+      walletId: wallet.id
+    }
+  })
+  const res = out.data.lnInvoiceCreate
+  const errors = res.errors
+  if (errors && errors.length > 0) {
+    throw new Error('failed to pay invoice ' + errors.map(e => e.code + ' ' + e.message).join(', '))
```
ekzyis, line 58: error should say `failed to create invoice`

wallets/blink/server.js
```diff
@@ -0,0 +1,62 @@
+import { withTimeout } from '@/lib/time'
+import { getScopes, SCOPE_READ, SCOPE_RECEIVE, SCOPE_WRITE, getWallet, request } from 'wallets/blink/common'
+import { msatsToSats } from '@/lib/format'
+export * from 'wallets/blink'
+
+export async function testCreateInvoice ({ apiKeyRecv, currencyRecv }) {
+  const strict = true
```
ekzyis, line 7: same here

wallets/blink/client.js
```diff
@@ -1,10 +1,23 @@
-import { galoyBlinkUrl } from 'wallets/blink'
+import { getScopes, SCOPE_READ, SCOPE_RECEIVE, SCOPE_WRITE, getWallet, request } from 'wallets/blink/common'
 export * from 'wallets/blink'
 
 export async function testSendPayment ({ apiKey, currency }, { logger }) {
-  currency = currency ? currency.toUpperCase() : 'BTC'
   logger.info('trying to fetch ' + currency + ' wallet')
+  const strict = false
```
ekzyis, line 6: I think this was only for debugging

ekzyis (commented):
Didn't test, only looked over the code. Left some comments.

Btw, maybe should have mentioned this earlier, but have you seen `scripts/genwallet.sh`? I created it to add new wallets easily but maybe it doesn't work that well for enhancing an existing wallet.

wallets/blink/common.js
riccardobl: the rationale here was that, since we re-export everything that is exported from index
`export * from 'wallets/blink'`
if there are many helpers/constants maybe it is better to put them elsewhere so that they are not re-exported, not sure if this makes sense since we use webpack, but felt cleaner this way

riccardobl:
> Btw, maybe should have mentioned this earlier, but have you seen `scripts/genwallet.sh`?

Cool script, I didn't see it before :100:

lib/validate.js
```diff
@@ -748,14 +748,32 @@ export const nwcSchema = object().shape({
...
+    }),
   currency: string()
     .transform(value => value ? value.toUpperCase() : 'BTC')
-    .oneOf(['USD', 'BTC'], 'must be BTC or USD')
-})
+    .oneOf(['USD', 'BTC'], 'must be BTC or USD'),
+  currencyRecv: string()
+    .transform(value => value ? value.toUpperCase() : 'BTC')
+    .oneOf(['BTC'], 'must be BTC')
+}, ['apiKey', 'apiKeyRecv'])
```
ekzyis, line 776: You need to add `...autowithdrawSchemaMembers` here for validation of autowithdraw settings to work, see `LNDAutowithdrawSchema` for example. I did that mistake in NWC, LNbits and phoenixd, see #1487,

huumn:
@ekzyis is this good to merge. I don't have Blink to QA and I'm not familiar enough with the wallet attachment code to purely reason about the changes.

wallets/blink/client.js
```diff
@@ -1,10 +1,22 @@
...
+  const scopes = await getScopes(apiKey)
+  if (!scopes.includes(SCOPE_READ)) {
+    throw new Error('missing READ scope')
+  }
+  if (!scopes.includes(SCOPE_WRITE)) {
+    throw new Error('missing WRITE scope')
+  }
+  if (scopes.includes(SCOPE_RECEIVE)) {
+    throw new Error('RECEIVE scope must not be present')
+  }
```
ekzyis, line 15: I would say this check is not needed for sending, we only need to check that we can pay invoices. Stackers already struggle with attaching wallets, we don't need to make it harder if not necessary.

wallets/blink/index.js
ekzyis, line 1:
```diff
diff --git a/wallets/blink/index.js b/wallets/blink/index.js
index b8e45e35..7966cc0a 100644
--- a/wallets/blink/index.js
+++ b/wallets/blink/index.js
@@ -20,7 +20,7 @@ export const fields = [
     name: 'currency',
     label: 'wallet type',
     type: 'text',
-    help: 'the blink wallet to use (BTC or USD for stablesats)',
+    help: 'the blink wallet to use for sending (BTC or USD for stablesats)',
     placeholder: 'BTC',
     clear: true,
     autoComplete: 'off',
@@ -41,7 +41,7 @@ export const fields = [
     name: 'currencyRecv',
     label: 'wallet type',
     type: 'text',
-    help: 'the blink wallet to use (BTC or USD for stablesats)',
+    help: 'the blink wallet to use for receiving (only BTC available)',
     value: 'BTC',
     clear: true,
     autoComplete: 'off',
```

wallets/blink/server.js
```diff
@@ -0,0 +1,61 @@
+import { withTimeout } from '@/lib/time'
+import { getScopes, SCOPE_READ, SCOPE_RECEIVE, SCOPE_WRITE, getWallet, request } from 'wallets/blink/common'
+import { msatsToSats } from '@/lib/format'
+export * from 'wallets/blink'
+
+export async function testCreateInvoice ({ apiKeyRecv, currencyRecv }) {
+  const scopes = await getScopes(apiKeyRecv)
+  if (!scopes.includes(SCOPE_READ)) {
+    throw new Error('missing READ scope')
+  }
```
ekzyis, line 10: Interesting. I thought `RECEIVE` scope might be enough but I tested it and the scopes query returned `null`. So I guess `READ` is indeed needed to check that we weren't given `WRITE` permissons. Funny.

wallets/blink/index.js
ekzyis: Also, the fields not visually being marked as read-only but not being able to be edited is a common UX bug that we have due to the way that we can't edit once attached (unless we enter other valid credentials).

I think basically every wallet field should have `editable: false` set because of this but I will check. Here's the patch for Blink:

```diff
diff --git a/wallets/blink/index.js b/wallets/blink/index.js
index 7966cc0a..392813cd 100644
--- a/wallets/blink/index.js
+++ b/wallets/blink/index.js
@@ -14,7 +14,8 @@ export const fields = [
     help: `you can get an API key from [Blink Dashboard](${galoyBlinkDashboardUrl}).\nPlease make sure to select ONLY the 'Read' and 'Write' scopes when generating this API key.`,
     placeholder: 'blink_...',
     optional: 'for sending',
-    clientOnly: true
+    clientOnly: true,
+    editable: false
   },
   {
     name: 'currency',
@@ -25,7 +26,8 @@ export const fields = [
     clear: true,
     autoComplete: 'off',
     optional: 'for sending',
-    clientOnly: true
+    clientOnly: true,
+    editable: false

   },
   {
@@ -35,7 +37,8 @@ export const fields = [
     help: `you can get an API key from [Blink Dashboard](${galoyBlinkDashboardUrl}).\nPlease make sure to select ONLY the 'Read' and 'Receive' scopes when generating this API key.`,
     placeholder: 'blink_...',
     optional: 'for receiving',
-    serverOnly: true
+    serverOnly: true,
+    editable: false
   },
   {
     name: 'currencyRecv',
```

ekzyis (commented):
Full review now, also tested with Blink app and attached send+recv.

1. read-only UX:

I noticed that the `readonly` input property for the receive currency is not correctly set. This patch should fix it:

```diff
diff --git a/pages/settings/wallets/[wallet].js b/pages/settings/wallets/[wallet].js
index 0343893f..0672a376 100644
--- a/pages/settings/wallets/[wallet].js
+++ b/pages/settings/wallets/[wallet].js
@@ -110,7 +110,7 @@ function WalletFields ({ wallet: { config, fields, isConfigured } }) {
         ...props,
         name,
         initialValue: config?.[name],
-        readOnly: isClient && isConfigured && editable === false && !!config?.[name],
+        readOnly: isClient && editable === false && ((isConfigured && !!config?.[name]) || !!props.value),
         groupClassName: props.hidden ? 'd-none' : undefined,
         label: label
           ? (
```

Also see patch regarding `editable: false` for other wallet fields.

2. Help text suggestions (see comment)

3. Too tight permissions for sending (see comment)

wallets/blink/server.js
riccardobl: i think it is also used to fetch the wallet id

ekzyis (commented):
This might be unrelated to the changes in this PR since there are no layout changes, but when I setup send+recv and refresh the page, I get a hydration error:

2024-10-22-212500_1920x1200_scrot

Will look into this now. Otherwise looks good

riccardobl:
yes, it is happening on my side too

riccardobl:
this is caused by using `clear:true` in clientOnly fields , I am still investigating the actual cause, but it doesn't seem related to this pr, it is just that this pr happens to use clear:true

ekzyis (approved):
This requires #1512 to be merged but otherwise is good

PR #1480, opened 2024-10-17
Fix infinite loop of loading wallet logs + center more button

ekzyis:
## Description

This fixes an infinite loop of reloading wallet logs.

Also moved `more` button from this

localhost_3000_settings_wallets_nwc(iPhone SE) (1)

to center like this:

localhost_3000_settings_wallets_nwc(iPhone SE)

I think this is more consistent with the logs layout and how other buttons for more look.

## Checklist

yes

`6`. Made sure logs for every wallet still load correctly.

yes, see screenshots

no

### DIFF
--- a/components/wallet-logger.js
+++ b/components/wallet-logger.js
@@ -12,10 +12,7 @@ import useIndexedDB from './use-indexeddb'
 import { SSR } from '@/lib/constants'
 
 export function WalletLogs ({ wallet, embedded }) {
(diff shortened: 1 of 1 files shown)


huumn:
This fixes the render loop but it causes the first page to get loaded twice:

Screenshot 2024-10-18 at 7 55 15 AM

ekzyis:
> This fixes the render loop but it causes the first page to get loaded twice (more just grabs page 1 again)

Fixed in https://github.com/stackernews/stacker.news/pull/1484

PR #1487, opened 2024-10-18
Fix missing autowithdraw settings validation

ekzyis:
## Description

The wallets which weren't the original receive wallets didn't validate autowithdraw settings (NWC, LNbits, phoenixd).

## Checklist

yes

`5`. I saved NWC, others should work too then

n/a

no

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -726,7 +726,8 @@ export const lnbitsSchema = object().shape({
         test: invoiceKey => adminKey !== invoiceKey,
         message: 'invoice key cannot be the same as admin key'
       })
-    })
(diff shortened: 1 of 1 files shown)


ekzyis:
I noticed this broke client config saves since they exclude the `autoWithdraw` members but are now required. Will fix in another PR.

huumn:
eta on fix? I told siggy I'd deploy the deletion fix. Should I pick it up?

ekzyis:
I've been thinking about a better fix but https://github.com/stackernews/stacker.news/pull/1499 should do it

PR #1488, opened 2024-10-18
Max total fee

### LINKED ISSUE #1383: Allow to set max base fee for withdrawals

I configured 1% as the max fee for autowithdrawals to my phoenixd node. I noticed that withdrawals smaller than 100 sats don't seem to work with phoenixd. I see `no route found` errors in the wallet logs. I strongly suspect this is because the ACINQ node takes a base fee of 1 sat for receives which means that 1% of anything smaller than 100 sats is not enough.

Since the channel between SN and ACINQ is not public, I tested this by creating a 1 sat invoice from phoenixd and paid it using my sats on SN and noticed that indeed, there was 1 sat as a routing fee:

2024-09-09-143748_743x282_scrot

When we switch to non-custodial, this means that I would receive fee credits for any zap that is smaller than 100 sats.

I am fine with paying 1 sat to receive 1 sat even if it means that I pay 50% as routing fees. That's just how lightning base fees work. I think we should allow to set a base fee and then use whatever is higher as the max fee: `max_fee = max(max_percentage_fee, max_base_fee)`

* don't care about small zaps becoming fee credits in the future if I continue to use phoenixd
* run node with private channel to SN again

**Additional context**

n/a

ekzyis:
## Description

This adds `max base fee` as a new autowithdrawal setting. This can be used to make small withdrawals possible without allowing a higher max fee rate for big withdrawals. 

This setting is set to 0 for users which have autowithdrawals enabled to keep existing fee settings and not suddenly allow higher fees without their interaction.

close #1383 based on #1487

## Screenshots

2024-10-18-190904_1920x1200_scrot

## Additional Context

- Does subtitle of network fees make sense?
- Does hint for max base fee make sense?
- Can we expect or do we want stackers to understand the difference between `max fee rate`, `max base fee` and pathfinding? 

## Checklist

yes, new db column added but is nullable

`6`. I checked that max fee is correctly calculated during autowithdrawal.

yes

no


huumn (commented):
It looks like you mean "max absolute fee" rather than "max base fee."

I think what we want is "max fee relative" and "max fee absolute" but I'm not sure that's pleb friendly.

huumn:
Maybe "max fee rate" and "max fee total"

huumn:
I think we also want the subtitle to say "we'll use whichever is higher" given the withdrawn amount.

ekzyis:
Updated text. Now looks like this:

2024-10-19-224942_1920x1200_scrot

huumn (commented):
This probably needs to account for #1499?

ekzyis:
> This probably needs to account for #1499?

Oh, I already accounted for the new `autoWithdrawMaxFeeTotal` in #1499 but not for `autoWithdrawThreshold`. Added now!

PR #1498, opened 2024-10-19
premature invoice deletion: fix interval in sql template

### LINKED ISSUE #1485: premature invoice deletion
https://stacker.news/items/727805

huumn:
fixes #1485 

It looks like prisma's interpretation of intervals in templates changed at some point. I didn't find anywhere else where we use intervals this way.

QA: `6`. made invoices, reproduced bug, then made some more invoices, ran job - no deletions, made one of the invoices artificially old, ran job - only old one deleted.

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -482,7 +482,7 @@ const resolvers = {
         FROM "Withdrawl"
         WHERE "userId" = ${me.id}
         AND id = ${Number(id)}
-        AND now() > created_at + interval '${retention}'
+        AND now() > created_at + ${retention}::INTERVAL
         AND hash IS NOT NULL
         AND status IS NOT NULL
       ), updated_rows AS (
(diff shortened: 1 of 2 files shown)


ekzyis (approved):
Also reproduced the bug. Makes sense that this syntax confuses Prisma. I also tested it with this query:

```js
await models.$queryRaw`
SELECT id, created_at
FROM "Withdrawl"
WHERE now() > created_at + interval '${'1 year'}'`
```

It fetched withdrawals that were clearly too young. Using `Prisma.sql` didn't explain how Prisma interprets this though:

```js
console.log(Prisma.sql`
SELECT id, created_at
FROM "Withdrawl"
WHERE now() > created_at + interval '${'1 year'}'
AND hash = '${'how does this look in Prisma.sql?'}'`.sql)
```

```sql
SELECT id, created_at
FROM "Withdrawl"
WHERE now() > created_at + interval '?'
AND hash = '?'
```

On a sidenote, I think we should also delete preimages since we delete payment hashes but they can be derived from the preimage or only delete bolt11s.

I think the question why we not only delete bolt11s came up before. The reason was was that payment hashes are sensitive because they can be used for correlation iirc. 

Following that reasoning, we should then also delete preimages.

huumn:
true we weren't storing the preimages in all cases before

PR #1503, opened 2024-10-20
Show Progress Bar on Back

### LINKED ISSUE #1471: Progress bar for back button
We display a yellow progress bar on navigation but don't do it for the back button.

tsmith123:
Fixes #1471 

**Observations before fixing**
- If a user clicks on a post from the homepage and then clicks back then the progress bar is now shown
- If a user click on a post and then clicks to somewhere else and then uses back then the progress bar is shown on every page

**Attempted Solution**
The issue lies with the passing of `shallow: true` in `_app.js`.

This PR offers the quick fix of `shallow: false` but I would assume this has implications elsewhere in the app?

### DIFF
--- a/components/nav/common.js
+++ b/components/nav/common.js
@@ -55,15 +55,7 @@ export function Back () {
   if (!back) return null
 
   return (
-    <a
-      role='button' tabIndex='0' className='nav-link p-0 me-2' onClick={() => {
-        if (back) {
-          router.back()
-        } else {
-          router.push('/')
-        }
-      }}
-    >
+    <a role='button' tabIndex='0' className='nav-link p-0 me-2' onClick={() => router.back()}>
       <BackArrow className='theme me-1 me-md-2' width={24} height={24} />
     </a>
   )
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -46,8 +46,8 @@ export default function MyApp ({ Component, pageProps: { ...props } }) {
   const router = useRouter()
 
   useEffect(() => {
-    const nprogressStart = (_, { shallow }) => !shallow && NProgress.start()
-    const nprogressDone = (_, { shallow }) => !shallow && NProgress.done()
+    const nprogressStart = (_, { shallow }) => NProgress.start()
+    const nprogressDone = (_, { shallow }) => NProgress.done()
 
     router.events.on('routeChangeStart', nprogressStart)
     router.events.on('routeChangeComplete', nprogressDone)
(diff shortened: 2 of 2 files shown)


components/nav/common.js
```diff
@@ -55,15 +55,7 @@ export function Back () {
...
   return (
-    <a
-      role='button' tabIndex='0' className='nav-link p-0 me-2' onClick={() => {
-        if (back) {
-          router.back()
-        } else {
-          router.push('/')
-        }
-      }}
-    >
```
tsmith123, line 66: As there's a check on line 55 for "truthy" 'back' value then no need to check again here

pages/_app.js
```diff
@@ -58,15 +58,14 @@ export default function MyApp ({ Component, pageProps: { ...props } }) {
...
     // to point to the same page but without SSR, ie ?nodata=true
     // this nodata var will get passed to the server on back/foward and
-    // 1. prevent data from reloading and 2. perserve scroll
+    // 1. prevent data from reloading and 2. preserve scroll
     // (2) is not possible while intercepting nav with beforePopState
     router.replace({
       pathname: router.pathname,
       query: { ...router.query, nodata: true }
-    }, router.asPath, { ...router.options, shallow: true }).catch((e) => {
+    }, router.asPath, { ...router.options, shallow: false }).catch((e) => {
```
tsmith123, line 66: This probably has implications elsewhere in the app?

huumn: yes this will cause the data to load twice on every page visit. no go with this change.

huumn (changes requested):
Given that shallow is intentional, I think you'll want to make `nprogressStart` and `nprogessDone` make an exception for back navigation if possible rather than changing the behavior of every router.replace operation.

tsmith123:
Thanks. The current behaviour is very strange though... if you go to a post from the homepage and then go to your profile page then click back twice it will show the progress bar for both backs.

huumn:
Are you saying we shouldn't have progress bars on back nav? Maybe this is the wrong solution to the right problem.

The problem we have is that nextjs wants to hit the server on back nav because we SSR content. Ideally nextjs wouldn't do that, but at least when this `useEffect` was written (probably nextjs v10) that wasn't possible.

tsmith123:
My last comment was just to say that if you navigate deeper into the app and then use the back button then you will see a progress bar. But if you just go to a post and then back then you won't - that's the strange behaviour.

I did look into `router.back()` when I first started looking at the issue but it doesn't offer a `shallow` option I'm afraid.

PR #1507, opened 2024-10-23
Fantasy wallet refactor

huumn:
This is ready for review.

TODO:

- [x] improve validation (do per field validation and compose into yup schema)
    - [x] test clientside
    - [x] test serverside
    - [x] document changes
    - [x] add validation for vaultEntries
- [x] debug config checking
- [x] debug vault post-changes
- [x] debug device sync flow
- [ ] test all send wallets
- [ ] test all receive wallets
- [x] enable/disable all wallets
- [x] test autowithdraw
- [x] test p2p zaps


api/resolvers/vault.js
```diff
@@ -0,0 +1,75 @@
...
+
+      const k = await models.vault.findUnique({
+        where: {
+          key,
+          userId: me.id
+        }
+      })
+      return k
+    },
+    getVaultEntries: async (parent, { keysFilter }, { me, models }, info) => {
```
huumn, line 17: We really only call this when the passphrase changes so that we can decrypt and reencrypt vault entries.

api/resolvers/vault.js
```diff
@@ -0,0 +1,75 @@
...
+              }
+            : undefined
+        }
+      })
+      return entries
+    }
+  },
+  Mutation: {
+    // atomic vault migration
+    updateVaultKey: async (parent, { entries, hash }, { me, models }) => {
```
huumn, line 35: This migrates the vault, storing all reencrypted entries and the new key hash - atomically.

api/resolvers/wallet.js
```diff
@@ -506,6 +527,15 @@ const resolvers = {
       }
       return { id }
     },
+    setWalletPriority: async (parent, { id, priority }, { me, models }) => {
```
huumn, line 530: setting wallet priority is now an isolated mutation so that we don't have to revalidate everything.

components/fee-button.js
```diff
@@ -93,7 +93,10 @@ function sortHelper (a, b) {
   }
 }
 
-export function FeeButtonProvider ({ baseLineItems = {}, useRemoteLineItems = () => null, children }) {
+const DEFAULT_BASE_LINE_ITEMS = {}
```
huumn, line 96: these non primitive defaults can cause rerenders

components/nav/common.js
```diff
@@ -263,7 +263,7 @@ export default function LoginButton () {
 
 function LogoutObstacle ({ onClose }) {
   const { registration: swRegistration, togglePushSubscription } = useServiceWorker()
-  const wallets = useWallets()
+  const { wallets } = useWallets()
```
huumn, line 266: the changes to wallets might've broke reseting client wallets

components/vault/use-vault-configurator.js
```diff
@@ -0,0 +1,160 @@
...
+        setKey(null)
+        setKeyHash(null)
+      } catch (e) {
+        toaster.danger('error clearing vault ' + e.message)
+      }
+    }
+  })
+
+  const disconnectVault = useCallback(async () => {
+    beforeDisconnectVault?.()
```
huumn, line 66: `beforeDisconnectVault` desyncs wallets into localstorage again

components/vault/use-vault-configurator.js
```diff
@@ -0,0 +1,160 @@
...
+            throw new Error('wrong passphrase')
+          }
+          toaster.danger(error.graphQLErrors[0].message)
+        }
+      })
+
+      setKey(vaultKey)
+      setKeyHash(vaultKey.hash)
+      await set('key', vaultKey)
+      onVaultKeySet?.(encrypt).catch(console.error)
```
huumn, line 105: `onVaultKeySet` syncs any wallets stored in `localStorage` then removes them.

components/wallet-card.js
```diff
@@ -57,9 +49,9 @@ export default function WalletCard ({ wallet, draggable, onDragStart, onDragEnte
               </Badge>)}
         </Card.Subtitle>
       </Card.Body>
-      <Link href={`/settings/wallets/${wallet.name}`}>
+      <Link href={`/settings/wallets/${wallet.def.name}`}>
         <Card.Footer className={styles.attach}>
-          {wallet.isConfigured
+          {isConfigured(wallet)
```
huumn, line 54: A lot of the mutatable wallet members were replaced with pure function calls

fragments/wallet.js
```diff
@@ -106,92 +107,72 @@ mutation removeWallet($id: ID!) {
   removeWallet(id: $id)
 }
 `
-
 // XXX [WALLET] this needs to be updated if another server wallet is added
+export const WALLET_FIELDS = gql`
```
huumn, line 111: Now new wallets only have to modify this fragment

lib/validate.js
```diff
@@ -33,159 +30,6 @@ export async function ssValidate (schema, data, args) {
...
-
-export async function walletValidate (wallet, data) {
-  if (typeof wallet.fieldValidation === 'function') {
-    return await formikValidate(wallet.fieldValidation, data)
-  } else {
-    return await ssValidate(wallet.fieldValidation, data)
-  }
-}
-
-addMethod(string, 'or', function (schemas, msg) {
```
huumn, line 53: These monkey patches were moved to `lib/yup`

lib/validate.js
```diff
@@ -363,56 +187,44 @@ export function advSchema (args) {
...
-  }),
-  cert: hexOrBase64Validator,
-  ...autowithdrawSchemaMembers
+export const vaultEntrySchema = key => object({
+  key: string().required('required').matches(key, `expected ${key}`),
+  iv: string().required('required').hex().length(24, 'must be 24 characters long'),
+  value: string().required('required').hex().min(2).max(1024 * 10)
 })
 
-export const CLNAutowithdrawSchema = object({
```
huumn, line 393: Now we expect a `validate` member to be present on each of a wallet's `fields`

lib/wallet.js
```diff
@@ -1,17 +0,0 @@
-export function fieldToGqlArg (field) {
```
huumn, line 1: This file's contents were moved to `wallets/graphql.js`

lib/yup.js
```diff
@@ -0,0 +1,200 @@
...
+        string().required('secret required').trim().matches(/^[0-9a-fA-F]{64}$/, 'secret must be 64 hex chars').validateSync(secret)
+      } catch (err) {
+        return context.createError({ message: err.message })
+      }
+      return true
+    }
+  })
+})
+
+addMethod(array, 'equalto', function equals (
```
huumn, line 168: This is for validating `vaultEntries`

pages/settings/index.js
```diff
@@ -77,6 +77,11 @@ export function SettingsHeader () {
             <Nav.Link eventKey='mutes'>muted stackers</Nav.Link>
           </Link>
         </Nav.Item>
+        <Nav.Item>
+          <Link href='/settings/passphrase' passHref legacyBehavior>
+            <Nav.Link eventKey='passphrase'>device sync</Nav.Link>
+          </Link>
+        </Nav.Item>
```
huumn, line 84: `device sync` is its own tab in settings

wallets/common.js
```diff
@@ -0,0 +1,163 @@
...
+
+export function canSend ({ def, config }) {
+  return !!def.sendPayment && isSendConfigured({ def, config })
+}
+
+export function canReceive ({ def, config }) {
+  return def.fields.some(f => f.serverOnly) && isReceiveConfigured({ def, config })
+}
+
+export function siftConfig (fields, config) {
```
huumn, line 100: everything that needs to be saved/validated is returned in the appropriate member

wallets/errors.js
```diff
@@ -0,0 +1,22 @@
+export class InvoiceCanceledError extends Error {
```
huumn, line 1: These errors in `components/payment.js` were causing a cyclical dep so I yanked them out.

wallets/validate.js
```diff
@@ -0,0 +1,110 @@
...
+  let schema = composeWalletSchema(walletDef, serverSide, skipGenerated)
+
+  if (canReceive({ def: walletDef, config: data })) {
+    schema = schema.concat(autowithdrawSchemaMembers)
+  }
+
+  await schema.validate(data, yupOptions)
+
+  const casted = schema.cast(data, { assert: false })
+  if (topLevel && walletDef.validate) {
```
huumn, line 25: We added an optional top level validation to wallet modules so an entire wallet could be validated in aggregate.

wallets/validate.js
```diff
@@ -0,0 +1,110 @@
...
+
+    if (clientOnly && serverSide) {
+      // For server-side validation, accumulate clientOnly fields as vaultEntries
+      vaultEntrySchemas[optional ? 'optional' : 'required'].push(vaultEntrySchema(name))
+    } else {
+      acc[name] = createFieldSchema(name, validate)
+
+      if (!optional) {
+        acc[name] = acc[name].required('required')
+      } else if (requiredWithout) {
```
huumn, line 80: This switch implements the `or` logic to accommodate accepting either send or receive.

wallets/validate.js
```diff
@@ -0,0 +1,110 @@
...
+}
+
+function composeWalletSchema (walletDef, serverSide, skipGenerated) {
+  const { fields } = walletDef
+
+  const vaultEntrySchemas = { required: [], optional: [] }
+  const schemaShape = fields.reduce((acc, field) => {
+    const { name, validate, optional, generated, clientOnly, requiredWithout } = field
+
+    if (generated && skipGenerated) {
```
huumn, line 68: generated fields (LNC dh keys) are selectively skipped until we know they've been generated (ie after testing a payment/connection).

wallets/lnbits/index.js
```diff
@@ -1,28 +1,55 @@
...
+          return true
+        })
   },
   {
     name: 'invoiceKey',
     label: 'invoice key',
     type: 'password',
     optional: 'for receiving',
     serverOnly: true,
-    editable: false
```
huumn, line 17: All fields are editable again (except for LNC's passphrase).

wallets/index.js
```diff
@@ -1,452 +1,237 @@
...
-  Enabled: 'Enabled',
-  Locked: 'Locked',
-  Error: 'Error'
-}
+const WalletsContext = createContext({
+  wallets: []
+})
 
-export function useWallet (name) {
+function useLocalWallets () {
```
huumn, line 17: This handles the array of `wallets` derived from localstorage (if they exist).

wallets/index.js
```diff
@@ -1,452 +1,237 @@
...
-}
-
-function extractClientConfig (fields, config) {
-  return extractConfig(fields, config, true)
-}
-
-function extractServerConfig (fields, config) {
-  return extractConfig(fields, config, false)
-}
+const walletDefsOnly = walletDefs.map(w => ({ def: w, config: {} }))
```
huumn, line 50: We pull wallet definitions into an array so they can be treated like localWallets and serverWallets ... which is primarily useful for displaying them in the wallet grid.

wallets/index.js
```diff
@@ -1,452 +1,237 @@
...
-      if (valid) {
-        if (priorityOnly) {
-          setClientConfig(newClientConfig)
-        } else {
-          try {
-            // XXX: testSendPayment can return a new config (e.g. lnc)
-            const newerConfig = await wallet.testSendPayment?.(newConfig, { me, logger })
-            if (newerConfig) {
-              newClientConfig = Object.assign(newClientConfig, newerConfig)
+  }, [me?.privates?.walletsUpdatedAt, me?.privates?.vaultKeyHash, refetch])
```
huumn, line 68: walletsUpdatedAt is updated in a trigger on `Wallet` and `VaultEntry`

wallets/index.js
```diff
@@ -1,452 +1,237 @@
...
+      const wallets = []
+      for (const w of data.wallets) {
+        const def = getWalletByType(w.type)
+        const { vaultEntries, ...config } = w
+        if (isActive) {
+          for (const { key, iv, value } of vaultEntries) {
+            try {
+              config[key] = await decrypt({ iv, value })
+            } catch (e) {
+              console.error('error decrypting vault entry', e)
```
huumn, line 83: We leave fields that we fail to decrypt blank.

wallets/index.js
```diff
@@ -1,452 +1,237 @@
...
+    const merged = {}
+    for (const wallet of [...walletDefsOnly, ...localWallets, ...serverWallets]) {
+      merged[wallet.def.name] = {
+        def: wallet.def,
+        config: {
+          ...merged[wallet.def.name]?.config,
+          ...Object.fromEntries(
+            Object.entries(wallet.config ?? {}).map(([key, value]) => [
+              key,
+              value ?? merged[wallet.def.name]?.config?.[key]
```
huumn, line 109: If there's a `value` for `key` in a lower priority wallet, but the higher priority config (e.g. server) couldn't be decrypted, we keep `value` for `key`.

wallets/index.js
```diff
@@ -1,452 +1,237 @@
...
 
-    try {
-      await client.mutate({
-        mutation: REMOVE_WALLET,
-        variables: { id: walletId }
-      })
-    } finally {
-      client.refetchQueries({ include: ['WalletLogs'] })
-      refetchConfig()
+  const unsyncLocalWallets = useCallback(() => {
```
huumn, line 152: this is called when device sync is intentionally disconnected ... this way it preserves the config, ie reset does not necessarily result in data loss.

riccardobl:
Mhh, there is a conflict between synched and unsynched sessions.

repo:
- Log-in from two sessions:  **session1** and **session2**
- Enable device sync in **session1** only
- Attach a nwc sender in  **session1** 
- Attach a different nwc sender (i use the same url with &v=2) in **session2**

**session1** will detach and lose the sender config

components/nav/common.js
```diff
@@ -292,7 +292,7 @@ function LogoutObstacle ({ onClose }) {
               await togglePushSubscription().catch(console.error)
             }
 
-            await wallets.resetClient().catch(console.error)
+            removeLocalWallets().catch(console.error)
```
riccardobl, line 296: not async
```suggestion
            removeLocalWallets()
```

wallets/index.js
```diff
@@ -1,452 +1,237 @@
...
+  return (
+    <WalletsContext.Provider
+      value={{
+        wallets,
+        reloadLocalWallets,
+        setPriorities,
+        onVaultKeySet: syncLocalWallets,
+        beforeDisconnectVault: unsyncLocalWallets
+      }}
+    >
```
riccardobl, line 195: Missing ctx value

```suggestion
        removeLocalWallets
```

components/nav/common.js
riccardobl, line 296: We should call disconnectVault() here, so if the user logs out from a public computer or some other untrusted setup, it doesn't leave the vault key behind

huumn: good recs, thanks!

huumn:
> I've notice that if you have more devices with conflicting configs, the last one to join the vault overwrite the others, maybe from an UX standpoint it is better to do the other way around, so you can connect first the main device and then let the other merge the missing configs.

Nice catch. I didn't test this combo. An empty, unsynced config shouldn't overwrite a synced config.

huumn:
> I've notice that if you have more devices with conflicting configs, the last one to join the vault overwrite the others, maybe from an UX standpoint it is better to do the other way around, so you can connect first the main device and then let the other merge the missing configs.

A generalized description of this problem is that, as a result of this pr, an unsynced wallet will not be aware of the vault entries so - when there aren't receive creds on save, it deletes the server wallet. However an unsynced wallet should be allowed to remove the receive creds.

The clearest solution likely introduces a new mutation to only remove receive creds, but that's not very easy because our wallet mutations are generated.

Anyway, that's where I'm at on this problem. Thanks for finding it!

huumn:
> I've notice that if you have more devices with conflicting configs, the last one to join the vault overwrite the others, maybe from an UX standpoint it is better to do the other way around, so you can connect first the main device and then let the other merge the missing configs.

I think we're going to let it fly as is. You're right though that the first device might be their primary and have more important config. I just don't think it's likely these conflicts happen much in practice, this pr has taken me long enough, and we plan to excavate most of this code again to refine the attachment flows.

PR #1510, opened 2024-10-23
Fix more button shown after logs deleted

ekzyis:
## Description

If a `more` button was shown and all wallet logs are deleted, the button was still shown.

## Checklist

yes

`5`. I deleted wallet logs and more button disappeared.

n/a

no

### DIFF
--- a/components/wallet-logger.js
+++ b/components/wallet-logger.js
@@ -248,5 +248,10 @@ export function useWalletLogs (wallet, initialPage = 1, logsPerPage = 10) {
     loadLogs()
   }, [wallet])
 
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


huumn (changes requested):
Does this need to be done with a side effect.

Writing a `clearLogs` function should do this in a more maintainable way, shouldn't it?

ekzyis:
The issue I had before was that `deleteLogs` from `useWalletLogger` is in a context where `setHasMore` is not available.

But I replaced the side effect by wrapping `setLogs` now since `deleteLogs` calls `setLogs` where I then have access to `setHasMore`.

PR #1511, opened 2024-10-23
Fix wallet validation blocking saves

### LINKED ISSUE #1501: Wallets that are failing can’t be disabled due to test invoice, validation
As stated in #1500, my wallet which ordinarily works is failing in my current network context. I’d like to simply disable it temporarily to use qr but I can’t disable it because it can only be saved if the wallet works - which it doesn’t at the moment.

ekzyis:
## Description

This fixes #1501 and the following unreported issue:

If you had WebLN enabled and then disabled the extension but kept it enabled as a wallet on SN, priority changes to that wallet were blocked because client validation via schema (silently) failed.

This was fixed by 355feb61 which makes the behavior of priority changes consistent for wallets stored on the server and client.

Other changes:

- 5a09b2b2 disables the WebLN wallet if `window.webln` is no longer available. I think that's a good idea.

- I replaced the local storage call to disable/enable wallets on triggers (like on `webln:enabled`) with a regular `saveConfig` call like we do for priority changes. Validation is also unnecessary in that case, so I renamed `priorityOnly` to `skipValidation` in cee7cbb2.

Since cee7cbb2 still runs validation on regular save so didn't fix #1501, I updated `saveConfig` to always skip validation if the wallet is about to get disabled in 5d0fc923.

## Additional Context

- The unreported issue will probably not exist in #1507 so feel free to close this if we can wait for the fix in it
- I believe #1501 is also fixed in #1460

## Checklist

yes

`7`. I was able to change WebLN wallet priority even though WebLN extension was disabled. I was still able to attach send+recv wallets. When I now disable a wallet, validation on client and server is skipped.

n/a

no

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -30,9 +30,9 @@ function injectResolvers (resolvers) {
     const resolverName = generateResolverName(w.walletField)
     console.log(resolverName)
 
-    resolvers.Mutation[resolverName] = async (parent, { settings, priorityOnly, ...data }, { me, models }) => {
+    resolvers.Mutation[resolverName] = async (parent, { settings, skipValidation, ...data }, { me, models }) => {
       // allow transformation of the data on validation (this is optional ... won't do anything if not implemented)
-      if (!priorityOnly) {
+      if (!skipValidation) {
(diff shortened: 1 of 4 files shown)


huumn:
I think this is superseded by #1507

PR #1515, opened 2024-10-24
Fix unknown query refetch attempt on priority change

ekzyis:
## Description

Fix #1504 

## Additional Context

> [!NOTE]
> This conflicts with #1511 because of the rename of `priorityOnly` to `skipValidation`

## Checklist

**Are your changes backwards compatible? Please answer below:**

yes

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**

`6`. I verified that wallet logs are still refetched after save but no longer if priority is changed.

**For frontend changes: Tested on mobile? Please answer below:**

n/a

**Did you introduce any new environment variables? If so, call them out explicitly here:**

no

PR #1516, opened 2024-10-24
supercharged wallet logs

### LINKED ISSUE #1380: Surface all relevant payment problems/successes in wallet logs
I feel like, especially on the receiving side, we aren't logging everything we should be.

It's probably worth spending a day and really going over every opportunity to add something to wallet logs and do it.

ekzyis:
## Description

Close #1380

based on #1510, #1511, #1515 

TODO:

- [x] wallet logger interface: replace `addWalletLog` with logger interface
- [x] responsive receive logs: don't refetch receive logs after save, poll during save
- [x] more (receive) logs: incoming payments; amount of payments; invoice info; info around errors
- [x] test various errors on unhappy paths
- [x] fix: quick zaps in succession that fail might cause the QR modal to miss some invoices (fixed in #1557)
- [x] fix: quick zaps can cause `failed to create stacker invoice, falling back to SN invoice Error: value is required` (fixed in 44b169a5)
- [x] ~fix: invoice description of unwrapped invoices isn't stored~ _mhh, can't reproduce anymore_

## Video

Current status (44b169a5):

https://github.com/user-attachments/assets/7bf7afc8-19c3-4b53-b54d-408efa3d43f1

## Additional Context

- we currently only log `wallet attached` / `wallet details updated` and `wallet enabled` / `wallet disabled` for receives so if you only attach spending credentials, there's no such log. ideally, we would only log these once, no matter if send and/or recv was attached. but how without creating a mess? previously, we logged them separately but that was noisy.

## Checklist

yes

`8`. Tried really hard to make sure I don't miss any possible error but I probably missed something anyway. However, if wallets work fine I am pretty sure this code is solid. Here's a video of two zapping each other (NWC<>LNbits) really hard with no issues, see video section.

yes, looks good to me:

localhost_3000_wallet_logs(iPhone SE) (6)

no


ekzyis:
Current status: a2db358a

* on successful autowithdrawal:

```
1m	[lnbits]	OK	↙ payment received: 1,000,000 msats
1m	[lnbits]	INFO	created invoice for 1,000,000 msats
1m	[lnbits]	INFO	↙ incoming payment: 1,000,000 msats
```

* if wallet failed to create invoice:

```
6s	[lnbits]	ERROR	failed to create invoice: test error
6s	[lnbits]	INFO	↙ incoming payment: 1,000,000 msats
```

* if wallet does not support msats:

```
4m	[lnbits]	OK	↙ payment received: 9,000 msats
4m	[lnbits]	WARN	wallet does not support msats
4m	[lnbits]	INFO	created invoice for 9,000 msats
4m	[lnbits]	INFO	↙ incoming payment: 9,900 msats
```

* if withdrawal itself failed:

```
1m	[lnbits]	ERROR	withdrawal failed: no route found
1m	[lnbits]	INFO	created invoice for 1,000,000 msats
1m	[lnbits]	INFO	↙ incoming payment: 1,000,000 msats
```

TODO:

- [ ] p2p zap logs on error and success
- [ ] error handling (HTTP vs JSON)
- [ ] logs that includes details like payment hash, link to invoice etc. when clicked/uncollapsed (?)

ekzyis:
How "rich log" looks so far:

https://github.com/user-attachments/assets/e97bd791-bae9-4d18-ac7e-9a6e1ce14292

(not really rich yet since it's just text, no links to anywhere)

I am not super happy with `+` and `-` as indicators that the log can be expanded / collapsed but the SVGs that we use in `ContextAwareToggle` don't align well in the table. Same for these unicode arrows: 🢓, 🢒 (as you can see the arrow-down is not vertically aligned)

ekzyis:
Current status (a610b933):

https://github.com/user-attachments/assets/1a9b1390-03ff-4a1b-b391-2b998165f891

I believe there is a bug where we don't store the invoice description since it's empty in the database even though the stored bolt11 does contain a description (see missing invoice description in last receiver log).

api/resolvers/wallet.js
```diff
@@ -663,16 +666,34 @@ const resolvers = {
...
-export const addWalletLog = async ({ wallet, level, message }, { models }) => {
-  try {
-    await models.walletLog.create({ data: { userId: wallet.userId, wallet: wallet.type, level, message } })
-  } catch (err) {
-    console.error('error creating wallet log:', err)
+export const walletLogger = ({ wallet, models }) => {
+  // server implementation of wallet logger interface on client
+  const log = (level) => async (message, context = {}) => {
+    try {
+      await models.walletLog.create({
```
ekzyis, line 673: no `console.log` here since I think wallets logs inside the server logs would be more noise than signal

api/resolvers/wallet.js
```diff
@@ -54,20 +54,25 @@ function injectResolvers (resolvers) {
...
       return await upsertWallet({
-        wallet: {
-          field: walletDef.walletField,
-          type: walletDef.walletType
-        },
+        wallet,
         testCreateInvoice:
           walletDef.testCreateInvoice && validateLightning && canReceive({ def: walletDef, config: data })
-            ? (data) => walletDef.testCreateInvoice(data, { me, models })
+            ? (data) => walletDef.testCreateInvoice(data, { logger, me, models })
```
ekzyis, line 69: we pass `logger` to `testCreateInvoice` of the wallets but we only use the logger for NWC. So far, I saw no need to add wallet-specific logging to the other wallets.

lib/url.js
```diff
@@ -213,6 +213,13 @@ export function parseNwcUrl (walletConnectUrl) {
   return params
 }
 
+export function assertContentTypeJson (res) {
+  const contentType = res.headers.get('content-type')
+  if (!contentType || !contentType.includes('application/json')) {
+    throw new Error(`POST ${res.url}: ${res.status} ${res.statusText}`)
```
ekzyis, line 219: no more gigantic error messages about how we couldn't parse the HTML body as JSON when a reverse proxy throws a gateway timeout

wallets/lnc/client.js
```diff
@@ -26,7 +26,7 @@ async function disconnect (lnc, logger) {
       }, 50)
       logger.info('disconnected')
     } catch (err) {
-      logger.error('failed to disconnect from lnc', err)
+      logger.error('failed to disconnect from lnc: ' + err)
```
ekzyis, line 29: `logger` no longer supports a variable number of arguments because second argument is now context

worker/wallet.js
```diff
@@ -294,7 +292,7 @@ export async function checkWithdrawal ({ data: { hash, withdrawal, invoice }, bo
 
     const fee = Number(wdrwl.payment.fee_mtokens)
     const paid = Number(wdrwl.payment.mtokens) - fee
-    const [{ confirm_withdrawl: code }] = await serialize([
+    const [[{ confirm_withdrawl: code }]] = await serialize([
```
ekzyis, line 295: for some reason the return type changed at some point

worker/wallet.js
```diff
@@ -305,35 +303,18 @@ export async function checkWithdrawal ({ data: { hash, withdrawal, invoice }, bo
     ], { models })
     if (code === 0) {
       notifyWithdrawal(dbWdrwl.userId, wdrwl)
-      if (dbWdrwl.wallet) {
-        // this was an autowithdrawal
-        const message = `autowithdrawal of ${
-          numWithUnits(msatsToSats(paid), { abbreviate: false })} with ${
-          numWithUnits(msatsToSats(fee), { abbreviate: false })} as fee`
-        await addWalletLog({ wallet: dbWdrwl.wallet, level: 'SUCCESS', message }, { models })
```
ekzyis, line 313: we don't need this because we log success/errors using `.then` and `.catch` on `payViaPaymentRequest`

ekzyis:
Okay, I don't know how to cause route errors from SN to recipient with our current local setup. I tried the following:

1. Raising channel fees but they don't apply in our local lightning network since nodes only pay fees for channels from other nodes but that's not the case in our local setup for wrapped invoices. The sender uses their own channel to pay SN and SN uses its own channel to pay recipient so no fees apply. I'd have to add a node that does nothing but route a payment between SN and recipient to get fees. I didn't want to setup a whole lightning node just to test route errors. Since it's also a circular payment, I didn't consider to close channels since they will be used for the incoming payment, too.[^1]

[^1]: Mhh, writing this out gave me an idea: if I use `stacker_lnd` for NWC and `stacker_cln` for LNbits, maybe I can simply close all channels between `sn_lnd` and `stacker_lnd` to trigger route errors :thinking: 

2. I thought I could maybe patch `nostr-wallet-connect-lnd` again or something but since I want to trigger a payment but make it fail on the receiver side (as far as the sender can tell), I would have to intercept the HTLC inside the receiving node to cause some failure. I didn't want to mess with any LND internals just to test route errors.

3. I tried to use this patch but seems like local pathfinding is faster than literally 0ms?

```diff
diff --git a/worker/paidAction.js b/worker/paidAction.js
index cb76a267..2bfe40b0 100644
--- a/worker/paidAction.js
+++ b/worker/paidAction.js
@@ -255,7 +255,7 @@ export async function paidActionForwarding ({ data: { invoiceId, ...args }, mode
       lnd,
       request: bolt11,
       max_fee_mtokens: String(maxFeeMsats),
-      pathfinding_timeout: LND_PATHFINDING_TIMEOUT_MS,
+      pathfinding_timeout: 0,
       max_timeout_height: maxTimeoutHeight
     })
```

So I wasn't really able to test any errors after the wrapped invoice was paid except simply not calling `payViaPaymentRequest` for forwarding at all. Since I wanted to have invoices expire quick for a quick feedback cycle, I patched some more (innocent) code:

```diff
diff --git a/wallets/nwc/server.js b/wallets/nwc/server.js
index ce222087..3860bc66 100644
--- a/wallets/nwc/server.js
+++ b/wallets/nwc/server.js
@@ -32,7 +32,7 @@ export async function createInvoice (
     params: {
       amount: msats,
       description,
-      expiry
+      expiry: 15
     }
   }, { logger })
   return result.invoice
diff --git a/wallets/wrap.js b/wallets/wrap.js
index b470587b..a2f991ec 100644
--- a/wallets/wrap.js
+++ b/wallets/wrap.js
@@ -118,15 +118,17 @@ export default async function wrapInvoice (bolt11, { msats, description, descrip
     }
 
     // validate the expiration
-    if (new Date(inv.expires_at) < new Date(Date.now() + INCOMING_EXPIRATION_BUFFER_MSECS)) {
-      throw new Error('Invoice expiration is too soon')
-    } else if (new Date(inv.expires_at) > new Date(Date.now() + MAX_EXPIRATION_INCOMING_MSECS)) {
-      // trim the expiration to the maximum allowed with a buffer
-      wrapped.expires_at = new Date(Date.now() + MAX_EXPIRATION_INCOMING_MSECS - INCOMING_EXPIRATION_BUFFER_MSECS)
-    } else {
-      // give the existing expiration a buffer
-      wrapped.expires_at = new Date(new Date(inv.expires_at).getTime() - INCOMING_EXPIRATION_BUFFER_MSECS)
-    }
+    // if (new Date(inv.expires_at) < new Date(Date.now() + INCOMING_EXPIRATION_BUFFER_MSECS)) {
+    //   throw new Error('Invoice expiration is too soon')
+    // } else if (new Date(inv.expires_at) > new Date(Date.now() + MAX_EXPIRATION_INCOMING_MSECS)) {
+    //   // trim the expiration to the maximum allowed with a buffer
+    //   wrapped.expires_at = new Date(Date.now() + MAX_EXPIRATION_INCOMING_MSECS - INCOMING_EXPIRATION_BUFFER_MSECS)
+    // } else {
+    //   // give the existing expiration a buffer
+    //   wrapped.expires_at = new Date(new Date(inv.expires_at).getTime() - INCOMING_EXPIRATION_BUFFER_MSECS)
+    // }
+
+    wrapped.expires_at = inv.expires_at
 
     // get routing estimates
     const { routingFeeMsat, timeLockDelay } =
diff --git a/worker/paidAction.js b/worker/paidAction.js
index 052f4e75..c06f9159 100644
--- a/worker/paidAction.js
+++ b/worker/paidAction.js
@@ -249,13 +249,13 @@ export async function paidActionForwarding ({ data: { invoiceId, ...args }, mode
     console.log('forwarding with max fee', maxFeeMsats, 'max_timeout_height', maxTimeoutHeight,
       'accept_height', acceptHeight, 'expiry_height', expiryHeight)
 
-    payViaPaymentRequest({
-      lnd,
-      request: bolt11,
-      max_fee_mtokens: String(maxFeeMsats),
-      pathfinding_timeout: LND_PATHFINDING_TIMEOUT_MS,
-      max_timeout_height: maxTimeoutHeight
-    }).catch(console.error)
+    // payViaPaymentRequest({
+    //   lnd,
+    //   request: bolt11,
+    //   max_fee_mtokens: String(maxFeeMsats),
+    //   pathfinding_timeout: LND_PATHFINDING_TIMEOUT_MS,
+    //   max_timeout_height: maxTimeoutHeight
+    // }).catch(console.error)
   }
 }
```

While doing this, I noticed the following: **If invoices expire earlier than `LND_PATHFINDING_TIMEOUT` * 2 (= 1 minute), the incoming invoice will never be canceled because the `checkInvoice` schedule will run out of retries before it hits this 2x timeout escape hatch:**

https://github.com/stackernews/stacker.news/blob/4cc3780eca1f1390bdd331f0e418b9a7efd4347e/worker/wallet.js#L278-L288

Since invoices can't be that short in prod, this can't happen in prod but I still wanted to raise awareness about this interplay between job retries and pathfinding timeout.

To still have a quick feedback cycle, I decreased `LND_PATHFINDING_TIMEOUT` to 5 seconds:

```diff
diff --git a/lib/constants.js b/lib/constants.js
index ad327236..8968d3bd 100644
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -72,7 +72,7 @@ export const OLD_ITEM_DAYS = 3
 export const ANON_FEE_MULTIPLIER = 100
 export const SSR = typeof window === 'undefined'
 export const MAX_FORWARDS = 5
-export const LND_PATHFINDING_TIMEOUT_MS = 30000
+export const LND_PATHFINDING_TIMEOUT_MS = 5000
 export const LNURLP_COMMENT_MAX_LENGTH = 1000
 export const RESERVED_MAX_USER_ID = 615
 export const GLOBAL_SEED = USER_ID.k00b
```

This means I was only able to test following cases:

1. payer cancels invoice

https://github.com/user-attachments/assets/9aaee7f0-29e8-4df8-b509-9cf51df23ab2

2. incoming invoice expires before it's forwarded (see patches above):

https://github.com/user-attachments/assets/642b76dd-8a3f-4a97-961a-8f06fefbe3c4

3. successful autowithdrawal:

https://github.com/user-attachments/assets/03579d35-ff60-4091-9ab8-f7bdef45c533

However, since the (error) logging is within the payment flow, it should work nicely with it.

ekzyis:
Mhh, if I zap too fast (like 3 quick zaps is already too much), I get Prisma errors which kill the worker:

```
error creating wallet log: PrismaClientKnownRequestError:
    Invalid `prisma.walletLog.create()` invocation:


    Timed out fetching a new connection from the connection pool. More info: http://pris.ly/d/connection-pool (Current connection pool timeout: 10, connection limit: 2)
        at Mn.handleRequestError (/app/node_modules/@prisma/client/runtime/library.js:121:7753)
        at Mn.handleAndLogRequestError (/app/node_modules/@prisma/client/runtime/library.js:121:7061)
        at Mn.request (/app/node_modules/@prisma/client/runtime/library.js:121:6745)
        at async l (/app/node_modules/@prisma/client/runtime/library.js:130:9633)
        at async file:///app/api/resolvers/wallet.js:694:7
        at async transition (file:///app/worker/paidAction.js:283:7)
        at async models.$transaction.isolationLevel (file:///app/worker/paidAction.js:69:20)
        at async Proxy._transactionWithCallback (/app/node_modules/@prisma/client/runtime/library.js:130:8000)
        at async transitionInvoice (file:///app/worker/paidAction.js:37:33)
        at async paidActionForwarded (file:///app/worker/paidAction.js:264:10)
        at async checkWithdrawal (file:///app/worker/wallet.js:294:14)
        at async EventEmitter.<anonymous> (file:///app/worker/wallet.js:226:9) {
      code: 'P2024',
      clientVersion: '5.20.0',
      meta: { modelName: 'WalletLog', connection_limit: 2, timeout: 10 }
    }
    error running hodl_invoice_updated PrismaClientKnownRequestError:
    Invalid `prisma.invoice.findUnique()` invocation:


    Timed out fetching a new connection from the connection pool. More info: http://pris.ly/d/connection-pool (Current connection pool timeout: 10, connection limit: 2)
        at Mn.handleRequestError (/app/node_modules/@prisma/client/runtime/library.js:121:7753)
        at Mn.handleAndLogRequestError (/app/node_modules/@prisma/client/runtime/library.js:121:7061)
        at Mn.request (/app/node_modules/@prisma/client/runtime/library.js:121:6745)
        at async l (/app/node_modules/@prisma/client/runtime/library.js:130:9633)
        at async checkInvoice (file:///app/worker/wallet.js:127:17)
        at async EventEmitter.<anonymous> (file:///app/worker/wallet.js:105:9) {
      code: 'P2024',
      clientVersion: '5.20.0',
      meta: { modelName: 'Invoice', connection_limit: 2, timeout: 10 }
    }
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

PrismaClientKnownRequestError:
Invalid `prisma.invoice.findUnique()` invocation:


Timed out fetching a new connection from the connection pool. More info: http://pris.ly/d/connection-pool (Current connection pool timeout: 10, connection limit: 2)
    at Mn.handleRequestError (/app/node_modules/@prisma/client/runtime/library.js:121:7753)
    at Mn.handleAndLogRequestError (/app/node_modules/@prisma/client/runtime/library.js:121:7061)
    at Mn.request (/app/node_modules/@prisma/client/runtime/library.js:121:6745)
    at async l (/app/node_modules/@prisma/client/runtime/library.js:130:9633)
    at async checkInvoice (file:///app/worker/wallet.js:127:17)
    at async EventEmitter.<anonymous> (file:///app/worker/wallet.js:105:9) {
  code: 'P2024',
  clientVersion: '5.20.0',
  meta: { modelName: 'Invoice', connection_limit: 2, timeout: 10 }
}

Node.js v18.20.4
Failed running 'worker/index.js'
```

Hopefully this is indeed the last TODO.

worker/wallet.js
ekzyis: I changed this to log in `paidActionForwarded` etc.

ekzyis:
Regarding the occasional `value is required` error during wrapping: I noticed that `getBlockHeight` sometimes returns `undefined`. I think that happens when the cache got stale and then we use a `stale-then-revalidate` strategy which means we return `undefined` and fetch the new value in the background.

api/resolvers/wallet.js
```diff
@@ -406,7 +411,7 @@ const resolvers = {
             userId: me.id,
             wallet: type ?? undefined,
             createdAt: {
-              gte: from ? new Date(Number(from)) : undefined,
+              gt: from ? new Date(Number(from)) : undefined,
```
ekzyis, line 414: make sure we don't fetch the last log over and over again. maybe not significant in terms of bandwidth but I think non-overlapping pages are better.

components/wallet-logger.js
```diff
@@ -27,8 +30,15 @@ export function WalletLogs ({ wallet, embedded }) {
...
       </div>
-      <div className={`${styles.logTable} ${embedded ? styles.embedded : ''}`}>
+      <div className={`${styles.tableContainer} ${embedded ? styles.embedded : ''}`}>
         <table>
+          <colgroup>
+            <col span='1' style={{ width: '1rem' }} />
+            <col span='1' style={{ width: '1rem' }} />
+            <col span='1' style={{ width: '1rem' }} />
+            <col span='1' style={{ width: '100%' }} />
+            <col span='1' style={{ width: '1rem' }} />
```
ekzyis, line 40: This is kind of a hack to use as much space as possible for the actual log message and push `+` to the end. Instead of `width: '1rem'` I could have even used `width: 0px` since table columns seem to reserve width for the space they need independent of what I set here :thinking: 

Maybe I should have used a CSS grid when I created the wallet logs but table also works now ...

That reminds me I considered to add alternating colors for the rows but that messed with the context since the context rows should stay the same color.

wallets/server.js
```diff
@@ -41,90 +42,125 @@ export async function createInvoice (userId, { msats, description, descriptionHa
...
-          description: wallet.user.hideInvoiceDesc ? undefined : description,
-          descriptionHash,
-          expiry
-        }, walletFull[walletField]), 10_000)
-
-      const bolt11 = await parsePaymentRequest({ request: invoice })
       if (BigInt(bolt11.mtokens) !== BigInt(msats)) {
         if (BigInt(bolt11.mtokens) > BigInt(msats)) {
-          throw new Error(`invoice is for an amount greater than requested ${bolt11.mtokens} > ${msats}`)
+          throw new Error('invoice invalid: amount too big')
```
ekzyis, line 78: we don't need to add specific info into the log message anymore since the exact values can be derived from the context now

prisma/schema.prisma
```diff
@@ -245,6 +245,7 @@ model WalletLog {
   wallet    WalletType
   level     LogLevel
   message   String
+  context   Json?      @db.JsonB
```
huumn, line 248: I'm curious why you used JSON here? Does context vary a lot or do you anticipate that it will?

huumn (approved):
It's really nice! I couldn't fault it and there a lots of neat bits to it, ie the polling and the click to expand.

The only thing I did notice is that the use of `JSON` context suggests an opportunity to establish a relationship between invoices/withdrawals/forwards and wallet log rows - one that we might be able to use in other places (like showing the logs relating to a specific action/invoice/withdrawal).

For now, I think not having this relationship is fine because we don't have any immediate plans for it. But, it's worth flagging - generically - because these relationships allow us to build even more comprehensive views of what's going on and people tend to really like that.

prisma/schema.prisma
ekzyis: I anticipated that we might want to add more to it and I didn't think that relationships would be important for logs. If a log has a wrong relationship because the database didn't enforce it, it will lose relevance over time because logs become less relevant over time.

For example, we could add `invoiceId` and don't have to care about migrating previous logs in the database with JSON. The client will have to care about old logs for a while until we delete them. We don't yet but this is what I anticipated we will do at some point.

ekzyis:
You reminded me that I thought it would be cool if I click on `zap failed` in my notification and it sends me to the relevant log line. I think we would need relationships for that and JSON wouldn't work because it only allows to go from log to somewhere else.

Issue #1496, opened 2024-10-18
make lnurl-pay and lightning address send to attached wallets


huumn:
Additionally, we'll want:

- option to receive CCs if attached wallet fails
- option to have SN proxy for the receiver for a 10% fee

riccardobl:
do we consider this solved by https://github.com/stackernews/stacker.news/pull/1570 ?

huumn:
I'll close as it's now #1591

Issue #1494, opened 2024-10-18
attached wallet fallbacks on failure in order of priority


huumn:
For zaps, this is more complicated than it sounds because either party in a p2p zap can cause it to fail and we only want to "fallback" for the failing side.

Also, if all fallbacks fail for a zap, should the last resort be cowboy credits?

ekzyis:
> Also, if all fallbacks fail for a zap, should the last resort be cowboy credits?

This is how it currently works in #1688. I think we mentioned this might be a setting but it's not for now.

huumn:
I don't think it will end up needing to be a setting, but we can see how people react over time.

PR #1523, opened 2024-10-27
receiver and sender fallback

riccardobl:
The PR implements fallbacks on receiver, lnurlp for attached wallets and does some refactoring to the wallet interface.

Closes https://github.com/stackernews/stacker.news/issues/1496
Closes https://github.com/stackernews/stacker.news/issues/1494
partially solves https://github.com/stackernews/stacker.news/issues/1558 by cancelling and creating a new invoice before showing the qr code

Changelog:
- **made pessimistic actions retryable**
- **implemented receiver fallback**
- **implemented sender fallback** 

-------------------
## Overview

This PR enhances payment reliability between sender and receiver wallets by adding fallback mechanisms.

The first part ensures that if a receiver’s wallet fails to generate a valid and wrappable invoice, the system will automatically attempt to create the invoice with the next available wallet, repeating this process until a valid invoice is obtained.

The second part addresses situations where a sender might have difficulty paying the invoice. Instead of simply trying a single payment path, the PR introduces a retry system that cycles through different sender-receiver wallet pairs. This way, if the first pair can’t connect, the system will increment an attempt counter and try another receiver wallet to find a working path. This approach helps avoid payment failures caused by both wallets being online but unable to connect directly.

If these attempts still don’t succeed, the system will prioritize fee credits or reward sats. ie. if a  p2p payment was prioritized initially by the action but fails, the system will reprioritize internal payment options. This prioritization allows the payment to go through automatically if sufficient funds are available. If these methods do not work, the system will fallback to the next paymentMethod in the action, and generate a new invoice, which the client displays as a QR code for the user to complete the payment manually.

## P2P preference flow

When a paid action is tried, the serverside code will go through the payment methods supported by the action, following the priority order. This behavior can be mutated by passing:
- `forceInternal` : to use **only** the internal payment methods that are supported by the action (currently only FEE_CREDIT)
- `prioritizeInternal` : to move the internal payment methods at the beginning in the priority list, making them the preferable methods while keeping all the others as fallbacks.

The paid action mutation returns a `retriable` flag that signals the sender it can retry the action using `retryPaidMutation` without changing anything, but just by increasing the `attempt` counter. This means the server side code has other ways to make the the transaction happen that were not tried yet. In the scope of this PR it means it has more p2p attachments to try.

Whenever a payment fails, the clientside code queries the paidAction to check if there is an invoiceForward's withdrawl status
-  if this is present the clientside code can determine that the payment went through on his side but failed to be forwarded on the other side, this will make the clientside code increase the `attempt` counter  and retry if `retriable` is true, to let the serverside code use the next p2p attachment on the receiving side
- if this is **not** present the clientside code will mark the sender as `failed` and retry with the next sender without increasing the `attempt` counter. This in turn will make the serverside code reuse the same receiving attachment.

If all the senders have failed or the retriable flag is false (all the receiver have failed), the clientside code will fallback to retry the payment using `prioritizeInternal` to hopefully complete it with FEE_CREDITS, if this fails too an invoice will be returned and passed to the qr code payment.

Note: this is not explicit in the code, but it is an effect of the logic: if we previously tried all the receivers and the forward failed for all of them, we will have an attempt counter high enough to exceed the number of receiving wallets, this will skip the p2p method, as it should, since there is no way to pay, even by qr code.

## Non P2P flow (not internal priority)

The paid actions is tried, the response is returned with a retriable flag always set to false and the invoice to pay.
The clientside code tries to pay with every sender, once all the senders have failed it fallbacks to use `prioritizeInternal`, if this fails, an invoice is returned for the qr code

## Non P2P flow (internal priority)

The paid actions is tried, the response is returned with a retriable flag always set to false and the invoice to pay.
The clientside code tries to pay with every sender, once all the senders have failed it fallbacks to use `prioritizeInternal`(FIXME: this is useless, it attempts the same thing twice, it should just skip instead)

 if this fails, an invoice is returned for the qr code.

## Note on invoice reuse

Every time a new attempt is made, the old invoice is cancelled and a new invoice is regenerated

-------------------

### setup
- stacker1:  
    -  fee credits: 1000
    -  sending attachments `[good, bad]`
    - receiving attachments` []`
  
- stacker2:
    - fee credits: 0
    - sending attachments: `[bad, good]`
    - receiving attachments: `[]`
  
- stacker3:
  - fee credits: 1000
  - sending attachments:` [bad,bad]`
  - receiving attachments` []`

- stacker4:
    - fee credits: 0
    - sending attachments:` []`
    - receiving attachments:` [good, bad]`

- stacker5:
  - fee credits: 0
  - sending attachments: `[]`
  - receiving attachments:` [bad, good]`
  
- stacker6: 
    - fee credits: 0
    - sending attachments: `[bad]`
    - receiving attachments:` [bad, bad]`

- stacker7:
    - fee credits: 0
    - sending attachments:` []`
    - receiving attachments: `[]`
        

- stacker8:
   - fee credits: 1 000 000
   - sending attachments: `[]`
   - receiving attachments: `[]`

###  Zap
| **send**<br>**receive** |stacker1 | stacker2 | stacker3| stacker4|
|-----|----|----|----|----|
|stacker4| [x] p2p |  [x] p2p|  [x] cc |- |
|sacker5| [x] p2p |  [x] p2p|  [x] cc | [x] qr|
|stacker6| [x] cc|  [x ] cc|  [x] cc |  [x] qr |
|stacker7| [x] cc|  [x] cc| [x] cc |  [x] qr  |     

### post result
- [x] stacker1 posts spending fee credits
-  [x] stacker2 posts spending from attached wallet
- [x] stacker3 posts spending  fee credits
- [x] stacker4  shows qr 
- [x] stacker5  shows qr
- [x] stacker6 fails, shows qr on manual retry

### LNURLP
-  [x] stacker3: cc
- [x] stacker4: p2p
- [x] stacker5: p2p
- [x] stacker6: cc

### Territory
- [x] stacker1 pays from attached wallet
- [x] stacker3 pays from qr code *(bug: if you cancel the qr code it goes to error 404)*
- [x] stacker8 pays from fee credits


riccardobl:
on hold until https://github.com/stackernews/stacker.news/pull/1507 is merged, i will rebase any change on top of the new code

components/notifications.js
```diff
@@ -484,13 +489,19 @@ function Invoicification ({ n: { invoice, sortTime } }) {
...
     <div className='fw-bold text-info'>
       <Check className='fill-info me-1' />{numWithUnits(n.earnedSats + n.withdrawl.satsFeePaid, { abbreviate: false, unitSingular: 'sat was ', unitPlural: 'sats were ' })}
       {n.withdrawl.p2p || n.withdrawl.autoWithdraw ? 'sent to your attached wallet' : 'withdrawn from your account'}
       <small className='text-muted ms-1 fw-normal' suppressHydrationWarning>{timeSince(new Date(n.sortTime))}</small>
       {(n.withdrawl.p2p && <Badge className={styles.badge} bg={null}>p2p</Badge>) ||
       (n.withdrawl.autoWithdraw && <Badge className={styles.badge} bg={null}>autowithdraw</Badge>)}
+      {n.withdrawl.invoiceForward?.[0]?.invoice?.comment &&
+        <small className='d-block ms-4 ps-1 mt-1 mb-1 text-muted fw-normal'>
+          <Text>{n.withdrawl.invoiceForward[0].invoice.comment}</Text>
+          {payerSig}
```
riccardobl, line 503: with p2p lnurlp the notification will be a withdraw notification instead of deposit, so we implement the code to show invoice comments on withdraw too

lib/constants.js
```diff
@@ -182,3 +185,11 @@ export const LONG_POLL_INTERVAL = Number(process.env.NEXT_PUBLIC_LONG_POLL_INTER
 export const EXTRA_LONG_POLL_INTERVAL = Number(process.env.NEXT_PUBLIC_EXTRA_LONG_POLL_INTERVAL)
 
 export const ZAP_UNDO_DELAY_MS = 5_000
+
+// the fee for the zap sybil service
+export const ZAP_SYBIL_FEE_PERCENT = 30n
+export const MAX_FEE_ESTIMATE_PERMILE = 25n // the maximum fee relative to outgoing we'll allow for the fee estimate
```
riccardobl, line 191: reduce bigint <-> number conversions

lib/validate.js
```diff
@@ -536,3 +515,83 @@ export const deviceSyncSchema = object().shape({
...
+}
+
+/**
+ * @param {number|bigint} x
+ * @returns {number|bigint}
+ */
+export const toPositive = (x) => {
+  if (typeof x === 'bigint') return toPositiveBigInt(x)
+  return toPositiveNumber(x)
+}
```
riccardobl, line 597: Just helpers to strictly handle bigints

pages/api/lnurlp/[username]/pay.js
```diff
@@ -30,22 +29,23 @@ export default async ({ query: { username, amount, nostr, comment, payerdata: pa
...
+        description = 'zap'
         descriptionHash = createHash('sha256').update(noteStr).digest('hex')
       } else {
         res.status(400).json({ status: 'ERROR', reason: 'invalid NIP-57 note' })
         return
       }
     } else {
-      description = user.hideInvoiceDesc ? undefined : `Funding @${username} on stacker.news`
+      description = `Funding @${username} on stacker.news`
+      description += comment ? `: ${comment}` : '.'
```
Comment (riccardobl) [inline pages/api/lnurlp/[username]/pay.js:40]: we don't need to handle invoice privacy here, it is done by the paid action

wallets/wrap.js
```diff
@@ -112,11 +129,6 @@ export default async function wrapInvoice (bolt11, { msats, description, descrip
       wrapped.description = inv.description
     }
 
-    if (me?.hideInvoiceDesc) {
-      wrapped.description = undefined
-      wrapped.description_hash = undefined
-    }
-
```
riccardobl, line 119: this is handled upstream by the create*Invoice code in wallets/server

worker/autowithdraw.js
```diff
@@ -42,7 +42,15 @@ export async function autoWithdraw ({ data: { id }, models, lnd }) {
...
   if (pendingOrFailed.exists) return
 
-  const { invoice, wallet } = await createInvoice(id, { msats, description: 'SN: autowithdrawal', expiry: 360 }, { models })
+  const { invoice, wallet } = await createInvoice(id,
+    {
+      msats,
+      description: 'SN: autowithdrawal',
+      expiry: 360,
+      fallbackToSN: false, // if the user doesn't have an attached wallet, there is nothing to withdraw to
+      direct: true // no need to wrap this one
```
riccardobl, line 51: this is to disable wrapping, that would otherwise happen automatically because the destination is an user wallet

wallets/server.js
```diff
@@ -11,120 +11,309 @@ import * as blink from 'wallets/blink/server'
...
+
+  // we are creating an invoice paid to an user
+  const receiverUser = await models.user.findUnique({ where: { id: receiverUserId } })
+  if (!receiverUser) throw new Error('user not found')
+
+  // respect receiver's privacy settings
+  if (receiverUser.hideInvoiceDesc) {
+    description = undefined
+    descriptionHash = undefined
+  }
```
riccardobl, line 107: receiver privacy

components/use-paid-mutation.js
```diff
@@ -23,33 +25,86 @@ export function usePaidMutation (mutation,
...
+          console.log('paid with wallet', wallet.def.name)
+          return { invoice, response }
+        } catch (err) {
+          walletErrors.push(err)
+          console.log('could not pay with wallet', wallet.def.name, 'will try another one ...')
+        }
+      }
+
+      try {
+        await invoiceHelper.cancel(invoice)
```
riccardobl, line 62: What if somehow the wallet failed but is still  trying to pay? (maybe an improper timeout?). Safer to wait for the invoice to be cancelled, so we don't double invoice.

components/use-paid-mutation.js
```diff
@@ -23,33 +25,86 @@ export function usePaidMutation (mutation,
...
+        console.log('could not pay with any wallet, and the invoice is not retriable, will fallback to another method')
+        break
+      }
+    }
+
+    // we try one last time using fee credits
     try {
-      return await waitForWalletPayment(invoice, waitFor)
+      console.log('could not pay with any wallet, will try with fee credits...')
+      const retry = await retryPaidAction({ variables: { invoiceId: parseInt(invoice.id), forceFeeCredits: true } })
```
riccardobl, line 83: we use fee credits only if the p2p payment is not possible

components/use-paid-mutation.js
```diff
@@ -23,39 +25,113 @@ export function usePaidMutation (mutation,
...
+    let response = originalResponse
+
+    if (!me) {
+      if (!invoice) {
+        setInnerResult(r => addPayError(new Error('You must be logged in'), r))
+        throw new Error('You must be logged in')
+      }
+      await waitForQrPayment(invoice, null, { persistOnNavigate, waitFor })
+      return { invoice, response }
+    }
```
riccardobl, line 62: anons gets the qr code right away

api/paidAction/index.js
```diff
@@ -95,271 +128,285 @@ export default async function performPaidAction (actionType, args, context) {
...
-    throw new Error(`retryPaidAction - action does not support optimism ${actionType}`)
+  let retriable = false
+  if (forceRetry) {
+    retriable = true
+    context.retriedWithFeeCredits = false
+    context.retriedManually = false
+  } else {
+    if (failedInvoice.retriable) retriable = true
+    context.retriedWithFeeCredits = !failedInvoice.retriableWithFeeCredits
+    context.retriedManually = !failedInvoice.retriableManually
```
riccardobl, line 282: should this be a single state with an enum? Maybe not, the 3 flags are not necessarily dependent to each other and we don't need to follow a specific flow.

api/paidAction/index.js
```diff
@@ -95,271 +128,285 @@ export default async function performPaidAction (actionType, args, context) {
...
-  if (!me) {
-    throw new Error(`retryPaidAction - must be logged in ${actionType}`)
+  if (!failedInvoice) {
+    throw new Error(`retryPaidAction - missing invoice ${actionType}`)
   }
 
-  if (!action.supportsOptimism) {
-    throw new Error(`retryPaidAction - action does not support optimism ${actionType}`)
+  let retriable = false
+  if (forceRetry) {
```
riccardobl, line 275: the code limits retries so that after a while we know that everything we tried failed, and we consider the action dead (both client and serverside).
However only the user knows if maybe it is a temporary issue with its node or something like that, so there is this `forceRetry` flag that the retry button can use in the ui to restart the cycle over.

api/paidAction/index.js
```diff
@@ -95,271 +128,285 @@ export default async function performPaidAction (actionType, args, context) {
...
 
-  const { msatsRequested, actionId } = failedInvoice
-  context.cost = BigInt(msatsRequested)
+  const { msatsRequested, actionId, actionArgs } = failedInvoice
+  context.cost = msatsRequested
   context.actionId = actionId
-  const invoiceArgs = await createSNInvoice(actionType, args, context)
+  context.retryForInvoice = failedInvoice
+  context.forceFeeCredits = forceFeeCredits
+  context.actionAttempt = failedInvoice.actionAttempt + 1 // next attempt
```
riccardobl, line 304: this is used to cycle through the wallets

api/resolvers/notifications.js
```diff
@@ -234,10 +234,14 @@ export default {
...
             AND "confirmedAt" IS NOT NULL
-            AND "isHeld" IS NULL
             AND "actionState" = 'PAID'
             AND "actionType" = 'TRANSFER'
             AND created_at < $2
+            AND NOT EXISTS (
+              SELECT 1
+              FROM "InvoiceForward"
+              WHERE "InvoiceForward"."invoiceId" = "Invoice".id
+            )
```
riccardobl, line 244: if there is an InvoiceForward, it must be a p2p transfer, this is to avoid showing the notification twice (for deposit + p2p forward)

api/paidAction/boost.js
```diff
@@ -34,12 +34,12 @@ export async function perform ({ invoiceId, sats, id: itemId, ...args }, { me, c
...
 
 export async function retry ({ invoiceId, newInvoiceId }, { tx, cost }) {
-  await tx.itemAct.updateMany({ where: { invoiceId }, data: { invoiceId: newInvoiceId, invoiceActionState: 'PENDING' } })
   const [{ id, path }] = await tx.$queryRaw`
     SELECT "Item".id, ltree2text(path) as path
     FROM "Item"
     JOIN "ItemAct" ON "Item".id = "ItemAct"."itemId"
-    WHERE "ItemAct"."invoiceId" = ${newInvoiceId}::INTEGER`
+    WHERE "ItemAct"."invoiceId" = ${invoiceId}::INTEGER`
+  await tx.itemAct.updateMany({ where: { invoiceId }, data: { invoiceId: newInvoiceId, invoiceActionState: 'PENDING' } })
```
riccardobl, line 42: This is just getting the stuff we need using the old `invoiceId` before `tx.itemAct.updateMany()`, since if the action is being converted from invoiced (eg. p2p) to invoiceless (eg. fee credit) the code will unbind the `itemAct` from any invoice and we won't have anything to fetch the data.

api/paidAction/index.js
```diff
@@ -30,75 +34,166 @@ export const paidActions = {
...
-      throw new Error(`Invalid action type ${actionType}`)
+      throw new Error(`ynvalid action type ${actionType}`)
     }
 
-    if (!me && !paidAction.anonable) {
-      throw new Error('You must be logged in to perform this action')
-    }
+    context.me = context.me ? await context.models.user.findUnique({ where: { id: context.me.id } }) : undefined
+    context.actionAttempt = context.actionAttempt ?? 0
+    context.forceInternal = context.forceInternal ?? false
```
riccardobl, line 77: this is currently used by things like auto territory renewal to force the code to use only internal payments (fee credits and reward sats)

api/paidAction/index.js
```diff
@@ -30,75 +34,166 @@ export const paidActions = {
...
+    context.prioritizeInternal = context.prioritizeInternal ?? false
```
riccardobl, line 78: Contrary to forceInternal, this will just prioritize reward sats and fee credits. 
This is used by the client side code after every sending wallet failed to pay every receiving wallet, so it will try to do an internal transaction, if this fails too they payment will fallback to the other methods in order, this should give another invoice that is then presented to the user with a qr code.

api/paidAction/index.js
```diff
@@ -30,75 +34,166 @@ export const paidActions = {
...
+
+    context.cost = await paidAction.getCost(args, context) ?? 0n
+    if (context.cost < 0n) throw new Error('cost cannot be negative')
 
-    context.me = me ? await models.user.findUnique({ where: { id: me.id } }) : undefined
-    context.cost = await paidAction.getCost(args, context)
+    context.sybilFeePercent = paidAction.getSybilFeePercent ? await paidAction.getSybilFeePercent(args, context) : undefined
+    if (context.sybilFeePercent !== undefined && context.sybilFeePercent < 0n) throw new Error('sybil fee percent cannot be negative')
```
riccardobl, line 84: actions can set their sybil fee

api/paidAction/index.js
```diff
@@ -30,75 +34,166 @@ export const paidActions = {
...
 
-    // special case for zero cost actions
-    if (context.cost === 0n) {
+    context.description = context.me?.hideInvoiceDesc ? undefined : await paidAction.describe(args, context)
+    context.descriptionHash = paidAction.describeHash ? await paidAction.describeHash(args, context) : undefined
+    context.supportedPaymentMethods = paidAction.paymentMethods ?? (paidAction.getPaymentMethods ? await paidAction.getPaymentMethods(args, context) : undefined)
```
riccardobl, line 88: Optionally actions can export a getter: `getPaymentMethods` that is used if `paymentMethod` field is not exported.  Maybe we can change payment priorities for some actions depending on user preference?

api/paidAction/index.js
```diff
@@ -30,75 +34,166 @@ export const paidActions = {
...
 
-    context.me = me ? await models.user.findUnique({ where: { id: me.id } }) : undefined
-    context.cost = await paidAction.getCost(args, context)
+    context.sybilFeePercent = paidAction.getSybilFeePercent ? await paidAction.getSybilFeePercent(args, context) : undefined
+    if (context.sybilFeePercent !== undefined && context.sybilFeePercent < 0n) throw new Error('sybil fee percent cannot be negative')
 
-    // special case for zero cost actions
-    if (context.cost === 0n) {
+    context.description = context.me?.hideInvoiceDesc ? undefined : await paidAction.describe(args, context)
+    context.descriptionHash = paidAction.describeHash ? await paidAction.describeHash(args, context) : undefined
```
riccardobl, line 87: actions can return a description hash

api/paidAction/index.js
```diff
@@ -107,283 +202,285 @@ export default async function performPaidAction (actionType, args, context) {
...
+      args
+    }, subContext)
     return {
-      invoice,
-      result: await action.perform?.({ invoiceId: invoice.id, ...args }, context),
-      paymentMethod: 'OPTIMISTIC'
+      invoice: invoiceEntry,
+      result: await performAction(invoiceEntry, paidAction, args, subContext),
+      paymentMethod: 'OPTIMISTIC',
+      retriableWithWallet: invoiceData.retriable
```
riccardobl, line 276: when retriableWithWallet is false, it means we have tried all the receiving wallets, this doesn't enforce anything, it is just a flag that lets the clientside code know it should probably use another methods (eg. prioritizeInternal or qr code). The user can still retry if he wishes so, using the retry button.

api/paidAction/index.js
```diff
@@ -107,283 +202,285 @@ export default async function performPaidAction (actionType, args, context) {
...
-  context.me = await models.user.findUnique({ where: { id: me.id } })
-
-  const { msatsRequested, actionId } = failedInvoice
-  context.cost = BigInt(msatsRequested)
-  context.actionId = actionId
-  const invoiceArgs = await createSNInvoice(actionType, args, context)
+export async function retryPaidAction ({ invoiceId, forceInternal, resetAttempts, prioritizeInternal }, context) {
+  const { models, lnd, boss, me } = context
 
   return await models.$transaction(async tx => {
```
riccardobl, line 307: I am not sure if running everything inside the transaction is overkill, but i suppose we don't want a race condition where the invoice is cancelled while the action is still being retried.

api/paidAction/index.js
```diff
@@ -107,283 +202,285 @@ export default async function performPaidAction (actionType, args, context) {
...
-    })
-
-    // create a new invoice
-    const invoice = await createDbInvoice(actionType, args, context, invoiceArgs)
-
-    return {
-      result: await action.retry({ invoiceId: failedInvoice.id, newInvoiceId: invoice.id }, context),
-      invoice,
-      paymentMethod: 'OPTIMISTIC'
+    const subContext = { ...context, tx }
```
riccardobl, line 308: prevent unexpected context leaking

api/paidAction/index.js
```diff
@@ -107,283 +202,285 @@ export default async function performPaidAction (actionType, args, context) {
...
+      throw new Error(`actions is not in a retriable state: ${failedInvoice.actionState}`)
     }
-  })
 
-  if (pendingInvoices >= MAX_PENDING_PAID_ACTIONS_PER_USER) {
-    throw new Error('You have too many pending paid actions, cancel some or wait for them to expire')
-  }
-}
+    // cancel the previous invoice
+    await finalizeHodlInvoice({ data: { hash: failedInvoice.hash }, lnd, models, boss })
```
riccardobl, line 323: Cancel the previous invoice
I am not sure if running everything in a transaction is overkill, but i suppose we don't want a race condition where the invoice being cancelled while the action is still retrying

api/paidAction/index.js
```diff
@@ -107,283 +202,286 @@ export default async function performPaidAction (actionType, args, context) {
...
+      throw new Error(`missing invoice ${actionType}`)
+    }
 
-  await assertBelowMaxPendingInvoices(context)
+    const { msatsRequested, actionId, actionArgs } = failedInvoice
+    subContext.cost = msatsRequested
+    subContext.actionId = actionId
+    subContext.retryForInvoice = failedInvoice
+    subContext.forceInternal = forceInternal
+    subContext.actionAttempt = failedInvoice.actionAttempt + 1 // next attempt
```
riccardobl, line 340: We track attempts like this to cycle through the receiving wallets.
Maybe a receiving wallet works fine for us, but the sender cannot pay the invoice (bad channels or whatever), so we just use the next receiving wallet when the sender retries the payment. This is where actionAttempt is used, to progress to the next wallet.

api/paidAction/index.js
```diff
@@ -107,283 +202,286 @@ export default async function performPaidAction (actionType, args, context) {
...
+    subContext.prioritizeInternal = prioritizeInternal
 
-  const description = await paidActions[actionType].describe(args, context)
-  const { invoice: bolt11, wallet } = await createUserInvoice(userId, {
-    // this is the amount the stacker will receive, the other 3/10ths is the sybil fee
-    msats: cost * BigInt(7) / BigInt(10),
-    description,
-    expiry: INVOICE_EXPIRE_SECS
-  }, { models })
+    if (resetAttempts) {
```
riccardobl, line 343: this is passes when the user clicks the `retry` button: we reset the cycle and we start over

api/resolvers/notifications.js
```diff
@@ -346,6 +363,11 @@ export default {
           "Invoice"."actionType" = 'POLL_VOTE' OR
           "Invoice"."actionType" = 'BOOST'
         )
+        AND EXISTS (
+          SELECT 1
+          FROM "ItemAct"
+          WHERE "ItemAct"."invoiceId" = "Invoice".id
+        )
```
riccardobl, line 370: when we have a failed invoiced action that it is then retried successfully with a fee credit or reward sats payment, we have this situation where there is a failed invoice that used to be bound to this action but it is not anymore. 
This check is to exclude it from notifications. Maybe there are other ways, but i am unsure.

api/resolvers/paidAction.js
```diff
@@ -46,28 +48,11 @@ export default {
...
-
-      if (invoice.actionState !== 'FAILED') {
-        if (invoice.actionState === 'PAID') {
-          throw new Error('Invoice is already paid')
-        }
-        throw new Error(`Invoice is not in failed state: ${invoice.actionState}`)
-      }
-
-      const result = await retryPaidAction(invoice.actionType, { invoice }, { models, me, lnd })
-
```
riccardobl, line 67: i moved all this code inside paidAction/index.js retryPaidAction

components/use-paid-mutation.js
```diff
@@ -23,39 +25,127 @@ export function usePaidMutation (mutation,
...
+    let invoiceUsed = false
+
+    // ensures every invoice is used only once
+    const refreshInvoice = async () => {
+      if (invoiceUsed) {
+        const retry = await retryPaidAction({ variables: { invoiceId: parseInt(invoice.id) } })
+        response = retry.data?.retryPaidAction
+        invoice = response?.invoice
+        invoiceUsed = true
+      } else invoiceUsed = true
```
riccardobl, line 62: ensure we are getting a fresh invoice everytime it is presented to a wallet or to the user.

components/use-paid-mutation.js
```diff
@@ -75,62 +165,52 @@ export function usePaidMutation (mutation,
...
+        setInnerResult({ data, ...rest })
+      }).catch(e => {
+        console.error('usePaidMutation: failed to pay invoice', e)
+        // onPayError is called after the invoice fails to pay
+        // useful for updating invoiceActionState to FAILED
+        onPayError?.(e, client.cache, { data })
+        setInnerResult(r => addPayError(e, r))
+      })
+
+      if (wait) await p
```
riccardobl, line 205: i've de-duplicated the code, but the logic should be the same

fragments/notifications.js
```diff
@@ -186,8 +186,16 @@ export const NOTIFICATIONS = gql`
...
+            satsFeePaid           
+            invoiceForward {
+              invoice {
+                id
+                nostr
+                comment
+                lud18Data
+              }
+            }
+          }
```
riccardobl, line 198: need this for the p2p lnurlp comments

api/paidAction/downZap.js
```diff
@@ -35,12 +35,12 @@ export async function perform ({ invoiceId, sats, id: itemId }, { me, cost, tx }
...
 
 export async function retry ({ invoiceId, newInvoiceId }, { tx, cost }) {
-  await tx.itemAct.updateMany({ where: { invoiceId }, data: { invoiceId: newInvoiceId, invoiceActionState: 'PENDING' } })
   const [{ id, path }] = await tx.$queryRaw`
     SELECT "Item".id, ltree2text(path) as path
     FROM "Item"
     JOIN "ItemAct" ON "Item".id = "ItemAct"."itemId"
-    WHERE "ItemAct"."invoiceId" = ${newInvoiceId}::INTEGER`
+    WHERE "ItemAct"."invoiceId" = ${invoiceId}::INTEGER`
+  await tx.itemAct.updateMany({ where: { invoiceId }, data: { invoiceId: newInvoiceId, invoiceActionState: 'PENDING' } })
```
riccardobl, line 43: same as https://github.com/stackernews/stacker.news/pull/1523/files#r1835496400

api/paidAction/index.js
```diff
@@ -33,72 +36,162 @@ export const paidActions = {
...
-        } else if (paymentMethod === PAID_ACTION_PAYMENT_METHODS.OPTIMISTIC) {
-          return await performOptimisticAction(actionType, args, context)
         }
+      } else if (paymentMethod === PAID_ACTION_PAYMENT_METHODS.P2P) {
+        try {
+          const optimistic =
+            paymentMethod === PAID_ACTION_PAYMENT_METHODS.OPTIMISTIC ||
+            // if action supports optimism and its a logged p2p action, we can try to be p2p and optimistic
+            (me && paymentMethod === PAID_ACTION_PAYMENT_METHODS.P2P &&
+              supportedPaymentMethods.includes(PAID_ACTION_PAYMENT_METHODS.OPTIMISTIC))
```
riccardobl, line 158: This is to ensure p2p actions stay pessimistic unless the actions supports PAID_ACTION_PAYMENT_METHODS.OPTIMISTIC.

api/paidAction/index.js
```diff
@@ -107,283 +200,286 @@ export default async function performPaidAction (actionType, args, context) {
...
-
-  return await models.$transaction(async tx => {
-    context.tx = tx
-
-    const invoice = await createDbInvoice(actionType, args, context, invoiceArgs)
-
+async function performOptimisticAction (invoiceData, actionType, paidAction, args, context) {
+  const { models, actionId, me } = context
+  const performInvoicedAction = async tx => {
+    const subContext = { ...context, tx, optimistic: true }
```
riccardobl, line 262: prevent accidental context leaking

api/paidAction/index.js
```diff
@@ -107,283 +200,286 @@ export default async function performPaidAction (actionType, args, context) {
...
-  if (!action.retry) {
-    throw new Error(`retryPaidAction - action does not support retrying ${actionType}`)
-  }
-
-  if (!failedInvoice) {
-    throw new Error(`retryPaidAction - missing invoice ${actionType}`)
+async function performPessimisticAction (invoiceData, actionType, paidAction, args, context) {
+  const { models, actionId, me } = context
+  const performInvoicedAction = async tx => {
+    const subContext = { ...context, tx, optimistic: false }
```
riccardobl, line 284: prevent accidental context leaking

api/paidAction/index.js
```diff
@@ -107,283 +200,286 @@ export default async function performPaidAction (actionType, args, context) {
...
-        },
-        data: paymentMethod === 'REWARD_SATS'
-          ? { msats: { decrement: cost } }
-          : { mcredits: { decrement: cost } }
-      })
-    }
+async function performInternalAction (cost, actionType, paidAction, paymentMethod, args, context) {
+  const { me, models } = context
+  const run = async tx => {
+    const subContext = { ...context, tx }
```
riccardobl, line 215: prevent accidental context leaking

api/paidAction/index.js
```diff
@@ -107,283 +200,286 @@ export default async function performPaidAction (actionType, args, context) {
...
-  const action = paidActions[actionType]
-
-  const result = await models.$transaction(async tx => {
-    context.tx = tx
+async function performAction (invoiceEntry, paidAction, args, context) {
+  const { retryForInvoice } = context
+  if (retryForInvoice && paidAction.retry) {
+    return await paidAction.retry({ invoiceId: retryForInvoice.id, newInvoiceId: invoiceEntry?.id }, context)
+  } else {
+    return await paidAction.perform?.({ invoiceId: invoiceEntry?.id, ...args }, context)
```
riccardobl, line 208: if the action can be retried -> we retry
otherwise we just rerun it

api/paidAction/itemCreate.js
```diff
@@ -149,13 +149,15 @@ export async function perform (args, context) {
 }
 
 export async function retry ({ invoiceId, newInvoiceId }, { tx }) {
+  const res = (await tx.$queryRaw`
+    SELECT *, ltree2text(path) AS path, created_at AS "createdAt", updated_at AS "updatedAt"
+    FROM "Item" WHERE "invoiceId" = ${invoiceId}::INTEGER`
+  )[0]
+  res.invoiceId = newInvoiceId
```
riccardobl, line 156: same as https://github.com/stackernews/stacker.news/pull/1523/files#r1835496400

api/paidAction/pollVote.js
```diff
@@ -42,11 +42,10 @@ export async function perform ({ invoiceId, id }, { me, cost, tx }) {
 }
 
 export async function retry ({ invoiceId, newInvoiceId }, { tx }) {
+  const { pollOptionId } = await tx.pollVote.findFirst({ where: { invoiceId } })
```
riccardobl, line 45: same as https://github.com/stackernews/stacker.news/pull/1523/files#r1835496400

api/paidAction/zap.js
```diff
@@ -61,12 +61,12 @@ export async function perform ({ invoiceId, sats, id: itemId, ...args }, { me, c
...
 
 export async function retry ({ invoiceId, newInvoiceId }, { tx, cost }) {
-  await tx.itemAct.updateMany({ where: { invoiceId }, data: { invoiceId: newInvoiceId, invoiceActionState: 'PENDING' } })
   const [{ id, path }] = await tx.$queryRaw`
     SELECT "Item".id, ltree2text(path) as path
     FROM "Item"
     JOIN "ItemAct" ON "Item".id = "ItemAct"."itemId"
-    WHERE "ItemAct"."invoiceId" = ${newInvoiceId}::INTEGER`
+    WHERE "ItemAct"."invoiceId" = ${invoiceId}::INTEGER`
+  await tx.itemAct.updateMany({ where: { invoiceId }, data: { invoiceId: newInvoiceId, invoiceActionState: 'PENDING' } })
```
riccardobl, line 69: same as https://github.com/stackernews/stacker.news/pull/1523/files#r1835496400

components/qr.js
```diff
@@ -16,20 +17,25 @@ export const qrImageSettings = {
...
+      const usableWallets = wallets.filter(w => !w.def.isAvailable || w.def.isAvailable())
+        .filter(w => w.config?.enabled && canSend(w))[0]
+      if (automated && usableWallets.length > 0) {
+        for (const wallet of usableWallets) {
+          try {
+            await wallet.sendPayment(value)
+            break
+          } catch (e) {
+            console.log(e?.message)
+          }
```
riccardobl, line 33: Try every wallet until one successfully pays

components/use-item-submit.js
```diff
@@ -101,6 +101,7 @@ export function useRetryCreateItem ({ id }) {
   const [retryPaidAction] = usePaidMutation(
     RETRY_PAID_ACTION,
     {
+      alwaysShowQROnFailure: true,
```
riccardobl, line 104: when the retry button is pressed, we always fallback to a qr code as last resort

components/use-paid-mutation.js
```diff
@@ -23,39 +25,126 @@ export function usePaidMutation (mutation,
...
+      return { invoice, response }
+    }
+
+    // we try every sender<->receiver combination
+    while (true) {
+      await refreshInvoice()
+      if (!invoice) return { invoice, response }
+
+      for (const wallet of usableWallets) {
+        try {
```
riccardobl, line 92: we try to pay with every sending wallet

components/use-paid-mutation.js
```diff
@@ -23,39 +25,126 @@ export function usePaidMutation (mutation,
...
-        (!alwaysShowQROnFailure && Date.now() - start > 1000) ||
-        err instanceof InvoiceCanceledError ||
-        err instanceof InvoiceExpiredError) {
-        // bail since qr code payment will also fail
-        // also bail if the payment took more than 1 second
-        // and cancel the invoice if it's not already canceled so it can be retried
-        invoiceHelper.cancel(invoice).catch(console.error)
-        throw err
+      console.log('could not pay with any wallet, will try with an internal payment...')
+      const retry = await retryPaidAction({ variables: { invoiceId: parseInt(invoice.id), prioritizeInternal: true } })
```
riccardobl, line 114: In the previous step we failed to pay the action with whatever preferred method it had, so now we try to reprioritize FEE_CREDITS and REWARD_SATS so that they are tried first (if the action supports them).
If this fails to, we will get a new invoice out of this call, this invoice can be used in the qr code

wallets/server.js
```diff
@@ -11,120 +11,231 @@ import * as blink from 'wallets/blink/server'
...
-
-      const invoice = await withTimeout(
-        createInvoice({
+      await checkWallet(models, walletFull, walletField, walletType)
+
+      console.log('use wallet', wallet.type)
+      if (hodl || wrapped) {
+        // we always wrap p2p invoices unless it is a direct non-hodl invoice
+        const wrappedMsats = msats
+        const innerMsats = BigInt(msats) * (100n - BigInt(sybilFeePercent ?? ZAP_SYBIL_FEE_PERCENT)) / 100n
```
riccardobl, line 120: this is to avoid bigint<->number conversions

wallets/server.js
```diff
@@ -11,120 +11,231 @@ import * as blink from 'wallets/blink/server'
...
+        const invoice = await withTimeout(
+          createInvoice({
+            msats: toPositiveNumber(innerMsats),
+            description,
+            descriptionHash,
+            expiry
+          }, walletFull[walletField]), 10_000)
+        await checkInvoice(models, wallet, invoice, innerMsats)
+        if (!lnd) throw new Error('no lnd provided for wrapping')
+        const { invoice: wrappedInvoice, maxFee } = await wrapInvoice(invoice,
```
riccardobl, line 132: we do wrapping here, so if for whatever reason this wallet has some incompatibility with our wrapping code, we fallback to the next one

wallets/wrap.js
```diff
@@ -53,9 +67,12 @@ export default async function wrapInvoice (bolt11, { msats, description, descrip
 
     // validate incoming amount
     if (msats) {
-      msats = toPositiveNumber(msats)
-      if (outgoingMsat * ZAP_SYBIL_FEE_MULT > msats) {
+      msats = toPositiveBigInt(msats)
+      const msatsCheck = outgoingMsat * 100n / (100n - zapSybilFeePercent)
```
riccardobl, line 71: avoid biging<->number conversions

wallets/wrap.js
```diff
@@ -161,8 +173,8 @@ export default async function wrapInvoice (bolt11, { msats, description, descrip
     }
 
     // validate the fee budget
-    const minEstFees = toPositiveNumber(routingFeeMsat)
-    const outgoingMaxFeeMsat = Math.ceil(msats * MAX_FEE_ESTIMATE_PERCENT)
+    const minEstFees = toPositiveBigInt(routingFeeMsat)
+    const outgoingMaxFeeMsat = msats * MAX_FEE_ESTIMATE_PERMILE / 1000n
```
riccardobl, line 177: avoid bigint<->number conversions

wallets/server.js
```diff
@@ -11,120 +11,231 @@ import * as blink from 'wallets/blink/server'
...
         level: 'ERROR',
         message: `creating invoice for ${description ?? ''} failed: ` + error
       }, { models })
     }
   }
-
   throw new Error('no wallet available')
 }
+
+export const addWalletLog = async ({ wallet, level, message }, { models }) => {
```
riccardobl, line 185: this was moved here to avoid cyclic dependencies with this file and the resolver

api/paidAction/index.js
huumn: what do you mean by "leaking"? memory or something else?

huumn:
Just a heads up: I probably won't have a first pass review of fallbacks until end of tomorrow. 

Questions in the meantime:
1. how are you keeping track of the wallet "pairs" that you mention?
2. are we actually trying every combination of wallet pairs?
    - afaict we'd want to 
        - iterate through sending wallets until we find one that can send SN money
        - iterate through receiving wallets until we find one that SN can send money to
    - is that what this does?
3. how are you dealing with the fee credit retry issue you brought up in chat? You mentioned creating a fake invoice in chat and that my alternative wouldn't work. Did you continue with the fake invoice approach?

api/paidAction/index.js
riccardobl: It is not a memory leak. It is a problem with the current code too, the way this context object is passed around in paidAction and then used to store internal values can lead to unexpected inner context leaking. 
eg. let's say you pass the context from a resolver to two paid actions

```js
something: async (parent, args, context) => {
    performPaidAction('ACTION1', {}, context)
    performPaidAction('ACTION2', {}, context)
}

```


action2 will get a context that has a bunch of random field from the action1 execution.

This never happens in our code, it is just to illustrate the issue, but in specific in our codebase, it could happen that one payment method fails in the `for (const paymentMethod of supportedPaymentMethods)` loop and when the next one is tried, it gets a dirty context with the left overs from the previous attempt. In our specific case it could end up having a closed or closing transaction in the tx field .
Making a shallow copy of the context like this, fixes this issue while keeping the current way of passing the context, and prevents future hard to debug issues related to context leaking.

riccardobl:
> are we actually trying every combination of wallet pairs?
> * afaict we'd want to
>   
>   * iterate through sending wallets until we find one that can send SN money
>   * iterate through receiving wallets until we find one that SN can send money to
> * ie sending and receiving fallbacks should operate independently, rather than as pairs - is that what this does?

The code calls the paidAction and receives an invoice, then tries to pay it with every sending wallet, then it calls retryPaidAction , gets a new invoice, tries to pay it and so on. So yes, it iterates through all the possible combinations, like this:

```
recv1 - send1
recv1 - send2
recv1 - sendN
recv2 - send1
recv2 - send2
recv2 - sendN
recvN - send1
recvN - send2
recvN - sendN
...
```

> * how are you keeping track of the wallet "pairs" that you mention?

it is tracked by increasing an attempt counter, this walks through the receiving wallets array:

https://github.com/stackernews/stacker.news/blob/b14e43aede16c231b4d9e92950041cfa960fe94c/wallets/server.js#L92-L95

every time the code calls retryPaidAction, the attempt counter is increased by 1, unless the user clicks manually on a retry button, in that case it resets to 0 (resetAttempts=true)


> how are you dealing with the fee credit retry issue you brought up in chat? You mentioned creating a fake invoice in chat and that my alternative wouldn't work. Did you continue with the fake invoice approach?

i've managed to find some solutions that didn't require this approach, by making the paidActions' retry() function retrieve the database entries associated to the failedInvoice before reassociating them to the new invoiceId, that in case of fee credits is null like here: https://github.com/stackernews/stacker.news/pull/1523/files#r1835496400

and then i had to patch the notification code here:  https://github.com/stackernews/stacker.news/pull/1523/files#r1835509613


> can you elaborate more on what you meant in chat?

since i am new to the code base, i am not 100% sure that the changes  mentioned above are enough to account for all the side effects of deassociating the entries from a paid action from their failed invoice. 
Up until now the code used to create either `actions with an invoice` or `actions without an invoice`, but never an `action that used to have an invoice but now it doesn't anymore`, so i am not sure if this is introducing side effects i am not aware of

huumn:
> So yes, it iterates through all the possible combinations, like this:

Because the sending and receiving routes will succeed and fail independently of each other, I don't think we need to walk through all combinations. e.g. if send1 succeeds at push money to SN, then we only need to retry recv1 through recvN. If recv1 through recvN are all failing, send2 won't succeed, nor send3.

Does that make sense? Does the code take this into account?

> it is tracked by increasing an attempt counter, that is used to walk through the receiving wallets array on every retry

Using a counter to walk through combos seems kind of fragile, but it's not obvious how it could be improved until I've looked at the code more.

> since i am new to the code base, i am not 100% sure that the changes mentioned above are enough to account for all the side effects of dissociating the entries created by the paid actions from their failed invoice. Up until now the code used to create either actions with an invoice or actions without an invoice, but never an action that used to have an invoice but now it doesn't anymore, so i am not sure if this is introducing side effects i am not aware of.

Oh i see! That should be straightforward for me to verify. When I finish up my review on lnurlp stuff, i'll take a look.

riccardobl:
> > So yes, it iterates through all the possible combinations, like this:
> 
> Because the sending and receiving routes will succeed and fail independently of each other, I don't think we need to walk through all combinations. e.g. if send1 succeeds at push money to SN, then we only need to retry recv1 through recvN. If recv1 through recvN are all failing, send2 won't succeed to get money to recv1 through recvN, nor will send3.
> 
> Does that make sense? Does the code take this into account?

you have a point, the code doesn't take that into account currently

wallets/server.js
ekzyis: this function was removed in #994

riccardobl:
Rebased to master

>Because the sending and receiving routes will succeed and fail independently of each other, I don't think we need to walk through all combinations. e.g. if send1 succeeds at push money to SN, then we only need to retry recv1 through recvN. If recv1 through recvN are all failing, send2 won't succeed to get money to recv1 through recvN, nor will send3.

> Does that make sense? Does the code take this into account?

now this is taken into account. 

The paid action mutation will comunicate to the client the invoiceForward's withdrawl status and a retriable boolean flag.

when the mutation returns a withdrawl status that is set but not CONFIRMED
 - if retriable is true: the client will call retryPaidMutation passing an increasing 'attempt' counter -> this will walk to the next receiver in the priority sorted list
 - if retriable is false: it means there are no more receivers, the client will fallback to a fee credit payment 

if the mutation doesn't have a withdrawl status
- it means the sender failed: the client will mark the sender as failed and try the next one without increasing the attempt counter
- when every sender has failed, the client will not retry but instead fall back to a fee credit payment

if the fee credit fails: the client will fallback to the qr code using the last `attempt` counter value.

Note: this is not explicit in the code, but it is an effect of the logic: if we previously tried all the receivers and the forward failed for all of them, we will have an attempt counter high enough to exceed the number of receiving wallets, this will skip the p2p method, as it should, since there is no way to pay, even by qr code.

api/paidAction/index.js
```diff
@@ -41,243 +41,246 @@ export default async function performPaidAction (actionType, args, incomingConte
...
-      // additionalpayment methods that logged in users can use
-      if (me) {
-        if (paymentMethod === PAID_ACTION_PAYMENT_METHODS.FEE_CREDIT) {
-          try {
+    for (const paymentMethod of supportedPaymentMethods) {
+      try {
+        console.log(`trying payment method ${paymentMethod}`)
+        switch (paymentMethod) {
+          case PAID_ACTION_PAYMENT_METHODS.FEE_CREDIT: {
+            if (!me || (me.msats ?? 0n) < cost) break // if anon or low balance skip
```
huumn, line 103: The original code that looks for `violates check` was there for a reason. The way you've rewritten this suppresses errors that aren't related to payments.

The switch statement is nice, but be careful causing regressions by changing code that you don't understand - and more importantly, code that isn't directly related to your changes. 

Stuff like this is what causes me to do full rewrites.

components/use-item-submit.js
huumn: The reason this isn't the default is that payments don't fail immediately. What's the behavior here when the payment fails 2 minutes later?

Does this prevent a QR from popping up on another, unrelated page 2 minutes later?

riccardobl: I thought, that since this is happening after pressing manually on the retry button, the user was going to wait for it. But you are right it is bad ux unless there is a modal or something that tells the user to wait. 
I've restored the previous behavior, better not add more unrelated stuff to this pr

ekzyis (commented):
I didn't run the code yet, but this is my review so far from looking at most of the code

api/paidAction/index.js
```diff
@@ -45,248 +45,266 @@ export default async function performPaidAction (actionType, args, incomingConte
...
+      context.supportedPaymentMethods = forcedPaymentMethods
+    } else if (prioritizeInternal) {
+      // prefer internal payment methods over the others (if they are supported)
+      const priority = {
+        [PAID_ACTION_PAYMENT_METHODS.FEE_CREDIT]: -2
+        // add other internal methods here
+      }
+      context.supportedPaymentMethods = context.supportedPaymentMethods.sort((a, b) => {
+        return priority[a] - priority[b]
+      })
```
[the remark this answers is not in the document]
riccardobl: it seems like they will be implemented as a separated payment method in ex-nov-5 pr, that's why i made the distinction

https://github.com/stackernews/stacker.news/blob/nov-5/api/paidAction/itemUpdate.js#L9-L13

api/paidAction/index.js
```diff
@@ -45,248 +45,266 @@ export default async function performPaidAction (actionType, args, incomingConte
...
-    if (!me && !paidAction.anonable) {
+    // add context properties
+    context.me = context.me ? await models.user.findUnique({ where: { id: context.me.id } }) : undefined
+    context.cost = await paidAction.getCost(args, context)
+    context.sybilFeePercent = await paidAction.getSybilFeePercent?.(args, context)
+    context.attempt = context.attempt ?? 0 // how many times the client thinks it has tried
+    context.forceInternal = context.forceInternal ?? false // use only internal payment methods
+    context.prioritizeInternal = context.prioritizeInternal ?? false // prefer internal payment methods
+    context.description = context.me?.hideInvoiceDesc ? undefined : await paidAction.describe?.(args, context)
+    context.descriptionHash = await paidAction.describeHash?.(args, context)
```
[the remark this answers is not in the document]
riccardobl: right, that's some left over from the lnurlp pr extraction

wallets/server.js
```diff
@@ -1,20 +1,20 @@
 // import server side wallets
-import * as lnd from '@/wallets/lnd/server'
-import * as cln from '@/wallets/cln/server'
-import * as lnAddr from '@/wallets/lightning-address/server'
-import * as lnbits from '@/wallets/lnbits/server'
-import * as nwc from '@/wallets/nwc/server'
-import * as phoenixd from '@/wallets/phoenixd/server'
-import * as blink from '@/wallets/blink/server'
+import * as lnd from '@wallets/lnd/server'
```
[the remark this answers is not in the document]
riccardobl: yeah, i've messed up the merge, should be fine now

api/paidAction/index.js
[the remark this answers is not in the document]
ekzyis: Mhh, I see. @huumn, does it make sense to add `REWARD_SATS` as a payment method? I understood it as an "implementation detail" of the `FEE_CREDIT` payment method where they are converted into CCs just-in-time. In that case, isn't it unnecessary as a separate payment method?

components/poll.js
```diff
@@ -103,7 +103,7 @@ function PollResult ({ v, progress }) {
   )
 }
 
-export function usePollVote ({ query = POLL_VOTE, itemId }) {
+export function usePollVote ({ query = POLL_VOTE, itemId, ...options }) {
```
ekzyis, line 106: Is `...options` needed? I don't see `usePollVote` called with anything else than `query` or `itemId`.

api/paidAction/index.js
```diff
@@ -45,248 +45,265 @@ export default async function performPaidAction (actionType, args, incomingConte
...
-  const { msatsRequested, actionId } = failedInvoice
-  const retryContext = {
-    ...incomingContext,
-    optimistic: true,
-    me: await models.user.findUnique({ where: { id: me.id } }),
-    cost: BigInt(msatsRequested),
-    actionId
+  if (failedInvoice.actionState !== 'FAILED') {
+    // you should cancel the invoice before retrying the action!
+    throw new Error(`actions is not in a retriable state: ${failedInvoice.actionState}`)
```
ekzyis, line 273: Why does the client have to cancel the invoice? Can't we do it here?

Canceling on the client is an additional round-trip and the code might be simpler if retrying implicitly cancels the existing invoice.

What do you think?

ekzyis (commented):
I looked through the sending code in detail now and tested the fallback using LNbits that fails and NWC that succeeds. I noticed that the client cancels the invoice. Is there a specific reason for that?

I still need to test more, especially the receiving part.

wallets/server.js
```diff
@@ -25,27 +25,40 @@ export default [lnd, cln, lnAddr, lnbits, nwc, phoenixd, blink, lnc, webln]
...
     const logger = walletLogger({ wallet, models })
 
     try {
       logger.info(
         `↙ incoming payment: ${formatSats(msatsToSats(msats))}`,
         {
-          amount: formatMsats(msats)
-        })
+          amount: formatMsats(toNumber(msats))
+        }) // TODO add fee info?
```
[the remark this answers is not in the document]
ekzyis: Ah, I see. I don't think we need to specifically point out the sybil fee here.

components/qr.js
```diff
@@ -16,20 +17,25 @@ export const qrImageSettings = {
...
 
   useEffect(() => {
     async function effect () {
-      if (automated && wallet) {
-        try {
-          await wallet.sendPayment(value)
-        } catch (e) {
-          console.log(e?.message)
+      const usableWallets = wallets.filter(w => !w.def.isAvailable || w.def.isAvailable())
+        .filter(w => w.config?.enabled && canSend(w))[0]
```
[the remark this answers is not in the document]
ekzyis: I am sorry, this was wrong, see https://github.com/stackernews/stacker.news/pull/1523#discussion_r1845894549

components/qr.js
```diff
@@ -21,10 +21,8 @@ export default function Qr ({ asIs, value, useWallet: automated, statusVariant,
 
   useEffect(() => {
     async function effect () {
-      const usableWallets = wallets.filter(w => !w.def.isAvailable || w.def.isAvailable())
-        .filter(w => w.config?.enabled && canSend(w))[0]
-      if (automated && usableWallets.length > 0) {
-        for (const wallet of usableWallets) {
+      if (automated && wallets.length > 0) {
+        for (const wallet of wallets) {
```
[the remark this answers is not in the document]
riccardobl: No worries, this part of the code has been rebased so many times that I've lost track of the changes as well

api/paidAction/index.js
riccardobl: Hmm, I did consider implementing it like you've suggested, i've kept it this way because it is what the original code does.

But thinking about this in more depth, i think you are right, and that probably the transaction isolation should be serializable, because there is a race condition here (and in the original code), where two transactions could start with a committed 'failed' invoice state, but then end up with two different results.

What do you think?

ekzyis (changes requested):
Did you test this with multiple zaps in-flight? I can consistently reproduce a transaction timeout if I zap more than three times in quick succession:

https://github.com/user-attachments/assets/1d18b000-01d8-4e2f-8810-9b7316edf395

I am not sure what is causing this, but my first guess would be that there is a race condition that results in a deadlock. 

This shouldn't touch any of the new receiving code since I am zapping someone who has no wallets attached.

Maybe it helps if you focus on getting the sending part right before caring about the receiving part.

Also noticed an issue with QR payments since there, the invoice is not refreshed.

I didn’t look further into the receiving part because I am already lacking confidence in the sending part.

components/use-paid-mutation.js
```diff
@@ -23,39 +25,144 @@ export function usePaidMutation (mutation,
...
+          // SN received the payment but couldn't forward it, we try another receiver with the same sender
+          if (!response.retriable) { // we are out of receivers
+            console.log('the receiver wallet failed to receive the payment, but we exhausted all options')
+            break
+          }
+          console.log('the receiver wallet failed to receive the payment, will try another one')
+          attempt++
+        } else {
+          // SN didn't receive the payment, so the sender must have failed
+          senderWallet.failed = true
```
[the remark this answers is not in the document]
riccardobl: sorry i don't understand this comment, what is the issue here? Javascript always passes objects as reference

components/qr.js
riccardobl: The logic on use-paid-mutation is there because we are cancelling and recreating the invoices.
In this case we don't do anything like that, so i think we won't need the complex retry logic here, a lightning invoice can be paid only once, right?

components/use-paid-mutation.js
```diff
@@ -23,39 +25,144 @@ export function usePaidMutation (mutation,
...
+    let response = originalResponse
+    let invoiceUsed = false
+
+    const cancelInvoice = async () => {
+      try {
+        invoiceUsed = true
+        await invoiceHelper.cancel(invoice)
+        console.log('old invoice canceled')
+      } catch (err) {
+        console.error('could not cancel old invoice', err)
```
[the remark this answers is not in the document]
riccardobl: i am not sure, won't this throw if the invoice was cancelled twice or is expired too?

riccardobl:
> I am not sure what is causing this, but my first guess would be that there is a race condition that results in a deadlock.
> 
> This shouldn't touch any of the new receiving code since I am zapping someone who has no wallets attached.
> 
> Maybe it helps if you focus on getting the sending part right before caring about the receiving part.
> 
> Also noticed an issue with QR payments since there, the invoice is not refreshed.
> 
> I didn’t look further into the receiving part because I am already lacking confidence in the sending part.

I don't think it is a deadlock, but the attachment being slow. 
In your test, the first wallet fails, so you are getting into the `retryPaidAction` logic, that runs everything into a transaction to rollback the changes if the attempt fails, if the attachment is slow this causes the transaction to timeout.
I will think to a fix.

riccardobl:
> This will attempt to pay the same invoice again. So afaict, this will require the same retry logic as in usePaidMutation.
>
> Since the logic in there isn't easily extractable to apply it here (as it is written, it requires to be in the scope of usePaidMutation), I think we need to rewrite the retry logic so we can reuse it here or QR codes can't use sender fallbacks.

The logic on use-paid-mutation is there because we are cancelling and recreating the invoices.
In this case we don't do anything like that, so i think we won't need the complex retry logic here, a lightning invoice can be paid only once, right?

components/use-paid-mutation.js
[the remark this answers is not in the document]
ekzyis: Canceling the previous invoice is pretty important since we don't want them to accidentally pay us twice. This code doesn't run exclusively for p2p zaps where we wouldn't be able to settle incoming invoices ourselves without the receiver settling first; **this runs for every paid action**. Our node would absolutely settle an old invoice even though they already paid for the action.

ekzyis:
> We don't know where this invoice is coming from in the qr logic, so how do we ask for another one?

Why do we not know where the invoice is coming from? Yes, the client can't tell but the backend can, that's why you need the same refresh logic.

riccardobl:
> Why do we not know where the invoice is coming from? Yes, the client can't tell but the backend can, that's why you need the same refresh logic.
> 
> >

I am very confused. If this QR code is used in use-paid-mutation, what is the point of trying to paying with attached wallets if this is shown only after all the attached wallet fails?
If it is used elsewhere, we can't use the retry logic from the paidActions. Am i missing something?

components/use-paid-mutation.js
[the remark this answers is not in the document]
riccardobl: So, this is an issue for optimistic actions, that can fail after the invoice is settled, but not for pessimistic actions that use hodl invoice and set the FAILED state on settling, is this correct?

ekzyis:
I am saying that qr.js contains code to pay with an attached wallet and thus it probably makes sense to also have fallbacks here for a consistent experience. I am not saying that we should use the code in qr.js in use-paid-mutation.

Maybe we don't need paying from attached wallets for QR codes anymore if a QR code is only shown if all wallets already failed, but I would need to check myself first. All I saw is that there was a loop that would pay the same invoice again.

riccardobl:
> Maybe we don't need paying from attached wallets for QR codes anymore if a QR code is indeed only shown if all wallets already failed in all cases, but I would need to check myself first. All I saw is that there was a loop that would pay the same invoice again.

From what i can tell, the QR code wallet payment is used only in `pages/invoices/[id].js`.
To me, this seems like a generic interface to pay invoices.

ekzyis:
Superseded by #1642 and #1688. Closing.

Issue #1524, opened 2024-10-28
Show original code in errors


ekzyis:
### Describe the problem you're trying to solve

We have production sourcemaps enabled via `productionBrowserSourceMaps` but when a stacker encounters an error, the stack trace does only include minified code.

I suspect we need to do some more to go from error to where that error happened in the source code.

### Describe the solution you'd like

If one clicks on `copy error information`, it shows a stack trace that points to source code.

### Describe alternatives you've considered

n/a

### Additional context

https://stacker.news/items/743444

riccardobl:
how can i trigger this error on purpose in the local dev environment?

riccardobl:
This should work: https://github.com/stackernews/stacker.news/pull/1528

The error reported in the thread resolves to:

```
Decoded stacktrace:
    
@webpack://_N_E/node_modules/mathjax-full/js/input/tex/base/BaseConfiguration.js:79:0
env@webpack://_N_E/node_modules/mathjax-full/js/input/tex/MapHandler.js:92:0
@webpack://_N_E/node_modules/mathjax-full/js/input/tex/TexParser.js:143:0
@webpack://_N_E/node_modules/mathjax-full/js/input/tex/TexParser.js:76:0
@webpack://_N_E/node_modules/mathjax-full/js/input/tex.js:118:0
@webpack://_N_E/node_modules/mathjax-full/js/core/MathItem.js:56:0
document@webpack://_N_E/node_modules/mathjax-full/js/core/MathDocument.js:119:40
math@webpack://_N_E/node_modules/mathjax-full/js/core/MathDocument.js:171:0
@webpack://_N_E/node_modules/mathjax-full/js/core/MathItem.js:52:0
@webpack://_N_E/node_modules/mathjax-full/js/core/MathDocument.js:329:0
value@webpack://_N_E/node_modules/rehype-mathjax/lib/create-renderer.js:49:46
text@webpack://_N_E/node_modules/rehype-mathjax/lib/create-plugin.js:264:0
node@webpack://_N_E/node_modules/unist-util-visit-parents/lib/index.js:346:0
grandparents@webpack://_N_E/node_modules/unist-util-visit-parents/lib/index.js:363:0
grandparents@webpack://_N_E/node_modules/unist-util-visit-parents/lib/index.js:363:0
@webpack://_N_E/node_modules/unist-util-visit-parents/lib/index.js:304:0
[...]

PR #1528, opened 2024-10-31
decode minified stacktrace

riccardobl:
## Description
This pull request addresses [issue #1524](https://github.com/stackernews/stacker.news/issues/1524). It adds a helper function that fetches and utilizes the source .map files to decode the stack trace and implements it when a user clicks on "Copy Error Information."

## Checklist

yes

9 - This implementation looks for a specific stack trace pattern, but I am unsure if it may vary. However, it does return the original stack trace intact, so any failure in pattern matching will not lead to losing information in the report.

### DIFF
--- a/components/error-boundary.js
+++ b/components/error-boundary.js
@@ -6,7 +6,7 @@ import copy from 'clipboard-copy'
 import { LoggerContext } from './logger'
 import Button from 'react-bootstrap/Button'
 import { useToast } from './toast'
-
+import { decodeMinifiedStackTrace } from '@/lib/stacktrace'
 class ErrorBoundary extends Component {
   constructor (props) {
     super(props)
@@ -27,7 +27,7 @@ class ErrorBoundary extends Component {
   getErrorDetails () {
     let details = this.state.error.stack
     if (this.state.errorInfo?.componentStack) {
-      details += `\n\nComponent stack:${this.state.errorInfo.componentStack}`
+      details += `\n\nComponent stack:\n    ${this.state.errorInfo.componentStack}`
     }
     return details
   }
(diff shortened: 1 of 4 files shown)


ekzyis:
Mhh, I just read the [NextJS docs about `productionBrowserSourceMaps`](https://nextjs.org/docs/pages/api-reference/next-config-js/productionBrowserSourceMaps) again. I read:

> When the `productionBrowserSourceMaps` option is enabled, the source maps will be output in the same directory as the JavaScript files. Next.js will automatically serve these files when requested.

So this means these source maps aren't automatically requested on errors? I hoped we just need to use NextJS more / properly instead of writing a custom stack trace decoder. But fine for me if there's no better solution. Will pull code and test today.

riccardobl:
> Mhh, I just read the [NextJS docs about `productionBrowserSourceMaps`](https://nextjs.org/docs/pages/api-reference/next-config-js/productionBrowserSourceMaps) again. I read:
> 
> > When the `productionBrowserSourceMaps` option is enabled, the source maps will be output in the same directory as the JavaScript files. Next.js will automatically serve these files when requested.
> 
> So this means these source maps aren't automatically requested on errors? I hoped we just need to use NextJS more / properly instead of writing a custom stack trace decoder. But fine for me if there's no better solution. Will pull code and test today.

I might be wrong, but i did some digging and my understanding is that they are just there ( ie. https://a.stacker.news/_next/static/chunks/pages/_app-f2a569540a712597.js.map ) for whatever logging/monitoring tool or code you want to use.

huumn:
This should be a huge improvement and hopefully someday we'll have a better means of alerting us to clientside errors/context.

PR #1539, opened 2024-11-05
fix #1538

### LINKED ISSUE #1538: New wallet validation code has cyclic dependency

### Describe the bug

Needs https://github.com/stackernews/stacker.news/pull/1537 or it won't reach the code that causes the issue.
Typing anything into the blink attachment will cause a cyclic dependency exception in the yup schema:

```
Unhandled Runtime Error
Error: Cyclic dependency, node was:"apiKeyRecv"

Source
wallets/validate.js (104:39) @ shape

  102 |   // we use Object.keys(schemaShape).reverse() to avoid cyclic dependencies in Yup schema
  103 |   // see https://github.com/jquense/yup/issues/176#issuecomment-367352042
> 104 |   const composedSchema = Yup.object().shape(schemaShape, Object.keys(schemaShape).reverse()).concat(Yup.object({
      |                                       ^
  105 |     enabled: Yup.boolean(),
  106 |     priority: Yup.number().min(0, 'must be at least 0').max(100, 'must be at most 100')
  107 |   }))
```

### Screenshots

image
### Steps To Reproduce

1. open blink wallet attachment with the current master
2. type anything in the first field

### Expected behavior

should not throw an exception

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

huumn:
fix #1538

afaict this fixes the cyclical dep issue. Breaking cyclical deps in yup requires pairs of tie breakers, e.g. `[['apiKey', 'apiKeyRecv'],['currency', 'currencyRecv']]`. Before #1479, we were coincidentally tie breaking using an array of all field names (with sometimes more than 2 members).

I did not read the issue fully before that describes this escape hatch: https://github.com/jquense/yup/issues/176#issuecomment-367352042

QA: `6`. I couldn't test blink in full, but it passes validation now if one of field pairs are set and doesn't pass if neither of them are. I also disabled validation on the client to test it on the server (it also fails if both `requiredWithout` fields aren't set, and succeeds if one is set).

### DIFF
--- a/wallets/blink/index.js
+++ b/wallets/blink/index.js
@@ -15,8 +15,8 @@ export const fields = [
     validate: string()
       .matches(/^blink_[A-Za-z0-9]+$/, { message: 'must match pattern blink_A-Za-z0-9' }),
     help: `you can get an API key from [Blink Dashboard](${galoyBlinkDashboardUrl}).\nPlease make sure to select ONLY the 'Read' and 'Write' scopes when generating this API key.`,
-    optional: 'for sending',
-    requiredWithout: 'apiKeyRecv'
+    requiredWithout: 'apiKeyRecv',
+    optional: 'for sending'
   },
   {
     name: 'currency',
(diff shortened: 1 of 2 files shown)


riccardobl (commented):
This fixes it, thanks.
But now empty fields do not use the default values, eg. if i type nothing into the currency field it will error saying "configuration must be able to send or receive" because the currency field is empty, while it should be set to BTC by the validation
```.transform(value => value ? value.toUpperCase() : 'BTC')```

riccardobl (approved):
With this and: https://github.com/stackernews/stacker.news/pull/1540 
everything is back to work as expected :+1:

PR #1540, opened 2024-11-05
add explicit support for default values in input fields

riccardobl:
We use the yup validation and value to set default values, but this seems fragile and prone to break as result of side effects from other code changes. 

This pr adds an explicit defaultValue property to inputs.


huumn:
lol you beat me to it

Issue #1547, opened 2024-11-06
Production build can't find module `lightning`


huumn:
I suspect this is caused by #1521 given `lib/lnd.js` didn't exist before then.

This is going to be tricky to reproduce in dev, but it'll require getting an `npm run build` going then attempting to run it.

```
Nov  6 01:57:31 ip-172-31-81-59 web: ⨯ Error: Cannot find module 'lightning'
Nov  6 01:57:31 ip-172-31-81-59 web: Require stack:
Nov  6 01:57:31 ip-172-31-81-59 web: - /var/app/staging/lib/lnd.js
Nov  6 01:57:31 ip-172-31-81-59 web: at Module._resolveFilename (node:internal/modules/cjs/loader:1140:15)
Nov  6 01:57:31 ip-172-31-81-59 web: at /var/app/current/node_modules/next/dist/server/require-hook.js:55:36
Nov  6 01:57:31 ip-172-31-81-59 web: at Function.resolve (node:internal/modules/helpers:188:19)
Nov  6 01:57:31 ip-172-31-81-59 web: at c (/var/app/current/.next/server/chunks/4926.js:3935:447)
Nov  6 01:57:31 ip-172-31-81-59 web: at /var/app/current/.next/server/chunks/4926.js:1936:2990 {
Nov  6 01:57:31 ip-172-31-81-59 web: code: 'MODULE_NOT_FOUND',
Nov  6 01:57:31 ip-172-31-81-59 web: requireStack: [ '/var/app/staging/lib/lnd.js' ]
Nov  6 01:57:31 ip-172-31-81-59 web: }
```

ekzyis:
Wild guess, but it's related to relative paths which are different in the production build?

huumn:
That would make sense. I just don't know where the relative path is coming from.

riccardobl:
This is due to the way the code gets the absolute path of the protobuf files 
```
  const lightningModulePath = req.resolve('lightning')
  const pathForProto = proto => join(lightningModulePath, protosDir, proto)
```

likely webpacks does some module mangling in production to remove unused code and somehow we end up without this module that is needed at runtime.

Imo we need to revert back to the privoxy workaround or just for the ln-service and lightning repos

riccardobl:
I've come up with a workaround https://github.com/stackernews/stacker.news/pull/1548 , but i can't test it fully in the dev environment.

PR #1521, opened 2024-10-25
TOR patch

riccardobl:
## Description

This pr adds tor to the dev environment , generates an onion address and fixes some issues that could cause tor 
 to not work as expected in production.

Closes #1493  
Closes #1531

## LND and gRPC-js ✓
To connect to an onion address using grpc-js in ln-service, [the only available solution](https://github.com/alexbosworth/lightning/issues/130) appears to be setting the environment variable GRPC_PROXY or HTTPS_PROXY. However, this forces all gRPC traffic through Tor, which is not ideal.

~~To address this, I set up Privoxy in front of Tor and configured it to route only .onion traffic through the network. This HTTP (gRPC) only proxy runs alongside the Tor HTTP proxy on port 7051.~~ this is now handled by a patched authenticatedLndGrpc that set  grpc.enable_http_proxy

## CLN ✓✓
cln worked out of the box once the hidden service was connected.

## LNBITS ✓✓
~~lnbits did not work out of the box. I had to replace the use of fetch with cross-fetch to make it functional, as the native fetch lacks support for agents. With this change, the connection works, but I'm still unable to connect successfully since the lnbits instance is set up to use VoidWallet, which is unrelated to this PR.~~
worked out of the box once the lnd paths were restored

## New commands

I've added some new commands to sndev

- `sndev stacker_lnd get_onion `: returns the onion host:port for the stacker_lnd instance
- `sndev stacker_lnd get_cert` : returns the tls cert for stacker_lnd (it now needs to include the onion address so it cannot be static)
- `sndev tor get_onion`: returns the onion address (same as sndev stacker_lnd get_onion but without the lnd port)
- `sndev stacker_clncli get_onion`: returns the onion address (host:port)
- `sndev lnbits get_onion`: returns the onion address (host:port)

## Checklist

**Are your changes backwards compatible? Please answer below:**

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**

all the wallets connect and the testInvoice succeed ~~(except lnbits, see [LNBITS](#LNBITS) )~~

**Did you introduce any new environment variables? If so, call them out explicitly here:**

GRPC_PROXY

PR #1548, opened 2024-11-06
TOR patch V2 (re: #1521)

riccardobl:
Reapplies #1521 but patches authenticatedLndGrpc using an external module https://github.com/stackernews/proxied-lnd-service. 

This should make webpack happy and close https://github.com/stackernews/stacker.news/issues/1547


huumn:
Sorry that this path I had us go down is causing so many problems. 

It's probably forking the whole thing so it's easier to apply patches to the proto files (which do tend to change).

Bos might also accept a PR to provide additional GRPC params.

riccardobl:
superseded by https://github.com/stackernews/stacker.news/pull/1551

PR #1551, opened 2024-11-07
fix lightning module resolution

huumn:
## Description

This reverts the revert of #1521. It aims to fix the module resolution we're seeing when locating `.proto` files for `lightning` by locating them before `webpack` runs and storing them in an environment variable `LIGHTNING_MODULE_PATH`.

After spending most of the day trying to write my own version of nextjs's module resolution, I realized we were just using `tsx` weirdly by making modules of our own directories. Not doing that allows `require.resolve` to work in `worker` and avoid some of the weird tricks we used to deal with CJS deps in the worker.

I still need to test if this works in production, but it should because `require.resolve` - when not using webpack - returns an absolute path.


riccardobl (commented):
Seems good to me.  I've tested it in the dev environment.
Everything works as expected in worker, apis and wallet attachment.
LND tor connection is also working as expected.

I never would have thought that these little package.json files were causing all this trouble. Nice find!

huumn:
The commonjs/esm stuff in nodejs is an abomination! :) I have no idea why the community never managed to solve it.

docker/lnd/stacker/tls.cert
ekzyis, line 1: Was deleting these files an accident?

ekzyis: https://github.com/stackernews/stacker.news/pull/1554

PR #1552, opened 2024-11-07
listen for local storage wallet changes

### LINKED ISSUE #1550: New local sending wallets aren't seen by other tabs until reload

### Describe the bug

If I configure a new sending wallet, it is not used until I refresh the page.

### Screenshots

https://github.com/user-attachments/assets/2a11a836-f355-4307-8527-fb73d8b9cb96

### Steps To Reproduce

1. Configure sending wallet in one tab
2. Try to zap in other tab
3. You get a QR code

### Expected behavior

The wallet I just configured is automatically used.

### Logs

_No response_

### Device information

_No response_

### Additional context

~This is most likely caused by saves not triggering `useWallet` to return the updated wallet.~ It is not, it's caused by different tabs.

huumn:
## Description

Add a `storage` even listener to pickup wallet updates in other tabs. fix #1550

QA: `8`. Opened two tabs to nwc wallet, modified config and watched it change. Detached one, watched the other go blank. Reattached one and was able to zap in the other without reload.

### DIFF
--- a/wallets/index.js
+++ b/wallets/index.js
@@ -41,7 +41,17 @@ function useLocalWallets () {
   }, [wallets, setWallets, me?.id])
 
   useEffect(() => {
+    // listen for changes to any wallet config in local storage
+    // from any window with the same origin
+    const handleStorage = (event) => {
+      if (event.key.startsWith(getStorageKey(''))) {
+        loadWallets()
+      }
+    }
+    window.addEventListener('storage', handleStorage)
+
     loadWallets()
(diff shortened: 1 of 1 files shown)


ekzyis (approved):
Wow, didn't know there was a [`StorageEvent`](https://developer.mozilla.org/en-US/docs/Web/API/StorageEvent), that's awesome! Even works across windows:

https://github.com/user-attachments/assets/98ac2a6e-aead-4a3e-a02d-98e7e8ab4819

Maybe there is more fun to be had with it :eyes:

PR #1554, opened 2024-11-07
Fix missing TLS certificates for stacker_lnd

ekzyis:
## Description

I noticed that local NWC couldn't connect to LND anymore because of TLS errors after I rebased on d1175493.

## Checklist

yes

`9`. local NWC can connect to LND again

n/a

no


ekzyis:
You are right, `sndev delete && sndev start` fixed it. I thought I did that but guess I didn't :shrug:

PR #1555, opened 2024-11-07
Fix invoice.stopWaiting() cancels all pending invoices

ekzyis:
## Description

I have noticed that if you zap in quick succession and all zaps fail, only the invoice of the first zap is canceled and is shown as a QR code:

This is caused by `invoice.stopWaiting()` affecting all pending invoices.

## Videos

<details>
<summary>bug</summary>

https://github.com/user-attachments/assets/d7143d41-66ce-4a55-ac54-8b374069ca93

</details>

<details>
<summary>proof of fix</summary>

https://github.com/user-attachments/assets/b25791e4-ab2d-4cbe-aad1-6b659b913694

</details>

## Additional Context

I am not entirely sure yet why that happens since one would assume that `useInvoice` returns a controller for each call but maybe that's not actually the case? Probably related to `useMemo` even though it _does_ create a new `AbortController` for each invoice so it should return a separate `waitController` for each.

I will investigate why this fix works more.

## Checklist

`6`. I don't understand the fix 100% but pretty sure that fixes it since I tried to zap as many times as possible before the first QR code pops up to trigger the bug again.

n/a

no

### DIFF
--- a/components/payment.js
+++ b/components/payment.js
@@ -82,7 +82,8 @@ export const useInvoice = () => {
           }
         }, FAST_POLL_INTERVAL)
 
-        const abort = () => {
+        const abort = (invoiceId) => {
+          if (id !== invoiceId) return
           console.info(`invoice #${id}: stopped waiting`)
           resolve()
           clearInterval(interval)
@@ -92,7 +93,7 @@ export const useInvoice = () => {
       })
     }
 
-    controller.stop = () => controller.abort()
+    controller.stop = (id) => controller.abort(id)
 
     return controller
   }, [isInvoice])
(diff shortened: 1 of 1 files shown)


huumn:
`useInvoice` has always been a bit of a weird hook. It has no state. In OOP terms, it's pretty much a static class.

huumn:
AFAICT all `useInvoice` calls will use the same controller. That's the underlying cause.

huumn:
I think what we want to do is decouple invoice "control" (need one instance per invoice) from the invoice create/cancel (this can work like a static class).

Invoice control should probably just be a function call `invoiceController(isInvoice, id, waitFor)`

huumn:
If you want, I can pick this up. I think this code needs to refactored. There's quite a bit of weirdness - lots of stateless hooks.

ekzyis:
Okay, I think you're better at refactoring and making architectural decisions. I'll rebase my PR on this fix and see if I find anything else then.

PR #1557, opened 2024-11-08
fix invoice waiting

huumn:
## Description

This more directly addresses the underlying problem that #1555 solves. See the discussion there for more explaination.

Basically this creates a new invoice controller for every payment that's made rather than reusing the same one.

This can use a more serious refactor, but it's not worth it right now - it needs a real "meditation" to figure out how to improve it.

QA: `6`. I tested zaps and greped around for any use of `waitUntilPaid` which was very minimal.

@ekzyis can you confirm it fixes the issue you originally identified. I don't have a reproduction.

### DIFF
--- a/components/payment.js
+++ b/components/payment.js
@@ -1,4 +1,4 @@
-import { useCallback, useMemo } from 'react'
+import { useCallback } from 'react'
 import { useMe } from './me'
 import { gql, useApolloClient, useMutation } from '@apollo/client'
 import { useWallet } from '@/wallets/index'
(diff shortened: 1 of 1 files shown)


ekzyis:
> @ekzyis can you confirm it fixes the issue you originally identified. I don't have a reproduction.

it does!

ekzyis (approved):
functional programming ftw

ekzyis:
Going to merge so I can rebase on master

Issue #1495, opened 2024-10-18
redesign of attached wallet page


huumn:
It's unclear how *exactly* to improve it, but some of the feedback I've received is people expecting to see logos for specific wallets rather than the protocol name.

riccardobl:
What if we keep the current page the same for *manual/advanced setup*, and we add an additional wizard page where the user can chose specific wallets by name/logo and then run a guided setup that ultimately ends up setting one or more attached wallet?

If we make a setup specific to the wallet providers we can, for example, use their apis to grab ln-addresses when we can't generate a scoped receiver nwc tokens or at least tell the user exactly what we need and link to relevant documentation.

huumn:
I like the idea of wizard skip/"bypass" to just see a wall of wallets. Good idea.

ekzyis:
> What if we keep the current page the same for manual/advanced setup, and we add an additional wizard page where the user can chose specific wallets by name/logo and then run a guided setup that ultimately ends up setting one or more attached wallet?
>
> If we make a setup specific to the wallet providers we can, for example, use their apis to grab ln-addresses when we can't generate a scoped receiver nwc tokens or at least tell the user exactly what we need and link to relevant documentation.

I had the following in mind which I think is similar but differs in detail. The major difference is wallets vs protocols:

1. **We only show wallets** (currently: LNbits, Blink, LND, CLN, phoenixd), no protocols (currently: NWC, LNURL, WebLN[^1]) when you visit /settings/wallets

[^1]: I am not sure if LNC is a protocol or a wallet. Afaik, it can only be used with LND so maybe it should be part of LND's configuration for sending anyway?

Stackers want to attach wallets, not protocols. They aren't necessarily aware what protocols their wallets support. It has happened multiple times that stackers thought their wallet is not supported because it wasn't shown even though their wallet supports lightning addresses.

This means we need to show as many wallets as possible so stackers can find their wallet. We can probably limit ourselves to the most popular ones first and add any other wallet on request.

I think that we mixed wallets with protocols is becoming a problem. It was a good idea to start with protocols to cover as many wallets as possible at once but we then added "wallets" that are _actually_ wallets. We now need to consider the difference between wallets and protocols in our UI and code.

2. If you click on a wallet, a wallet wizard will start

Wallet wizards include specific guides. **A wallet wizard only configures this wallet, not any other "wallet". So if a wallet supports NWC and lightning addresses, this will configure NWC and lightning address only for this wallet.**

[...]

PR #1559, opened 2024-11-08
Wallet send+recv status derived from logs

### LINKED ISSUE #1490: Wallet status derived from wallet logs
As a follow on to #1380, we should add a status to every wallet that's derived from the last N wallet log lines.

e.g. orange if the last N lines were `warning`, red if the last line is `error`, etc.

ekzyis:
## Description

For now, this simply uses the most recent log that has `context.status` set to derive the status. `context.status` is set if `context.bolt11` is set to make sure only payments affect status.

Close #1490 Based on #1588 Related to #1495

TODO:

- [x] ~refactor workaround for conditional polling of wallet logs via `useInterval`~ ~_maybe it's not as dumb as I initially thought since alternatives look dumber_~ _no longer needed_
- [x] separate status for send and receive + icons
- [x] with icons for send+recv, we don't need the badges anymore and can put logos in there, see https://github.com/stackernews/stacker.news/issues/1495#issuecomment-2450151673
- [x] fix IDB bug where first page is never fetched (see #1588)
- [x] fix WebLN looks enabled when it isn't
- [x] ~fix logging payments (see https://github.com/stackernews/stacker.news/pull/1588#issuecomment-2474909027)~ _I decided to simply also use polling logs for wallet status so no code change needed._
- [x] fix reordering
- [x] fix disabling wallet removes sending credentials

## Screenshots

2024-11-13-220522_1920x1080_scrot

2024-11-13-222833_1920x1080_scrot

## Video

https://github.com/user-attachments/assets/ae62a6bc-baca-4402-92b9-62e2270c2da2

## Additional Context

1. I like the grey background for the icons in this:

2024-11-13-223506_173x199_scrot

more than this:

382002204-ecc2d9aa-2781-4eda-a4f8-1af5dfa81552

even though it started as an accident.

2. I am setting `context.recv = true` in the `walletLogger` (the function) such that every log in the db will have `recv: true` inserted in the database. We could instead rely on `!context.send` or simply add `context.recv` to all logs before returning them to the client.

3. Aspect ratio of logos are different but all images use `width='100%'` which means that an image with a higher height/width ratio appears bigger. Good example is the LND logo on the config page since it's more a square.

## Checklist

yes, `context.status` + `context.send` or `context.recv` is only used to derive wallet status. If no log has this set, the wallet status for send/recv only depends on if the wallet is configured.

`8`. See video for Q&A of LNbits. I also tested NWC. I also tested reordering and saving combinations of send/recv.

n/a

no


components/log-message.js
```diff
@@ -27,6 +27,14 @@ export default function LogMessage ({ showWallet, wallet, level, message, contex
 
   const style = hasContext ? { cursor: 'pointer' } : { cursor: 'inherit' }
   const indicator = hasContext ? (show ? '-' : '+') : <></>
+  const filteredCtx = context
```
ekzyis, line 30: we don't want to show `send` or `recv` in the context of a log message

components/wallet-buttonbar.js
```diff
@@ -11,7 +11,7 @@ export default function WalletButtonBar ({
   return (
     <div className={`mt-3 ${className}`}>
       <div className='d-flex justify-content-between'>
-        {isConfigured(wallet) &&
+        {isConfigured(wallet) && wallet.def.requiresConfig &&
```
ekzyis, line 14: I don't like that WebLN always showed `detach`. This fixes it by only showing it if there's actually something to delete.

wallets/lnbits/client.js
```diff
@@ -8,7 +8,7 @@ export async function testSendPayment ({ url, adminKey, invoiceKey }, { logger }
   url = url.replace(/\/+$/, '')
   await getWallet({ url, adminKey, invoiceKey })
 
-  logger.ok('wallet found')
+  logger.info('wallet found')
```
ekzyis, line 11: I didn't want this to affect the wallet status

ekzyis: I reverted this and instead use `context.status` now to determine which log messages should determine wallet status. This is only set for payments (= `context.bolt11` is set).

pages/settings/wallets/[wallet].js
```diff
@@ -67,10 +67,16 @@ export default function WalletSettings () {
...
+  const { card: { image, title, subtitle } } = wallet?.def || { card: {} }
+
   return (
     <CenterLayout>
-      <h2 className='pb-2'>{wallet?.def.card.title}</h2>
-      <h6 className='text-muted text-center pb-3'><Text>{wallet?.def.card.subtitle}</Text></h6>
+      {image
+        ? typeof image === 'object'
+          ? <img {...image} alt={title} className='pb-2' />
+          : <img src={image} width='33%' alt={title} className='pb-2' />
```
Comment (huumn) [inline pages/settings/wallets/[wallet].js:77]: I'm not sure this img path thing will work in prod ... they might need to go in `./public`.

pages/settings/wallets/[wallet].js
```diff
@@ -67,10 +67,16 @@ export default function WalletSettings () {
     }
   }, [wallet.def])
 
+  const { card: { image, title, subtitle } } = wallet?.def || { card: {} }
```
Comment (huumn) [inline pages/settings/wallets/[wallet].js:70]:
```suggestion
  const { card: { image, title, subtitle } = {} } = wallet?.def
```

Are there wallet defs without cards now?

huumn (approved):
Very clean. Nice work!

I'm still huffy about context being denormalized but it's probably helping use iterate on the design faster.

huumn:
Screenshot 2024-11-17 at 2 47 24 PM

When doing UI stuff that involves colors make sure to test light mode.

pages/settings/wallets/[wallet].js
Comment (ekzyis) [reply pages/settings/wallets/[wallet].js:70]: Mhh, no, I just saw us using `wallet?.def` a lot so I wanted to assume `wallet?.def` can actually be `undefined` in some case here so I added `|| { card: {} }`. But your solution looks better and I didn't think much about this line.

Imo, we use optional chaining in a lot of places where it shouldn't be needed but it's a trade-off between failing fast (= crashing) and subtle bugs so I can see the point of using it even when in theory, it shouldn't be needed.

Comment (huumn) [reply pages/settings/wallets/[wallet].js:70]: Oh right, that would cause a crash. I got in the optional chaining habit back when I was learning react a few years ago. It'd probably be better to do explicit checks or something now.

ekzyis:
Oh noes :eyes: Should we give the cards always a dark background?

huumn:
I'm not sure. There are two options:

1. give the cards a medium grey color in light mode
2. support both dark/light images for wallet cards

ekzyis:
Went with the second option in 362bc736 since the logos were easy enough to convert to light mode:

https://github.com/user-attachments/assets/611924e2-5f9b-4cbe-9fbb-f8b6bc23abb4

PR #1565, opened 2024-11-09
Fix outlawed can show up on front page

ekzyis:
## Description

Not tested but I believe this is what was missing.

We probably only need the outlaw clauses since the other clauses shouldn't be relevant for the front page but I think they don't hurt (except the extra user query maybe).

But we also use `filterClause` for the mostly empty subs so I guess we should apply the same filtering here.

## Checklist

yes

`1`. It does not crash the server.

n/a

no

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -535,6 +535,7 @@ export default {
                       '"Item".bio = false',
                       ad ? `"Item".id <> ${ad.id}` : '',
                       activeOrMine(me),
+                      await filterClause(me, models, type),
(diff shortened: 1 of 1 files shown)


huumn:
> 1. It does not crash the server.

😆

ekzyis:
we gotta have standards

huumn:
lol Agreed, standards have to start somewhere

huumn (approved):
I tested this and it does the job. For perf reasons, I might choose to fix the bug in the materialized view, but if I don't manage to get around to that, we'll ship this.

Issue #1422, opened 2024-09-23
Add collapsable answers


SouthKoreaLN:
**Is your feature request related to a problem? Please describe.**
Nothing important, nothing frustrating. Just thought that if I want to keep posting puzzles like [this one](https://stacker.news/items/696880), it'd be nice to have a way to collapse an answer to hide spoilers. But then again, unlikely this will ever be the primary use of SN, so I won't care much if it does not get implemented :)

**Describe the solution you'd like**
Something like what is illustrated here: https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab

**Describe alternatives you've considered**
Alternative: just ask people not to look at the answers before replying.

kravhen:
Would something like a simple spoiler concealer like this work?
![image](https://github.com/user-attachments/assets/5885a727-968a-4076-adb3-8726957da298)
![sn-spoiler-tag](https://github.com/user-attachments/assets/4ed42165-f26a-470e-8b00-f3d6e18c27e5)

This would use the existing rehype stylers and CSS method for custom tags.

Other options to make it look like a real `<details>`+`<summary>` HTML element as in the gist seems like it'd require a clunkier solution that might not be necessary, like defining a Spoiler component or something.

For the purposes mentioned, the first option seems preferable to me.

Thoughts?

SouthKoreaLN:
This would look like it fits my purpose... as a user, this would just require me adding some tags around the part that needs to be concealed?

kravhen:
Yup, you would write it `<spoiler>`like this`</spoiler>`. Maybe `<hide>` to make it shorter.

SouthKoreaLN:
For me, this would work, if it works for larger blocks of text too.

huumn:
To support more use cases, I'd prefer a more generic approach like github's `details`. 

<details>
    <summary>like this</summary>

This was made with the following markdown:

```md
<details>
    <summary>like this</summary>

This was made with the following markdown:
[...]
</details>
```
</details>

SouthKoreaLN:
Darthcoin gave an interesting suggestion about having to [pay to view the answer](https://stacker.news/items/736461/r/south_korea_ln?commentId=736471). Not fully convinced yet as I was just thinking of adding an extra click before one sees the answer, but in the case this suggestion makes sense, it would be best then to have the sats go to the stacker who posted the hidden answer. This might add a layer of complexity though to this issue.

kravhen:
[...]

PR #1567, opened 2024-11-10
Add handling for details & summary tags (collapsible sections)

kravhen:
## Description
PR to solve #1422 
Handles `<details>` and `<summary>` tags to allow for users to have collapsible sections in text. 
Behavior is close to 1:1 with that of Github's handling.
Minor differences: 
- on github, you need a blank line (`\n\n`) to have a proper newline in the resulting text. In this implementation, all newlines are handled (`\n`).
- on github, markdown inside `<details>` is not handled. In this implementation, it is.

## Screenshots
The following text showing different structures will render as pictured on subsequent image:

```
Structure 1:

<details>no summary tags
lorem ipsum
and text on a new line</details>

----------

Structure 2:

<details>
<summary>text inside summary</summary>
lorem ipsum
</details>

----------

Structure 3:

<details>
    <summary>summary indentation single line</summary>
1. first thing
2. second thing 
3. third thing
</details>

----------

Structure 4:

<details>
  <summary>
    summary text here
  </summary>
  text inside details with *markdown* working **properly**
</details>

----------

Structure 5:

<details>
<summary>Shopping list</summary>
- Vegetables
- Fruits
- Fish
</details>

```
image

## Additional Context

I was able to get it to work very close to github-like behavior without the additional imports of
```
import { fromMarkdown } from 'mdast-util-from-markdown'
import { toHast } from 'mdast-util-to-hast'
```
except I couldn't get inner lists and inner markdown to work, noting that markdown does not work inside `<details>` on github, but lists do.

## Checklist

Yes. Existing tags (though unlikely to have been used since they were not handled and would appear as plain text) will now render properly.

7/10
Tested a variety of markdown inside `<details>` and various ways of structuring the tags: all tags on one line, a few lines, indentations, etc.
It works in bios. 
Tags remain as plaintext in post titles as expected.
Will QA some more while in draft.

Yes. Works on mobile as expected.

No.

## DRAFT
Creating this as draft for now for the following reasons:
- I would make a pass over styling to try to spruce it up a little visually, keeping consistent with the site's color scheme. I'm not a designer though so I'd try to make it slick but keep it subtle, simple.
- I can try to make newlines work as on github while keeping the markdown processing functionality, but am welcoming feedback if we rather it be truly 1:1 with behavior on github and forget markdown inside the tags.
- While merging this, I could also, if maintainers want, add a commit with the initial `<spoiler>` solution. That solution was little additional code and lean, leveraging the stylers already used for `<sub>` and `<sup>`. The use cases could be also be different. A spoiler for someone who wants to hide the winner of a sports match or a character dying in a show. The `<details>`/`<summary>` tags for breaking down bigger posts into smaller chunks or chapters, or also to hide answers like @SouthKoreaLN wished.


kravhen:
The implementation handles HTML `<details>` and `<summary>` tags within markdown content, transforming them into proper HTML elements while preserving their content and structure. It's particularly complex because it needs to handle various formats users might write these tags in, including multi-line content, nested tags, and mixed markdown/HTML content.

kravhen:
I learned a lot about merging and rebasing, but the commit history is a helluva mess due to trying different things and forgetting other things since I first opened this.

Will submit a clean PR.

PR #1570, opened 2024-11-10
Implement deposit as receive paidAction

riccardobl:
As title says.

This is pretty much done. I need to review everything again though and I'm concerned about the dual use and display of invoices/withdrawls (ie with/without invoiceforwards) ... it's probably worth giving forwards their own dedicated pages and notifications.

(Unfortunately, I noticed we were importing files from the `wallets` dir like it was a package, ie an implicit relative path, e.g. `import * from 'wallet/client', and couldn't stop myself from fixing it in the middle of this PR ... so if you're reviewing, ignore that.)

TODO:

- [x] prevent two outgoing payments with same hash ~~(requires implementing send paidAction)~~
    - send will require big changes best suited to another PR. See #1591
- [x] throw descriptive user facing error when the above constraint is hit
- [x] make sure accounts cannot drive their balance too high
- [x] test receive notifications
- [x] enhance invoice page to show whether p2p/autowithdraw and maybe which wallet received the payment
- [x] retest lnurlp
- [x] drop `create_invoice` function from postgres


api/paidAction/transfer.js
```diff
@@ -0,0 +1,57 @@
...
+  return targetUserId
+}
+
+export async function perform ({ invoiceId, sats, description, descriptionHash, comment, targetUserId }, { me, tx }) {
+  const invoice = await tx.invoice.update({
+    where: { id: invoiceId },
+    data: {
+      comment
+    }
+  })
```
riccardobl, line 25: this is adding the 'comment' to the invoice, since it is not part of the generic paidAction flow

api/resolvers/notifications.js
```diff
@@ -228,6 +228,23 @@ export default {
...
+            AND "actionState" = 'PAID'
+            AND "actionType" = 'TRANSFER'
+            AND created_at < $2
+            AND NOT EXISTS (
+              SELECT 1
+              FROM "InvoiceForward"
+              WHERE "InvoiceForward"."invoiceId" = "Invoice".id
+            )              
+            ORDER BY "sortTime" DESC
+            LIMIT ${LIMIT})`
```
riccardobl, line 246: the current code does not select these actions that are not associated with an item and not p2p, so we add this code here that does just that but only for the actionType=TRANSFER

api/resolvers/wallet.js
```diff
@@ -126,7 +126,11 @@ export async function getWithdrawl (parent, { id }, { me, models, lnd }) {
     },
     include: {
       user: true,
-      invoiceForward: true
+      invoiceForward: {
+        include: {
+          invoice: true
+        }
+      }
```
riccardobl, line 133: we need this to fetch the comment of the invoice to display it in the notification page

components/notifications.js
```diff
@@ -484,13 +490,19 @@ function Invoicification ({ n: { invoice, sortTime } }) {
...
       <Check className='fill-info me-1' />{numWithUnits(n.earnedSats + n.withdrawl.satsFeePaid, { abbreviate: false, unitSingular: 'sat was ', unitPlural: 'sats were ' })}
       {n.withdrawl.p2p || n.withdrawl.autoWithdraw ? 'sent to your attached wallet' : 'withdrawn from your account'}
       <small className='text-muted ms-1 fw-normal' suppressHydrationWarning>{timeSince(new Date(n.sortTime))}</small>
       {(n.withdrawl.p2p && <Badge className={styles.badge} bg={null}>p2p</Badge>) ||
       (n.withdrawl.autoWithdraw && <Badge className={styles.badge} bg={null}>autowithdraw</Badge>)}
+      {n.withdrawl.invoiceForward?.[0]?.invoice?.comment &&
+        <small className='d-block ms-4 ps-1 mt-1 mb-1 text-muted fw-normal'>
+          <Text>{n.withdrawl.invoiceForward[0].invoice.comment}</Text>
+          {payerSig}
+        </small>}
```
riccardobl, line 505: we make withdrawals notifications that are caused by p2p actions (such as lnurlp transfer), display their invoice comment  if they have one.

api/paidAction/transfer.js
```diff
@@ -0,0 +1,57 @@
...
+  await notifyDeposit(targetUserId, invoice)
```
huumn, line 26: Won't this notify them every time a deposit is *attempted* not every time a deposit *succeeds*?

api/paidAction/transfer.js
```diff
@@ -0,0 +1,57 @@
...
+  return { sats, targetUserId }
+}
+
+export async function describe ({ sats, description, descriptionHash }, context) {
+  return `SN: ${description ?? ''}`
+}
+
+export async function describeHash ({ sats, description, descriptionHash }, context) {
```
huumn, line 34: It looks like you added quite a few new exports from `paidAction` modules. Can you document those in the README?

pages/api/lnurlp/[username]/pay.js
```diff
@@ -30,22 +29,23 @@ export default async ({ query: { username, amount, nostr, comment, payerdata: pa
       // If there is an amount tag, it MUST be equal to the amount query parameter
       const eventAmount = note.tags?.find(t => t[0] === 'amount')?.[1]
       if (schnorr.verify(note.sig, note.id, note.pubkey) && hasPTag && hasETag && (!eventAmount || Number(eventAmount) === Number(amount))) {
-        description = user.hideInvoiceDesc ? undefined : 'zap'
+        description = 'zap'
```
Comment (huumn) [inline pages/api/lnurlp/[username]/pay.js:32]: why are we not respecting this setting anymore?

pages/api/lnurlp/[username]/pay.js
```diff
@@ -30,22 +29,23 @@ export default async ({ query: { username, amount, nostr, comment, payerdata: pa
...
         descriptionHash = createHash('sha256').update(noteStr).digest('hex')
       } else {
         res.status(400).json({ status: 'ERROR', reason: 'invalid NIP-57 note' })
         return
       }
     } else {
-      description = user.hideInvoiceDesc ? undefined : `Funding @${username} on stacker.news`
+      description = `Funding @${username} on stacker.news`
```
Comment (huumn) [inline pages/api/lnurlp/[username]/pay.js:39]: why are we not respecting this setting anymore?

components/notifications.js
```diff
@@ -464,7 +470,7 @@ function Invoicification ({ n: { invoice, sortTime } }) {
               if (disableRetry) return
               setDisableRetry(true)
               try {
-                const { error } = await retry({ variables: { invoiceId: parseInt(invoiceId) } })
+                const { error } = await retry({ variables: { invoiceId: parseInt(invoiceId), resetAttempts: true } })
```
huumn, line 473: Does this `resetAttempts` depend on the other pr?

api/paidAction/index.js
```diff
@@ -30,6 +31,7 @@ export const paidActions = {
   TERRITORY_BILLING,
   TERRITORY_UNARCHIVE,
   DONATE,
+  TRANSFER,
```
huumn, line 34: I wonder if `EXTERNAL_P2P` might be a better name? Transfer is just a bit ambiguous in the context of SN.

api/paidAction/transfer.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export async function getSybilFeePercent ({ sats, description, descriptionHash }, context) {
+  return 10n
+}
+
+export async function onPaid ({ invoice }, { tx }) {
+  const isP2P = !!invoice.invoiceForward
+  if (isP2P) return
+  if (!invoice?.msatsReceived) throw new Error('Not paid?')
```
huumn, line 45: Did something change where somethings calling `onPaid` isn't doing this check?

huumn (commented):
I just left some first pass questions as I get familiar.

This looks like it depends on the nov-5 pr. It might be nice to get this in before nov-5 so it can be tested independently and we aren't deploying a bunch of new code at once. To do that though, we'll want to add a setting that makes wrapping the invoice optional. The plan is to allow them to receive directly to their connected wallet or with wrapping for added privacy.

pages/api/lnurlp/[username]/pay.js
Comment (riccardobl) [reply pages/api/lnurlp/[username]/pay.js:32]: because it is now passed to a paid action, so it is redundant since paid actions already respect this settings internally when they create their invoices:

for SN invoices:

https://github.com/stackernews/stacker.news/blob/4675a2c29d1de4b43b34e1bdda32a3f7b83ac81b/api/paidAction/index.js#L299


and p2p: 

https://github.com/stackernews/stacker.news/blob/4675a2c29d1de4b43b34e1bdda32a3f7b83ac81b/wallets/server.js#L159

components/notifications.js
riccardobl: Yes, it is a left over from the extraction. I will clean it up

api/paidAction/index.js
riccardobl: good point, indeed transfer is a bad name

riccardobl:
> I just left some first pass questions as I get familiar.
> 
> This looks like it depends on the nov-5 pr. It might be nice to get this in before nov-5 so it can be tested independently and we aren't deploying a bunch of new code at once. To do that though, we'll want to add a setting that makes wrapping the invoice optional. The plan is to allow them to receive directly to their connected wallet or with wrapping for added privacy.

I can rebase to master, but if we don't wrap the invoice, afaik, we can't be sure it has been paid, so how do we keep the current dm-like ux of lnurl payments?
If we just tell the user to rely on its external wallet to receive and read these messages, imo the ux would be degraded quite a lot, these attached services often are not designed with this use-case in mind (eg. you don't get push notifications), and it would conflict with the `receiver privacy` settings, meaning the user might unexpectedly lose these messages entirely.

IMO the feature should be: 
- receive to your SN wallet: ie. the behavior in the current master
or
- receive to your attached wallet: always private and wrapped, for a fee, initially of the default sybil fee, but with the other PR reduced to 10%

This would keep the current ux intact, what do you think?

api/paidAction/index.js
riccardobl: on second thought, EXTERNAL_P2P might be as confusing, since this is not always a p2p transfer (if the user has no attached wallets for example). 
Maybe just EXTERNAL or LNURLP ?

ekzyis: It can still be considered p2p since they still need to pay any QR codes with own wallets.

However, if this is exclusively about LNURLP, why not call it LNURLP?

(I don't know, I didn't read the rest of the code, I'm just replying to this message)

api/paidAction/transfer.js
riccardobl: nope, you are right this check is useless

api/paidAction/transfer.js
riccardobl: since these are fields introduced by the other pr i will just leave them out for now

huumn:
> Even though, the high sybil fee is pretty bad, maybe should I extract also the custom sybil fee code from the other PR?
>
> What do you think?

Yes, I think we should build anything that's possible to build on top of master, on top of master. 

The nov-5 branch exists for the code we'll ship when the switch is made - not things that'd be useful before the switch. @ekzyis mentioned that the name of the branch was probably confusing. The intent is for nov-5 to mostly house the last final critical changes required to make the switch.

huumn:
If there are important changes from nov-5 you'd like to see in master before the switch, lmk and I'll add them in a separate PR.

huumn:
Thanks for updating it after the merge. I'll try to get the review done by the end of my day.

components/notifications.js
```diff
@@ -339,6 +339,11 @@ function InvoicePaid ({ n }) {
 
     if (id) payerSig += id
   }
+  return payerSig
+}
+
+function InvoicePaid ({ n }) {
+  const payerSig = getPayerSig(n.invoice.lud18Data)
```
huumn, line 346: For nonP2P at least payer sig is not set for me in dev (I used all lnurl fields): 

Screenshot 2024-11-11 at 5 58 03 PM

huumn (changes requested):
I still have more reviewing to do, but I wanted to give you some of the review earlier.

Also, as a note to myself, this PR effectively removes wallet balance limits (`create_invoice` rejects invoices that will bring account balances over 100k sats) and we'll need to solve that.

pages/api/lnurlp/[username]/pay.js
```diff
@@ -71,27 +71,20 @@ export default async ({ query: { username, amount, nostr, comment, payerdata: pa
     }
 
     // generate invoice
-    const expiresAt = datePivot(new Date(), { minutes: 5 })
-    const invoice = await createInvoice({
+    const actionResponse = await performPaidAction('LNURLP', {
+      sats: toPositiveNumber(msatsToSats(BigInt(amount))),
```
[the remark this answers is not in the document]
Comment (riccardobl) [reply pages/api/lnurlp/[username]/pay.js:75]: do you think we should change the paid action code to support msats?
Is there a situation where it makes sense to send with msats precision through lnurlp? 
Considering there are even attached wallets that do not support that.

riccardobl:
> Also, as a note to myself, this PR effectively removes wallet balance limits (`create_invoice` rejects invoices that will bring account balances over 100k sats) and we'll need to solve that.

this is something we wont' have to worry post ex-nov-5 update right? So what if we add a temporary fix in onPaid that just throws if the payment is not p2p and going over the limit?

components/notifications.js
riccardobl: right, i wasn't passing the lud18 data to the action. Fixed now

huumn:
How did you test p2p? 

p2p lnurlp from one stackerA (withdraw -> lightning address) to stackerB with p2p lnurl enabled does not work and cannot work with the code as is. This is because the stackerA creates paymentA to an invoice with hashA from `sn_lnd` to `sn_lnd`. When paymentA is received by `sn_lnd` it attempts to make paymentB to an invoice with hashA, but lnd rejects it because it can't have two inflight payments with the same hash (hashA).

huumn:
To ship this, we'll need to restrict paying from SN's custodial wallet to another stacker using lnurlp p2p.

riccardobl:
> How did you test p2p?
> 
> p2p lnurlp from one stackerA (withdraw -> lightning address) to stackerB with p2p lnurl enabled does not work and cannot work with the code as is. This is because the stackerA creates paymentA to an invoice with hashA from `sn_lnd` to `sn_lnd`. When paymentA is received by `sn_lnd` it attempts to make paymentB to an invoice with hashA, but lnd rejects it because it can't have two inflight payments with the same hash (hashA).

I've been testing it using `sndev fund` to pay the invoice, from the way you are describing this, it seems you use the withdraw feature? I didn't even consider that initially, because it was supposed to be merged after the nov-5 update.

> To ship this, we'll need to restrict paying from SN's custodial wallet to another stacker using lnurlp p2p.

Maybe we can check the origin header from the lnurlp endpoint or send something in the url params? Do you have a better idea?

huumn:
> Maybe we can check the origin header from the lnurlp endpoint or send something in the url params? Do you have a better idea?

We should have enough data in the database to determine that this is happening.

huumn:
I'll takeover this pr and figure this out. It's already loaded into my working memory and it'll be faster than more rounds of review. 

Solid work on this one! It's clear you're getting more familiar with the code base - trend is strong. Please have the fallback pr in good shape for review when you can.

worker/wallet.js
```diff
@@ -171,26 +169,6 @@ export async function checkInvoice ({ data: { hash, invoice }, boss, models, lnd
...
-      VALUES ('finalizeHodlInvoice', jsonb_build_object('hash', ${hash}), 21, true, ${expiresAt})`,
-      models.invoice.update({
-        where: { hash },
-        data: {
-          msatsReceived: Number(inv.received_mtokens),
-          expiresAt,
-          isHeld: true
-        }
-      })
-    ], { models })
```
huumn, line 193: This was dead code. Hold invoices were all create in paidAction.

worker/wallet.js
```diff
@@ -144,10 +144,8 @@ export async function checkInvoice ({ data: { hash, invoice }, boss, models, lnd
       return await paidActionPaid({ data: { invoiceId: dbInv.id, invoice: inv }, models, lnd, boss })
     }
 
-    // NOTE: confirm invoice prevents double confirmations (idempotent)
-    // ALSO: is_confirmed and is_held are mutually exclusive
-    // that is, a hold invoice will first be is_held but not is_confirmed
-    // and once it's settled it will be is_confirmed but not is_held
+    // XXX we need to keep this to allow production to migrate to new paidAction flow
+    // once all non-paidAction receive invoices are migrated, we can remove this
```
huumn, line 148: Shortly after this ships we'll be able to remove this code.

worker/paidAction.js
```diff
@@ -151,9 +151,17 @@ export async function paidActionPaid ({ data: { invoiceId, ...args }, models, ln
...
-      await paidActions[dbInvoice.actionType].onPaid?.({ invoice: dbInvoice }, { models, tx, lnd })
+      const updateFields = {
+        confirmedAt: new Date(lndInvoice.confirmed_at),
+        confirmedIndex: lndInvoice.confirmed_index,
+        msatsReceived: BigInt(lndInvoice.received_mtokens)
+      }
+
+      await paidActions[dbInvoice.actionType].onPaid?.({
+        invoice: { ...dbInvoice, ...updateFields }
+      }, { models, tx, lnd })
```
huumn, line 162: This was required so that `onPaid` could see `msatsReceived`

api/paidAction/receive.js
```diff
@@ -0,0 +1,70 @@
+import { PAID_ACTION_PAYMENT_METHODS } from '@/lib/constants'
+import { toPositiveBigInt } from '@/lib/validate'
+import { notifyDeposit } from '@/lib/webPush'
+import { numWithUnits, msatsToSats } from '@/lib/format'
+
+export const anonable = false
+
+export const paymentMethods = [
+  PAID_ACTION_PAYMENT_METHODS.P2P,
+  PAID_ACTION_PAYMENT_METHODS.OPTIMISTIC
```
huumn, line 10: Because only logged in stackers can receive, this action doesn't create new db records, and hold invoices should be avoided unless we want to do refunds, we don't allow pessimistic payment flows.

api/typeDefs/wallet.js
```diff
@@ -124,6 +124,11 @@ const typeDefs = `
     itemAct: ItemAct
   }
 
+  type InvoiceForward {
+    invoice: Invoice
+  }
```
huumn, line 129: If this invoice does not belong to the same user as the withdrawal, this reveals the sender.

For lnurlp and receives, these users are the same but for p2p zaps they are different. 

While the UI would not have exposed this privacy leak, the API does. e.g. If I receive a p2p zap, I can determine who zapped me by querying the API for my withdrawl and ask it to include the invoice that corresponds to it.

PR #1577, opened 2024-11-11
configurable sybil fee

riccardobl:
Actions can define their sybil fee for wrapped invoices

### DIFF
--- a/api/paidAction/README.md
+++ b/api/paidAction/README.md
@@ -171,6 +171,7 @@ All functions have the following signature: `function(args: Object, context: Obj
     - this is only used for p2p wrapped zaps currently
 - `describe`: returns a description as a string of the action
     - for actions that require generating an invoice, and for stackers that don't hide invoice descriptions, this is used in the invoice description
+- `getSybilFeePercent` (optional): returns the action sybil fee as a `BigInt` (eg. 30n for 30%). If unspecified, the default will be used
 
 #### Function arguments
 
--- a/api/paidAction/README.md
+++ b/api/paidAction/README.md
@@ -171,7 +171,7 @@ All functions have the following signature: `function(args: Object, context: Obj
     - this is only used for p2p wrapped zaps currently
 - `describe`: returns a description as a string of the action
     - for actions that require generating an invoice, and for stackers that don't hide invoice descriptions, this is used in the invoice description
-- `getSybilFeePercent` (optional): returns the action sybil fee as a `BigInt` (eg. 30n for 30%). If unspecified, the default will be used
+- `getSybilFeePercent` (optional): returns the action sybil fee as a `BigInt` (eg. 30n for 30%). If unspecified, the default (30n) will be used
 
 #### Function arguments
 
--- a/api/paidAction/README.md
+++ b/api/paidAction/README.md
@@ -167,11 +167,11 @@ All functions have the following signature: `function(args: Object, context: Obj
     - this function is called when an optimistic action is retried
     - it's passed the original `invoiceId` and the `newInvoiceId`
     - this function should update the rows created in `perform` to contain the new `newInvoiceId` and remark the row as `PENDING`
-- `invoiceablePeer`: returns the userId of the peer that's capable of generating an invoice so they can be paid for the action
+- `getInvoiceablePeer`: returns the userId of the peer that's capable of generating an invoice so they can be paid for the action
     - this is only used for p2p wrapped zaps currently
 - `describe`: returns a description as a string of the action
     - for actions that require generating an invoice, and for stackers that don't hide invoice descriptions, this is used in the invoice description
-- `getSybilFeePercent` (optional): returns the action sybil fee as a `BigInt` (eg. 30n for 30%). If unspecified, the default (30n) will be used
+- `getSybilFeePercent` (required if `getInvoiceablePeer` is implemented): returns the action sybil fee percent as a `BigInt` (eg. 30n for 30%)
 
 #### Function arguments
 
@@ -180,6 +180,7 @@ All functions have the following signature: `function(args: Object, context: Obj
 `context` contains the following fields:
 - `me`: the user performing the action (undefined if anonymous)
 - `cost`: the cost of the action in msats as a `BigInt`
+- `sybilFeePercent`: the sybil fee percent as a `BigInt` (eg. 30n for 30%)
 - `tx`: the current transaction (for anything that needs to be done atomically with the payment)
 - `models`: the current prisma client (for anything that doesn't need to be done atomically with the payment)
 - `lnd`: the current lnd client
--- a/api/paidAction/index.js
+++ b/api/paidAction/index.js
@@ -46,7 +46,7 @@ export default async function performPaidAction (actionType, args, context) {
     context.cost = await paidAction.getCost(args, context)
 
     context.sybilFeePercent = await paidAction.getSybilFeePercent?.(args, context)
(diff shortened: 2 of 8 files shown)


lib/constants.js
```diff
@@ -182,3 +182,7 @@ export const LONG_POLL_INTERVAL = Number(process.env.NEXT_PUBLIC_LONG_POLL_INTER
 export const EXTRA_LONG_POLL_INTERVAL = Number(process.env.NEXT_PUBLIC_EXTRA_LONG_POLL_INTERVAL)
 
 export const ZAP_UNDO_DELAY_MS = 5_000
+
+// the fee for the zap sybil service
+export const ZAP_SYBIL_FEE_PERCENT = 30n
+export const MAX_FEE_ESTIMATE_PERMILE = 25n // the maximum fee relative to outgoing we'll allow for the fee estimate
```
riccardobl, line 188: reduce bigints<->number conversions

lib/validate.js
```diff
@@ -533,3 +551,43 @@ export const deviceSyncSchema = object().shape({
...
+}
+
+/**
+ * @param {number|bigint} x
+ * @returns {number|bigint}
+ */
+export const toPositive = (x) => {
+  if (typeof x === 'bigint') return toPositiveBigInt(x)
+  return toPositiveNumber(x)
+}
```
riccardobl, line 593: helpers to do checked bigint<->number conversions

api/paidAction/README.md
```diff
@@ -171,6 +171,7 @@ All functions have the following signature: `function(args: Object, context: Obj
     - this is only used for p2p wrapped zaps currently
 - `describe`: returns a description as a string of the action
     - for actions that require generating an invoice, and for stackers that don't hide invoice descriptions, this is used in the invoice description
+- `getSybilFeePercent` (optional): returns the action sybil fee as a `BigInt` (eg. 30n for 30%). If unspecified, the default will be used
```
ekzyis, line 174: Can you mention here what the default is?

api/paidAction/index.js
```diff
@@ -45,6 +45,9 @@ export default async function performPaidAction (actionType, args, context) {
     context.me = me ? await models.user.findUnique({ where: { id: me.id } }) : undefined
     context.cost = await paidAction.getCost(args, context)
 
+    context.sybilFeePercent = paidAction.getSybilFeePercent ? await paidAction.getSybilFeePercent(args, context) : undefined
```
ekzyis, line 48: You can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) even for promises and function calls

api/paidAction/index.js
```diff
@@ -45,6 +45,9 @@ export default async function performPaidAction (actionType, args, context) {
...
+    if (context.sybilFeePercent != null && context.sybilFeePercent < 0n) throw new Error('sybil fee percent cannot be negative')
```
ekzyis, line 49: I don't like `!=`, it always looks like a bug in JS. Try to use `!==`, `!` (infix operator) or optional chaining for cases like this.

For example, in this case, you could use `context?.sybilFeePercent < 0n` since comparing a number to `null` or `undefined` always returns `false`:

```
$ node
Welcome to Node.js v20.18.0.
> null < 0n
false
> null > 0n
false
> undefined < 0n
false
> undefined > 0n
false
```

But I don't know if _that_ looks like a bug to other developers.

Also, if `sybilFeePercent` is not defined, shouldn't we apply the default instead of throwing?

ekzyis (commented):
not tested, just a quick look over a part of the code

api/paidAction/index.js
riccardobl: I disagree that  `!= null` looks like a bug.  It has a [defined behavior ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality ) and this is one of the correct use cases. Since the code is checking for both null and undefined.

This is  a safety check that throws if the sybil fee is negative. null or undefined fees are fine, in this case the default will be use in wrapInvoice.

ekzyis: JS defined behavior is usually a foot gun and when I see `!=`, it's a bug in 99% of the cases in my experience (and that's why you should look out for them, too). I know considering least common denominators between devs is annoying but it will also be annoying to always pause and wonder: "did whoever wrote this code knew what they were doing?"

What's wrong with `context?.sybilFeePercent < 0n`? Can we get behind that?

huumn:
I ended up refining this PR a bit - mostly to:
1. make it consistent with other parts of the code
2. follow through with some of the things your code was implying you wanted to do

I hope you don't mind, but it was a bit easier than explaining because a lot of my `paidAction` code is funky.

Things worth noting:
1. I used `getSybilFeePercent` in the way you implied you wanted to use it (something to be passed as context to `paidAction`s) and documented its use in the `context` arg
2. unrelated to your changes, but ripe for refactor, I created a `createWrappedInvoice` function to move this unrelated wrapping logic from `paidAction`
3. I renamed `invoiceablePeer` to be consistent with your improved/clearer function naming practice `getInvoiceablePeer`
4. I refined the function signature of `wrapInvoice` a tiny bit to be more consistent with related functions
5. I made `getSybilFeePercent` return values required if `getInvoiceablePeer` returns a value and check it closer to where it's used
6. Unrelated to your changes, I increased out max fee for wrapped invoices to 3% from 2.5% (i intended to do this before, but never got around to it)

worker/paidAction.js
```diff
@@ -112,8 +112,10 @@ async function transitionInvoice (jobName, { invoiceId, fromState, toState, tran
 async function performPessimisticAction ({ lndInvoice, dbInvoice, tx, models, lnd, boss }) {
   try {
     const args = { ...dbInvoice.actionArgs, invoiceId: dbInvoice.id }
-    const result = await paidActions[dbInvoice.actionType].perform(args,
-      { models, tx, lnd, cost: BigInt(lndInvoice.received_mtokens), me: dbInvoice.user })
+    const context = { tx, cost: BigInt(lndInvoice.received_mtokens) }
+    context.sybilFeePercent = await paidActions[dbInvoice.actionType].getSybilFeePercent?.(args, context)
+
+    const result = await paidActions[dbInvoice.actionType].perform(args, context)
     await tx.invoice.update({
```
huumn, line 119: This allows `getSybilFeePercent` to work generically for any pessimistic action now.

wallets/wrap.js
```diff
@@ -10,21 +9,28 @@ const INCOMING_EXPIRATION_BUFFER_MSECS = 300_000 // the buffer enforce for the i
 const MAX_OUTGOING_CLTV_DELTA = 500 // the maximum cltv delta we'll allow for the outgoing invoice
 export const MIN_SETTLEMENT_CLTV_DELTA = 80 // the minimum blocks we'll leave for settling the incoming invoice
 const FEE_ESTIMATE_TIMEOUT_SECS = 5 // the timeout for the fee estimate request
+const MAX_FEE_ESTIMATE_PERCENT = 3n // the maximum fee relative to outgoing we'll allow for the fee estimate
```
riccardobl, line 12: isn't it better to progressively move all the constants under lib/constants so we can see and tweak all of them in a single file?

huumn: imho if the constants will only be used in one place, they should be defined in the same place they are used. It's a code locality thing.

The idiom I'm familiar with is "define a variable closest to where it is used"

PR #1585, opened 2024-11-12
fix: reply storage is updated with the new content on file upload

### LINKED ISSUE #1522: images in drafts are deleted when navigating to a link in preview

### Describe the bug

If inserting an image when drafting a post, the image is deleted if you move to preview mode and then click a link to another page.  

### Screenshots

_No response_

### Steps To Reproduce

1. click "post", then choose discussion post.
2. Go to text box "write mode and insert a photo
3. enter the url to an external page in the body of the draft text.
4. Switch to "preview", then click the link to the external page
5. The code for the inserted image has been deleted when you return to the "write" menu.

### Expected behavior

I expected the image to remain in the draft while checking external links before posting.

### Logs

_No response_

### Device information

OS Ubuntu Linux
Browser Firefox

### Additional context

_No response_

aegroto:
## Description

Images and other file uploads update the draft reply, fixes #1522

## Screenshots

Unfortunately it's difficult to show the difference using screenshots, maybe I can try with a screen recording later?

## Additional Context

Nothing to report.

## Checklist

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:** 8. Throughly tested manually on dummy posts, no automatic tests.

**For frontend changes: Tested on mobile? Please answer below:** Yes, they work.

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -131,16 +131,12 @@ export function InputSkeleton ({ label, hint }) {
   )
 }
 
-function setNativeValue (element, value) {
-  const valueSetter = Object.getOwnPropertyDescriptor(element, 'value').set
-  const prototype = Object.getPrototypeOf(element)
-  const prototypeValueSetter = Object.getOwnPropertyDescriptor(prototype, 'value').set
-
-  if (valueSetter && valueSetter !== prototypeValueSetter) {
-    prototypeValueSetter.call(element, value)
-  } else {
-    valueSetter.call(element, value)
-  }
+// fix https://github.com/stackernews/stacker.news/issues/1522
+// see https://github.com/facebook/react/issues/11488#issuecomment-558874287
+function setNativeValue (textarea, value) {
+  const setter = Object.getOwnPropertyDescriptor(window.HTMLTextAreaElement.prototype, 'value')?.set
+  setter?.call(textarea, value)
+  textarea.dispatchEvent(new Event('input', { bubbles: true, value }))
 }
 
 export function MarkdownInput ({ label, topLevel, groupClassName, onChange, onKeyDown, innerRef, ...props }) {
@@ -383,7 +379,6 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, onKe
                 const s3Keys = [...text.matchAll(AWS_S3_URL_REGEXP)].map(m => Number(m[1]))
                 updateUploadFees({ variables: { s3Keys } })
                 setSubmitDisabled?.(false)
-                innerRef.current.dispatchEvent(new Event('change', { bubbles: true, value: text }))
               }}
               onError={({ name }) => {
                 let text = innerRef.current.value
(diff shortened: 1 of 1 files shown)


ekzyis:
Hey, thanks for the PR! Videos are nice, especially for UI/UX stuff but no need here anymore, I saw that this fixes it:

https://github.com/user-attachments/assets/d836f2f7-70e2-4a72-bbb7-30697f38a291

This is a nice way to fix it as simple as possible but ideally, I'd like to see a fix that [triggers](https://developer.mozilla.org/en-US/docs/Web/Events/Creating_and_triggering_events) a [change event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) so we can reuse the logic in this function:

https://github.com/stackernews/stacker.news/blob/d1c770dbbc832cb28ad5e301fc11ae0e2bc2f8b9/components/form.js#L489-L499

I tried the following when #1522 was created but for some reason, the callback doesn't run (I just tried again).

```diff
diff --git a/components/form.js b/components/form.js
index a67853f8..c38d6142 100644
--- a/components/form.js
+++ b/components/form.js
@@ -370,6 +370,7 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, onKe
                 const s3Keys = [...text.matchAll(AWS_S3_URL_REGEXP)].map(m => Number(m[1]))
                 updateUploadFees({ variables: { s3Keys } })
                 setSubmitDisabled?.(false)
+                innerRef.current.dispatchEvent(new Event('input', { bubbles: true, value: text }))
               }}
               onError={({ name }) => {
                 let text = innerRef.current.value
```

Can you check if you can find out how to get this to work or at least why it doesn't work? It would be a cleaner solution because it reuses code and makes changing the input via image upload consistent with any other change since there's more code in that callback than just `window.localStorage.setItem`

components/form.js
```diff
@@ -370,6 +373,7 @@ export function MarkdownInput ({ label, topLevel, groupClassName, onChange, onKe
                 const s3Keys = [...text.matchAll(AWS_S3_URL_REGEXP)].map(m => Number(m[1]))
                 updateUploadFees({ variables: { s3Keys } })
                 setSubmitDisabled?.(false)
+                window.localStorage.setItem(storageKey, text)
```
ekzyis, line 376: Since apparently `storageKey` can be undefined, you should also check for that case here (like is done in `onChangeInner`).

aegroto: Sorry, you are right, I totally missed it.

aegroto:
Hello, thanks for the detailed review! I acknowledge that a solution calling the an event would be better. Just wondering, may moving the whole logic the local storage from MarkdownInput to the parent component, in the onChange event, work for you?

aegroto:
I have converted the code to work with events by following this snippet: https://github.com/facebook/react/issues/10135#issuecomment-314441175

The issue may be related to the fact that change events are dropped if the text is not changed in practice, therefore the setNativeValue function uses some sort of reflection to bypass this mechanism.

I admit this looks kind of hacky to me and would prefer a React-based solution instead, but if you'd like to work with events this is a deal.

ekzyis (changes requested):
Wow, awesome that you found out why it wasn't working!

Added some suggestions how this could be even improved.

Requesting changes because the Formik handler should definitely be called imo.

aegroto:
Thanks for your feedback, I have integrated your changes and I confirm everything is still working as expected.

ekzyis (approved):

PR #1588, opened 2024-11-13
Fix first IDB operation never executed

ekzyis:
## Description

I noticed that if you only queue a single IDB operation like I want to do in #1559 (I only want to fetch the first page once without polling), it never resolves or rejects. The wrapped operation is never executed because `db` passed to `processQueue` in `queueOperation` is not defined yet.

I fixed this using a ref but I probably also need to add a `setInterval` to wait until `idb.current` is defined before running `processQueue` else there might be a race condition. And imo the name of `queueOperation` implies that the operation is queued to wait until IDB is ready.

- [x] ~TODO: do this ^~ _there actually is no race condition, using a ref completes the idea behind `processQueue` afaict_

The bug can be reproduced on master (d1c770db) using the following patch and noticing that `running queued page fetch ...` is never logged to the console when you visit /settings/wallets:

```diff
diff --git a/components/use-indexeddb.js b/components/use-indexeddb.js
index eac82872..0b538763 100644
--- a/components/use-indexeddb.js
+++ b/components/use-indexeddb.js
@@ -249,7 +249,9 @@ function useIndexedDB ({ dbName, storeName, options = DEFAULT_OPTIONS, indices =
   }, [queueOperation, storeName])
 
   const getPage = useCallback((page = 1, pageSize = 10, indexName = null, query = null, direction = 'next') => {
+    console.log('queue page fetch')
     return queueOperation((db) => {
+      console.log('running queued page fetch ...')
       return new Promise((resolve, reject) => {
         const transaction = db.transaction(storeName, 'readonly')
         const store = transaction.objectStore(storeName)
diff --git a/components/use-interval.js b/components/use-interval.js
index 04b528ce..df093d1d 100644
--- a/components/use-interval.js
+++ b/components/use-interval.js
@@ -2,7 +2,11 @@ import { useEffect } from 'react'
 
 function useInterval (cb, ms, deps) {
   return useEffect(() => {
-    const interval = setInterval(cb, ms)
+    const interval = setInterval(() => {
+      if (cb() === false) {
+        clearInterval(interval)
+      }
+    }, ms)
     return () => clearInterval(interval)
   }, deps)
 }
diff --git a/components/wallet-logger.js b/components/wallet-logger.js
index b4465a06..298f5546 100644
--- a/components/wallet-logger.js
+++ b/components/wallet-logger.js
@@ -195,7 +195,7 @@ function tag (walletDef) {
   return walletDef.shortName || walletDef.name
 }
 
-export function useWalletLogs (wallet, initialPage = 1, logsPerPage = 10) {
+export function useWalletLogs (wallet, { initialPage = 1, logsPerPage = 10, poll = true } = {}) {
   const [logs, _setLogs] = useState([])
   const [page, setPage] = useState(initialPage)
   const [hasMore, setHasMore] = useState(true)
@@ -312,7 +312,8 @@ export function useWalletLogs (wallet, initialPage = 1, logsPerPage = 10) {
 
   useInterval(() => {
     loadNew().catch(console.error)
-  }, 1_000, [loadNew])
+    return poll
+  }, 1_000, poll ? [loadNew] : [])
 
   return { logs, hasMore: !loading && hasMore, loadMore, setLogs, loading }
 }
diff --git a/wallets/index.js b/wallets/index.js
index 42a78247..2bf082d7 100644
--- a/wallets/index.js
+++ b/wallets/index.js
@@ -5,7 +5,7 @@ import { useApolloClient, useMutation, useQuery } from '@apollo/client'
 import { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react'
 import { getStorageKey, getWalletByType, Status, walletPrioritySort, canSend, isConfigured, upsertWalletVariables, siftConfig, saveWalletLocally } from './common'
 import useVault from '@/components/vault/use-vault'
-import { useWalletLogger } from '@/components/wallet-logger'
+import { useWalletLogger, useWalletLogs } from '@/components/wallet-logger'
 import { decode as bolt11Decode } from 'bolt11'
 import walletDefs from 'wallets/client'
 import { generateMutation } from './graphql'
@@ -68,6 +68,8 @@ export function WalletsProvider ({ children }) {
   const [serverWallets, setServerWallets] = useState([])
   const client = useApolloClient()
 
+  useWalletLogs(null, { poll: false })
+
   const { data, refetch } = useQuery(WALLETS,
     SSR ? {} : { nextFetchPolicy: 'cache-and-network' })
```

## Additional Context

I wonder how this worked before or how we didn't notice it before I added polling? Did I break something in #1516?

## Checklist

yes

`6`. This fixes the issue I've seen.

n/a

no

### DIFF
--- a/components/use-indexeddb.js
+++ b/components/use-indexeddb.js
@@ -9,7 +9,7 @@ const DEFAULT_INDICES = []
 const DEFAULT_VERSION = 1
 
 function useIndexedDB ({ dbName, storeName, options = DEFAULT_OPTIONS, indices = DEFAULT_INDICES, version = DEFAULT_VERSION }) {
-  const [db, setDb] = useState(null)
+  const db = useRef(null)
   const [error, setError] = useState(null)
   const [notSupported, setNotSupported] = useState(false)
   const operationQueue = useRef([])
@@ -19,15 +19,15 @@ function useIndexedDB ({ dbName, storeName, options = DEFAULT_OPTIONS, indices =
     setError(error)
   }, [])
 
-  const processQueue = useCallback((db) => {
-    if (!db) return
+  const processQueue = useCallback(() => {
+    if (!db.current) return
 
     try {
       // try to run a noop to see if the db is ready
-      db.transaction(storeName)
+      db.current.transaction(storeName)
       while (operationQueue.current.length > 0) {
         const operation = operationQueue.current.shift()
-        operation(db)
+        operation(db.current)
       }
     } catch (error) {
       handleError(error)
@@ -55,11 +55,11 @@ function useIndexedDB ({ dbName, storeName, options = DEFAULT_OPTIONS, indices =
           const database = event.target.result
           database.onversionchange = () => {
             database.close()
-            setDb(null)
+            db.current = null
             handleError(new Error('Database is outdated, please reload the page'))
           }
-          setDb(database)
-          processQueue(database)
+          db.current = database
+          processQueue()
         }
       }
 
@@ -81,9 +81,7 @@ function useIndexedDB ({ dbName, storeName, options = DEFAULT_OPTIONS, indices =
 
     return () => {
       isMounted = false
-      if (db) {
-        db.close()
-      }
+      db.current?.close()
     }
   }, [dbName, storeName, version, indices, options, handleError, processQueue])
 
@@ -106,7 +104,7 @@ function useIndexedDB ({ dbName, storeName, options = DEFAULT_OPTIONS, indices =
       }
 
       operationQueue.current.push(wrappedOperation)
-      processQueue(db)
+      processQueue()
     })
   }, [processQueue, db, notSupported, error])
 


ekzyis:
> The wrapped operation is never executed because db passed to `processQueue` in `queueOperation` is not defined yet.

To explain a little further why this doesn't work as expected since the PR description doesn't really explain the underlying reason, only the symptoms:

I see that the idea behind `queueOperation` was that either `db` is already defined which means `processQueue` will immediately dequeue and run the operation or the operation will stay queued and will run when we call `processQueue(db)` again when db is set.

However, this doesn't work as expected since we might have already set the db and called `processQueue` with no operation queued yet. When we then attempt to fetch the first page, `db` is still undefined in the scope of the function so it will not be able to run the operation until next render but then it's already too late if we never run `processQueue` again.

This also means there actually isn't a race condition here, using a ref fixes this exact problem of stale state in scope.

ekzyis:
Oh, just realized in #1559 that this might have broken wallet logs for some reason since I get this error now in the console when I try to zap with LNbits:

```
[lnbits] ↗ sending payment: 1,000 sats
IndexedDB error: DOMException: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.
    at eval (webpack-internal:///./components/use-indexeddb.js:33:24)
    at eval (webpack-internal:///./components/use-indexeddb.js:116:13)
    at new Promise (<anonymous>)
    at eval (webpack-internal:///./components/use-indexeddb.js:106:16)
    at eval (webpack-internal:///./components/use-indexeddb.js:125:16)
    at eval (webpack-internal:///./components/wallet-logger.js:350:23)
    at eval (webpack-internal:///./components/wallet-logger.js:421:13)
    at Object.ok (webpack-internal:///./components/wallet-logger.js:428:46)
    at Object.eval [as sendPayment] (webpack-internal:///./wallets/index.js:365:20)
```

ekzyis:
I have decided to simply also poll logs for wallet status. Then the wallet status also automatically updates on errors even if you don't refresh the page.

PR #1589, opened 2024-11-13
feat: recent unpaid bounties selection

### LINKED ISSUE #1586: Filter out paid bounties

The recent/bounties endpoint does not let remove paid bounties. It could be beneficial to list only bounties that are still open.

Add an additional filtering option to show only unpaid bounties.

_No response_

### Additional context

_No response_

aegroto:
## Description

Added an option to show only recent unpaid bounties. Fixes #1586

## Screenshots

The current *bounties* selection shows both paid and unpaid bounties:

immagine

The introduced *unpaid bounties* option filters out paid bounties and show only available ones:

immagine

## Additional Context

I have some concerns, mainly due to to my lack of experience in the project:

1. This is the first selection option that includes a space inside. May it be a source of problems in the future?
2. The additional selection has been implemented as a separate option in the dropdown menu. Is there a way to add it as an additional option, such as a checkbox, while showing the "bounties" results? That would be interesting but as of now the query is merely done in terms of the route.
3. The option has been added to the ITEM_TYPES constant. Should it be added to the ITEM_TYPES_USER as well? Where is this other constant used?

## Checklist

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:** 8, I have tested it manually but were unable to add automated tests.

**For frontend changes: Tested on mobile? Please answer below:** Yes, everything works.


ekzyis:
> The additional selection has been implemented as a separate option in the dropdown menu. Is there a way to add it as an additional option, such as a checkbox, while showing the "bounties" results? That would be interesting but as of now the query is merely done in terms of the route.

Yes, this should be possible and I would prefer this a lot more than having another dedicated option:

2024-11-13-231346_475x138_scrot

We don't use dynamic checkboxes anywhere but we have dynamic filters depending on what you selected in [top](https://stacker.news/top/posts/forever). Maybe the code for them is inspiring.

aegroto:
I have modified the widget to use a checkbox instead of a separate select option:

immagine

The dropdown is still redirected to "bounties", I believe the UX is fluid enough. Let me know what you think of this reworked version :)

ekzyis:
Thank you, I will take a look tomorrow

ekzyis:
Sorry, forgot about this yesterday. Will review now

components/recent-header.js
```diff
@@ -1,10 +1,36 @@
...
+
+function ActiveBountiesCheckbox ({ router, prefix }) {
+  const onChange = (e) => {
+    const route = e.target.checked ? 'bounties_active' : 'bounties'
+
+    router.push(prefix + `/recent/${route}`)
+  }
+
+  return (
+    <div className='mx-2 my-1'>
```
ekzyis, line 22: Checkbox isn't properly aligned with the dropdown:

2024-11-17-232405_1920x1080_scrot
2024-11-17-232401_1920x1080_scrot

This fixes it:

```suggestion
    <div className='mx-2 mb-2'>
```

components/recent-header.js
```diff
@@ -1,10 +1,36 @@
...
+  }
+
+  return type
+}
+
+function ActiveBountiesCheckbox ({ router, prefix }) {
+  const onChange = (e) => {
+    const route = e.target.checked ? 'bounties_active' : 'bounties'
+
+    router.push(prefix + `/recent/${route}`)
```
ekzyis, line 18: I would prefer to use a query parameter for the checkbox instead of a dedicated route to make it look cleaner:

```suggestion
    const u = new URL(prefix + '/recent/bounties', window.location.origin)
    if (e.target.checked) u.searchParams.set('active', true)
    router.push(u)
```

However, we then also need this patch to get the information to the backend without having to update the mutation just for this case:

```diff
diff --git a/pages/~/recent/[type].js b/pages/~/recent/[type].js
index 7e09afb6..1a21aa4b 100644
--- a/pages/~/recent/[type].js
+++ b/pages/~/recent/[type].js
@@ -10,7 +10,12 @@ import PageLoading from '@/components/page-loading'

 const staticVariables = { sort: 'recent' }
 const variablesFunc = vars =>
-  ({ includeComments: COMMENT_TYPE_QUERY.includes(vars.type), ...staticVariables, ...vars })
+  ({
+    includeComments: COMMENT_TYPE_QUERY.includes(vars.type),
+    ...staticVariables,
+    ...vars,
+    type: vars?.type === 'bounties' && vars?.active ? 'bounties_active' : (vars?.type || '')
+  })

 export const getServerSideProps = getGetServerSideProps({
   query: SUB_ITEMS,
```

components/recent-header.js
```diff
@@ -14,18 +40,23 @@ export default function RecentHeader ({ type, sub }) {
     : ITEM_TYPES
 
   type ||= router.query.type || type || 'posts'
+  type = mapTypeToSelectOption(type)
```
ekzyis, line 43: I believe this is no longer needed since the switch to the checkbox

components/recent-header.js
```diff
@@ -1,10 +1,36 @@
...
 import { Select } from './form'
 import { useRouter } from 'next/router'
 
+function mapTypeToSelectOption (type) {
+  switch (type) {
+    case 'bounties_active': return 'bounties'
+  }
+
+  return type
+}
```
ekzyis, line 12: no longer needed

ekzyis (changes requested):
Works great, thank you!

https://github.com/user-attachments/assets/99013b96-f9c9-4107-b8b1-35c811a832eb

However, I found some aesthetic issues in the frontend code, see suggestions.

aegroto:
Hello, I totally agree with your review. I should've used a query param since the beginning, that was the most straightforward solution, sorry for that. I will correct the code and update the PR as soon as I can.

aegroto:
Thanks for your suggestion, I have applied them to my branch and tested that everything keeps working. I have decided to refactor variablesFunc adding explicit conditionals to avoid the nested ternary operators, but let me know if you prefer to keep with the previous version.

ekzyis (approved):
> I have decided to refactor variablesFunc adding explicit conditionals to avoid the nested ternary operators, but let me know if you prefer to keep with the previous version.

Thank you, that was a good decision!

Looks good to me, I also tested again.

aegroto:
@ekzyis Thanks for your feedback. I am curious, why would prefer to use `replace` instead of `push`? Anyway, once you take the decision I will update the PR accordingly.

ekzyis (approved):

aegroto:
Hello, is there anything missing here and in my other pull requests? I am willing to correct them if necessary once I've got your feedback.

huumn:
@aegroto we've just been swamped with higher priority things right now. I'm sorry for the delay.

I will get to your PRs before the end of next week at the latest. Feel free to pickup more issues in the meantime.

Issue #1549, opened 2024-11-06
Support multiple NWC relays 


ekzyis:
### Describe the problem you're trying to solve

Wallets like [cashu.me](https://cashu.me) include multiple relays in their NWC connection string. We currently only use the first relay.

### Describe the solution you'd like

Parse all relays from NWC connection in `parseNwcUrl` using `url.searchParams.getAll('relay')` instead of `url.searchParams.get('relay')` and use all of them to publish requests and listen for responses.

https://github.com/stackernews/stacker.news/blob/4cc3780eca1f1390bdd331f0e418b9a7efd4347e/lib/url.js#L190-L214

### Additional context

Logging would become very noisy if we log all events for every relay individually like we do now for the single relay. We should consider to remove all logs inside the `nwcCall` function if they aren't needed. NWC is the only wallet that uses the `logger` inside its own functions. So maybe it's indeed not needed and we can simply rely on general wallet logs?

riccardobl:
this is the default behavior in ndk

Issue #1558, opened 2024-11-08
Timeout wallet payments


huumn:
Wallet payments can fail silently, e.g. the NWC consumer doesn't respond to a payment request, in which case we will wait until the invoice expires (10 minutes in most cases) - which is way too long. Stackers will assume we're unresponsive.

It's not as simple as timing out and displaying a QR code for the same invoice however. It's possible the QR code is paid and then the NWC consumer comes online and attempts to pay the same invoice. This is dangerous because once the QR code is paid, the preimage for the payment has been revealed, and the sats in the second payment can be stolen by nodes along the route that know the preimage.

To timeout safely, we'll want to cancel the original invoice, then generate a new invoice to display in the QR.

riccardobl:
this is solved by https://github.com/stackernews/stacker.news/pull/1523

ekzyis:
Can you extract the commit from #1523 in a separate PR? #1523 already has a lot of changes and we can merge a smaller PR faster.

riccardobl:
> Can you extract the commit from #1523 for this into a separate PR if it doesn't mess/isn't related too much with the other code in it? #1523 already has a lot of changes and we can merge and review a smaller PR faster.

I'll see what i can do, but the fix is a side effect of the different retrying mechanism. Anyway i am cleaning up the pr, so maybe we'll end up with fewer changes

riccardobl:
correction: my PR fixes only this part
> This is dangerous because once the QR code is paid, the preimage for the payment has been revealed, and the sats in the second payment can be stolen by nodes along the route that know the preimage.

the use-paid-mutation calls retryPaidAction that, with the changes in the PR, cancels the previous invoice and returns a new one for the QR code.

However it doesn't change the long timeout

ekzyis:
> Wallet payments can fail silently, e.g. the NWC consumer doesn't respond to a payment request, in which case we will wait until the invoice expires (10 minutes in most cases) - which is way too long. Stackers will assume we're unresponsive.

Since this ticket was created because of an unresponsive territory payment, can't we control the timeout simply via low invoice expiration for invoices that pay SN?

We can't do this for p2p payments because we don't control the outgoing invoice and expiry of incoming invoice must be higher and have a buffer. But stackers don't wait for these payments since they are handled in the background so I think failing fast is less important.

huumn:
We can. This issue is mostly about falling back to QR after a reasonable amount of time, so I'm not sure how that fits into what you're saying.

[...]

PR #1590, opened 2024-11-13
NDK

riccardobl:
## Description

Migrate nostr code to NDK.
I've kept it OOP like the original code.

- [x] nwc sender
- [x] nwc receiver
- [x] crosspost
- [x] zap note

closes https://github.com/stackernews/stacker.news/issues/1549

## Additional Context

## Checklist

yes

zap notes are not tested by they should work™

n/a

no


lib/nostr.js
```diff
@@ -17,155 +16,151 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
+      console.error('Crosspost error:', error)
+      return { error }
     }
   }
 }
 
+/**
+ * @type {Nostr}
+ */
+export default new Nostr()
```
riccardobl, line 162: this is the global singleton, connections to specific relays are handled with RelaySets and their life cycle is left to ndk to handle

riccardobl:
Logging for nwc is a bit tricky because NDK abstracts away everything and uses  [npm:debug](https://www.npmjs.com/package/debug) internally. I've got it to work by adding a custom `debug` logger that proxies to the wallet logger, but i am not sure if the verbosity is too high.

ekzyis (changes requested):
You mentioned you tested NWC send but when I try to attach, it fails with `pubkey not set`. This means that `pubkey` [here](https://github.com/nostr-dev-kit/ndk/blob/6b3ea8b298680bee654ba87107da09bc5a63cb54/ndk/src/nwc/index.ts#L94-L110) is not getting set:

```js
static async fromURI(ndk: NDK, uri: string): Promise<NDKNwc> {
    const u = new URL(uri);

    // validate protocol nostr+walletconnect
    if (u.protocol !== "nostr+walletconnect:") {
        throw new Error("Invalid protocol");
    }

    const nwc = new NDKNwc({
        ndk,
        pubkey: u.host ?? u.pathname,
        relayUrls: u.searchParams.getAll("relay") ?? [""],
        secret: u.searchParams.get("secret") ?? "",
    });

    return nwc;
}
```

It's related to different URL parsing on server vs client since NWC receive works:

server using `node`:

```
$ node
Welcome to Node.js v20.18.0.
> new URL("nostr+walletconnect://eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848?relay=wss%3A%2F%2Frelay.primal.net&secret=78dd6a6c93f02c0831ff1e74c8ae9a388144f67c10da8276f1b4e475ad9f980f")
URL {
  href: 'nostr+walletconnect://eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848?relay=wss%3A%2F%2Frelay.primal.net&secret=78dd6a6c93f02c0831ff1e74c8ae9a388144f67c10da8276f1b4e475ad9f980f',
  origin: 'null',
  protocol: 'nostr+walletconnect:',
  username: '',
  password: '',
  host: 'eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848',
  hostname: 'eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848',
  port: '',
  pathname: '',
  search: '?relay=wss%3A%2F%2Frelay.primal.net&secret=78dd6a6c93f02c0831ff1e74c8ae9a388144f67c10da8276f1b4e475ad9f980f',
  searchParams: URLSearchParams {
    'relay' => 'wss://relay.primal.net',
    'secret' => '78dd6a6c93f02c0831ff1e74c8ae9a388144f67c10da8276f1b4e475ad9f980f' },
  hash: ''
}
```

browser console:

2024-11-29-031600_1125x269_scrot

See how `u.host` is only set to the pubkey in `node`. In the browser, it's set to the empty string and since NDK uses `??`, it will pick the empty string.

But even if NDK would use `||`, the pubkey will be wrong because of the `//`. We got around that different URL parsing behavior by copying [how Alby does it](https://github.com/getAlby/js-sdk/blob/ee07a4c8d9c294eacf4198b48b3e7ed61fb9c5eb/src/NWCClient.ts#L224-L251). They replace the protocol with `http://` in which case it works in both environments:

server using `node`:

```
$ node
Welcome to Node.js v20.18.0.
> new URL("http://eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848?relay=wss%3A%2F%2Frelay.primal.net&secret=78dd6a6c93f02c0831ff1e74c8ae9a388144f67c10da8276f1b4e475ad9f980f")
URL {
  href: 'http://eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848/?relay=wss%3A%2F%2Frelay.primal.net&secret=78dd6a6c93f02c0831ff1e74c8ae9a388144f67c10da8276f1b4e475ad9f980f',
  origin: 'http://eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848',
  hostname: 'eed35e49c2fdb42819210d56ba4efdd1b3eaee92b271dbb6bfc28bd049526848',
  port: '',
  pathname: '/',
  search: '?relay=wss%3A%2F%2Frelay.primal.net&secret=78dd6a6c93f02c0831ff1e74c8ae9a388144f67c10da8276f1b4e475ad9f980f',
  searchParams: URLSearchParams {
    'relay' => 'wss://relay.primal.net',
    'secret' => '78dd6a6c93f02c0831ff1e74c8ae9a388144f67c10da8276f1b4e475ad9f980f' },
  hash: ''
}
```

browser console:

2024-11-29-031953_1111x277_scrot

See how `u.host` is now set to the pubkey in both cases.

However, that workaround doesn't work with NDK since it checks for the protocol.

And if I use `nostr+walletconnect:` instead of `nostr+walletconnect://`, I run into the same issue that NDK uses `??` instead of `||`.

So I am not sure how anyone is able to use NDK's NWC implementation in v2.10.5 of `@nostr-dev-kit/ndk` in the browser currently. This happens in Brave and Firefox.

The file where it uses `pubkey: u.host ?? u.pathname` also hasn't changed in 5 months.

How were you able to attach NWC send?

riccardobl:
Very strange... 
It works on firefoxdev and chrome  🤔

https://github.com/user-attachments/assets/64163cd3-3b93-49d5-806b-9d4fa6c256f1


and brave 173.89

but not on the regular firefox(???)
image

riccardobl:
But i remember this being an issue in the past., maybe it is an update they are rolling out or a security rule some setups don't have. I just saw it working and trusted the library to do the right thing.
Luckily NDK exposes the nwc class, so , to avoid surprises, i've reapplied the alby workaround modified for multiple relays.

ekzyis (commented):
Attaching NWC send+recv works now. I also tested zapping with NWC and since attaching recv works, the server was able to create the test invoice.

However, the worker doesn't seem to be able to connect to the relay. It blocks at

```js
await nwc.blockUntilReady(timeout)
```

until it times out. Increasing the timeout did not help, even if I increase it to 30 seconds in wallets/server.js:175.



 Can you check if that also happens on your side?

riccardobl:
I've enabled debug and there is this exception being thrown by the worker
```
2024-12-01T13:39:19.448Z ndk:pool:main Failed to connect to relay wss://relay.primal.net/ ReferenceError: WebSocket is not defined
```

turns out ndk expects WebSocket to be available in nodejs, the doc says to use a polyfill, but we don't have it on the worker.
So i've added this to worker/index.js
```js
import ws from 'isomorphic-ws'
if (typeof WebSocket === 'undefined') {
  global.WebSocket = ws
}
```

lib/nostr.js
```diff
@@ -17,155 +15,148 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
-      throw err
-    } finally {
-      clearInterval(interval)
-    }
+  /**
+   *
+   * @param {args} param0
+   * @param {string} [args.privKey] - private key to use for signing
+   * @param {boolean} [args.supportNip07] - whether to use NIP-07 signer if available
+   * @returns {NDKPrivateKeySigner | NDKNip46Signer | NDKNip07Signer | null} - a signer instance
```
ekzyis, line 60: nitpick: `| NDKNip46Signer` should have been part of #1636 and not included here already

but it's okay now

lib/nostr.js
```diff
@@ -17,155 +15,148 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
-      return await withTimeout(checkPromise, timeout)
-    } catch (err) {
-      this.close()
-      throw err
-    } finally {
-      clearInterval(interval)
-    }
+  /**
+   *
+   * @param {args} param0
```
ekzyis, line 57:
```suggestion
   * @param {Object} args
```

lib/nostr.js
```diff
@@ -17,155 +15,148 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
+   * @param {number} rawEvent.kind
+   * @param {number} rawEvent.created_at
+   * @param {string} rawEvent.content
+   * @param {Array<Array<string>>} rawEvent.tags
+   * @param {Object} context
+   * @param {string} context.privKey
+   * @param {NDKSigner} context.signer
+   * @returns {Promise<NDKEvent>}
+   */
+  /* eslint-disable camelcase */
```
ekzyis, line 79: This disables the rule for the remainder of the file. I think that's okay, so you can just put this line once somewhere at the top of the file.

lib/nostr.js
```diff
@@ -17,155 +15,148 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
+   * @param {string} rawEvent.content
+   * @param {Array<Array<string>>} rawEvent.tags
+   * @param {Object} context
+   * @param {Array<string>} context.relays
+   * @param {string} context.privKey
+   * @param {NDKSigner} context.signer
+   * @param {number} context.timeout
+   * @returns {Promise<NDKEvent>}
+   */
+  /* eslint-disable camelcase */
```
ekzyis, line 106: you only need  this once, see other comment

lib/nostr.js
```diff
@@ -17,155 +15,148 @@ export const DEFAULT_CROSSPOSTING_RELAYS = [
...
+   * @param {NDKSigner} context.signer
+   * @returns {Promise<NDKEvent>}
+   */
+  /* eslint-disable camelcase */
+  async sign ({ kind, created_at, content, tags }, { signer } = {}) {
+    const event = new NDKEvent(this.ndk)
+    event.kind = kind
+    event.created_at = created_at
+    event.content = content
+    event.tags = tags
```
ekzyis, line 85: Why not pass event members as second argument to the `NDKEvent` constructor?

ekzyis (changes requested):
Works well now! I tested paying with nwc, receiving autowithdrawals with nwc, errors and crossposting.

Only found [one issue with error handling](https://github.com/stackernews/stacker.news/pull/1590#discussion_r1867572607), the rest of the comments are only suggestions.

wallets/nwc/client.js
```diff
@@ -11,11 +11,8 @@ export async function testSendPayment ({ nwcUrl }, { logger }) {
...
 export async function sendPayment (bolt11, { nwcUrl }, { logger }) {
-  const result = await nwcCall({
-    nwcUrl,
-    method: 'pay_invoice',
-    params: { invoice: bolt11 }
-  },
-  { logger })
+  const nwc = await getNwc(nwcUrl)
+  const { error, result } = await nwc.payInvoice(bolt11)
+  if (error) throw new Error(error.code + ' ' + error.message)
```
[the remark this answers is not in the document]
riccardobl: what's the point of catching the exception here? If it throw inside payInvoice it will be just propagated back because we are awaiting it.
The  `if (error) throw new Error(error.code + ' ' + error.message)` makes sure that if ndk decides to start following their own type specs,  we won't leave any error unthrown.

package-lock.json
```diff
@@ -20320,6 +20543,47 @@
...
+    },
+    "node_modules/websocket-polyfill": {
+      "version": "0.0.3",
+      "resolved": "https://registry.npmjs.org/websocket-polyfill/-/websocket-polyfill-0.0.3.tgz",
+      "integrity": "sha512-pF3kR8Uaoau78MpUmFfzbIRxXj9PeQrCuPepGE6JIsfsJ/o/iXr07Q2iQNzKSSblQJ0FiGWlS64N4pVSm+O3Dg==",
+      "dependencies": {
+        "tstl": "^2.0.7",
+        "websocket": "^1.0.28"
+      }
+    },
```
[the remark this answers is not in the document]
riccardobl: must be something specific to our deployment, i've added the undefined check to be sure that if some other websocket polyfill starts working in the future, we don't override it.

wallets/nwc/client.js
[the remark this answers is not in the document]
ekzyis: > what's the point of catching the exception here? If it throws inside payInvoice it will be just propagated back because we are awaiting it.

It throws an object instead of a proper error message so we log `payment failed: [Object object]` currently

> The if (error) throw new Error(error.code + ' ' + error.message) makes sure that if ndk decides to start following their own type specs, we won't leave any error unthrown.

okay, we can keep this then but we still need `try/catch` currently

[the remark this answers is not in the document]
riccardobl: Ah! Ok, that's a pretty strange way to pass errors... i'm adding the recatch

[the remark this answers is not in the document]
riccardobl: done

wallets/nwc/index.js
```diff
@@ -33,61 +33,38 @@ export const card = {
...
-    throw new Error(`invalid ${method} response: missing error or result`)
-  } finally {
-    relay?.close()
-    logger?.info(`closed connection to ${relayUrl}`)
+    const { error, result } = await fun()
+    if (error) throw new Error(error.code + ' ' + error.message)
+    return result
+  } catch (e) {
+    if (e.error) throw new Error(e.error.code + ' ' + e.error.message)
+    throw e
```
ekzyis, line 62: I think we should only log the human readable message and use the code as fallback if it's not set for some reason:

```suggestion
    const { error, result } = await fun()
    if (error) throw new Error(error.message || error.code)
    return result
  } catch (e) {
    if (e.error) throw new Error(e.error.message || e.error.code)
    throw e
```

ekzyis (commented):
Just one last suggestion

ekzyis:
In case someone is wondering: I am keeping this branch up-to-date with master because I am building my code for #1558 on this.

Issue #1569, opened 2024-11-10
Uploaded videos don't load on Safari iOS


ekzyis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

See title

### Screenshots

Safari PWA on iOS:

![signal-2024-11-10-162050](https://github.com/user-attachments/assets/a1ab6793-d32b-47a4-b42c-64001241e537)

Brave on Desktop:

![2024-11-10-162006_1920x1080_scrot](https://github.com/user-attachments/assets/711816cf-a867-474c-8cf5-7d646b51c084)

### Steps To Reproduce

1. Upload video from other device because of #1568
2. View item on Safari iOS

(not confirmed)

### Expected behavior

Video is visible on Safari iOS

huumn:
It's likely a CSS issue. Safari is uniquely glitchy when it comes to video.

Soxasora:
> It's likely a CSS issue. Safari is uniquely glitchy when it comes to video.

I'm exploring also this issue, **you're right it is indeed a CSS issue**, specifically `text.module.css` at line 187
```
.p:has(> .mediaContainer) .mediaContainer
{
    display: flex;
    width: min-content; /* Safari doesn't like this */
    max-width: 100%;
}
```

This actually works on Safari for Mac
![Safari loading the video](https://github.com/user-attachments/assets/b325f53f-bb7e-4075-9dbb-c3e45e108384)


But on Safari for iOS (or PWA) the story is much different, it tries to render all the possible resolutions of the image that actually is a video, in fact `media-or-link.js` tries to load the video as img before creating a video element that only then will be populated with the video data.
I'm still investigating, for now I'm battling hydration errors as testing PWAs is its own hell

PR #1593, opened 2024-11-15
fix uploaded videos don't load on safari

Soxasora:
## Description

By removing an incompatible-between-renders `width: min-content`, uploaded videos now show and play on Safari for iOS and macOS, fixing #1569.

## Screenshots
Safari on macOS:
Safari on macOS

Safari on iOS/PWA:
Safari on iOS

## Additional Context
Removing `width: min-content` didn't cause any structural change and actually didn't do anything (?) other than satisfy Safari.

In the issue I talked about `media-or-link.js`
> rendering all possible resolutions of an image that actually is a video

I can confirm that this wasn't the problem and it was just that CSS line

## Checklist

Yes, this edit doesn't break functionality nor UX/UI

7, I tested on various browsers for various platforms, this edit didn't change whatsoever the past behavior on other browsers.

Tested mainly on iOS, videos now load effortlessly


ekzyis:
I can confirm videos now load on Safari iOS _sometimes_ but more often than not, they are detected as images that only take up space:

loads correctly as video:

signal-2024-11-15-213948

loads as image:

signal-2024-11-15-213951

For some reason, it initially seemed to work more as video (I think once there were no controls) than not but now it always loads as an image :thinking: 

Does this also happen on your end?

ekzyis (commented):
forgot to publish the above as a review

Soxasora:
> Does this also happen on your end?

Yes, that's why I held back this PR for so long, if you check the console you can see that it throws 422 (Unprocessable Entity) from imgproxy and Safari just stops there lol

ekzyis:
> > Does this also happen on your end?
> 
> Yes, that's why I held back this PR for so long, if you check the console you can see that it throws 422 (Unprocessable Entity) from imgproxy and Safari just stops there lol
> 
> edit: I think that in prod this shouldn't (?) happen

I updated my comment and added a patch for this, I think this is unrelated

Soxasora:
> I updated my comment and added a patch for this, I think this is unrelated. 

Tried it just now and it works which indicates that this may be inconsistent, could be the fact that it tries to load the video as image before replacing it with video?

ekzyis:
> Tried it just now and it works which indicates that this may be inconsistent, could be the fact that it tries to load the video as image before replacing it with video?

Yeah something like this might be the case but it's hard to tell without a browser console on mobile. USB debugging might be required for this issue.

I wouldn't be surprised if you need a Mac for this though ...

Soxasora:
> I wouldn't be surprised if you need a Mac for this though ...

I am using a Mac indeed because I didn't know better but you sparked my curiosity and there are tools like Inspect that should do this.

Adding to inconsistencies, it seems like that iOS stops at the image stage and macOS can reach the video stage (of the media-or-link pipeline)

huumn:
Cool! IIRC I put this there for image/video gridding. So I'll just need to verify this doesn't break that.

Soxasora:
> So I'll just need to verify this doesn't break that.

It does, sorry didn't think of uploading more videos

This is the behavior with min-content
with min-content

And this is the behavior without min-content
without min-content

I found out that with or without min-content, videos will be shown after applying @ekzyis patch. I suspect that Safari stops after rendering it as image, disabling min-content gave us the video just because it stopped Safari from calculating it on the image

With min-content and patch:
photo_2024-11-16_09-49-25

Soxasora:
I can confirm that Safari queues the video tag after the image has finished playing (???lol)

As a proof-of-concept I flipped the image-or-video logic and added a dynamic min-content.
Safari/PWA will load the video reliably and also respect min-content for content-sizing:
photo_2024-11-16_11-56-11

huumn:
For some reason, videos don't load for me in desktop safari now. Left if brave. Right is desktop safari.

Screenshot 2024-11-16 at 5 40 31 PM

huumn:
Nice work! Do you have a lightning address I can send sats to?

Soxasora:
Thanks! Here's the address: soxasora@blink.sv

PR #1595, opened 2024-11-15
Fix unpaid items are counted

ekzyis:
Fix this:

https://github.com/user-attachments/assets/2c6ecc00-5b8e-4401-841c-fab7b57be939

Tested with anon and logged in. Also tested with items that were paid with fee credits (`invoiceActionState IS NULL`).

### DIFF
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -945,7 +945,7 @@ export default {
         }
       })
     },
-    ncomments: async (user, { when, from, to }, { models }) => {
+    ncomments: async (user, { when, from, to }, { me, models }) => {
       if (typeof user.ncomments !== 'undefined') {
         return user.ncomments
       }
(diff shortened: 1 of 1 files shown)


ekzyis:
I just searched for `models.item.count` and also fixed it for `ncomments`. I also noticed that `auctionPosition` also calls `models.item.count` but it already uses `item.status` which is enough afaict.

api/resolvers/user.js
```diff
@@ -922,11 +922,12 @@ export default {
           createdAt: {
             gte,
             lte
-          }
+          },
+          ...(me?.id === user.id ? {} : { OR: [{ invoiceActionState: 'PAID' }, { invoiceActionState: null }] })
```
huumn, line 926: This can all be an `OR` I think:

e.g.

```
OR: [{userId: me.id}, { invoiceActionState: 'PAID' }, { invoiceActionState: null }]
```

Alternatively, we don't need to count unpaid items at all and it can just be:

```
OR: [{ invoiceActionState: 'PAID' }, { invoiceActionState: null }]
```

ekzyis: Decided to never count unpaid items

Issue #1258, opened 2024-07-08
Pull down to refresh on iOS


ekzyis:
**Is your feature request related to a problem? Please describe.**

This is a feature request that was requested [a long time ago](https://stacker.news/items/407120).

Pull down to refresh works on Android but not on iOS.

**Describe the solution you'd like**

Page refreshes on pull down.

**Describe alternatives you've considered**

n/a

**Additional context**

n/a

Soxasora:
https://github.com/user-attachments/assets/33c51bee-35ec-4b0a-85b0-1d35e529c837

refreshing 🍾
i'm going to make it more graceful, any tips?

note: Android has native support for pull to refresh, this either has to be enabled just for iOS or site-wide disabling Android's one

huumn:
You probably want this: https://www.apollographql.com/docs/react/data/refetching#clientrefetchqueries

PR #1599, opened 2024-11-17
feat: PWA pull to refresh

Soxasora:
## Description

Pull-to-refresh for iOS, completing #1258.

As `router.reload()` is not graceful and `router.replace(...)` adds "/~" to the path breaking functionality across the app, I've decided to use `router.push()` instead, reloading the same page.
This updates the browser's history but since we're talking about a PWA there's no history to be accessed from the user.

This component wraps `<Layout />` and is present in every page that make use of it.

## Screenshots

https://github.com/user-attachments/assets/46209199-ae9d-433b-8c5f-92ae9f073c6c

## Additional Context

At the moment it's enabled also for Android, disabling its native pull-to-refresh. If we want to enable it back we can pass `android=false`

Not being a master of creativity and CSS I want to know if you'd like to see something else instead of _refreshing..._, even though I think it's geeky in its own way

## Checklist

checkPWA prevents it from being triggered on anything that's not a PWA, also being a new component that has minimal CSS and loads only if the user is at the top of the page, I can say that it doesn't break functionality.

7, I don't have an Android phone and I would like to know if there are any problems with it, as said before we can disable or enable Android's PTR.
As it should work only in a PWA context, I also tested it on browsers

Yes, being a mobile-only feature

### DIFF
--- a/components/pull-to-refresh.js
+++ b/components/pull-to-refresh.js
@@ -2,26 +2,34 @@ import { useRouter } from 'next/router'
 import { useState, useRef, useEffect, useCallback } from 'react'
 import styles from './pull-to-refresh.module.css'
 
-export default function PullToRefresh ({ children }) {
+const PULL_THRESHOLD = 300
+const REFRESH_TIMEOUT = 500
+
+export default function PullToRefresh ({ children, android }) {
   const router = useRouter()
   const [isRefreshing, setIsRefreshing] = useState(false)
   const [pullDistance, setPullDistance] = useState(0)
   const [isPWA, setIsPWA] = useState(false)
+  const [isAndroid, setIsAndroid] = useState(false)
   const touchStartY = useRef(0)
   const touchEndY = useRef(0)
 
+  const checkPWA = () => {
+    const androidPWA = window.matchMedia('(display-mode: standalone)').matches
+    const iosPWA = window.navigator.standalone === true
+    setIsAndroid(androidPWA) // we need to know if the user is on Android to enable toggling its native PTR
+    setIsPWA(androidPWA || iosPWA)
+  }
+
   useEffect(() => {
-    const checkPWA = () => {
-      // android/general || ios
-      return window.matchMedia('(display-mode: standalone)').matches || window.navigator.standalone === true
-    }
-    setIsPWA(checkPWA())
+    checkPWA()
   }, [])
 
   const handleTouchStart = useCallback((e) => {
-    if (!isPWA || window.scrollY > 0) return // don't handle if the user is not scrolling from the top of the page
+    // don't handle if the user is not scrolling from the top of the page, is not on a PWA or if we want Android's native PTR
+    if (!isPWA || (isAndroid && !android) || window.scrollY > 0) return
     touchStartY.current = e.touches[0].clientY
-  }, [isPWA])
+  }, [isPWA, isAndroid, android])
 
   const handleTouchMove = useCallback((e) => {
     if (touchStartY.current === 0) return
@@ -31,12 +39,12 @@ export default function PullToRefresh ({ children }) {
 
   const handleTouchEnd = useCallback(() => {
     if (touchStartY.current === 0 || touchEndY.current === 0) return
-    if (touchEndY.current - touchStartY.current > 300) { // current threshold is 300, subject to change
+    if (touchEndY.current - touchStartY.current > PULL_THRESHOLD) {
       setIsRefreshing(true)
-      router.push(router.asPath)
+      router.push(router.asPath) // reload the same path
       setTimeout(() => {
         setIsRefreshing(false)
-      }, 500) // simulate loading time
+      }, REFRESH_TIMEOUT) // simulate loading time
     }
     setPullDistance(0) // using this to reset the message behavior
     touchStartY.current = 0 // avoid random refreshes by resetting touch
@@ -44,7 +52,8 @@ export default function PullToRefresh ({ children }) {
   }, [router])
 
   useEffect(() => {
-    if (!isPWA) return
+    // don't handle if the user is not on a PWA or if we want Android's native PTR
+    if (!isPWA || (isAndroid && !android)) return
     document.addEventListener('touchstart', handleTouchStart)
     document.addEventListener('touchmove', handleTouchMove)
     document.addEventListener('touchend', handleTouchEnd)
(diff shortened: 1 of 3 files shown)


components/pull-to-refresh.js
```diff
@@ -0,0 +1,95 @@
...
+  }, [])
+
+  const handleTouchStart = useCallback((e) => {
+    // don't handle if the user is not scrolling from the top of the page, is not on a PWA or if we want Android's native PTR
+    if (!isPWA || (isAndroid && !android) || window.scrollY > 0) return
+    touchStartY.current = e.touches[0].clientY
+  }, [isPWA, isAndroid, android])
+
+  const handleTouchMove = useCallback((e) => {
+    if (touchStartY.current === 0) return
```
ekzyis, line 35: Is this needed so the text doesn't get stuck in some cases? At least that's what I can tell starts happening when I remove this line. A comment in the code would be nice here.

components/pull-to-refresh.js
```diff
@@ -0,0 +1,95 @@
...
+    touchEndY.current = e.touches[0].clientY
+    setPullDistance(touchEndY.current - touchStartY.current)
+  }, [])
+
+  const handleTouchEnd = useCallback(() => {
+    if (touchStartY.current === 0 || touchEndY.current === 0) return
```
ekzyis, line 41: Same reason here I suppose

components/pull-to-refresh.js
```diff
@@ -0,0 +1,95 @@
+import { useRouter } from 'next/router'
+import { useState, useRef, useEffect, useCallback } from 'react'
+import styles from './pull-to-refresh.module.css'
+
+const PULL_THRESHOLD = 300
```
ekzyis, line 5: 100 as twice the threshold when `pull to refresh` starts showing might be a better threshold.

I would also call this `REFRESH_THRESHOLD` since `PULL_THRESHOLD` sounds like that's when we show the first message.

components/pull-to-refresh.js
```diff
@@ -0,0 +1,95 @@
...
+    <div className={android ? styles.pullToRefreshContainer : ''}> {/* android prop if true disables its native PTR */}
+      <div
+        onTouchStart={handleTouchStart}
+        onTouchMove={handleTouchMove}
+        onTouchEnd={handleTouchEnd}
+      >
+        {pullDistance > 0 || isRefreshing
+          ? (
+            <>
+              <p className={`${styles.pullMessage} ${pullDistance > 50 || isRefreshing ? styles.fadeIn : ''}`}>
```
ekzyis, line 84: This suggests you want to show the message when `pullDistance > 50`, but above you're checking for `pullDistance > 0`. Isn't checking it at one place enough?

components/pull-to-refresh.js
```diff
@@ -0,0 +1,95 @@
...
+                {getPullMessage()}
+              </p>
+              {isRefreshing && <div className={styles.spacer} />}
```
ekzyis, line 87: I think we should also drop this since afaict, it only causes layout shift

components/pull-to-refresh.module.css
```diff
@@ -0,0 +1,25 @@
...
+
+.fadeIn {
+    opacity: 1;
+    top: 0;
+}
+
+.spacer {
+    position: relative;
+    margin-bottom: 32px;
+}
```
ekzyis, line 25: see other comment

components/pull-to-refresh.module.css
```diff
@@ -0,0 +1,25 @@
+.pullToRefreshContainer {
+    overscroll-behavior: contain;
+}
+
+.pullMessage {
+    position: fixed;
+    left: 50%;
+    top: -50px;
+    transform: translateX(-50%);
+    font-size: 18px;
```
ekzyis, line 10:
```suggestion
    font-size: small;
```

ekzyis (approved):
Hey, this is very impressive! It works great and the code is very clean. Stackers (myself included) are going to love this! Additionally, this was a breeze to review. Thank you so much for this! Looking forward to more contributions from you :eyes: 

Now to the review:

I tested this on desktop by simply manually setting `isPWA` to `true` and then using the mobile emulator. I also tested that the refresh actually works by loading new comments that I created in a new tab.

I only have a few UX and CSS nitpicks:

1. is it possible to prevent scrolling the window down when the `pull to refresh` shows up and I want to abort?

https://github.com/user-attachments/assets/378a4f32-06e6-468a-af34-7dbc75ba045e

2. Why did you add the spacer during refresh? Afaict it does nothing except create layout shift. Therefore I think we should also remove it, it's even more cleaner that way. So the meme should have actually been this:

image

3. the pull message is a little big so I think we should use `font-size: small` in the `pullMessage` CSS class.

4. the distance to pull down might be a little too much (almost half of the screen) but we can roll this out and then see if we need to adjust it.

5. pulling down also works when I start to pull down from the bottom of the screen when I am scrolled all the way to the top. Not sure if that's how other apps do it, I think you need to start at the top?

After dropping `refreshing ...`, the spacer and making the text smaller, the UI/UX is this:

https://github.com/user-attachments/assets/5d7ad248-834b-4467-a33f-f82c748c6059

Regarding Android: I haven't tested it on Android but since this works so well, I think we can keep your code enabled on Android to have a consistent experience.

If anything is weird, it's very easy to go back to the native behavior on Android.

Since these are all just nitpicks and it's already as-is a huge improvement, this is approved from my side :white_check_mark: 

<sub>@huumn is the final boss though</sub>

components/pull-to-refresh.js
```diff
@@ -0,0 +1,95 @@
...
+  const checkPWA = () => {
+    const androidPWA = window.matchMedia('(display-mode: standalone)').matches
+    const iosPWA = window.navigator.standalone === true
+    setIsAndroid(androidPWA) // we need to know if the user is on Android to enable toggling its native PTR
+    setIsPWA(androidPWA || iosPWA)
+  }
+
+  useEffect(() => {
+    checkPWA()
+  }, [])
```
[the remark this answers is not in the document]
Soxasora: Oops 😳

components/pull-to-refresh.js
Soxasora: Yes! I thought I was going crazy because between refreshes the text would get stuck, noted thanks :P

components/pull-to-refresh.js
Soxasora: Touch can be finicky with JS, I wanted to make sure that we can't encounter weird behavior

components/pull-to-refresh.js
Soxasora: Noted, you're right it would make much more sense!

components/pull-to-refresh.js
Soxasora: Ok so, I think that stayed there during multiple versions of this code, IIRC before placing touch checks this was a workaround to get me through refreshes and avoid unintended messages

components/pull-to-refresh.js
Soxasora: And it doesn't look clean

Soxasora:
> > is it possible to prevent scrolling the window down when the pull to refresh shows up and I want to abort?
> 
> a solution for this might be to use [the behavior in your first video](https://github.com/stackernews/stacker.news/issues/1258#issuecomment-2480071902) where the whole page moves down when you pull down

The behavior you're encountering should be uniquely present on desktop as it doesn't allow overscroll like mobile devices, if you notice in my two videos, the whole page moves down when you pull down :P

wait- now that I think about it, to make it theoretically work on Android I conditionally disable overscroll, we would get the same result you get on desktop... we definitely need to go with your UX/UI changes in order to make it pretty for both devices

Soxasora:
https://github.com/user-attachments/assets/2fc2a22d-fd46-4a02-b7c8-f52c58b388fc

well it looks super cool.

I moved `overscroll-behavior-y` into the body to avoid double refreshes, applied changes and corrected some typos

Soxasora:
i forgot to memoize the messages 🫣
with calculated margin we can stay safe across devices, didn't cross my mind, thanks @huumn!

PR #1601, opened 2024-11-18
feat: carousel integration with history navigation

### LINKED ISSUE #1508: Embedded picture UX improvment
When I am looking at an embedded picture (like in, I clicked on it, and it's filling the screen), I intuitively click "back" on my android phone to get back to the post where this picture was posted. However, because of the current choice in UX, it brings me back to the main feed instead.

Just bring me back to the post, not the feed.

I could probably break the habit and just click away from the picture to make it go away and stay in the post. That's how it currently works.

aegroto:
## Description

Carousel's modal now closes when navigating backward through the browser's history, going back to the post instead of returning to the main page. Fixes #1508.

## Screenshots

I made no screenshots but it should be pretty easy to test. I can make a video if needed.

## Additional Context
I wasn't able to use  `router.push` as suggested in the original issue and had to stick to the native APIs, [which is documented by NextJS as an alternative to router methods as well](https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#using-the-native-history-api).

## Checklist

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:** 8, I tested the changes manually and with various content.


components/carousel.js
```diff
@@ -104,10 +105,14 @@ function CarouselOverflow ({ originalSrc, rel }) {
 }
 
 export function CarouselProvider ({ children }) {
+  const pathname = usePathname()
+
   const media = useRef(new Map())
   const showModal = useShowModal()
 
   const showCarousel = useCallback(({ src }) => {
+    window.history.pushState(null, '', pathname)
```
ekzyis, line 114: This isn't a [shallow push](https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating#shallow-routing). On backwards navigation, this closes the carousel by reloading the page. We do not want to reload the page, we only want to close the carousel on backwards navigation.

ekzyis (changes requested):

components/carousel.js
aegroto: I have not been able to use `router.push` because it was not updating the history in anyway. I don't know the reason, but I believe it may be related to the path being identical to the last one, and I guess the router drops duplicate entries.

I didn't understand that the push had to be unique for the entire carousel. I will work on this as soon as possible.

huumn: Adding a query param might work:

`router.push({ pathname: router.pathname, query: { carousel: true, ...router.query}}, router.asPath, { shallow: true }}`

aegroto: That's what I've been trying to do but it looks like the carousel is immediately closed after a 'router.push' or 'router.replace' call. This does not happen when calling native `window' APIs though.

huumn: If you used `shallow: true` and it did that, then I'd guess something is rerendering that shouldn't.

aegroto: I believe this is due to line 73 of `modal.js`:
```
router.events.on('routeChangeStart', maybeOnClose)
```

The push call triggers this event, which then calls maybeOnClose. This makes sense as the native window API is not trigger router's events.

aegroto:
I have been exploring this issue in a while and I confirm that the reason `router.push` is not viable is line 73 of modal.js:

```
router.events.on('routeChangeStart', maybeOnClose)
```

On the other hand, I have not found any solid way to solve it using native APIs either. An idea may be to track state pushes and add a trigger on back events, doing as much backs as pushes caused by carousel's opening. However, this looks hacky.

@ekzyis @huumn I believe that the best thing to do should be to re-engineer how the carousel works, may using [Parallel routes](https://nextjs.org/docs/app/building-your-application/routing/parallel-routes) or, more simply, giving control to the parent component over the modal's show/hide mechanism, avoiding the involve of routing. Let me know your thoughts about this, also because if this issue needs a rewrite I kindly ask you to consider upgrading it from *easy* to *medium*.

aegroto:
I have committed a version of the fix which works in case of multiple carousel shows as well:

https://github.com/user-attachments/assets/fa6c8908-54c5-4293-86f1-0e03021154cd

The code overrides the `close` function of the carousel, triggering a 'back' navigation when the carousel is closed. It is still relying on native history APIs, unfortunately a router's shallow navigation immediately closes the carousel, presumably due to the listener I have highlighted in the previous comments.

Before marking this PR ready for review I would like to know if it's fine for your to use window's native APIs, then maybe explore a solution which relies on router in the future.

PR #1605, opened 2024-11-18
fix: hover on boost button (Issue 1566)

### LINKED ISSUE #1566: Boost icon can get stuck in hover style

### Describe the bug

If you move your cursor fast enough over the boost icon, it can get stuck in the hover style until you hover over it again.

This has been annoying me for a while but now I found out how to reproduce it.

### Video

https://github.com/user-attachments/assets/c14e8190-0c18-4bd0-812e-aae3d204c03c

### Steps To Reproduce

See video. The right speed is key.

### Expected behavior

Boost icon can not get stuck in hover style

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

AndreaDiazCorreia:
## Fix boost button hover state bug

Fixes #1566

### Description
Fixed an issue where the boost button's hover state could get stuck when moving the mouse quickly over it. The hover effect would remain even when the mouse was outside the button area until the user hovered over it again.

### Changes
- Restructured the boost button component to match the same event handling pattern as the upvote button
- Added `LongPressable` and `ActionTooltip` components to provide consistent hover behavior
- Improved hover state management by using the same event handling structure that works reliably in the upvote component

### Testing
- Rapidly move mouse over boost button - hover state should not get stuck
- Move mouse away from boost button - hover effect should disappear immediately
- Click and interact with boost button - all functionality should work as before


huumn (changes requested):
`LongPressable` doesn't make sense in this case - there is not long press action on boost. 

In general, it looks like you copied a lot of logic over from upvote that doesn't apply to boost like `nextColor`. Please remove all code that does not apply to boost.

It might help by figuring out the root cause of the issue.

AndreaDiazCorreia:
Hi! I removed some unused code. For some reason the `LongPressable` component fixes the bug by providing more robust pointer event handling. This seems to work because `LongPressable` uses a consistent event system (pointer events instead of mixing mouse/touch events) and handles event propagation at the container level, which prevents the hover state from getting stuck when the mouse moves quickly.

Changes made:
- Removed unnecessary `ActionTooltip`
- Kept `LongPressable` for event handling
- Maintained original color and style logic

huumn:
I ended up solving this without the long pressible hack as I requested in https://github.com/stackernews/stacker.news/commit/5dac2f2ed0517120bec47f452de44b63a3983e23.

Do you have a lightning address I can send some sats to for trying?

AndreaDiazCorreia:
Cool, I'll check what you changed to fix it. Here's my Lightning address: andrea@lawallet.ar

PR #1606, opened 2024-11-18
Feat: New syntax highlight system based on shiki

### LINKED ISSUE #1242: Syntax highlighting

Code blocks don't use syntax highlighting. This makes code hard to read.

If you use \`\`\`[language] instead of simply \`\`\` to start a code block, it should highlight the syntax as the selected language like Github does. Here with python as an example:

```python
def foo():
  bar = 1
  return bar
```

n/a

**Additional context**

We already use `react-syntax-highlighter` but we only use `language='text'`:

https://github.com/stackernews/stacker.news/blob/78520b787b928756c81700ab5b3ac16170ac9084/components/text.js#L134-L136

However, it requires more than passing the language since if I manually set it to `python`, it does not do syntax highlighting.

aegroto:
## Description

I have implemented a novel code highlight widget based on [Shiki](https://shiki.style/), which authors claim to [keep the bundle size small](https://shiki.style/packages/next#performance). Fixes #1242.

## Screenshots

Light mode:

immagine

Dark mode:

immagine

## Additional Context

I am undecided about which themes to adopt. The complete list is provided [here](https://shiki.style/themes).

## Checklist

**Are your changes backwards compatible? Please answer below:** This removes any reference to react-syntax-highligher, replacing it with Shiki. However, I guess no working code depends on this.

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:** 8, I have manually tested with different languages and with no at all on both comments and posts, checking if the preview was updated as well.

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:** Yes.

**Did you introduce any new environment variables? If so, call them out explicitly here:** No.


aegroto:
@huumn Thanks for your feedback. I have removed the dependency on directly setting the HTML.

I have also fixed a bug that blocked the adoption of darkMode. However, it looks like the useDarkMode hook is not triggering a re-render until the webpage is refreshed. Could you please check if my usage of the hook is incorrect? Otherwise there may be a bug preventing dark mode updates to be propagated through to the widgets.

ekzyis:
The missing dark mode update was fixed in https://github.com/stackernews/stacker.news/pull/1603

aegroto:
@ekzyis Thanks for your update, I have merged last changes to the master and I confirm dark mode switch works now! I believe the code is ready for review.

huumn:
Did you use a bundle analyzer to make sure this isn't bloating the bundle? I can do that myself, but I won't merge it until I find the time to do it. 

~80ish new transitive deps seems like a lot for a nice-to-have thing. I'll still pay the bounty even if it's not worth the deps, but I just wanted to call that out.

aegroto:
@huumn Thanks for your feedback. I totally understand your concern and I totally agree to check the actual bundle size. Have you got any reference tool you'd like me to use?

huumn:
Oh I didn't even think of doing it in CI. In the past I used https://www.npmjs.com/package/@next/bundle-analyzer

If you want to do something that works in a github action, I'd gladly pay `medium` for it.

aegroto:
Thanks. I will take a look into the various bundle analyzers and will do the analysis for this version as soon as I can. I will also try to include one that can be added as an action to the repo.

aegroto:
@huumn Sorry for the ping, I'd need a feedback. Which command should be used to build the bundle? I have tried using the standard "npm run build" and I always get some errors, mainly due to the lack of environment variables. I have tried doing it in a Docker container as well but had no luck. Using this Dockerfile:

```
# syntax=docker/dockerfile:1

FROM node:18.20.4-bullseye

RUN git clone https://github.com/stackernews/stacker.news

ENV NODE_ENV=production

WORKDIR /stacker.news

RUN npm ci
RUN npm run build
CMD ["/bin/bash"]
```

I end up having a bunch of these errors:

```
unhandledRejection TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:405:5)
    at new URL (node:internal/url:676:13)
    at /stacker.news/.next/server/chunks/4926.js:1:43211
    at 69496 (/stacker.news/.next/server/chunks/4926.js:1:43728)
    at t (/stacker.news/.next/server/webpack-runtime.js:1:127)
    at /stacker.news/.next/server/chunks/4926.js:1015:5058
    at t.a (/stacker.news/.next/server/webpack-runtime.js:1:866)
    at 95639 (/stacker.news/.next/server/chunks/4926.js:1015:4951)
    at t (/stacker.news/.next/server/webpack-runtime.js:1:127)
    at /stacker.news/.next/server/chunks/4926.js:3986:461 {
  type: 'TypeError',
  input: 'undefined',
  code: 'ERR_INVALID_URL'
}
```

As said, it looks like a missing environment variable, however I am explicitly setting NODE_ENV, I guess it should be enough.
Full logs: https://gist.github.com/aegroto/e8a8ee732ceec17efc5cf662cfa6073a

huumn:
We usually do builds in production, so I don't know off the top of my head.

But, you might be able to simply copy `.env.development` to `.env.production` and get farther. 

If you have more questions that aren't exclusively related to this PR, please as them in the discussion area.

aegroto:
I have designed an action that compares the bundle size of pull requests and applied this in my fork to evaluate the impact of these modifications on the current master: https://github.com/aegroto/stacker.news/pull/4#issuecomment-2514983823

It should be tuned to the dimension of the bundle. However, it seems like Shiki increases the bundle size of 20 KiB over 1.61 MB, let me know if you find this acceptable.

PR #1609, opened 2024-11-19
Fix 'Cannot mix BigInt and other types' on zaps

ekzyis:
## Description

In #1570, `sybilFeePercent` was removed from the context. This broke zaps with fee credits because `sybilFeePercent` was now undefined.

## Additional Context

`sybilFeePercent` is now always included in any action context. 

## Checklist

yes

`6`. I tested zaps with fee credits and pessimistic zaps via anon.

n/a

no

### DIFF
--- a/api/paidAction/index.js
+++ b/api/paidAction/index.js
@@ -56,7 +56,8 @@ export default async function performPaidAction (actionType, args, incomingConte
     }
     const context = {
       ...contextWithMe,
-      cost: await paidAction.getCost(args, contextWithMe)
+      cost: await paidAction.getCost(args, contextWithMe),
+      sybilFeePercent: await paidAction.getSybilFeePercent?.(args, contextWithMe)
     }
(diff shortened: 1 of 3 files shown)


api/paidAction/index.js
```diff
@@ -186,8 +187,7 @@ async function beginPessimisticAction (actionType, args, context) {
 async function performP2PAction (actionType, args, incomingContext) {
   // if the action has an invoiceable peer, we'll create a peer invoice
   // wrap it, and return the wrapped invoice
-  const { cost, models, lnd, me } = incomingContext
-  const sybilFeePercent = await paidActions[actionType].getSybilFeePercent?.(args, incomingContext)
+  const { cost, sybilFeePercent, models, lnd, me } = incomingContext
```
ekzyis, line 190: The incoming context now already includes `sybilFeePercent`

PR #1615, opened 2024-11-19
Fix mentions for pessimistic actions

ekzyis:
## Description

If we're running an item creation pessimistically, the worker didn't include `models` in the context when calling `performPessimisticAction`. If the text included a mention, the query to fetch the mentioned users would then fail.

## Video

https://github.com/user-attachments/assets/480ae087-8868-4c2f-b74b-b77e3ed852d6

## Checklist

yes

`5`. Anon can now mention stackers.

n/a

no


huumn:
those should be using `tx` instead. I'll fix it.

PR #1621, opened 2024-11-20
Always check res.ok and content-type header

ekzyis:
## Description

I have noticed that I forgot some response assertions in #1516.

I also noticed that it makes more sense to _always_ check the content-type header if we're going to call `res.json()`.

Therefore, if we know the response might include error details in JSON, we first check the header. If we know this isn't the case, we first check if the response was ok before we check the header. If we would always check first if the response is ok, we might miss error details.

## Additional Context

https://stacker.news/items/773703

## Checklist

yes

`7`. I tested successful withdrawals with:

- LNbits
- NWC
- ~CLN~ _local certificate needs an update_

n/a

no

### DIFF
--- a/wallets/blink/common.js
+++ b/wallets/blink/common.js
@@ -1,3 +1,5 @@
+import { assertContentTypeJson, assertResponseOk } from '@/lib/url'
+
 export const galoyBlinkUrl = 'https://api.blink.sv/graphql'
 export const galoyBlinkDashboardUrl = 'https://dashboard.blink.sv/'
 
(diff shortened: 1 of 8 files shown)


wallets/blink/common.js
```diff
@@ -37,15 +39,14 @@ export async function request (authToken, query, variables = {}) {
...
-    if (res.status === 401) {
-      throw new Error('unauthorized')
-    } else {
-      throw new Error('API responded with HTTP ' + res.status)
-    }
-  }
+
+  // consume response body to avoid memory leaks
+  // see https://github.com/nodejs/node/issues/51162
+  res.text().catch(() => {})
```
ekzyis, line 45: @riccardobl I remember you mentioned memory leaks if the body is not consumed as mentioned in https://github.com/nodejs/node/issues/51162. But the ticket is closed so I am not sure if that's a problem? Afaict, the response body will be garbage-collected.

If this is still a problem, shouldn't we make sure the body is consumed for every fetch on the server?

ekzyis: Oh, I think I added a bug since I think I consume the body here so I no longer have it for `res.json()`.

huumn:
You should be able to test CLN without a cert. I think I made that change to test against polar.

wallets/blink/common.js
ekzyis: > This is still a problem and afaik we do consume the body on all fetches on the server.

Mhh, but not in the error handling I added in #1516. We can throw before we consume it.

ekzyis:
> You should be able to test CLN without a cert. I think I made that change to test against polar.

Ah ok, it works with the onion but only with the cert, else I get

```
ERROR	failed to create test invoice: request to http://hmsykvji73qkrzxnjxmnzas3uvync7sbnubrmijclr3tu46r3ve2nuyd.onion:3010/v1/invoice failed, reason: socket hang up
```

PR #1627, opened 2024-11-21
Wallet filters

ekzyis:
## Description

This adds filters to the wallet page so we can filter by send or receive.

It also stores the filter state in the query params.

## Video

https://github.com/user-attachments/assets/7e545bb6-47b0-49e7-a096-9fd661f86846

## Additional context

See comments

## Checklist

yes

`8`. Tested filters, tested route replacement and reload and that backward navigation is not prevented by using `router.replace` instead of `router.push`.

yes

no

### DIFF
--- a/pages/settings/wallets/index.js
+++ b/pages/settings/wallets/index.js
@@ -7,6 +7,9 @@ import { useCallback, useState } from 'react'
 import { useIsClient } from '@/components/use-client'
 import WalletCard from '@/components/wallet-card'
 import { useToast } from '@/components/toast'
+import BootstrapForm from 'react-bootstrap/Form'
+import RecvIcon from '@/svgs/arrow-left-down-line.svg'
+import SendIcon from '@/svgs/arrow-right-up-line.svg'
 
 export const getServerSideProps = getGetServerSideProps({ authRequired: true })
 
@@ -16,6 +19,7 @@ export default function Wallet ({ ssrData }) {
   const isClient = useIsClient()
   const [sourceIndex, setSourceIndex] = useState(null)
   const [targetIndex, setTargetIndex] = useState(null)
+  const [filter, setFilter] = useState({ send: false, receive: false })
 
   const reorder = useCallback(async (sourceIndex, targetIndex) => {
     const newOrder = [...wallets.filter(w => w.config?.enabled)]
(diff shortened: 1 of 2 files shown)


styles/wallet.module.css
```diff
@@ -1,10 +1,13 @@
 .walletGrid {
   display: grid;
-  grid-template-columns: repeat(auto-fit, minmax(160px, 1fr));
+  grid-template-columns: repeat(auto-fill, 160px);
```
ekzyis, line 3: Using the same width as a card will be (see fixed width below) prevents a grid layout shift if there aren't enough cards for the first row.

styles/wallet.module.css
```diff
@@ -1,10 +1,13 @@
...
   grid-gap: 20px;
-  justify-items: center;
-  align-items: center;
```
ekzyis, line 6: these had no effect afaict

pages/settings/wallets/index.js
```diff
@@ -17,6 +21,12 @@ export default function Wallet ({ ssrData }) {
   const [sourceIndex, setSourceIndex] = useState(null)
   const [targetIndex, setTargetIndex] = useState(null)
 
+  const router = useRouter()
+  const [filter, setFilter] = useState({
+    send: router.query.send === 'true' || false,
+    receive: router.query.receive === 'true' || false
```
riccardobl, line 27:
```suggestion
    send: router.query.send === 'true',
    receive: router.query.receive === 'true'
```
forgotted debug code?

pages/settings/wallets/index.js
```diff
@@ -65,6 +75,13 @@ export default function Wallet ({ ssrData }) {
     }
   }, [sourceIndex, reorder, onReorderError])
 
+  const onFilterChange = useCallback((key) => {
+    return e => {
+      setFilter({ ...filter, [key]: e.target.checked })
+      router.replace({ query: { ...router.query, [key]: e.target.checked } }, undefined, { shallow: true })
+    }
```
riccardobl, line 82:
```suggestion
    setFilter((oldFilter) => ({
        ...oldFilter,
        [key]: e.target.checked
    }))
```
avoid state states

pages/settings/wallets/index.js
```diff
@@ -65,6 +75,13 @@ export default function Wallet ({ ssrData }) {
     }
   }, [sourceIndex, reorder, onReorderError])
 
+  const onFilterChange = useCallback((key) => {
+    return e => {
+      setFilter({ ...filter, [key]: e.target.checked })
```
riccardobl, line 80:
```suggestion
    setFilter((oldFilter) => ({
        ...oldFilter,
        [key]: e.target.checked
    }))
```
avoid stale states

riccardobl:
Not a big deal , but I've noticed a small regression the layout responsiveness, check this

before
image


after
image

styles/wallet.module.css
```diff
@@ -1,10 +1,13 @@
...
+}
+
+.walletFilters {
   margin-top: 3rem;
+  margin-left: 0.2rem;
+  margin-bottom: 0.5rem;
```
riccardobl, line 10: selecting and deselecting a filter quickly causes the word to be selected by the browser
image

maybe we should add
```
  user-select: none;
```
here

pages/settings/wallets/index.js
ekzyis: Oh, no, I just added the router stuff later and didn't realize the `|| false` is unnecessary, thanks

ekzyis:
> Not a big deal , but, unless this is intentional (but it seems a bit off with a centered layout), I've noticed a small regression in the layout responsiveness, check this

Ohh, I only tested on iPhone SE dimensions, good catch, this was not intended!

I only intended to prevent this behavior on desktop if the first row can't be filled:

```diff
 .walletGrid {
   display: grid;
-  grid-template-columns: repeat(auto-fit, minmax(160px, 1fr));
+  grid-template-columns: repeat(auto-fill, 160px);
```

https://github.com/user-attachments/assets/ce5ae07f-118c-4bb8-83ab-6b4c10e526dd

But this can't happen on mobile, so on mobile, the grid should continue to use `minmax(160px, 1fr)`. I'll add a media query.

riccardobl:
Thank you for looking into it.
There are still some edge cases, see below, but it is hard to catch all of them with a grid layout. 
I've tested with all the common resolutions in north america and europe (stats from https://gs.statcounter.com/screen-resolution-stats) and it looks fine, so i think we are good.

image
image

riccardobl (approved):

PR #1629, opened 2024-11-21
Filter Bounties By Paid and Unpaid

tsmith123:
## Description

This PR adds a sub filter to the recent/bounties page so that paid and unpaid bounties can be shown.

When the user selects the sub filter, it appends a query string value to the url and passes this value to the item resolver. If the user selects "all" from the filter then the query string value will be removed.

**DRAFT**
This is a draft PR because the SQL queries don't yet work. I'm not sure how to know whether a bounty has been paid which is why the code is commented out in the PR. I've tried to use `Item.bountyPaidTo` but it results in an error.

## Screenshots
Screenshot 2024-11-21 at 06 41 11
Screenshot 2024-11-21 at 06 41 21

## Additional Context

My test environment doesn't have any posts with bounties and therefore my local testing has been limited. I don't know how to add sats to my test account so I can't create bounty posts.

Is it possible to preload my test account with sats to help with testing?


api/resolvers/item.js
```diff
@@ -299,6 +299,12 @@ function typeClause (type) {
     case 'bios':
       return ['"Item".bio = true', '"Item"."parentId" IS NULL']
     case 'bounties':
+      // if (subType === 'paid') {
+      //   return ['"Item".bounty IS NOT NULL', '"Item".bountyPaidTo IS NOT NULL', '"Item"."parentId" IS NULL']
+      // } else if (subType === 'unpaid') {
+      //   return ['"Item".bounty IS NOT NULL', '"Item".bountyPaidTo IS NULL', '"Item"."parentId" IS NULL']
+      // }
+
```
tsmith123, line 307: @huumn Is `bountyPaidTo` a field because I tried the above but it returns an error. Is this "the way" to know whether a bounty has been paid?

ekzyis:
This filter has already been added in #1589

tsmith123:
Oh bugger. No one had assigned it. Mine does work slightly differently however as it makes a new request to fetch unpaid and paid rather than filtering the ones on screen already. Maybe some of it is usable.

tsmith123:
Ok, only looked very quickly. If you refresh the page does it preserve the filter?

aegroto:
I admit I have not tried, but should it? You are actually refreshing the "bounties" page so it may make sense to reset the filter on every hard refresh.

api/resolvers/item.js
ekzyis: You need to use quotation marks for case-sensitive SQL fields (see surrounding code)

ekzyis:
I am sorry for your unnecessary work. Next time please check if there was a PR linked. The link also mentioned it will close the issue:

2024-11-21-154301_854x86_scrot

I have already reviewed that #1589 contains 99% of what we need.

PR #1630, opened 2024-11-21
HOTFIX: restore missing link between lnurlp and nip57

riccardobl:
The nip57 job relies on the dbInvoice desc field to get the metadata it needs to publish the zap receipts
```js
    const desc = JSON.parse(inv.desc)
    const ptag = desc.tags.filter(t => t?.length >= 2 && t[0] === 'p')[0]
    const etag = desc.tags.filter(t => t?.length >= 2 && t[0] === 'e')[0]
    const atag = desc.tags.filter(t => t?.length >= 2 && t[0] === 'a')[0]
    const relays = desc.tags.find(t => t?.length >= 2 && t[0] === 'relays').slice(1)
```
however at some point there must have been a regression that caused this link to be broken, since this field is not being used anymore.

This PR adds an optional `internalDescription` context field that when passes is used to set the dbInvoice.desc field (that could be different than the servedInvoice.description field due to privacy settings).

This is an hotfix to restore the functionality with minimal changes, to avoid conflicts, since this code is being worked elsewhere.

### DIFF
--- a/api/paidAction/index.js
+++ b/api/paidAction/index.js
@@ -351,7 +351,8 @@ async function createDbInvoice (actionType, args, context) {
     actionOptimistic: optimistic,
     actionArgs: args,
     expiresAt,
-    actionId
+    actionId,
+    desc: context.internalDescription ?? servedInvoice.description
   }
 
   let invoice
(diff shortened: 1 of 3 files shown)


huumn (commented):
Nice catch!

I think we'll want to pass this to `RECEIVE`'s `perform` as a `noteStr` arg like we do with `lud18Data`. What do you think?

riccardobl:
Sure, why not. I thought the original intention was to have the desc field set for all the invoices. I've applied your suggestion in the last commit.

PR #1631, opened 2024-11-22
Use invoice.cancelledAt to determine if invoice expired or was canceled by user

ekzyis:
## Description

If an invoice expires, LND emits a new invoice update event with `is_canceled` set. We use that event in `checkInvoice` to mark the invoice as cancelled in our database.

However, this means that we can't really tell if an invoice was canceled because it expired or because it was actually canceled by the user. For LND, this makes no difference, but imo, this makes a difference in our UI because QR codes change their state from "expired" to "cancelled".

I think this state change is bad UX. If the user didn't cancel the invoice themselves (by closing the modal), the status should stay at expired and the modal should not close.

I fixed this by adding a new column `Invoice.cancelledAt` so we can now check `invoice.expiresAt < invoice.cancelledAt`. If this is true, we know the invoice was canceled because it expired.

## Video

<details>
<summary>existing behavior on master</summary>

https://github.com/user-attachments/assets/f84649d1-9628-414c-b1d9-0e4f41e23b2f

</details>

<details>
<summary>new (fixed) behavior (this PR)</summary>

https://github.com/user-attachments/assets/14dd3241-85da-45dc-b60c-ecb5c8e1cfea

</details>

## Additional checklist

I have seen sometimes on master how the invoice expired but it was never cancelled in the database. However, I wasn't able to properly debug using `console.trace` etc. since it never seemed to happen when I was ready to debug it. LND might not always emit an event if an invoice expired or we're missing events.

I have also seen this rare bug in this PR.

## Checklist

yes, I only added a new column but didn't drop `invoice.cancelled`. We can drop it in a future migration if we want.

`6`. QR codes that expire stay open and continue to show `expired`.

n/a

no

### DIFF
--- a/components/countdown.js
+++ b/components/countdown.js
@@ -43,7 +43,7 @@ export function CompactLongCountdown (props) {
                 ? ` ${props.formatted.hours}:${props.formatted.minutes}:${props.formatted.seconds}`
                 : Number(props.formatted.minutes) > 0
                   ? ` ${props.formatted.minutes}:${props.formatted.seconds}`
-                  : Number(props.formatted.seconds) > 0
+                  : Number(props.formatted.seconds) >= 0
                     ? ` ${props.formatted.seconds}s`
                     : ' '}
           </>
--- a/api/typeDefs/wallet.js
+++ b/api/typeDefs/wallet.js
@@ -108,6 +108,7 @@ const typeDefs = `
     bolt11: String!
     expiresAt: Date!
     cancelled: Boolean!
+    cancelledAt: Date
     confirmedAt: Date
     satsReceived: Int
     satsRequested: Int!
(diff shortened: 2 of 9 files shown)


components/countdown.js
```diff
@@ -43,7 +43,7 @@ export function CompactLongCountdown (props) {
                 ? ` ${props.formatted.hours}:${props.formatted.minutes}:${props.formatted.seconds}`
                 : Number(props.formatted.minutes) > 0
                   ? ` ${props.formatted.minutes}:${props.formatted.seconds}`
-                  : Number(props.formatted.seconds) > 0
+                  : Number(props.formatted.seconds) >= 0
```
ekzyis, line 46: Since I removed the state update after the countdown runs on, I am using `>=` here now so the timer doesn't disappear when timer ran out but the next poll is pending.

Maybe I shouldn't have removed the state update but this makes the code simpler (less state) + it exposes the rare bug by getting stuck at 0 seconds.

PR #1633, opened 2024-11-22
Fix: progress bar shown on back navigation through pathname check

### LINKED ISSUE #1471: Progress bar for back button

aegroto:
## Description

Progress bar is shown when navigating back in the history. Fixes #1471.

## Videos

**Current behaviour (master)**
The progress bar is not showing in the upper-left corner while navigating back.

https://github.com/user-attachments/assets/5d307dca-7fa7-4206-b0b6-a5e577dbdc98

**Proposed behaviour (this PR)**
The progress bar is shown in the upper-left corner while navigating back if the pathname changes.

https://github.com/user-attachments/assets/320b49a5-59e8-48f6-928b-185e897330c1

## Additional Context

The fix is based on checking, in case of shallow routing, if the pathname changes. If so, the progress bar is shown, otherwise it is not. I think this makes sense as we do not want to show progress bar if we are navigating in the same page. If this is not desired, instead, it would make sense to not check the 'shallow' value at all.

## Checklist

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:** 8, manual testing on both master and PR branch


huumn:
There was an issue where the progress bar wouldn't stop in some cases. I think it's because useEffect would run before the route finished changing, racing against the callback. I fixed it by making progress.done() unconditional. Other than that, I think this should work, thanks!

aegroto:
@huumn I have noted that you have replaced pathname with router.pathname. Did you check if the progress bar works only when the query params change? Because I guess router.pathname does not. Excuse me if this is the intended behaviour, this is just a doubt that has come into my mind.

huumn:
It works when query params change and when they stay the same but the rest of the path changes.

PR #1636, opened 2024-11-22
Nip46 auth with NDK

### LINKED ISSUE #864: Login with Nostr NIP-46

riccardobl:
## Description

Nip46 auth using NDK (reimplementation of https://github.com/stackernews/stacker.news/pull/1311)

Depends on https://github.com/stackernews/stacker.news/pull/1590 closes #864 

This PR updates the behavior of the "Login with Nostr" button by adding an intermediary page that presents the following login options:
- **Login with extension**:  the current behavior using the browser extension eg. alby
- **Login by pasting a bunker:// token**
- **Login by using a nip-05 address**

# Screenshots

image
image
image

# Additional Context

The login by qr code/nostrconnect:// has been removed due to the thing we discussed. 

I've been testing using https://nsec.app/

## Checklist

yes

yes , tested both bunker:// , nip-05 and extension to ensure there is no regression.

Yes

no


ekzyis:
I went ahead and rebased this branch #1590, I hope you don't mind.

However, I wonder if this in draft because it depends on #1590 even though #1590 itself was ready for review? Or is this in draft because the changes between this PR and #1590 (`git diff ndk..nip46V2`) aren't ready for review yet?

Also, can you use the PR template which mentions how you tested this (which remote signer etc.) or if there was anything unclear during implementation? This helps a lot with the review.

components/nostr-auth.js
```diff
@@ -74,83 +89,186 @@ export function NostrAuth ({ text, callbackUrl, multiAuth }) {
...
-          })
-        } catch (e) {
-          throw new Error('authorization failed', e)
-        }
-      } catch (e) {
-        if (!mounted) return
-        console.log('nostr auth error', e)
-        setExtensionError({ message: `${text} failed`, details: e.message })
+  // authorize user
+  const auth = useCallback(async (nip46token) => {
```
riccardobl, line 105: Since NDK has this nice signer abstraction, i've made this single callback to handle both nip46 and nip07 logins.

components/nostr-auth.js
```diff
@@ -74,83 +89,186 @@ export function NostrAuth ({ text, callbackUrl, multiAuth }) {
...
+                    <Row>
+                      <Col>
+                        <ul>
+                          <li>
+                            <a href='https://nsec.app/'>Nsec.app</a><br />
+                            available for: chrome, firefox, and safari
+                          </li>
+                          <li>
+                            <a href='https://app.nsecbunker.com/'>nsecBunker</a><br />
+                            available as: SaaS or self-hosted
```
riccardobl, line 214: the original PR had more signers in this list, but i've found they have issues or don't seem to be very well maintained, so i've decided to keep only these two

ekzyis (changes requested):
1. **URL parsing issue**

This has the same issue as the NDK PR had: NDK only looks at `url.hostname` to set the bunker pubkey but this doesn't work in Firefox 127.0.2 for example.

To test further, I manually set `bunkerPubkey` to the correct value before I return the signer (see video below).

2. **Confusing modal**

I noticed that the modal button text is confusing: "confirm" sounds like you have to go yourself to the bunker and confirm the request, and _then_ click on confirm in SN but clicking on confirm in SN opens the bunker in a new tab and closes the modal. The SN screen then only shows a spinner without any information to the user what is going on.

Similar issue with "cancel":  you click on it, it also just closes the modal but continues to show the spinner. I'd expect it to completely cancel the login process.

I think this could be fixed by not using a modal. Why did you choose to use one here? If you want to give a link to the bunker, you can just provide a link to the bunker without a modal, no? I think there should never be just a spinner on a screen without telling the user if we're waiting for them to do something or if we're waiting for something else to happen. 

3. **No modal shown sometimes**

When I login and logout, I sometimes get no modal. I think this is related to using `signer.once("authUrl", challengeResolver)` and when there was no page load between login and logout. The signer doesn't fire again on the next login.

Here's a video where this happens (02:12 - 02:24):

https://github.com/user-attachments/assets/edbe8258-d5b1-4962-9e9c-ee6678872aba

4. **Two signatures required per login**

Having to confirm two requests to login is also unfortunate but I think you did that to keep the login flow the same for all signers so this PR can stay simple. We can keep it like this for this PR but this isn't ideal UX so I think we will want to change this at some point. I think this means the Nip47Signer needs to run on the server?

5. **Crossposting**

Afaict, this doesn't work with crossposting. it throws "no way to sign this event, please provide a signer or private key".

riccardobl:
> 4. Two signatures required per login

what do you mean? There is only one signature per login

> I think this means the Nip47Signer needs to run on the server?

that would expose user secrets

> 5. Crossposting

This pr is only for login, as the feature request was only for that. 
Crossposting needs more work that includes having a way to set which signer to use and store the secret to the vault.

riccardobl:
> 3. No modal shown sometimes
> When I login and logout, I sometimes get no modal. I think this is related to using signer.once("authUrl", challengeResolver) and when there was no page load between login and logout. The signer doesn't fire again on the next login.

the callback is set everytime auth() is called, so it doesn't need a page reload to work.
I don't understand what you are doing exactly to test this. 
Are you trying to login twice with the same nip46 token?

https://github.com/user-attachments/assets/ae8713a7-ccbf-400b-abae-afbf9c502693

ekzyis:
> what do you mean? There is only one signature per login

Sorry, I meant two confirmations in the signer required. I think the solution would be to use [NIP-98](https://github.com/nostr-protocol/nips/blob/master/98.md) since permission to sign `kind:27235` is already included in the basic permissions. So we can ask the signer to sign `kind:27235` without another confirmation and send that signed event to the server instead of our custom event. 

> that would expose user secrets

Even though my idea about running the signer on the server is no longer relevant thanks to NIP-98, which user secrets would this expose since this is remote signing? Do you mean the secrets in the bunker URL?

> This pr is only for login, as the feature request was only for that.
Crossposting needs more work that includes having a way to set which signer to use and store the secret to the vault.

Ah okay, makes sense. I forgot I was using firefox where I didn't have the nostr extension so I thought this broke crossposting but it didn't.

> Are you trying to login twice with the same nip46 token?

Oh yes, that was the issue. Works with fresh tokens. Is it possible to handle this? Only showing a spinner with no info is bad UX (related to my question why you decided to use a modal).

riccardobl:
> 1. URL parsing issue

should be fixed now, i've added the same workaround


> 2. Confusing modal

i've replaced the modal with this

image

riccardobl:
> Sorry, I meant two confirmations in the signer required.

you mean the two popups that comes up in nsec.app? 
I think this is either an ux problem in nsec.app or NDK is not using nip46 to its full extent, because it is possible to ask for permissions in the [connect method](https://github.com/nostr-protocol/nips/blob/master/46.md#methodscommands)

> Even though my idea about running the signer on the server is no longer relevant thanks to NIP-98, which user secrets would this expose since this is remote signing? Do you mean the secrets in the bunker URL?

yes the server could steal the bunker secret

NIP-98 seems another auth method that is unrelated to NIP-46

> Oh yes, that was the issue. Works with fresh tokens. Is it possible to handle this? Only showing a spinner with no info is bad UX (related to my question why you decided to use a modal).

we can only add an arbitrary timeout, it is the signer (nsec.app) that decides to ignore these events instead of replying with an error

riccardobl:
> NDK is not using nip46 to its full extent

i think it's ndk, see
```javascript
  return new Promise((resolve, reject) => {
      const connectParams = [this.userPubkey ?? ""];
      if (this.secret) connectParams.push(this.secret);
      if (!this.bunkerPubkey) throw new Error("Bunker pubkey not set");
      this.rpc.sendRequest(
        this.bunkerPubkey,
        "connect",
        connectParams,
        24133,
        (response) => {
          if (response.result === "ack") {
            this.getPublicKey().then((pubkey) => {
              this._user = new NDKUser({ pubkey });
              resolve(this._user);
            });
          } else {
            reject(response.error);
          }
        }
      );
    });
```
it should include a third `optional_requested_permissions` connectParams

ekzyis:
> you mean the two popups that comes up in nsec.app?

yes

> I think this is either an ux problem in nsec.app or NDK is not using nip46 to its full extent, because it is possible to ask for permissions in the [connect method](https://github.com/nostr-protocol/nips/blob/master/46.md#methodscommands)

I think it's an issue with how we're using NIP-46 for our authentication. We don't need to ask for additional permissions, we're just not using the permissions we're given.

> NIP-98 seems another auth method that is unrelated to NIP-46

NIP-98 is the same that we currently do (sign an event) but standardized so that we don't have to ask for permission to sign the event since that permission was already granted

> we can only add an arbitrary timeout, it is the signer (nsec.app) that decides to ignore these events instead of replying with an error

We could show a message like "are you sure you used a fresh token?" after the timeout but not actually abort the login because the user might just be slow with pressing "confirm" in their signer

riccardobl:
is it defined anywhere that kind 27235 is supposed to be a default permissions or is it a choice of nsec.app?
We can change kind of the event, in the end we just need the signature, but isn't it an arbitrary choice of the signer?

lib/nostr.js
```diff
@@ -187,48 +179,10 @@ export function nostrZapDetails (zap) {
...
-    }
-
-    return { successfulRelays, failedRelays, noteId }
-  } catch (error) {
-    console.error('Crosspost error:', error)
-    return { error }
+// workaround NDK url parsing issue (see https://github.com/stackernews/stacker.news/pull/1636)
+class NDKNip46SignerURLPatch extends NDKNip46Signer {
+  connectionTokenInit (connectionToken) {
+    connectionToken = 'http' + connectionToken.substring(connectionToken.indexOf('://'))
```
ekzyis, line 185: Can this be anything else than `bunker://`? If not, I think this is easier to read and does the same:

```suggestion
    connectionToken = connectionToken.replace('bunker://', 'http://')
```

ekzyis:
> Is it defined anywhere that kind 27235 is supposed to be a default permissions or is it a choice of nsec.app?
We can change kind of the event, in the end we just need the signature, but isn't it an arbitrary choice of the signer?

I don't know, but it's at least a standard.

> EDIT: changed to use kind 27235

Thanks, much better now!

riccardobl:
> NIP-46 is only remote signing, it's not actually an auth method. NIP-98 is the auth method. It's the same that we currently do (sign an event) but standardized so that we don't have to ask for permission to sign the event since that permission was already granted

NIP-98 requires its own flow [see](https://github.com/nostr-protocol/nips/blob/master/98.md) it is basically nostr basic auth, i don't think it is worth implementing all that for us, so i've just added some generic metadata to make it "pass" just in case some signer is doing some extra validation

ekzyis (commented):
Some CSS suggestions:

```diff
diff --git a/components/nostr-auth.js b/components/nostr-auth.js
index 697853af..e056516b 100644
--- a/components/nostr-auth.js
+++ b/components/nostr-auth.js
@@ -138,13 +138,13 @@ export function NostrAuth ({ text, callbackUrl, multiAuth }) {
 
   return (
     <>
-      <h3 className='w-100 pb-2'>{status.title ? status.title : ((text || 'Login') + ' with Nostr')}</h3>
+      <h3 className='w-100 mb-0 pb-3 text-center'>{status.title ? status.title : ((text || 'Login') + ' with Nostr')}</h3>
       {status.error && <NostrError message={status.msg} />}
       {status.loading
         ? (
           <>
-            <Moon className='spin fill-grey' width='50' height='50' />
-            <div className='text-muted pt-4 pb-4 w-100'>{status.msg}</div>
+            <Moon className='spin fill-grey' width='24' height='24' />
+            <div className='text-muted py-3 w-100 text-center'>{status.msg}</div>
             {status.button && (
               <Button
                 className='w-100' variant='primary'

```

<details>
<summary>before</summary>

2024-12-08-213629_1920x1200_scrot

</details>

<details>
<summary>after</summary>

2024-12-08-213701_1920x1200_scrot

</details>

How does the NIP-05 login work and how can I test it? I see that [NIP-46](https://github.com/nostr-protocol/nips/blob/master/46.md) mentions this now:

>  nip05 login is removed

Should we drop it and only mention `bunker://`?

riccardobl:
Added the timeout 

image

riccardobl:
> Some CSS suggestions:

wouldn't this make it look a bit odd compared to the other login screens and the general sn layout that has the main alignment to left?

image

riccardobl:
> How does the NIP-05 login work and how can I test it? I see that [NIP-46](https://github.com/nostr-protocol/nips/blob/master/46.md) mentions this now:

you can use this from the same nsec.app page 
image

lib/nostr.js
ekzyis: @riccardobl did you miss this suggestion?

ekzyis (approved):
> wouldn't this make it look a bit odd compared to the other login screens and the general sn layout that is generally aligned to left?

Mhh, I don't know. I just felt that the spinner in the center with everything else not in the center looks odd. But maybe it's just me or it's just too big for me.

So please ignore if you don't agree, @huumn will have the last word anyway.

Going to approve this now since my comment about the timeout implementation and the wording of the hint are mostly just nitpicks.

components/nostr-auth.js
```diff
@@ -1,71 +1,78 @@
...
+  const [status, setStatus] = useState({
+    msg: '',
+    error: false,
+    loading: false,
+    title: undefined,
+    button: undefined
+  })
+
+  const [suggestion, setSuggestion] = useState(null)
+  const suggestionTimeout = useRef(null)
```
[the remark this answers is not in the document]
riccardobl: The clearSuggestionTimer in the finally block is to avoid causing an useless state update if we are showing the error, the clearSuggestionTimer on unmount is to not hold the context (up to 10 seconds) from the garbage collector.
Generally it is a good practice to make code that clear after itself.

Issue #853, opened 2024-02-18
Support PDF uploads


huumn:
`imgproxy` allows for conversion of PDFs to images, so add support for PDF uploads.

ekzyis:
> PDF only? What about epubs?

-- https://stacker.news/items/429663

Since I don't see EPUB specifically mentioned in the improxy docs[^1], I think if we want to support epubs, we need to convert them to PDF first.

[^1]: https://docs.imgproxy.net/image_formats_support

brymut:
Still working on this @Soxasora ?

PR #1637, opened 2024-11-22
Feat: PDF support via Imgproxy

Soxasora:
## Description

Completes #853.
These changes allows for PDFs to be uploaded and processed by Imgproxy. At the moment it retrieves the first page of an uploaded PDF to be displayed as a thumbnail, sets the quality, the color background and DPI per [Imgproxy options](https://docs.imgproxy.net/usage/processing#page).

## Screenshots

n/a

## Additional Context

This is an exclusive Imgproxy PRO feature, other than following the documentation, AFAIK there's no way for me to test this functionality. Therefore i'll draft this PR until we're sure it works.

## Checklist

Being an addition with processing and detection happening only after image/video, it shouldn't affect existing uploads

2. 
- PDFs do upload and can be accessed
- Images and videos completely work as before
- Functionality cannot be tested as this is an imgproxy pro only feature and I don't have access to it atm.

Color background for PDF is set to be white

### DIFF
--- a/components/file-upload.js
+++ b/components/file-upload.js
@@ -21,7 +21,9 @@ export const FileUpload = forwardRef(({ children, className, onSelect, onUpload,
   const s3Upload = useCallback(async file => {
     const element = file.type.startsWith('image/')
       ? new window.Image()
-      : document.createElement('video')
+      : file.type.startsWith('video/')
+        ? document.createElement('video')
+        : document.createElement('object') // pdf
 
     file = await removeExifData(file)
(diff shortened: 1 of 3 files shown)


Soxasora:
> >    PDF only? What about epubs?
> 
> -- https://stacker.news/items/429663

epubs can be converted to PDF via [epub-parser](https://www.npmjs.com/package/@gxl/epub-parser). If support is still wanted we need to evaluate adding this library

PR #1639, opened 2024-11-22
fix LNC attachment and docs

riccardobl:
It seems the import broke at some point. This PR makes the LNC attachment  """"work"""" again.

Also, the doc needed a refresh after the sndev refactoring.


huumn:
Thanks!

PR #1640, opened 2024-11-22
refactor: Check darkmode in useWallets

ekzyis:
## Description

`useWallets` is a more appropriate location to check which image we should use.

## Checklist

yes

`8`. The correct image is still used in light and dark mode.

### DIFF
--- a/components/wallet-card.js
+++ b/components/wallet-card.js
@@ -7,8 +7,6 @@ import { Status, isConfigured } from '@/wallets/common'
 import DraggableIcon from '@/svgs/draggable.svg'
 import RecvIcon from '@/svgs/arrow-left-down-line.svg'
 import SendIcon from '@/svgs/arrow-right-up-line.svg'
-import useDarkMode from './dark-mode'
-import { useEffect, useState } from 'react'
 
 const statusToClass = status => {
   switch (status) {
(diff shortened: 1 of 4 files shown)


huumn:
> useWallets is a more appropriate location to check which image we should use.

Can you explain why that is the case? Isn't this mixing data and presentation and when we modify wallet cards, I now need to modify configuration loading?

Might it be better to just have a hook for this? e.g. `useWalletImage(wallet)`

ekzyis:
> Can you explain why that is the case? Isn't this mixing data and presentation and when we modify wallet cards, I now need to modify configuration loading?

Mhh, I didn't see this as "mixing data and presentation", only as "returning the data the rest of the code needs". With doing the darkmode check in `useWallets`, other code doesn't need to be aware of this detail.

> Might it be better to just have a hook for this? e.g. `useWalletImage(wallet)`

Yeah this sounds better though and related to https://github.com/stackernews/stacker.news/pull/1643#discussion_r1855255109

huumn:
> Mhh, I didn't see this as "mixing data and presentation", only as "returning the data the rest of the code needs". 

I had similar feelings about the way `status` was included in `useWallets`, but I didn't want to block that and knew we can easily refactor one violation out in the future. I definitely feel the urge myself to start doing this too though. I struggled with this a lot during my refactor.

ekzyis:
> I had similar feelings about the way `status` was included in `useWallets`, but I didn't want to block that and knew we can easily refactor one violation out in the future. I definitely feel the urge myself to start doing this too though. I struggled with this a lot during my refactor.

lol, just created a PR for this: #1646

PR #1642, opened 2024-11-23
Sender fallbacks

ekzyis:
## Description

Partially solve #1494 for sender side

Based on #1651 #1660

## TODO

- [x] fix payment check in `sendPayment`: wallets that are enabled but can only receive will throw `SenderError` instead of `WalletNotConfiguredError`
- [x] handle aggregated wallet errors in QR code
- [x] fix cache not updated with latest invoice on payment success or error
- [x] fix p2p zap retries create new invoices from SN, not from receiver
- [x] ~fix hmacs lost~ _doesn't matter, anon is not affected_
- [x] fix zap retry in notifications not updating to paid when paid

**Manual Tests**

- [x] comment create: first wallet fails, second does not fail
- [x] zap (optimistic action): first wallet fails, second does not fail
- [x] zap: no wallet succeeds
- [x] zap: first wallet succeeds
- [x] zap: no wallet available
- [x] zap: pay with fee credits (shouldn't be affected at all, just to make sure it still works)
- [x] zap: anon zaps (shouldn't be affected at all, just to make sure it still works)
- [x] territory create (pessimistic action): first wallet fails, second does not fail
- [x] territory create: no wallet succeeds
- [x] territory create: first wallet succeeds
- [x] territory create: no wallet available
- [x] territory create: pay with fee credits

## Screenshots

how errors from one or multiple wallets will now be shown:

2024-11-26-233406_1920x1080_scrot

## Additional Context

I think the only reason we only allowed retries for optimistic actions was because retries weren't necessary yet for pessimistic actions.

We also should probably move all code from components/payment.js into wallets/payment.js but I didn't to not create more noisy changes than necessary.

## Checklist

tbd

tbd

tbd

no, not planned

### DIFF
--- a/api/paidAction/index.js
+++ b/api/paidAction/index.js
@@ -304,14 +304,6 @@ export async function retryPaidAction (actionType, args, incomingContext) {
     throw new Error(`retryPaidAction - must be logged in ${actionType}`)
   }
 
-  if (!action.paymentMethods.includes(PAID_ACTION_PAYMENT_METHODS.OPTIMISTIC)) {
-    throw new Error(`retryPaidAction - action does not support optimism ${actionType}`)
-  }
-
-  if (!action.retry) {
-    throw new Error(`retryPaidAction - action does not support retrying ${actionType}`)
-  }
-
   if (!failedInvoice) {
     throw new Error(`retryPaidAction - missing invoice ${actionType}`)
   }
@@ -319,7 +311,7 @@ export async function retryPaidAction (actionType, args, incomingContext) {
   const { msatsRequested, actionId, actionArgs } = failedInvoice
   const retryContext = {
     ...incomingContext,
-    optimistic: true,
+    optimistic: failedInvoice.actionOptimistic,
     me: await models.user.findUnique({ where: { id: me.id } }),
     cost: BigInt(msatsRequested),
     actionId
@@ -345,7 +337,7 @@ export async function retryPaidAction (actionType, args, incomingContext) {
     const invoice = await createDbInvoice(actionType, actionArgs, context)
 
     return {
-      result: await action.retry({ invoiceId: failedInvoice.id, newInvoiceId: invoice.id }, context),
+      result: await action.retry?.({ invoiceId: failedInvoice.id, newInvoiceId: invoice.id }, context),
       invoice,
       paymentMethod: 'OPTIMISTIC'
     }
--- a/api/paidAction/index.js
+++ b/api/paidAction/index.js
@@ -308,10 +308,10 @@ export async function retryPaidAction (actionType, args, incomingContext) {
     throw new Error(`retryPaidAction - missing invoice ${actionType}`)
   }
 
-  const { msatsRequested, actionId, actionArgs } = failedInvoice
+  const { msatsRequested, actionId, actionArgs, actionOptimistic } = failedInvoice
   const retryContext = {
     ...incomingContext,
-    optimistic: failedInvoice.actionOptimistic,
+    optimistic: actionOptimistic,
     me: await models.user.findUnique({ where: { id: me.id } }),
     cost: BigInt(msatsRequested),
     actionId
@@ -339,7 +339,7 @@ export async function retryPaidAction (actionType, args, incomingContext) {
     return {
       result: await action.retry?.({ invoiceId: failedInvoice.id, newInvoiceId: invoice.id }, context),
       invoice,
-      paymentMethod: 'OPTIMISTIC'
+      paymentMethod: actionOptimistic ? 'OPTIMISTIC' : 'PESSIMISTIC'
     }
   }, { isolationLevel: Prisma.TransactionIsolationLevel.ReadCommitted })
 }
--- a/api/paidAction/index.js
+++ b/api/paidAction/index.js
@@ -317,7 +317,29 @@ export async function retryPaidAction (actionType, args, incomingContext) {
     actionId
   }
 
-  const invoiceArgs = await createSNInvoice(actionType, actionArgs, retryContext)
+  let invoiceArgs
+  const invoiceForward = await models.invoiceForward.findUnique({
+    where: { invoiceId: failedInvoice.id },
+    include: {
+      wallet: true,
+      invoice: true,
+      withdrawl: true
+    }
+  })
+  if (invoiceForward) {
+    // TODO: receiver fallbacks
+    // use next receiver wallet if forward failed
+    const { userId } = invoiceForward.wallet
+    const { invoice: bolt11, wrappedInvoice: wrappedBolt11, wallet, maxFee } = await createWrappedInvoice(userId, {
+      msats: failedInvoice.msatsRequested,
+      feePercent: await action.getSybilFeePercent?.(actionArgs, retryContext),
+      description: await action.describe?.(actionArgs, retryContext),
+      expiry: INVOICE_EXPIRE_SECS
+    }, retryContext)
+    invoiceArgs = { bolt11, wrappedBolt11, wallet, maxFee }
+  } else {
+    invoiceArgs = await createSNInvoice(actionType, actionArgs, retryContext)
+  }
 
   return await models.$transaction(async tx => {
     const context = { ...retryContext, tx, invoiceArgs }
--- a/api/paidAction/index.js
+++ b/api/paidAction/index.js
@@ -326,9 +326,10 @@ export async function retryPaidAction (actionType, args, incomingContext) {
       withdrawl: true
     }
   })
-  if (invoiceForward) {
-    // TODO: receiver fallbacks
-    // use next receiver wallet if forward failed
+  // TODO: receiver fallbacks
+  // use next receiver wallet if forward failed (we currently immediately fallback to SN)
+  const failedForward = invoiceForward?.withdrawl && invoiceForward.withdrawl.actionState !== 'CONFIRMED'
+  if (invoiceForward && !failedForward) {
     const { userId } = invoiceForward.wallet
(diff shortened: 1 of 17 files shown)


wallets/payment.js
```diff
@@ -0,0 +1,156 @@
...
+  //
+  // we do this here anyway since we need the logger for each wallet.
+  // we ensure hooks are always called in the same order by sorting the wallets by name.
+  //
+  // we don't use the return value of useWallets here because it is empty on first render
+  // so using it would change the order of the hooks between renders.
+  //
+  // see https://react.dev/reference/rules/rules-of-hooks
+  const loggers = walletDefs
+    .sort((def1, def2) => def1.name.localeCompare(def2.name))
```
ekzyis, line 25: mhhh, I don't really need this `sort` since I can depend on `walletDefs` not changing it's order :thinking:

wallets/index.js
```diff
@@ -128,7 +128,7 @@ export function WalletsProvider ({ children }) {
       }
     }
 
-    // sort by priority, then add status field
+    // sort by priority
```
ekzyis, line 131: a change I forgot in #1646

api/paidAction/index.js
```diff
@@ -244,28 +244,20 @@ export async function retryPaidAction (actionType, args, incomingContext) {
...
   }
 
-  if (!action.paymentMethods.includes(PAID_ACTION_PAYMENT_METHODS.OPTIMISTIC)) {
-    throw new Error(`retryPaidAction - action does not support optimism ${actionType}`)
-  }
-
-  if (!action.retry) {
-    throw new Error(`retryPaidAction - action does not support retrying ${actionType}`)
-  }
-
```
ekzyis, line 254: So far, I think it's okay to remove these checks to allow retries for pessimistic actions but I will make more sure

components/use-paid-mutation.js
```diff
@@ -32,23 +33,36 @@ export function usePaidMutation (mutation,
...
-        // bail since qr code payment will also fail
-        // also bail if the payment took more than 1 second
-        // and cancel the invoice if it's not already canceled so it can be retried
-        invoiceHelper.cancel(invoice).catch(console.error)
+      walletError = null
+      if (err instanceof WalletError) {
+        walletError = err
+        // wallet payment error handling always creates a new invoice to retry
+        // unless no wallet was even able to attempt a payment
+        if (err.newInvoice) walletInvoice = err.newInvoice
```
ekzyis, line 47: this is maybe the weirdest part of this code: using an error to pass the new invoice to the calling context

api/paidAction/index.js
ekzyis: I see no problem with removing these. Retries for optimistic actions work like before and pessimistic actions support retries as expected now.

api/paidAction/index.js
```diff
@@ -244,22 +244,14 @@ export async function retryPaidAction (actionType, args, incomingContext) {
     throw new Error(`retryPaidAction - must be logged in ${actionType}`)
```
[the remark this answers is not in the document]
ekzyis: No, they have no attached wallets and can only pay QR codes, at least for now

components/use-paid-mutation.js
ekzyis: > Can't you extract the part that recreates the invoice or return it from waitForWalletPayment ?

I don't think so. Even if I extract it, I still need to pass the latest invoice to the calling context in some way. I can't use return since that would mean the wallet payment succeeded which isn't the case here.

I do return the latest invoice on success though, see the `waitForPayment` call for pessimistic actions below.

ekzyis:
Oh, I didn't test p2p zaps. Thanks, will check and fix!

components/use-paid-mutation.js
riccardobl: It might seem unimportant in the context this function is used, but these are things that build up technical debt, since now you have a function that leaks invoices, unless you remember that unexpectedly  the error contains a new resource you need to cancel if you don't need it, that's why passing around **new** data in an error field is generally bad practice, errors should report only what and why something failed.

But I think you can extract the logic and make a proper error object simply by not recreating the last invoice but just cancelling it and setting a `failedInvoice` field in the error. Then when you catch it in `waitForPayment` you can recreate it with `await invoiceHelper.retry(error.failedInvoice)`  only if it is needed.
 eg. if   `skipQr` is true, you won't even need to retry it.

Thoughts?

ekzyis: Oh, I had it like this before but it looked awkward to do something different on the last payment attempt error (not create new invoice) so I changed it to this with less code.

But the argument about "new data" is a good one, I'll change it back to only save the failed invoice in the error and the caller can retry if they want so. Thanks!!

ekzyis: Did this in 96f76a9.

I realized that I can filter for only the wallets that are actually able to pay so the logic if there is another wallet we can try is indeed basically just `if (i === wallets.length - 1)`. For some reason I didn't think the filter would work so I didn't even try and I think that's why I also didn't do it like this before.

This means the checks in `sendPayment` aren't really required anymore so maybe we should remove them :thinking:

ekzyis:
Need to change some more stuff like the flow when no wallets is available for 96f76a9. Will put in draft until this is ready for review again.

wallets/lnc/client.js
```diff
@@ -15,20 +13,12 @@ export async function testSendPayment (credentials, { logger }) {
...
-      const msg = err.message || err.toString?.()
-      if (msg.includes('invoice expired')) throw new InvoiceExpiredError(hash)
-      if (msg.includes('canceled')) throw new InvoiceCanceledError(hash)
-      throw err
-    }
+    const lnc = await getLNC(credentials, { logger })
+    const { paymentError, paymentPreimage: preimage } = await lnc.lnd.lightning.sendPaymentSync({ payment_request: bolt11 })
+    if (paymentError) throw new Error(paymentError)
+    if (!preimage) throw new Error('No preimage in response')
+    return preimage
```
ekzyis, line 21: I removed the `try/catch` here since we already have error handling in `waitForPayment` (which also checks for canceled or expired invoices) and in `sendPayment` which wrapps `wallet.def.sendPayment`.

I think this was just old code that we no longer need.

ekzyis:
> this implementation of retryPaidAction doesn't work for p2p actions, it will turn them to pessimistic actions that pay SN.

Fixed now. The code now retries the same receiver wallet if the forward didn't fail. If the forward failed, we immediately fallback to the SN node.

huumn:
Cool, I'm going to try to refactor out the array of `useWalletLogger()`, so please don't force push.

huumn:
Main thing I did in [b608fb6](https://github.com/stackernews/stacker.news/pull/1642/commits/b608fb6848ac071cffe06c2b7c115f955484aac8) was 

1. factor out the wallet-specific code from the generic "helper" logic
2. turn the wallet-specific code it into a hook that returns a "factory", i.e. factory produces a logger given a wallet

I'm not sure how generically we'll be able to apply this pattern but it works here.

wallets/index.js
```diff
@@ -218,34 +215,13 @@ export function useWallets () {
...
-      const message = err.message || err.toString?.()
-      logger.error(`payment failed: ${message}`, { bolt11 })
-      throw err
-    }
-  }, [wallet, logger])
-
-  if (!wallet) return null
-
-  return { ...wallet, sendPayment }
+export function useWalletsWithPayments () {
```
ekzyis, line 221: mhh, I would call this `useSendWallets` since this does not return wallets with the `sendPayment` function

components/use-paid-mutation.js
```diff
@@ -81,15 +98,33 @@ export function usePaidMutation (mutation,
...
+          }
           onPaid?.(client.cache, { data })
         }).catch(e => {
           console.error('usePaidMutation: failed to pay invoice', e)
+          if (e.invoice) {
+            // update the failed invoice for the Apollo cache update
+            data = {
+              [dataKey]: {
+                ...data[dataKey],
+                invoice
```
ekzyis, line 124: oh, this should be

```suggestion
                invoice: e.invoice
```

components/use-paid-mutation.js
```diff
@@ -99,14 +134,15 @@ export function usePaidMutation (mutation,
...
           if (!response.result) {
             // if the mutation didn't return any data, ie pessimistic, we need to fetch it
             const { data: { paidAction } } = await getPaidAction({ variables: { invoiceId: parseInt(invoice.id) } })
             // create new data object
             // ( hmac is only returned on invoice creation so we need to add it back to the data )
             data = {
-              [Object.keys(data)[0]]: {
+              [dataKey]: {
                 ...paidAction,
                 invoice: { ...paidAction.invoice, hmac: invoice.hmac }
```
ekzyis, line 147: this probably does no longer correctly set the hmac since `invoice` is no longer the original invoice

huumn:
Is this good to merge? I noticed you added another `todo`

It might be a consequence of my refactor, but because state is kept in memory, retries don't really work after navigation.

Also sorry about the renaming and `Error` refactors. It helped me understand what was going on to rename some stuff and encapsulate others.

wallets/index.js
huumn: oh good point

huumn:
I looked into the zap retry with fallback cache issue some. It's not clear why it's no longer working but these retry cache updates were the hardest part of my initial retry work. The item act retry and item retry are quite similar though - odd it's just affecting one of them.


The only thing I can thing of is that the retry mutation in useWalletPayment might be conflicting with the cache update

huumn (approved):
Feel free to merge into master if there aren't any issues remaining.

PR #1646, opened 2024-11-24
More wallet hooks: useWalletStatus & useWalletSupport

ekzyis:
## Description

I noticed this

> It's also going to be tempting to start making useWallets a monolith again, but that's part of what made the code so buggy. Meaning, it's going to be convenient to push all code into useWallets for anything that uses useWallets but it's going to be inconvenient to maintain.

-- https://github.com/stackernews/stacker.news/pull/1643#discussion_r1855255109

also applies to what I did in #1559. This PR removes that code from `useWallets` and moves it into their own wallet hooks.

Based on #1638, #1640

## Additional Context

`useWalletSupport` is a bit of a ridiculous "hook" but I think it's indeed better to have it separated so we can use it only where we need it (which is only inside the `WalletCard` component).

## Checklist

yes

- `7`. Wallet cards still show correct status and support.

n/a

no

### DIFF
--- a/components/wallet-card.js
+++ b/components/wallet-card.js
@@ -8,18 +8,13 @@ import DraggableIcon from '@/svgs/draggable.svg'
 import RecvIcon from '@/svgs/arrow-left-down-line.svg'
 import SendIcon from '@/svgs/arrow-right-up-line.svg'
 import { useWalletImage } from '@/components/wallet-image'
-
-const statusToClass = status => {
-  switch (status) {
-    case Status.Enabled: return styles.success
-    case Status.Disabled: return styles.disabled
-    case Status.Error: return styles.error
-    case Status.Warning: return styles.warning
-  }
-}
+import { useWalletStatus, statusToClass } from '@/components/wallet-status'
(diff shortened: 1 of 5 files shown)


components/wallet-card.js
```diff
@@ -8,18 +8,13 @@ import DraggableIcon from '@/svgs/draggable.svg'
...
-    case Status.Warning: return styles.warning
-  }
-}
+import { useWalletStatus, statusToClass } from '@/components/wallet-status'
+import { useWalletSupport } from '@/components/wallet-support'
 
 export default function WalletCard ({ wallet, draggable, onDragStart, onDragEnter, onDragEnd, onTouchStart, sourceIndex, targetIndex, index }) {
   const image = useWalletImage(wallet)
+  const status = useWalletStatus(wallet)
+  const support = useWalletSupport(wallet)
```
huumn, line 17: Oh boy that's like a breathe of fresh air.

ekzyis:
Keeping this as a draft until #1638 is merged and I rebased this again on master

Issue #879, opened 2024-02-25
YouTube "clip" embed


huumn:
We embed youtube videos but not clips which are becoming more common. example: https://stacker.news/items/438611

TheMhv:
This is harder than it seems.

The path /clip/<clipId> is not compatible with embed. To make it work as an embed, you must have 2 parameters that are not shared in this path, `videoId` (which is the default embed parameter) and `clipt` (some video time parameter, a hash that does not exist as a standard )

But there is a "solution", you can make a request directly to the link and get the embed link that is shown by the metadata tags `og:video:url`, `og:video:secure_url` or `twitter:player`.

The problem is that Youtube has a CORS policy that does not accept requests from other servers, so it is not possible to make a request directly through the code, the request would have to go through a reverse proxy server that bypasses cors

https://stackoverflow.com/questions/74697206/what-format-is-the-data-in-the-clipt-parameter
https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/5344

PR #1652, opened 2024-11-26
support youtube clip embedding

riccardobl:
Closes https://github.com/stackernews/stacker.news/issues/879

### DIFF
--- a/api/resolvers/embed.js
+++ b/api/resolvers/embed.js
@@ -1,0 +1,26 @@
+import jsdom from 'jsdom'
+
+export default {
+  Query: {
+    fetchEmbedMeta: async (parent, { source }, { models, me }) => {
+      const { hostname, pathname } = new URL(source)
+      if (hostname.endsWith('youtube.com') && pathname.includes('/clip')) {
+        const html = await fetch(source).then(res => res.text())
+        const dom = new jsdom.JSDOM(html)
+        const meta = dom.window.document.querySelector('meta[property="og:video:url"]')
+        if (meta) {
+          const content = meta.getAttribute('content')
+          const clipId = content.match(/clip=([^&]+)/)[1]
+          const videoId = content.match(/embed\/([^?]+)/)[1]
+          const clipt = content.match(/clipt=([^&]+)/)[1]
+          return {
+            clipId,
+            clipt,
+            videoId
+          }
+        }
+      }
+      return {}
+    }
+  }
+}
--- a/components/embed.js
+++ b/components/embed.js
@@ -5,6 +5,8 @@ import styles from './text.module.css'
 import { Button } from 'react-bootstrap'
 import { TwitterTweetEmbed } from 'react-twitter-embed'
 import YouTube from 'react-youtube'
+import { useLazyQuery } from '@apollo/client'
+import gql from 'graphql-tag'
 
 function TweetSkeleton ({ className }) {
   return (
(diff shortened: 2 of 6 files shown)


lib/url.js
```diff
@@ -144,6 +144,20 @@ export function parseEmbedUrl (href) {
...
+      if (pathname.includes('/clip')) {
+        const id = pathname.split('/').slice(-1).join()
+        return {
+          provider: 'youtube',
+          id: null,
+          meta: {
+            videoId: null,
+            clipId: id,
+            clipt: null,
+            metaSource: href
```
riccardobl, line 157: metaSource is used to specify an external source where to find metadata

components/embed.js
```diff
@@ -117,10 +119,32 @@ const SpotifyEmbed = function SpotifyEmbed ({ src, className }) {
...
+  const [meta, setMeta] = useState(initialMeta)
+
+  const [fetchMeta] = useLazyQuery(gql`  
+    query FetchEmbedMeta($source: String!) {
+      fetchEmbedMeta(source: $source)
+    }`)
+
+  useEffect(() => {
+    let abort
+    if (initialMeta.metaSource) {
```
riccardobl, line 135: if there is a metaSource url, use the fetchEmbedMeta to pull the extra metadata

api/resolvers/embed.js
```diff
@@ -0,0 +1,48 @@
...
+export default {
+  Query: {
+    fetchEmbedMeta: async (parent, { provider, args }, { models, me }) => {
+      const hash = crypto.createHash('sha256').update(JSON.stringify({ provider, args })).digest('hex')
+
+      let meta = (await models.embedMeta.findUnique({
+        where: { hash }
+      }))?.meta
+
+      if (meta?.updatedAt && (Date.now() - meta.updatedAt.getTime()) / 1000 > EMBED_META_CACHE_DURATION_S) {
```
riccardobl, line 15: Don't cache this forever, so that the system recovers at some point if youtube does breaking changes

ekzyis (commented):
Is there something wrong with `cachedFetcher` from `lib/fetch`?

I thought we discussed that we want to use the same caching implementation like we do for [price](https://github.com/stackernews/stacker.news/blob/7cb2aed9db0ce1b95ae67e33857b6ad2149f705e/api/resolvers/price.js), [blockheight](https://github.com/stackernews/stacker.news/blob/master/api/resolvers/blockHeight.js), [fees](https://github.com/stackernews/stacker.news/blob/master/api/resolvers/chainFee.js) etc.

Or wasn't this changed yet so this isn't actually ready for review yet?

riccardobl:
> Is there something wrong with `cachedFetcher` from `lib/fetch`?
> 
> I thought we discussed that we want to use the same caching implementation like we do for [price](https://github.com/stackernews/stacker.news/blob/7cb2aed9db0ce1b95ae67e33857b6ad2149f705e/api/resolvers/price.js), [blockheight](https://github.com/stackernews/stacker.news/blob/master/api/resolvers/blockHeight.js), [fees](https://github.com/stackernews/stacker.news/blob/master/api/resolvers/chainFee.js) etc.
> 
> Or wasn't this changed yet so this isn't actually ready for review yet?

cacheFetcher uses in-memory cache, it is good for price, blockheight and fees because they all are a single values that comes from a single source, but using it to store clips metadata can potentially make the cache grow too large. Using the database and its memory management capabilities makes more sense in this case, also because this cache can last for a very long time (even for ever if we want to), since these values should never change unless youtube updates its internals.

huumn:
I like that you created a table for embed meta!

Rather than doing a "lazy load" of embed meta, I think it might be better to do it as job in the worker like we do with images/video, so that the client doesn't need to be aware of the embed meta. That method has worked well for us. 

We can also add a trigger to denormalize the embed meta into its `"Item"` so that we don't have to join (or do an additional query) when items are queried.

(I'd prefer images/videos be normalized like your embed code and denormalize in a trigger like this, but that's another issue.)

Thoughts?

riccardobl:
I've moved the logic into a worker.
The system is extensible by registering a function of type `MetaFetcher` to a provider key in `lib/embedFetcher/MetaFetchers`.

worker/fetchEmbedMeta.js
```diff
@@ -0,0 +1,92 @@
...
+  const text = item.text || item.url
+  if (!text) return
+
+  const urls = findUniqueUrls(text) // extract every url from the text
+  if (!urls.length) return
+
+  const embeds = urls
+    .map(url => parseEmbedUrl(url)) // parse every unique url as a potential embed (initial url based deduplication)
+    .filter(embed => !!embed) // if it is not an embed, we can skip it
+    .reduce((acc, embed) => { // deduplicate embeds by id and provider (they might have different urls but be resolved to the same embed)
```
riccardobl, line 17: This is to avoid to fetch the same thing more than once.

> Why would someone want to embed the same thing more than once? 

Idk to abuse the system, or meme, either way, better to stop it here rather than firing multiple fetches

huumn (commented):
This will probably sit until after jan 3. 

We haven't had anyone share a youtube clip in months, so on that level this may not be worth it for youtube clips alone, but it'll be worth merging if the logic is also useful for other stuff. Hard to tell atm, but let's keep it open until we have more time for nice-to-haves.

PR #1653, opened 2024-11-26
feat: clustered ffmpeg transcoding

Soxasora:
## Description

Clustered FFmpeg transcoding for video uploads

Proof-of-concept with Node and Express acting as a master server that queue transcoding jobs to be picked up by its defined workers.
It enables transcoding for video uploads streamlining compatibility across devices.

## Screenshots
n/a

## Additional Context

Transcoding requires some time and it can add a noticeable overhead on video uploads. The result video is a slimmed, color corrected mp4 file.

note: this is **not** a well-tested feature and _nowhere near ready_ for production, I halted development on this to explore other solutions
note 2: this transcodes **every** video file to mp4, could also allow only video/quicktime mime type to act as a last-resort option

## Checklist

Doesn't interfere with video playback, it's a new step while uploading that only happens when the file is a video.

5.
Tried with 4 workers and 8 simultaneous uploads, enqueueing correctly the jobs until workers were available.
An 18 MB 720p mov file was correctly transcoded to 13 MB 720p mp4 file.

```
# .env.development
# ffmpeg
NEXT_PUBLIC_FFMPEG_URL=http://localhost:3002
FFMPEG_PORT=3002
FFMPEG_WORKERS=4
```


ekzyis:
This would be really cool, so I don't have to cry anymore when I can't watch videos uploaded from iShit devices

example: https://stacker.news/items/926344

ekzyis:
Another example: https://stacker.news/items/1093341

I see video player controls but it's just a still image:

2025-08-25-120258_1920x1080_scrot

I downloaded the raw file and my browser couldn't play it because it's Apple

```
$ file 105641
105641: ISO Media, Apple QuickTime movie, Apple QuickTime (.MOV/QT)
```

Soxasora:
I won't lie, I stopped this PR because I felt I wasn't ready enough to approach a solution like this. It's a common example of how you can get theory right but practice wrong.

If we want to have our own ffmpeg solution, then we need to have the best approach! For example, back then, I theorized a way to cut costs by having AWS Lambdas take care of multiple ffmpeg jobs, spawning instances only when needed and that's already a good start.

My main worry was that this clustered solution would leech resources from SN, but maybe I'm overestimating the impact.

Anyway, I want to resuscitate this and finally break free from the _QuickTime jail_.

PR #1654, opened 2024-11-26
use flexbox for wallet card header and make logos more consistent

riccardobl:
## Description

Change the wallet card to use flexbox to show the header icons.
Make logos size more consistent.

## Screenshots

Before
image

After 
image

## Additional Context

The flexbox header make  it is easier to use the entire space, eg. to put stuff on the left, used in https://github.com/stackernews/stacker.news/pull/1602


ekzyis (commented):
Thanks, looks great!

I only have three nitpicks:

1. formatting (see suggestions)
2. I am not sure if all these CSS properties are required. However, I think we already had some unnecessary CSS properties before so this is not a regression.

3. I think we also should do something for the logos in the configuration page since they can also get quite large:

2024-11-27-025915_1920x1080_scrot

For example, we could do something like this:

```diff
diff --git a/pages/settings/wallets/[wallet].js b/pages/settings/wallets/[wallet].js
index b3bcca62..98cc2bf5 100644
--- a/pages/settings/wallets/[wallet].js
+++ b/pages/settings/wallets/[wallet].js
@@ -19,6 +19,7 @@ import validateWallet from '@/wallets/validate'
 import { ValidationError } from 'yup'
 import { useFormikContext } from 'formik'
 import { useWalletImage } from '@/components/wallet-image'
+import styles from '@/styles/wallet.module.css'
 
 export const getServerSideProps = getGetServerSideProps({ authRequired: true })
 
@@ -72,7 +73,7 @@ export default function WalletSettings () {
   return (
     <CenterLayout>
       {image
-        ? <img {...image} className='pb-3 px-2 mw-100' />
+        ? <img {...image} className={styles.walletBanner} />
         : <h2 className='pb-2'>{wallet.def.card.title}</h2>}
       <h6 className='text-muted text-center pb-3'><Text>{wallet.def.card.subtitle}</Text></h6>
       <Form
diff --git a/styles/wallet.module.css b/styles/wallet.module.css
index 62d1da47..df6c34f0 100644
--- a/styles/wallet.module.css
+++ b/styles/wallet.module.css
@@ -49,6 +49,11 @@
   line-height: 1;
 }
 
+.walletBanner {
+  max-width: 256px;
+  max-height: 100px;
+  padding-bottom: 1rem;
+}
 
 .badge {
   color: var(--theme-grey) !important;
```

2024-11-27-025936_1920x1080_scrot

What do you think?

riccardobl:
> 2\. I am not sure if all these CSS properties are required. However, I think we already had some unnecessary CSS properties before so this is not a regression.

which properties do you find confusing?


> I think we also should do something for the logos in the configuration page since they can also get quite large:

done, i've changed your proposal with  `max-width: min(256px, 100vw)` so it never goes out of the screen edges

ekzyis:
> which properties do you find confusing?

Afaict, these CSS properties don't do anything. Not sure if they even did before:

```diff
diff --git a/styles/wallet.module.css b/styles/wallet.module.css
index cc56a26e..0bfee57a 100644
--- a/styles/wallet.module.css
+++ b/styles/wallet.module.css
@@ -22,29 +22,21 @@

 .indicators {
   display: flex;
-  align-items: center;
-  justify-content: center;
   column-gap: 0.2rem;
   margin-left: auto;
 }

 .cardMeta {
   width: 100%;
-  margin: 0;
   display: flex;
   padding: 10px;
-  line-height: 0;
-  align-items: center;
   position: absolute;
-  top: 0;
-  left: 0;
 }

 .walletLogo {
   max-width: 100%;
   max-height: 40%;
   margin: auto;
-  text-align: center;
   font-size: 1.5rem;
   line-height: 1;
 }
```

> done, i've changed your proposal with max-width: min(256px, 100vw) so it never goes out of the screen edges

Cool!

PR #1667, opened 2024-11-30
fix can't upload from iOS camera/mov files

### LINKED ISSUE #1568: Can't use "record video" as upload option on Safari iOS

### Describe the bug

If you click on Upload, you can upload an existing file or take a photo or video.

If you choose video, the resulting video can't be uploaded because it uses `.mov`.

### Screenshots

signal-2024-11-10-162656

### Steps To Reproduce

1. Click on "upload file"
2. Select "Take Photo or Video"
3. Record video and try to upload

### Expected behavior

We support MOV format

### Logs

_No response_

### Device information

- OS: iOS
- Browser: Safari

### Additional context

_No response_

Soxasora:
## Description

Allows H264 `video/quicktime` uploads.
This is a follow-up to the #1617 PR that fixed #1568 core problem (not being able to upload from iOS/Safari).

MOV files being containers for MPEG-4 should be supported on the majority of devices, and iOS recordings are encoded in H264 or HEVC.

**But Firefox doesn't support HEVC.[^1]**

### _So what do we do_
When a video is exported in Safari or PWA, it gets transcoded into H264 automatically by iOS eliminating (at the moment) the need for dedicated transcoding. 
Since there's no way to detect a video codec without an external library and Firefox before uploading throws an error if it doesn't support the video[^2] we could assume that if an error is generated and the file is video/quicktime then it's a codec error, throwing a custom toaster error that says:

> codec might not be supported, check video settings

## Screenshots

n/a will provide later

## Additional Context

n/a

## Checklist

**Are your changes backwards compatible? Please answer below:** Doesn't affect uploads of other files and content already uploaded. 

6. I tried on a plethora of Android devices and it worked without any hassle **IF H264 ENCODED**. After extensive tests I found out that HEVC is the sole reason why mov files are not generally supported, thankfully iOS transcodes it to H264 for us so they work great!
Needless to say that Apple devices and browsers are supported and also browsers such as Firefox, Chromium-based browsers.

[^1]: Firefox is rolling out HEVC support (on Windows) with the use of Windows Media Foundation, after paying for it on Windows Store.
[^2]: Throws NS_ERROR_DOM_MEDIA_METADATA_ERR

### DIFF
--- a/components/file-upload.js
+++ b/components/file-upload.js
@@ -24,6 +24,7 @@ export const FileUpload = forwardRef(({ children, className, onSelect, onUpload,
       : document.createElement('video')
 
     file = await removeExifData(file)
+    file = await adjustMov(file) // adjust mov files to mp4
 
     return new Promise((resolve, reject) => {
       async function onload () {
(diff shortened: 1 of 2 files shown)


Soxasora:
I updated this PR as iOS automatically transcodes mov files to H264 if HEVC, there's no need for renaming or changing mimetype (also was spotty on Firefox)

PR #1674, opened 2024-12-02
Fix missing invoice retry before showing QR code

ekzyis:
## Description

This adds a new property `retry` to `waitForQrPayment`. If this is set, the `waitForQrPayment` will first retry the invoice before showing it to the user.

To keep the code simple, invoices can now also be canceled without a hmac as long as it's your own invoice.

## Checklist

yes

tbd

n/a

no


ekzyis:
As discussed in our meeting, I will not continue with this PR since it's not a big problem that we're still trying to pay with attached wallets in the background while we show the QR code. Receiver fallbacks (#1494) are more important currently.

If I remember correctly, it was also mentioned that the UX is not clear that clicking on pending cancels the payment in the background just to show a new invoice in the QR code. If you then close the QR code again, the wallets no longer attempt the payment in the background but only when you click on `retry payment` again, at which point it goes back to `pending`.

I personally think these UI state transitions (`retry payment` -> `pending` -> `QR code` -> ...) can be a good UX. But because implementing the necessary cache updates to make this work well is kind of complicated/messy, we should rethink how we do these cache updates before we attempt any such thing.

However, there's an issue with this that already exists and this PR doesn't fix: `pending` does not necessarily mean the wallets are attempting to pay since if you refresh the page, any payments attempts no longer happen but it still shows `pending`.

Fixing this would probably be the better first step to improve the user-facing code around retries.

PR #1677, opened 2024-12-03
feat: carousel back action integrated into browser history

### LINKED ISSUE #1508: Embedded picture UX improvment

aegroto:
## Description

*This is a cleaner version of [this PR](https://github.com/stackernews/stacker.news/issues/1508) as that branch contained a lot of commits and was diverging from master.*

Carousel's modal now closes when navigating backward through the browser's history, going back to the post instead of returning to the main page. Fixes #1508.

## Screenshots

I have committed a version of the fix which works in case of multiple carousel shows as well:

https://github.com/user-attachments/assets/fa6c8908-54c5-4293-86f1-0e03021154cd

## Additional Context
I wasn't able to use  `router.push` as suggested in the original issue and had to stick to the native APIs, [which is documented by NextJS as an alternative to router methods as well](https://nextjs.org/docs/app/building-your-application/routing/linking-and-navigating#using-the-native-history-api).

The code overrides the `close` function of the carousel, triggering a 'back' navigation when the carousel is closed. It is still relying on native history APIs, unfortunately a router's shallow navigation immediately closes the carousel, presumably due to the listener I have highlighted in the previous comments.

**Before marking this PR ready for review I would like to know if it's fine for your to use window's native APIs, then maybe explore a solution which relies on router in the future.**

## Checklist

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:** 8, I tested the changes manually and with various content.

### DIFF
--- a/components/carousel.js
+++ b/components/carousel.js
@@ -5,6 +5,7 @@ import ArrowRight from '@/svgs/arrow-right-line.svg'
 import styles from './carousel.module.css'
 import { useShowModal } from './modal'
 import { Dropdown } from 'react-bootstrap'
+import { usePathname } from 'next/navigation'
 
 function useSwiping ({ moveLeft, moveRight }) {
   const [touchStartX, setTouchStartX] = useState(null)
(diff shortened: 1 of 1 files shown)


aegroto:
@huumn @ekzyis Sorry for the ping but I'd need a feedback on this PR once you've got some time, just to be aware if this solution relying on the direct APIs is fine for you.

huumn:
We are occupied with high priority issues. We will get to this when we have time.

Issue #1439, opened 2024-09-27
Multi-Factor Authentication


SatsAllDay:
**Is your feature request related to a problem? Please describe.**
Not a problem, but accounts would be more secure if they could opt-in to multi factor authentication. This is especially useful for stackers who either:
1. have a large balance
2. have attached wallets with large balances
3. own territories
4. have a well-known reputation

**Describe the solution you'd like**
I'd like to be able to configure MFA in the form of (at least) TOTP on my account. I think this could be used in the following ways:
1. Required on login after first factor is complete
2. Required when performing "high risk" actions like transferring a territory, or transacting over a certain (configurable?) threshold

I said ideally TOTP because MFA over SMS is bad, and email is only slightly better. It would be best to be able to use an open source authenticator app.

**Describe alternatives you've considered**
There's no real alternative that I can think of for things like territory ownership, but for balance concerns, you can link to an external wallet with budgets to help minimize risk.

**Additional context**
[Discussed here](https://stacker.news/items/702424/r/WeAreAllSatoshi?commentId=702887)

ekzyis:
TOTPs would be great! I also like passkeys.

PR #1680, opened 2024-12-04
Login 2fa (totp)

riccardobl:
## Description

Add optional TOTP 2fa for login.
Partially solves: https://github.com/stackernews/stacker.news/issues/1439

## Screenshots
Screenshot_20241204_155305
Screenshot_20241204_153539

## Additional Context

next-auth does not support 2fa natively. 
There are several ways to implement it, but i've come up with this custom solution that is tailored to sn.

The goals of this implementation are:
- don't make it feel like an hack
     - don't ask for TOTP before the login (eg. enter the totp code if you have one)
     - don't ask the user to login twice
- make it work for every present and future credential provider 
- use graphql apis like the rest of the site
- don't mess with next-auth internals
- make it extensible for other 2fa methods

To achieve this i've implemented 2fa as two middlewares:
-  pageGuard: that redirects the user to the 2fa prompt 
-  sessionGuard: that forces the session to 'anon'  until 2fa is resolved

Most of the logic is contained in `/lib/auth2fa`

As reference i've used the github 2fa flow.

## Checklist

yes

6

not yet

no

### DIFF
--- a/lib/auth2fa.js
+++ b/lib/auth2fa.js
@@ -1,0 +1,177 @@
+import { TOTP } from 'otpauth'
+import { NextResponse } from 'next/server'
+import { getToken, encode } from 'next-auth/jwt'
+
+/**
+ * Get totp provider
+ */
+function getTotp ({ label, secret } = {}) {
+  return new TOTP({
+    issuer: 'stacker.news',
+    label,
+    digits: 6,
+    period: 30,
+    secret
+  })
+}
+
+/**
+ * Check if the given token is valid within the window
+ * @param {Object} param0
+ * @param {string} param0.secret - the totp secret
+ * @param {string} param0.token - the totp token
+ * @returns {boolean} - true if the token is valid
+ */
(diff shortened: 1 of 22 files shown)


lib/auth2fa.js
```diff
@@ -0,0 +1,177 @@
...
+ * @param {args} param0
+ * @param {string} param0.label - the totp label (eg. usernames)
+ * @returns {Object} - the totp secret
+ * @returns {string} base32 - the base32 secret
+ * @returns {string} uri - the totp uri
+ */
+export function generateTotpSecret ({ label = 'stacker.news - login' }) {
+  const totp = getTotp({ label })
+  return {
+    base32: totp.secret.base32,
```
riccardobl, line 42: this one is the setup-key in google authenticator

huumn (commented):
I like the design goals of this and the code looks pretty straightforward. 

This will need very careful focused review and QA from both @ekzyis and I. At least on my part, I likely won't get to deep review until after jan 3.

PR #1679, opened 2024-12-04
Add routing node to test payment forwarding

ekzyis:
## Description

This PR attempts to add a routing node so we can test forward failures.

TODO:

- [ ] fix `lnd` can't find route to `sn_lnd` even though both are connected via `cln`

## Additional Context

See commit message of 0e0f9f0cd6d93c94d090022342fe729adcb3b2d4.

## Checklist

**Are your changes backwards compatible? Please answer below:**

tbd

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**

tbd

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:**

no

**Did you introduce any new environment variables? If so, call them out explicitly here:**

tbd

PR #1688, opened 2024-12-05
Receiver fallbacks

### LINKED ISSUE #1494: attached wallet fallbacks on failure in order of priority

ekzyis:
## Description

Receiver fallbacks are implemented by making `getInvoiceableWallets` aware of previously failed wrapped invoices. If the id of such a failed invoice is passed to it (`predecessorId`), it will skip any wallets for which a forward was already attempted.

For this, a column `predecessorId` was added to the `Invoice` table which points to the previously failed invoice.

With this link back to the "predecessor invoice", we can check which wallet was already tried and skip that wallet in [this recursive query](https://github.com/stackernews/stacker.news/pull/1688/files#diff-0c0ce28094ad375fbda380b3a0f54c1ed18075e887652acecce6b51359eb9cf8R118-R154).

Since we want to retry with the same sender wallet on receiver errors, the client is made aware of receiver errors via the existing invoice polling. If status is `FAILED_FORWARD` or the invoice is held + cancelled, the invoice polling will throw `WalletReceiverError` now. We need to check both since it's dependent on timing what the client will see first (`FAILED_FORWARD` transitions almost immediately to `FAILED`).

If there is no more receiver wallet to try, we fallback to an invoice that pays SN and credits the receiver with CCs. 

close #1494 ~based on #1679~ based on #1700

## TODOs

- [x] don't log failed forwards as sender errors
- [ ] make sure we avoid infinite retry loops if there ever is a bug?

## Tests

- [x] sender with 1 good wallet and no CCs pays receiver with 2 bad wallets: falls back to CCs
- [x] sender with 1 bad and 1 good wallet and no CCs pays receiver with 2 bad wallets: falls back to CCs
- [x] paying with CCs (even though both have wallets attached)
- [x] sender with 1 bad and 1 good wallet and no CCs pays receiver with 2 good wallets: p2p zap success to first good receiver wallet 

## Additional Context

- I am testing failed forward using this patch:

```diff
diff --git a/worker/paidAction.js b/worker/paidAction.js
index 70396ddb..9d711b0f 100644
--- a/worker/paidAction.js
+++ b/worker/paidAction.js
@@ -268,7 +268,7 @@ export async function paidActionForwarding ({ data: { invoiceId, ...args }, mode
     payViaPaymentRequest({
       lnd,
       request: bolt11,
-      max_fee_mtokens: String(maxFeeMsats),
+      max_fee_mtokens: String(0),
       pathfinding_timeout: LND_PATHFINDING_TIMEOUT_MS,
       max_timeout_height: maxTimeoutHeight
     }).catch(console.error)
```

- The properties `wallet`, `message`, `reason` and `invoice`  of the custom errors can probably be moved into the parent class but that should be refactored in a separate PR imo.  

- https://github.com/stackernews/stacker.news/pull/1688#discussion_r1878331597

## Checklist

yes, this only adds a nullable database column

`7`. See tests above and I looked very closely at the code for extended periods of time

n/a

no


wallets/server.js
```diff
@@ -110,18 +110,37 @@ export async function createWrappedInvoice (userId,
...
+          SELECT "Invoice"."id", "Invoice"."failedInvoiceId"
+          FROM "Invoice"
+          JOIN "Retries" ON "Invoice"."id" = "Retries"."failedInvoiceId"
+          WHERE "Invoice"."actionState" = 'RETRYING'
+        )
+        SELECT
+          "Wallet"."id"
+        FROM "Retries"
+        JOIN "InvoiceForward" ON "InvoiceForward"."invoiceId" = "Retries"."id"
+        JOIN "Wallet" ON "Wallet"."id" = "InvoiceForward"."walletId"
```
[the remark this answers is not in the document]
ekzyis: Using `WHERE "InvoiceForward"."withdrawlId" IS NOT NULL` now

riccardobl:
Won't this cause unwanted sats -> CC conversion? 
If all the receivers fail, i think it should try to pay with CC -> CC, not  LN -> CC

ekzyis:
> I think it should prioritize p2p if the receiver has attached wallets.

I think zaps should be paid in the interest of the sender by default. Therefore, without a setting where the sender can express their preference that depends on if the receiver can receive real sats, I think we should try CCs first.

I think it would also be confusing if we try CCs first for posts and replies but not for zaps.[^1] Preferring sats over CCs should be opt-in imo so this would need a setting which I think we can add later if there's enough demand from users. We shouldn't assume too much about user preferences and I think receiver fallbacks without this setting are valuable enough by themselves already.

So currently, I think it's not worth it to add the complexity of two different payment flows (one for SN fees, one for zaps[^2]) where one even now _also_ includes a sending dependency on receiving next to the already existing dependency of receiving on sending  (a receiver cannot avoid to get paid in CCs if the sender only has CCs). The dependency of the receiver on the sender is already tricky UX-wise (need to explain to them that their wallet isn't broken but what they receive depends on the sender etc) so I'd rather not make sending also depend on receiving unless there is enough user demand for such a thing.

[^2]: was this handled in your PR?

Thanks for flagging though, this wasn't on my mind! I'll look into which changes would be required in theory to accommodate this if/when we need it while I review my implementation.

I am also sorry if I missed some existing discussion between @huumn and you about this. I am pretty sure I did. Could be that I am also disagreeing with @huumn regarding this here :eyes: 

[^1]: I assume for posting fees it's clear that everybody will want to try CCs first and we won't need a setting?

> otherwise i think the system will end up becoming almost CC exclusive.

30% of each zap goes into rewards (which become sats) so the half life of a CC is almost exactly two zaps: 1 * 0.7 * 0.7 = 0.49 < 0.50. 

Therefore, I think how users zap each other doesn't matter much for the amount of CCs in the system, it matters more how many use attached wallets. If everybody has an attached wallet, CCs will cease to exist quickly enough.

huumn:
#1678 tries p2p first if the sender and receiver has an attached wallet. If the receiver wallet fails, then I think the sender should just do CCs (either those they already have, or jit purchasing them).

Either way, it shouldn't require a massive overhaul to change the priority of payment methods. Sender and receiver fallbacks should be pretty well encapsulated from payment priority.

wallets/server.js
```diff
@@ -110,18 +110,48 @@ export async function createWrappedInvoice (userId,
...
+          SELECT "Invoice"."id", "Invoice"."predecessorId"
+          FROM "Invoice"
+          JOIN "Retries" ON "Invoice"."id" = "Retries"."predecessorId"
+          WHERE "Invoice"."actionState" = 'RETRYING'
+        )
+        SELECT
+          "Wallet"."id"
+        FROM "Retries"
+        JOIN "InvoiceForward" ON "InvoiceForward"."invoiceId" = "Retries"."id"
+        JOIN "Wallet" ON "Wallet"."id" = "InvoiceForward"."walletId"
```
ekzyis, line 151: This join is useless, I can simply select `"InvoiceForward"."walletId"`

components/use-invoice.js
```diff
@@ -16,14 +16,21 @@ export default function useInvoice () {
...
       throw new InvoiceExpiredError(data.invoice)
     }
 
-    if (cancelled || actionError) {
+    const failed = cancelled || actionError
+
+    // failed forwards might already have been finalized (actionState === 'FAILED') so we also check for failed held payments
+    if (actionState === 'FAILED_FORWARD' || (failed && isHeld)) {
+      throw new WalletReceiverError(data.invoice)
+    }
```
ekzyis, line 31: It is unfortunate that invoice polling can confusingly also return a wallet error now but I think that's not too bad for the sake of keeping this simple.

Alternative would be to manually query the invoice on payment errors to check for failed forwards but I think the flow is easier to understand if the receiver error is thrown here.

wallets/payment.js
```diff
@@ -24,44 +26,72 @@ export function useWalletPayment () {
...
+          await invoiceHelper.cancel(latestInvoice)
+        }
 
-        // try next wallet if the payment failed because of the wallet
-        // and not because it expired or was canceled
-        if (err instanceof WalletError) {
-          aggregateError = new WalletAggregateError([aggregateError, err], latestInvoice)
-          continue
+        // only create a new invoice if we will try to pay with a wallet again
+        const retry = paymentError instanceof WalletReceiverError || i < wallets.length - 1
```
ekzyis, line 87: Mhh, in theory, I could remove `paymentError instanceof WalletReceiverError` since I already check for this above and then deduct 1 from `i` so `i < wallets.length - 1` would still be true on receiver errors even if it's the last sender wallet. But I kind of like to explicitly mention the case of receiver errors here :thinking:

ekzyis:
Converted to draft because this requires a rebase on #1700

ekzyis:
Okay, superficially tested again based on #1700 and worked as expected, this is ready for review again.

wallets/server.js
```diff
@@ -110,18 +110,47 @@ export async function createWrappedInvoice (userId,
...
+          SELECT "Invoice"."id", "Invoice"."predecessorId"
+          FROM "Invoice"
+          JOIN "Retries" ON "Invoice"."id" = "Retries"."predecessorId"
+          WHERE "Invoice"."actionState" = 'RETRYING'
+        )
+        SELECT
+          "InvoiceForward"."walletId"
+        FROM "Retries"
+        JOIN "InvoiceForward" ON "InvoiceForward"."invoiceId" = "Retries"."id"
+        WHERE "InvoiceForward"."withdrawlId" IS NOT NULL
```
ekzyis, line 151: Mhhh, I think `WHERE "Retries"."isHeld" = 't'` instead of `"InvoiceForward"."withdrawlId" IS NOT NULL` would also work and would not require a join :thinking:

huumn: we can check `actionError` to be sure

ekzyis: It is currently not set for forward failures but I can change that if you meant that

huumn: I have a fix I'm working on locally. I just need to make sure I didn't regress it.

huumn: I don't know if this changed or was always there, but fees are being deducted from p2p zap receives and shouldn't be

Screenshot 2024-12-10 at 1 45 58 PM

huumn: Then my notification adds the fee too

Screenshot 2024-12-10 at 1 47 49 PM

huumn (approved):
I fixed the sender error checking by returning `fowardStatus` which exposes the status of the withdrawal to the sender. 

Other than the logging issues (which I suspect are unrelated) this works well. I'm ***very*** happy with `predecessor` recursive query - it's reliable and simple.

components/use-invoice.js
```diff
@@ -16,20 +16,21 @@ export default function useInvoice () {
...
-      throw new InvoiceCanceledError(data.invoice, actionError)
+    const failed = cancelled || actionError
+
+    if (failed && (forwardStatus && forwardStatus !== 'CONFIRMED')) {
+      throw new WalletReceiverError(data.invoice)
     }
 
-    // write to cache if paid
-    if (actionState === 'PAID') {
-      client.writeQuery({ query: INVOICE, variables: { id }, data: { invoice: data.invoice } })
```
huumn, line 32: This was a stray optimization I made at one point. It's not necessary.

wallets/server.js
ekzyis: > but fees are being deducted from p2p zap receives and shouldn't be

Mhh, I thought I fixed that at some point. If the wallet created a 70 sat invoice, it should obviously receive 70 sats since senders pay network fees.

> Then my notification adds the fee too

This is also wrong, right? Imo, the notification should show how much you received and if you click on it, the fees are shown separately.

So for a 100 sat zap, it should show "70 sats were zapped directly to your attached wallet" and if you click into it, it will show "70 sats with 1 sat routing fee".

Does this make sense / do you agree?

ekzyis:
> I fixed the sender error checking by returning `fowardStatus` which exposes the status of the withdrawal to the sender.

Should we also use that in the recursive query so client and server definitely have the same view about receiver errors?

wallets/server.js
huumn: SN pays the fee from SN to the receiver. So it shouldn't show any fee.

huumn:
> Should we also use that in the recursive query so client and server definitely have the same view about receiver errors?

Up to you

PR #1697, opened 2024-12-10
Set defaults for env vars in docker-compose.yml

ekzyis:
## Description

This makes it possible to bypass `sndev` if we want to run some `docker-compose` calls natively. This is currently not possible because `docker-compose` doesn't load the environment file and throws errors like this:

```
WARN[0000] The "CPU_SHARES_LOW" variable is not set. Defaulting to a blank string.
error while interpolating services.mailhog.cpu_shares: failed to cast to expected type: strconv.ParseInt: parsing "": invalid syntax
```

_While writing this, I realized I could have used `--env-file` just like `sndev` does ... oh well, but maybe it's still good to have defaults in the file?_

## Additional Context

I (thought I) needed this to debug CLN container build, see https://github.com/stackernews/stacker.news/pull/1679#issuecomment-2530588844

## Checklist

yes, according to [this link](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) which I found [here](https://unix.stackexchange.com/a/389659), this shell parameter expansion is POSIX compliant

`8`. I can now run `docker-compose build` without it complaining about unset parameters.

n/a

no

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,7 @@ services:
       - db:/var/lib/postgresql/data
     labels:
       CONNECT: "localhost:5431"
-    cpu_shares: "${CPU_SHARES_IMPORTANT}"
+    cpu_shares: "${CPU_SHARES_IMPORTANT:-1024}"
   tor:
     build:
       context: ./docker/tor
@@ -49,7 +49,7 @@ services:
     restart: unless-stopped
     volumes:
       - tordata:/tordata/
-    cpu_shares: "${CPU_SHARES_LOW}"
+    cpu_shares: "${CPU_SHARES_LOW:-256}"
     env_file: *env_file
     healthcheck:
       <<: *healthcheck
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -61,8 +61,8 @@ services:
     build:
       context: ./
       args:
-        - UID=${CURRENT_UID}
-        - GID=${CURRENT_GID}
+        - UID=${CURRENT_UID:-1000}
+        - GID=${CURRENT_GID:-100}
     restart: unless-stopped
```
ekzyis, line 66: Used the return values of `id -u` and `id -g` on my machine with a single user

ekzyis:
Okay, no, I think this is a bad idea to attempt to copy the default values from .env.development into the docker compose file since it's either not consistent or I have to copy all of them.

I am missing a lot of them here since I only added the ones to get `docker-compose build --no-cache cln` to work. Closing.

Issue #1317, opened 2024-08-19
Bolt12 wallet receives


huumn:
- add lndk to sndev environment
- possibly fork payment lndk to get the payment hash of the payment request so it can be wrapped in a bolt11 for the sender
- this might require other modifications to receive wallets as they have expectation of a bolt11

m0wer:
Looks like this is already implemented on https://github.com/stackernews/stacker.news/pull/1727 very cool! Pending merge.

PR #1702, opened 2024-12-10
Bolt12 dev environment

riccardobl:
## Description

This solves the first point of https://github.com/stackernews/stacker.news/issues/1317 by adding an eclair node as receiver and lndk as sender.

## Additional Context

Before deciding to use Eclair, I tried various configurations with CLN as the receiver, connected directly to LND and also with a CLN or LND routing node in between. I couldn’t get it to work reliably with those setups. Eclair, however, worked pretty much out of the box after making a few changes to the bitcoind configuration.

## Testing
```
$ bash sndev cli eclair tipjarshowoffer
<OFFER>

$ bash sndev cli sn_lndk get-invoice <OFFER> 10000
<INVOICE>

$ bash sndev cli sn_lndk pay-invoice <INVOICE-HEX> 10000
```


huumn:
The changes I made:

- moved eclair, lndk, and cln to `wallets` docker compose profile which seemed more fitting (and downloads/builds for lndk and eclair are pretty slow)
- specified x86 platform for lndk ([no build support for arm on linux](https://github.com/lndk-org/lndk/blob/bce93885f5fc97f3ceb15dc470117e10061de018/Cargo.toml#L62))
- fixed lndk port conflict with router_lnd

PR #1709, opened 2024-12-11
Fix pending forwards considered paid by client

### LINKED ISSUE #1707: Client considers some possible invoice states of failed forwards as paid

### Describe the bug

Polling a wrapped invoice at the wrong time leads the client to believe the invoice was paid because `inv.satsReceived > 0` and the invoice polling code detected no error.

### Screenshots

https://github.com/user-attachments/assets/c7933cdb-0a29-45d2-b9f1-e6deb482088b

### Steps To Reproduce

1. p2p zap someone
2. get unlucky
3. see invoice was considered paid but the forward actually failed 

### Expected behavior

any invoice that is in some failed state is detected as failed and not as paid by the client

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

ekzyis:
## Description

Fix #1707 

This excludes pending forwards from ever being considered for something by the client.

I think this is a change that should really only affect the bug regarding failed forwards and doesn't cause any regression during other payments.

## Additional Context

Maybe the server should set a `pending` field in the invoice which the client can use to know that it shouldn't use this state to determine something about the invoice? I don't like how the client becomes more aware of individual transitions because it _might_ see them just so it can ignore them. 

## Checklist

yes

`7`.  Tested with failed forwards, successful forwards and no forwards. The bug is easy to trigger without these changes and this patch:

```diff
diff --git a/wallets/payment.js b/wallets/payment.js
index 33baa717..db8c9883 100644
--- a/wallets/payment.js
+++ b/wallets/payment.js
@@ -123,7 +123,7 @@ function invoiceController (inv, isInvoice) {
           clearInterval(interval)
           signal.removeEventListener('abort', abort)
         }
-      }, FAST_POLL_INTERVAL)
+      }, 100)

       const abort = () => {
         console.info(`invoice #${inv.id}: stopped waiting`)
diff --git a/worker/paidAction.js b/worker/paidAction.js
index 70396ddb..9d711b0f 100644
--- a/worker/paidAction.js
+++ b/worker/paidAction.js
@@ -268,7 +268,7 @@ export async function paidActionForwarding ({ data: { invoiceId, ...args }, mode
     payViaPaymentRequest({
       lnd,
       request: bolt11,
-      max_fee_mtokens: String(maxFeeMsats),
+      max_fee_mtokens: String(0),
       pathfinding_timeout: LND_PATHFINDING_TIMEOUT_MS,
       max_timeout_height: maxTimeoutHeight
     }).catch(console.error)
```

n/a

no

### DIFF
--- a/components/use-invoice.js
+++ b/components/use-invoice.js
@@ -4,8 +4,6 @@ import { InvoiceCanceledError, InvoiceExpiredError, WalletReceiverError } from '
 import { RETRY_PAID_ACTION } from '@/fragments/paidAction'
 import { INVOICE, CANCEL_INVOICE } from '@/fragments/wallet'
 
-const PENDING_FORWARD_STATES = ['PENDING_HELD', 'FORWARDING']
-
 export default function useInvoice () {
   const client = useApolloClient()
   const [retryPaidAction] = useMutation(RETRY_PAID_ACTION)
@@ -18,7 +16,7 @@ export default function useInvoice () {
       throw error
     }
 
-    const { cancelled, cancelledAt, actionState, actionError, expiresAt, forwardStatus } = data.invoice
+    const { cancelled, cancelledAt, actionError, expiresAt, forwardStatus } = data.invoice
 
     const expired = cancelledAt && new Date(expiresAt) < new Date(cancelledAt)
     if (expired) {
@@ -35,12 +33,7 @@ export default function useInvoice () {
       throw new InvoiceCanceledError(data.invoice, actionError)
     }
 
-    // never let check pass if a forward is pending
-    // see https://github.com/stackernews/stacker.news/issues/1707
-    const pendingForward = PENDING_FORWARD_STATES.includes(actionState)
-    const check = that(data.invoice) && !pendingForward
-
-    return { invoice: data.invoice, check }
+    return { invoice: data.invoice, check: that(data.invoice) }
   }, [client])
 
   const cancel = useCallback(async ({ hash, hmac }) => {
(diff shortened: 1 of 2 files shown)


components/use-invoice.js
```diff
@@ -16,24 +18,29 @@ export default function useInvoice () {
...
     const expired = cancelledAt && new Date(expiresAt) < new Date(cancelledAt)
     if (expired) {
       throw new InvoiceExpiredError(data.invoice)
     }
 
-    const failed = cancelled || actionError
-
-    if (failed && (forwardStatus && forwardStatus !== 'CONFIRMED')) {
+    const failedForward = forwardStatus && forwardStatus !== 'CONFIRMED'
+    if (failedForward) {
```
ekzyis, line 29: Removing `failed` from the condition for failed forwards makes this resolve a little faster since the withdrawal can already have failed before the invoice itself is marked as failed, see video in https://github.com/stackernews/stacker.news/issues/1707#issuecomment-2535829482.

huumn:
Why not just change `waitFor = i => actionState === 'PAID'` in the case of send wallet being present?

ekzyis:
I thought it's okay to assume that `useInvoice` is only used in the case of send wallets so we can hide the complexity in there and that's also why I thought that is reasonable limiting of `waitFor`.

> ... `waitFor` should maybe be refactored somehow though. It's not clear where the function is coming from and most of the time we're using a default param from somewhere else in the code base.

I agree, good time to also refactor `useInvoice` as you mentioned in the past. But I would rather fix this without a refactor first so this doesn't block a release.

I'll fix this in `useAct` the way you described it. I'll keep `waitFor` in `useQrPayment` as-is since it doesn't affect this bug but is related to #1712.

PR #1714, opened 2024-12-11
Don't expose p2p zap receivers to routing fees

### LINKED ISSUE #1704: On p2p zaps, receivers are exposed to routing fees they aren't paying
See: https://github.com/stackernews/stacker.news/pull/1688#discussion_r1878731563

ekzyis:
## Description

Fix #1704 

## Checklist

yes

`8`. Tested successful forwards and autowithdrawals. Expected amounts are shown in notifications and wallet logs. 

n/a

no

### DIFF
--- a/worker/paidAction.js
+++ b/worker/paidAction.js
@@ -316,13 +316,11 @@ export async function paidActionForwarded ({ data: { invoiceId, withdrawal, ...a
   }, { models, lnd, boss })
 
   if (transitionedInvoice) {
-    const { bolt11, msatsPaid, msatsFeePaid } = transitionedInvoice.invoiceForward.withdrawl
-    // the amount we paid includes the fee so we need to subtract it to get the amount received
-    const received = Number(msatsPaid) - Number(msatsFeePaid)
+    const { bolt11, msatsPaid } = transitionedInvoice.invoiceForward.withdrawl
 
     const logger = walletLogger({ wallet: transitionedInvoice.invoiceForward.wallet, models })
     logger.ok(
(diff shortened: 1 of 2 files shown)


worker/paidAction.js
```diff
@@ -316,13 +316,11 @@ export async function paidActionForwarded ({ data: { invoiceId, withdrawal, ...a
   }, { models, lnd, boss })
 
   if (transitionedInvoice) {
-    const { bolt11, msatsPaid, msatsFeePaid } = transitionedInvoice.invoiceForward.withdrawl
-    // the amount we paid includes the fee so we need to subtract it to get the amount received
-    const received = Number(msatsPaid) - Number(msatsFeePaid)
+    const { bolt11, msatsPaid } = transitionedInvoice.invoiceForward.withdrawl
```
huumn, line 319: Crap. I think the problem here is that I broke the assumption with p2p stuff that `msatsPaid = msatsPaying + fee = payment.mtokens`. I think we'll want to fix that but I don't recall if we need to change stuff elsewhere.

huumn (commented):
It looks like I broke a convention we were using.

worker/paidAction.js
huumn: It looks like all the withdrawal code no longer includes the `msatsFeePaid` in the `msatsPaid`. I guess we can update any violations of this *new* convention.

huumn: Actually looks like this is a false alarm. Must've always been this way and this was wrong somehow.

PR #1724, opened 2024-12-14
Remove unused wallet context args

ekzyis:
## Description

We were passing `models` and `me` into the wallet test functions. I think they were used in the past for logging but we no longer need them. 

I also removed `logger` from the NWC functions since they don't use it.

## Additional Context

Maybe I am overdoing it with small PRs but it's nice to get small changes out of the way so I can focus on the relevant stuff in #1558.

## Checklist

yes

`5`. I looked at all `testSendPayment` and `testCreateInvoice` function signatures and saw none of them use these arguments.

Regarding removal of `logger`: removing an unused argument doesn't break any function calls in JS since you can always call a function with as many arguments as you want, any arguments that aren't defined as parameters are simply not used.

n/a

no

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -65,7 +65,7 @@ function injectResolvers (resolvers) {
         wallet,
         testCreateInvoice:
           walletDef.testCreateInvoice && validateLightning && canReceive({ def: walletDef, config: data })
-            ? (data) => walletDef.testCreateInvoice(data, { logger, me, models })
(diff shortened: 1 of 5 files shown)


huumn:
> Maybe I am overdoing it with small PRs but it's nice to get small changes out of the way so I can focus on the relevant stuff in https://github.com/stackernews/stacker.news/issues/1558.

You aren't. It's easier to review one "subject" at a time.

Issue #1598, opened 2024-11-17
Make wallet log context normal


huumn:
Currently we store an anything-goes JSON object as the wallet log context. We shouldn't:

1. the context object is undocumented (ie not even documented in code in one place)
2. migrating the context object over time (which is inevitable) is going to be messy
3. its keys need to iterated over/filtered in JS to tell if the wallet log has utility in whatever function is using wallet logs
4. context keys can't be properly indexed in the database without making a bigger mess
5. given the above, it's not worth developing views based on the wallet log's relationship to any data stored in context (e.g. viewing all the wallet logs related to an invoice payment)
6. people are going to store random denormalized data in the context object over time because its the path of least resistance for them and pushes the burden of their decisions into the future 

Keeping vanity data in JSON blobs is fine, but we shouldn't be storing important data this way. The DB is the root of sanity in our architecture. Storing normalized data in it means we can access data declaratively and can *mostly* liberate ourselves from typechecking most of our code and writing tests for what/how data is being stored.

ekzyis:
Something to consider here is that we can only enforce relations for the receive logs since the send logs are stored on the client.

huumn:
Oh good point. I'm imagining we kind of loosely enforce the same data shape on the client as we do on the server like we do with configs.

ekzyis:
This was fixed by #1826

PR #1727, opened 2024-12-14
Bolt12 wallet attachment

riccardobl:


lib/bolt11-tags.js
```diff
@@ -0,0 +1,10 @@
+import { decode } from 'bolt11'
+
+export function isBolt11 (request) {
+  return request.startsWith('lnbc') || request.startsWith('lntb') || request.startsWith('lntbs') || request.startsWith('lnbcrt')
+}
+
+export function bolt11Tags (bolt11) {
+  if (!isBolt11(bolt11)) throw new Error('not a bolt11 invoice')
+  return decode(bolt11).tagsObject
+}
```
riccardobl, line 10: i've renamed `@/lib/bolt11` to `@/lib/bolt11-tags`

lib/bolt12-info.js
```diff
@@ -0,0 +1,33 @@
...
+
+const TYPE_DESCRIPTION = 10n
+const TYPE_PAYER_NOTE = 89n
+const TYPE_PAYMENT_HASH = 168n
+
+export function isBolt12 (invoice) {
+  return invoice.startsWith('lni1') || invoice.startsWith('lno1')
+}
+
+export function bolt12Info (bolt12) {
```
riccardobl, line 12: This is the bolt12 equivalent of the bolt11Tags helper.
Similar to the bolt11 counterpart, i've used a light parser just for the few tags we need in the ui, since the lndk parser needs access to the grpc endpoint.

I couldn't find an established pure js bolt12 parser, so i wrote a small tlv parser (`@/lib/tlv`) following the specs.

lib/lndk.js
```diff
@@ -0,0 +1,249 @@
...
+  // convert from lndk response to ln-service parsePaymentRequest output layout
+  let minCltvDelta
+  for (const path of invoice_contents.payment_paths) {
+    const info = path.blinded_pay_info
+    if (minCltvDelta === undefined || info.cltv_expiry_delta < minCltvDelta) {
+      minCltvDelta = info.cltv_expiry_delta
+    }
+  }
+
+  const out = {
```
riccardobl, line 118: This function converts the output from lndk to an object with a layout compatible with the output from ln-service.

I've followed the [doc](https://github.com/alexbosworth/ln-service/blob/158597a28c3b392b98a175801ad56ade535b3195/README.md?plain=1#L939) so it should match the output very closely, but honestly i don't like the fact that ln-service returns  so many values that mean the same thing with different types (eg, tokens, mtokens, safe_tokens...), i've implemented this conversion because it makes our code simpler with fewer changes, since we can handle both bolt11 and bolt12 invoices in the same way, but imo we should provide our own abstraction at some point, using only bigint msats and things that make more sense for our codebase.

lib/lndk.js
```diff
@@ -0,0 +1,249 @@
...
+  return new Promise((resolve, reject) => {
+    lndk.GetInvoice({
+      offer,
+      // expects msats https://github.com/lndk-org/lndk/blob/bce93885f5fc97f3ceb15dc470117e10061de018/src/lndk_offers.rs#L182
+      amount: toPositiveNumber(msats),
+      payer_note: description,
+      response_invoice_timeout: timeout
+    }, async (error, response) => {
+      if (error) return reject(error)
+      const bech32invoice = 'lni1' + bech32b12.encode(Buffer.from(response.invoice_hex_str, 'hex'))
```
riccardobl, line 163: Bolt12 invoices are not supposed to be presented to the user, so the spec doesn't define an human readable representation, instead it is left to apps and nodes to handle the representation how they see more fit. 
LNDK uses an hex string,  while CLN use bech32.
For consistency with bolt11, i've decided to convert them to bech32 like in CLN.

lib/lndk.js
huumn: I agree, ln-service tries to make the grpc interface nicer, but it isn't as well documented as the grpc interface so it ends up just being confusing most of the time.

lib/bolt11-tags.js
ekzyis: Why not put everything bolt11 related into lib/bolt11.js? Same for bolt12 stuff?

api/lnd/index.js
```diff
@@ -10,6 +11,11 @@ const lnd = global.lnd || authenticatedLndGrpc({
   macaroon: process.env.LND_MACAROON,
   socket: process.env.LND_SOCKET
 }).lnd
+installLNDK(lnd, {
```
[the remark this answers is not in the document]
ekzyis: You can then also remove all these checks:

```js
const lndk = lnd?.lndk
if (!lndk) throw new Error('lndk not installed, please use installLNDK')
```

lib/bolt11-tags.js
riccardobl: because ln-service uses 'fs' to load protobuf files  and it breaks webpack builds for the browser

api/resolvers/wallet.js
```diff
@@ -732,8 +738,8 @@ export const walletLogger = ({ wallet, models }) => {
   const log = (level) => async (message, context = {}) => {
     try {
       if (context?.bolt11) {
-        // automatically populate context from bolt11 to avoid duplicating this code
-        const decoded = await parsePaymentRequest({ request: context.bolt11 })
+        // automatically populate context from invoice to avoid duplicating this code
+        const decoded = await parseInvoice({ request: context.bolt11, lnd })
```
ekzyis, line 742: Mhh, I think ideally this would store the bolt12 offer as `bolt12` in the context instead of the bolt12 invoice as `bolt11`.

But I am not sure how easy it is to get the offer back from the invoice? But I think we can at least save it as `bolt12` here.

_I really have to do #1598 soon :eyes:_

wallets/server.js
```diff
@@ -45,31 +68,20 @@ export async function createInvoice (userId, { msats, description, descriptionHa
...
-        if (BigInt(msats) - BigInt(bolt11.mtokens) >= 1000n) {
-          throw new Error('invoice invalid: amount too small')
-        }
-
-        logger.warn('wallet does not support msats')
+      } else if (isBolt12Offer(invoice)) {
+        throw new Error('the wallet returned a bolt12 offer, but an invoice was expected')
       }
 
+      checkInvoice(invoice, { msats }, { lnd, logger })
```
ekzyis, line 84: `await` missing :eyes:

wallets/server.js
```diff
@@ -6,32 +6,55 @@ import * as lnbits from '@/wallets/lnbits/server'
...
+export default [lnd, cln, lnAddr, lnbits, nwc, phoenixd, blink, lnc, webln, bolt12]
 
 const MAX_PENDING_INVOICES_PER_WALLET = 25
 
-export async function createInvoice (userId, { msats, description, descriptionHash, expiry = 360 }, { predecessorId, models }) {
+async function checkInvoice (invoice, { msats }, { lnd, logger }) {
+  const parsedInvoice = await parseInvoice({ lnd, request: invoice })
+  logger.info(`created invoice for ${formatSats(msatsToSats(parsedInvoice.mtokens))}`, {
+    bolt11: invoice
+  })
```
ekzyis, line 33: I think this logger call should be in `createInvoice` immediately after `walletCreateInvoice` like it was before

wallets/server.js
```diff
@@ -6,32 +6,55 @@ import * as lnbits from '@/wallets/lnbits/server'
...
+
+export async function createInvoice (userId, { msats, description, descriptionHash, expiry = 360, supportBolt12 = true }, { predecessorId, models, lnd }) {
   // get the wallets in order of priority
   const wallets = await getInvoiceableWallets(userId, { predecessorId, models })
 
   msats = toPositiveNumber(msats)
 
   for (const { def, wallet } of wallets) {
     const logger = walletLogger({ wallet, models })
+    if (def.isBolt12OnlyWallet && !supportBolt12) continue
```
ekzyis, line 57: Do you still need this? You also check below for `supportBolt12` after the invoice was fetched

lib/lndk.js
```diff
@@ -0,0 +1,249 @@
...
+        return
+      }
+
+      resolve({
+        routingFeeMsat: toPositiveNumber(res.routing_fee_msat),
+        timeLockDelay: toPositiveNumber(res.time_lock_delay)
+      })
+    })
+  })
+}
```
ekzyis, line 249: I think this is the same as `estimateRouteFee` for bolt11 except it needs to parse the bolt12 request before. This should afterwards call `estimateRouteFee` imo.

worker/index.js
```diff
@@ -73,6 +74,11 @@ async function work () {
     macaroon: process.env.LND_MACAROON,
     socket: process.env.LND_SOCKET
   })
+  installLNDK(lnd, {
+    cert: process.env.LNDK_CERT,
+    macaroon: process.env.LNDK_MACAROON,
+    socket: process.env.LNDK_SOCKET
+  })
```
ekzyis, line 81: see other comment

ekzyis (commented):
Tested normal and automated withdrawals and they work well!

But I noticed the followings things I mentioned in comments that I'd like to see discussed before approval:

1. lib/ was meant to be files that the server and client can import. Can we use dynamic imports so we don't need to split bolt files by server and client and can have just one file per bolt?

2. how did you test `max_fee`?

3. duplicate checks via `if` and schema

4. function signature of `parseBolt12Request`

5. error handling via async callback 

6. add similar function for bolt tags like for `payInvoice`, `parseInvoice`, `estimateFees`?

7. Showing BOLT12 invoices to the user: I think ideally they would always only see the BOLT12 offer. That's also what other wallets do. Is that possible in this PR or should we keep it for another PR?

7. mutating the `lnd` object instead of passing `lndk` separately in context

9. LND custom options

I read in the LNDK docs that we need to [run LND with custom options](https://github.com/lndk-org/lndk?tab=readme-ov-file#running-lnd):

```
[protocol]
protocol.custom-message=513
protocol.custom-nodeann=39
protocol.custom-init=39
```

You've added these options in #1702.

Afaict, this means @huumn needs to configure LND differently in prod. You should have mentioned this since I think this wouldn't have worked as-is in prod. :eyes: 

10. other small stuff (see comments)

docker/lndk/Dockerfile
```diff
@@ -1,16 +1,18 @@
...
+RUN useradd -u 1000 -m lndk
 RUN mkdir -p /home/lndk/.lndk 
 COPY ["./tls-*", "/home/lndk/.lndk"]
 RUN chown 1000:1000 -Rvf /home/lndk/.lndk && \
     chmod 644 /home/lndk/.lndk/tls-cert.pem && \
     chmod 600 /home/lndk/.lndk/tls-key.pem
 
 USER lndk
-RUN curl --proto '=https' --tlsv1.2 -LsSf https://github.com/lndk-org/lndk/releases/download/v0.2.0/lndk-installer.sh | sh
+RUN curl --proto '=https' --tlsv1.2 -LsSf $INSTALLER_DOWNLOAD_URL/lndk-installer.sh | sh
```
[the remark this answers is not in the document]
riccardobl: this pull my lndk fork that supports max_fee, it needs to be an env variable because it is used internally by the lndk-installer script

lib/lndk.js
```diff
@@ -0,0 +1,249 @@
...
+  } = invoice_contents
+
+  // convert from lndk response to ln-service parsePaymentRequest output layout
+  let minCltvDelta
+  for (const path of invoice_contents.payment_paths) {
+    const info = path.blinded_pay_info
+    if (minCltvDelta === undefined || info.cltv_expiry_delta < minCltvDelta) {
+      minCltvDelta = info.cltv_expiry_delta
+    }
+  }
```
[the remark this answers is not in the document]
riccardobl: original code is On, proposed code is  O(n log n)

wallets/server.js
riccardobl: yes,  this early check is to exclude bolt12-only wallets when WE cannot use them, logging an error for a perfectly fine wallet just because someone is asking an invoice via lnurlp, would be confusing for the user

api/resolvers/wallet.js
riccardobl: i don't think the offer is available as part of the [invoice tlv](https://github.com/lightning/bolts/blob/master/12-offer-encoding.md#tlv-fields-for-invoice_request). 
Can we do all the cosmetic changes from bolt11->bolt12 on a separated pr? This one is already big, propagating all the naming would add even more changes. Unless you mean something else

lib/lndk.js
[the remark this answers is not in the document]
ekzyis: You should not optimize for performance if n is really small. But okay, we can keep it like this, I understand now what this is trying to do. It's not that big of a deal, it just took me a few seconds.

lib/bolt11-tags.js
riccardobl: yes, it is possible, but last time dynamic imports caused a lot of issue in production.
Are they supposed to work reliably after your patch @huumn ?

docker/lndk/Dockerfile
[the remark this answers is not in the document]
ekzyis: Oh, I missed the `stackernews/lndk`. Makes sense now!

lib/lndk.js
```diff
@@ -0,0 +1,249 @@
...
+  if (!lndk) throw new Error('lndk not installed, please use installLNDK')
+
+  return new Promise((resolve, reject) => {
+    lndk.GetInvoice({
+      offer,
+      // expects msats https://github.com/lndk-org/lndk/blob/bce93885f5fc97f3ceb15dc470117e10061de018/src/lndk_offers.rs#L182
+      amount: toPositiveNumber(msats),
+      payer_note: description,
+      response_invoice_timeout: timeout
+    }, async (error, response) => {
```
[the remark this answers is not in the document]
riccardobl: It should be fine if just wrap the entire call back in a try-catch

api/resolvers/wallet.js
ekzyis: > Can we do all the cosmetic changes from bolt11->bolt12 on a separated pr? This one is already big, propagating all the naming would add even more changes. Unless you mean something else

I just meant the user-facing stuff (so not just cosmetic imo), not renaming variables but yeah, we can do it in a separate PR. Just wanted to flag this.

lib/lndk.js
```diff
@@ -0,0 +1,249 @@
...
+      if (error) return reject(error)
+      const bech32invoice = 'lni1' + bech32b12.encode(Buffer.from(response.invoice_hex_str, 'hex'))
+
+      // sanity check
+      const parsedInvoice = await parseBolt12Request({ lnd, request: bech32invoice })
```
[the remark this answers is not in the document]
riccardobl: i can remove the option to pass an hex string here, if you don't like supporting both representations. But since this is not just adding or removing a prefix, but doing the bech32 encoding, i'd like to keep the encoding and decoding thing here, to make it a part of the sanity check.

lib/lndk.js
```diff
@@ -0,0 +1,249 @@
+import { msatsToSats, satsToMsats, toPositiveNumber } from '@/lib/format'
+import { loadPackageDefinition } from '@grpc/grpc-js'
+import LNDK_RPC_PROTO from '@/lib/lndkrpc-proto'
+import protobuf from 'protobufjs'
+import grpcCredentials from 'lightning/lnd_grpc/grpc_credentials'
+import { defaultSocket, grpcSslCipherSuites } from 'lightning/grpc/index'
```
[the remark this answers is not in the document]
riccardobl: grpcSslCipherSuites yes, defaultSocket no,  removing it

lib/lndk.js
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  const client = new OffersService(lndSocket, credentials, params)
+  lnd.lndk = client
+}
+
+const featureBitMap = {
+  0: { bit: 0, type: 'DATALOSS_PROTECT_REQ', is_required: true },
+  1: { bit: 1, type: 'DATALOSS_PROTECT_OPT', is_required: false },
+  3: { bit: 3, type: 'INITIAL_ROUTING_SYNC', is_required: true },
```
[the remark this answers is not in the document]
riccardobl: i've changed the feature bit map to check for even bits

lib/lndkrpc-proto.js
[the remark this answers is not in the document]
riccardobl: It's the main branch here: https://github.com/stackernews/lndk , i'll add a note

For testing i just changed it to check for > instead >= so that it would trigger with 0 fees https://github.com/riccardobl/lndk/commit/c1f95eaa61fac7d07436c99f41f9db7dd09b463e

lib/lndk.js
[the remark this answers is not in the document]
riccardobl: This is supposed to be a wrapper around lndk grpcs.
I think you are referring to the conversion between the lndk decodeInvoice output and the ln-service-like object, you are right it is better to move it to bolt12.js

api/lnd/index.js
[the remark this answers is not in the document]
riccardobl: Since the Tor patch, we’ve been creating this in /lib/lnd.js, but I understand your concerns.

I am not a fan of the proposed alternative of passing it through context, for the following reasons:
- it adds yet another field that we need to remember to pass through all the functions
- it depends on lnd being also passed for fee estimation
- it is error-prone and defeats the purpose of having a unified Bolt interface
- lndk doesn't implement the full bolt12 standard and there are plans to implement bolt12 natively in lnd (see [1]). So, if we make all the functions depends already only on lnd, we won't have to refactor anything outside of the bolt12 interface to switch to the native lnd implementation.

<details>
  <summary>[1]</summary>
https://github.com/lightningnetwork/lnd/issues/5594
https://github.com/lightningnetwork/lnd/pull/9369

image
</details>
 
So, to address your concerns, I’ve changed the code to use a WeakMap to maintain lndk instances until their corresponding lnd is GCed. This should address your main concern that, as far as i understand, is that in the remote eventuality that the lnd client decides to export a service called lndk, the previous code could end up shadowing it.

lib/bolt11-tags.js
huumn: I don't have the full context but ...

I don't recall them causing issues in prod, but we should only use dynamic imports for bundle size reasons, or if the file we're including is determined dynamically. Using it so we can put code in `lib` that doesn't need to be there, seems silly to me.

If something is server specific, I recommend putting it in `api` and possibly creating an `api/lib` for future such code too.

wallets/bolt12/index.js
```diff
@@ -0,0 +1,25 @@
...
+    clear: true,
+    serverOnly: true,
+    validate: string()
+  }
+]
+
+export const card = {
+  title: 'Bolt12',
+  subtitle: 'receive payments to a bolt12 offer',
+  image: { src: '/wallets/bolt12.svg' }
```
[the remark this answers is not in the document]
huumn: The logo is nice. Given we plan to overhaul the UI/UX flow of this, I don't think it matter much.

api/lnd/index.js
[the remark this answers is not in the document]
ekzyis: > So, to address your concerns, I’ve changed the code to use a WeakMap to maintain lndk instances until their corresponding lnd is GCed.

Can you please just pass `lndk` as we pass everything else? I don't want to review a new pattern of achieving the same thing in a different way for reasons that don't seem significant.

> This should address your main concern that, as far as i understand, is that in the remote eventuality that the lnd client decides to export a service called lndk, the previous code could end up shadowing it.

Yes but another big concern I had with the previous and now this code is that this is an example where you seem to go out of your way to deviate from our existing patterns, which are much easier to review since we’ve become familiar with their strengths and weaknesses over several years.

riccardobl:
needs to be retested because there were a lot of conflicts and dubious merges

riccardobl:
- changed lndk to be passed to every context instead of using a weakreference
- fixed some merge conflicts
- retested everything


should be ready for review

knocte:
ping?

ekzyis:
this is blocked by #2084 

but I will familiarize myself with the code this week and fix the conflicts

PR #1729, opened 2024-12-16
Fix missing logs on save

ekzyis:
## Description

On save, we didn't log if sending details were saved, updated or deleted.

## Checklist

yes

`7`. Tested this with all kinds of combinations between old and new config. 

n/a

no


api/resolvers/wallet.js
```diff
@@ -551,7 +551,7 @@ const resolvers = {
 
       const logger = walletLogger({ wallet, models })
       await models.wallet.delete({ where: { userId: me.id, id: Number(id) } })
-      logger.info('wallet detached')
+      logger.info('details for receiving deleted')
```
huumn, line 554: If this wallet is using device sync, isn't it possible there were no receive creds ever saved or deleted?

PR #1732, opened 2024-12-17
Fix send-only wallet save after device sync enabled

ekzyis:
## Description

_Using NWC as the example wallet here._

If you have a NWC send-only config for a wallet and then enable device sync, the mutation in `syncLocalWallets` only creates a row in `Wallet` and `VaultEntry` but not in `WalletNWC`.

If you then try to save again, the backend will see that there is a row for the wallet in `Wallet` and will try to update `WalletNWC`. This fails because there was no such row created during sync to server. 

This PR fixes this by always creating a row in `WalletNWC` if `Wallet` was created.

---

update: Okay, this breaks WebLN though as the comment `// client only wallets has no walletData` indicated why there was this check for existing wallet data. Added as TODO.

Maybe the fix is to also pass an empty receive config during local sync since that's what we also do on regular saves.

---

update: decided to use a [nested upsert](https://www.prisma.io/docs/orm/reference/prisma-client-reference#upsert-1) during `wallet.update` 

## Video of bug

https://github.com/user-attachments/assets/50eb45d8-81b5-4ff9-8aa7-c7625b240ee2

## Additional Context

- There is still a bug where if you now disconnect from the user vault, the row in `VaultEntry` is deleted but the row in `Wallet` isn't so you now have an enabled wallet with an empty config. Added as TODO.

## TODO

- [ ] ~fix empty enabled wallet after vault disconnect with send-only wallet~ _will fix in different PR because unrelated_

## Checklist

yes

`5`. I am not entirely sure if I tested every combination of send-only, send+recv, recv-only and device sync enabled/disabled with NWC but I tried and I saw no issues. That these changes fix the issue without breaking anything else also make sense to me.

Also tested enabling/disabling WebLN with and without device sync since a previous attempt broke it.

n/a

no

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -810,9 +810,9 @@ async function upsertWallet (
           ...(Object.keys(recvConfig).length > 0
             ? {
                 [wallet.field]: {
-                  update: {
-                    where: { walletId: Number(id) },
-                    data: recvConfig
+                  upsert: {
+                    create: recvConfig,
+                    update: recvConfig
                   }
                 }
               }
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -786,7 +786,7 @@ async function upsertWallet (
     }
   }
 
-  const { id, enabled, priority, ...walletData } = data
+  const { id, enabled, priority, ...recvConfig } = data
 
   const txs = []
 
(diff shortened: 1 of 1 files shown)


huumn:
I tried to figure out where this could go wrong, but I think it's right. 

I've been wondering if there's a way we can make this less crazy. I think part of the problem is that the relationship between vault entries and wallets isn't strong enough. Maybe we need to have a column in the wallets for each vault entry, even if it's just a foreign key. 

```prisma
model WalletNWC {
  id         Int      @id @default(autoincrement())
  walletId   Int      @unique
  wallet     Wallet   @relation(fields: [walletId], references: [id], onDelete: Cascade)
  createdAt  DateTime @default(now()) @map("created_at")
  updatedAt  DateTime @default(now()) @updatedAt @map("updated_at")
  nwcUrlRecv String?
  nwcUrlSendVaultEntryId Int? 
  nwcUrlSendVaultEntry VaultEntry @relation(fields: [nwcUrlSendVaultEntryId], references: [id], onDelete: SetNull)
}
```

ekzyis:
> I've been wondering if there's a way we can make this less crazy.

I think first we need to fix the frontend because of #1743 :eyes:

PR #1738, opened 2024-12-19
Fix nwc timeout not increased

ekzyis:
## Description

We intended to increase the wallet timeouts in https://github.com/stackernews/stacker.news/commit/faeefdc498cec342deb65dfc864f161fce26b9f9 and https://github.com/stackernews/stacker.news/commit/16b7160d3611c8f181b2babb9b3e68550f496d77 but NWC has its own timeout for the initial relay connection.

## Checklist

yes

`7`. This will now use whatever timeout is bigger for the initial relay connection.

n/a

no


huumn:
If we are going to use one timeout, I think we want `Math.min` because of the load balancer timeout.

huumn:
Closing because it looks like this bug was from long lived ndk connections getting in an indeterminate state. It also shouldn't take more than 15 seconds to open a websocket anyway.

PR #1739, opened 2024-12-19
Close relay connections after each NWC call

ekzyis:
## Description

In an attempt to fix relay issues, we're opening and closing relay connections after each nwc call. This is the same behavior as before we switched to NDK in #1590.

## Additional Context

As @riccardobl mentioned in the chat, it probably makes sense to keep the relay connections open on the server which this code doesn't do (yet). However, the previous code also didn't do that.

## Checklist

yes

tbd, will test more tomorrow

n/a

no


huumn:
I don't know if it fixes the issue, but send/receive still works, so I'm shipping it. Thank you!

PR #1740, opened 2024-12-19
Refine #1739 and fix regression causing nostr crossposts and login to not work

### LINKED ISSUE #1741: bug report nip46 login shows error "m.ZP.getSigner is not a function"

### Describe the bug

Trying to use nip46 login (nip05 or bunker url) gives the same error "m.ZP.getSigner is not a function"

### Screenshots

Screenshot from 2024-12-19 09-32-37

### Steps To Reproduce

1. Go to Login with Nostr.
2. Type nip05 (i.e. artur@nsec.app) or a bunker url.
3. Click "Login with token or NIP05"
4. Error message

### Expected behavior

Expected behavior is for signer popup to show up so I could confirm the connection.

### Logs

hook.js:608 TypeError: m.ZP.getSigner is not a function
    at nostr-auth.js:128:28

Source code line referred above:
      const signer = Nostr.getSigner({ nip46token, supportNip07: useExtension })

### Device information

Ubuntu 20, Chrome 116

### Additional context

_No response_

riccardobl:
This PR 
 -  fixes nostr logins and crossposts: Closes #1741 
 - and refines  #1739 by:
    - moving the close method inside the nostr class
    - adding static method to return the ndk singleton
    - disconnecting from the relay after crosspost on the client

### DIFF
--- a/components/nostr-auth.js
+++ b/components/nostr-auth.js
@@ -124,8 +124,9 @@ export function NostrAuth ({ text, callbackUrl, multiAuth }) {
       const k1 = data?.createAuth.k1
       if (!k1) throw new Error('Error generating challenge') // should never happen
 
+      const nostr = Nostr.get()
       const useExtension = !nip46token
-      const signer = Nostr.getSigner({ nip46token, supportNip07: useExtension })
+      const signer = nostr.getSigner({ nip46token, supportNip07: useExtension })
       if (!signer && useExtension) throw new Error('No extension found')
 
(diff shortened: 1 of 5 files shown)


ekzyis (commented):
Do we need a global instance for remote signing? Can't we always create a new instance and then destroy it with `nostr.close()`? I think it's confusing to use a global instance and individual instances in our code.

riccardobl:
Yes, we can use a new instance for remote signing.
The singleton is important only serverside, really.

PR #1742, opened 2024-12-19
Zebedee send and receive attachment

riccardobl:
## Description

Zebedee send and receive attachment

## Screenshots
image
image
image
image

## Additional Context

For sending attachments, I just use the Zebedee dashboard APIs, nothing strange there. 
But for receiving, I had to get creative since, as far as I know, Zebedee doesn’t have a receive-only API.

What I did was ask the user for their userId to fetch the invoice through the public LNURLp endpoint. However, since the userId is hard to find (I think you can only get it through the API), I added the option to input the ZBD Gamertag (or LN address), which is then converted to the userID.  This way, the attachment doesn’t break if the user changes their Gamertag.

## Checklist

yes

8

yes

no


wallets/zebedee/index.js
```diff
@@ -0,0 +1,42 @@
...
+export const fields = [
+  {
+    name: 'apiKey',
+    label: 'api key',
+    type: 'password',
+    optional: 'for sending',
+    help: `you can get an API key from [Zebedee Dashboard](${DASHBOARD_URL}) from \n\`Project->API->Live\``,
+    clientOnly: true,
+    requiredWithout: 'gamerTagId',
+    validate: string()
```
huumn, line 23: I suspect we could do a little better for validation.

Same for the gamer tag.

huumn (changes requested):
At a glance, this looks solid. The only thing that caught my eye is the lack of validation.

---------

Also, I don't think we need to worry about the gametag/name changing too much once we breakout send/receive stuff. In anticipation of that, it might be better to not even do receives in the same form and just let them use the lightning address attachment so that we don't have to do another quirky migration in the near future.

I mostly suspect this wallet attachment will be largely unused in which case we don't want it to get in the way of improving other stuff. lmk what you think

riccardobl:
> At a glance, this looks solid. The only thing that caught my eye is the lack of validation.
> 
> Also, I don't think we need to worry about the gametag/name changing too much once we breakout send/receive stuff. In anticipation of that, it might be better to not even do receives in the same form and just let them use the lightning address attachment so that we don't have to do another quirky migration in the near future.
> 
> I mostly suspect this wallet attachment will be largely unused in which case we don't want it to get in the way of improving other stuff. lmk what you think

i think you are probably right, the receiver concoction can be added later if there is demand, the ln address will work.

wallets/zebedee/index.js
riccardobl: I can see the zebedee api key today is an alphanumeric 32 characters long string, however there is nothing in the doc that implies this is and always be its shape, so i just added a reasonable min and max length. 
Are you happy with that?

Issue #1648, opened 2024-11-25
Signup button on iOS has no vertical margin


huumn:
![image](https://github.com/user-attachments/assets/8b7c5e2c-c2f7-4e44-b43c-68f06685e533)

PR #1745, opened 2024-12-19
fix: adjusts the navbar and signup button for mobile

cyphercosmo:
_Updated on `Fri Dec 20 23:09:21 2024 UTC`_

## Description

A simple change to address issue #1648 by removing a conflicting tailwind class and only passing it when it's necessary.

## Screenshots

image

## Checklist

Yes

10

Yes

No

### DIFF
--- a/components/nav/common.js
+++ b/components/nav/common.js
@@ -227,7 +227,7 @@ export function SignUpButton ({ className = 'py-0', width }) {
 
   return (
     <Button
-      className={classNames('align-items-center ps-2 py-1 pe-3', className)}
+      className={classNames('align-items-center ps-2 pe-3', className)}
       style={{ borderWidth: '2px', width: width || '150px' }}
       id='signup'
       onClick={() => handleLogin('/signup')}
@@ -359,7 +359,7 @@ export function LoginButtons ({ handleClose }) {
         <LoginButton />
       </Dropdown.Item>
       <Dropdown.Item className='py-1'>
-        <SignUpButton />
+        <SignUpButton className='py-1'/>
       </Dropdown.Item>
       <Dropdown.Item className='py-1'>
         <SwitchAccountButton handleClose={handleClose} />


huumn:
Awesome, but I'd rather make the button the right size. The effect of your change is to obstruct more pixels of what people want to see.

huumn (changes requested):
As I said in my other comment please

cyphercosmo:
I agree @huumn, on it.

cyphercosmo:
Please check @huumn , I found the best solution to be a mix of making the nav bar slightly bigger but also reducing the size of the button.

Any suggestions?

huumn (changes requested):
My suggestion is the same. We do not want the navbar larger. Just make the button smaller please.

cyphercosmo:
If I only make the button small it continues to touch the navbar edge as when you reported, I'm trying to not introduce custom CSS only for this button but rather I'm using the CSS framework you guys chose.

If you have something specific in mind how it should look please share, otherwise I prefer to not continue trying to guess exactly what I should do that will look decent.

ekzyis:
I think all this needs is to remove the vertical padding. This stops the button from touching the top and bottom:

https://github.com/user-attachments/assets/8f221d19-7738-4175-a002-86e9fa4d5b12

I also tried to apply some vertical margin but that makes the navbar height increase instead of decreasing the height of the button.

cyphercosmo:
Okay, I tried your approach @ekzyis, I noticed that the SignupButton had this `py-1` class hardcoded so I checked the references to this component and the only place except the sticky bar it's currently being used is when you click the anon icon, so I made the change to pass it explicitly in that case, checked it for both situations and it's working.

cyphercosmo (commented):
The reason I'm saying I removed a conflicting class is because the `className` variable defaults to `py-0` and no use of the component was passing the class name.

huumn:
Do you have a lightning address @cyphercosmo? I owe you some sats for this

cyphercosmo:
Hey @huumn ! 

Yes it's cyphercosmo@getalby.com.

Looking forward to contributing more, also feel free to ask if there is something important you need done.

PR #1748, opened 2024-12-20
Require invite limit

ekzyis:
## Description

The frontend says `invite limit` is optional but if you don't enter anything and submit, we throw 

> Variable "$limit" got invalid value ""; Int cannot represent non-integer value: ""

I believe this has been an issue even before #1649 since the validation schema was not changed in #1649 and I had to update it to use `.nullable()`.

Additionally, our code checked if `invite.limit` was defined everywhere except in the most critical part where it gets redeemed, lol.

## Checklist

yes

`6`. Testing creating and redeeming limited and unlimited invite links.

n/a

no

### DIFF
--- a/api/paidAction/inviteGift.js
+++ b/api/paidAction/inviteGift.js
@@ -21,7 +21,7 @@ export async function perform ({ id, userId }, { me, cost, tx }) {
     where: { id, userId: me.id, revoked: false }
   })
 
-  if (invite.giftedCount >= invite.limit) {
+  if (invite.limit && invite.giftedCount >= invite.limit) {
     throw new Error('invite limit reached')
   }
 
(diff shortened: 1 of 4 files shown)


huumn:
Given the way these are abused, I think the move is to make the limit required. 

Since we merged user specified invite codes, having a limit is even more important, because we'll no doubt have people creating invite links like `/invites/grayruby`.

ekzyis:
I see, makes sense, made the invite limit required for the mutation in f27be13 but didn't enforce it in the database because I am not sure how to deal with existing unlimited invites:

```
> SELECT COUNT(1) FROM "Invite" WHERE "limit" IS NULL;
256
> SELECT COUNT(1) FROM "Invite" WHERE "limit" IS NULL AND "revoked" = 'f';
179
```

I think we should first just make creating new unlimited invites impossible (this PR), announce that we will delete any unlimited invite in X weeks, then delete them and then make the `Invite.limit` column non-nullable.

Issue #1751, opened 2024-12-20
Daily stacked and spent summary notifications


huumn:
This has been requested many times.

Stackers want an overview of how much they earned relative to spent. 

This shouldn't be too hard to do and can be pretty minimal to start. We'd just grab `user_stats` for the prior day and make sure to add any rewards that were just given out (at midnight), then the notification could say "you stacked X sats and spent X sats."

huumn:
I personally want this so I can try to spend as much as I stacked every day.

m0wer:
Does this still need work? I've implemented it before realizing there already was https://github.com/stackernews/stacker.news/pull/1756 🤦

m0wer:
Thanks @Soxasora ! What's a “seed”? Seed funding round or something else?

Here is the code: https://github.com/stackernews/stacker.news/pull/2132

I would need some help or ideas on a simple test scenario. I've tried creating an email account and zapping and then change the Docker host time to one day in the future but that wasn't enough to get something on the past day for `user_stats_days`. I might be missing something. I tried running the job `materializedViewRefresher` and then the new `dailyStackedNotification`.

huumn:
Please hold off on this! I'm currently refactoring related code to make this a lot easier. I'm adding the `blocked` label in the meantime.

m0wer:
Should this be unblocked?

ekzyis:
No, this is still blocked by https://github.com/stackernews/stacker.news/pull/2084

PR #1756, opened 2024-12-22
feat: daily stacked and spent sats notification

Soxasora:
## Description

Daily stacked and spent sats summary addressing #1751 
It features a brief summary of how much an user has stacked (including rewards) and spent in 24h.

## Additional Context

## Checklist

TODO:
- [ ] refactor DailyStats to DailySats
- [ ] push notification
- [ ] cleanup
- [x] settings toggle for daily stats
- [x] notification component
- [x] notification resolver


worker/satSummary.js
Soxasora, line 1: realized I didn't push this before, sorry

huumn (commented):
This looks nice. People are going to like this.

I'll QA and UI review tomorrow.

huumn:
Heads up: I'm going to create a better seed to test this. I won't review until tomorrow.

> Sometimes, dailySatSummary runs before views-days updates the user_stats_days materialized view, causing it to fail in obtaining new data. To ensure the availability of new data, I decided to pair dailySatSummary with views-days since they both run at midnight anyway, I hope it's okay.

This is an okay way to do this if we are dependent on those views afaict.

--------------

I think there may be a more general UX issue with this approach though and it's tricky to solve. By depending on user_stats_days, this misses the rewards and territory revenue which are computed just after midnight. I think people will expect the notification to include those recent numbers instead of the ones from earlier in the day. For example:

As is, and correct me if I'm wrong, after midnight on 1/1:

1. this summary notification is shown with data from 12/31/2024 (including rewards/revenue earned from 12/30)
2. rewards and territory revenue are computed for 12/31/2024, on 1/1/2025, and notifications are shown
3. stacked/spent from (2) won't be summarized until 1/2/2025

------------------

The above won't block my approval and payment. It looks like you did all the important parts so I should have an easy time doing whatever is required to fix this ux thing.

Soxasora:
> By depending on user_stats_days, this misses the rewards and territory revenue which are computed just after midnight.

I noticed this behavior the time I tested territory fees, I was conflicted on whether doing something about it or leave it as-is because territory revenue was being given after the midnight, so I just waited for your opinion on it. 

I considered:
- Subtracting `msats_reward`, `msats_revenue` from `user_stats_days.msats_stacked` with a LEFT JOIN from `SubAct` and `Earn` to get the latest data

huumn:
I think the move might be to create a new view that's computed at 12:15a just for this purpose. I'm not sure there's a better way.

Soxasora:
Agree, it would also be its own pgboss job because we wouldn't need `views-days` anymore

huumn:
> edit: also I noticed that I have 2 commits in the future because I was testing with time, didn't rebase because it could mess with your tree

lol I was wondering why my comments on github weren't showing up. I think you found a github bug. [github trusts git a lot](https://github.com/huumn/git-remote-gitern-encrypted).

huumn:
Sorry, been slacking. I'm currently working on a seed for this.

PR #1757, opened 2024-12-22
configurable nostr signer

riccardobl:
## Description

## Additional Context

## Checklist

### DIFF
--- a/lib/nostr.js
+++ b/lib/nostr.js
@@ -34,15 +34,15 @@ export default class Nostr {
    * @type {NDK}
    */
   _ndk = null
-  static globalInstance = null
-  constructor ({ privKey, defaultSigner, relays, nip46token, supportNip07 = false, ...ndkOptions } = {}) {
+  static _globalInstance = null
+
+  constructor ({ relays, ...ndkOptions } = {}) {
     this._ndk = new NDK({
       explicitRelayUrls: relays,
       blacklistRelayUrls: RELAYS_BLACKLIST,
       autoConnectUserRelays: false,
       autoFetchUserMutelist: false,
       clientName: 'stacker.news',
-      signer: defaultSigner ?? this.getSigner({ privKey, supportNip07, nip46token }),
       ...ndkOptions
     })
   }
@@ -51,10 +51,10 @@ export default class Nostr {
    * @type {NDK}
    */
   static get () {
-    if (!Nostr.globalInstance) {
-      Nostr.globalInstance = new Nostr()
+    if (!Nostr._globalInstance) {
+      Nostr._globalInstance = new Nostr()
     }
-    return Nostr.globalInstance
+    return Nostr._globalInstance
   }
 
   /**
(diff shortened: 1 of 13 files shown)


pages/settings/index.js
```diff
@@ -197,6 +220,42 @@ export default function Settings ({ ssrData }) {
...
+                  const nostr = new Nostr()
+                  try {
+                    await callWithTimeout(async () => {
+                      const testSignerInstance = nostr.getSigner({
+                        userPreferences: { signer, signerType, signerInstanceKey }
+                      })
+                      if (testSignerInstance instanceof NDKNip46Signer) {
+                        testSignerInstance.once('authUrl', nostrAuthChallengeResolver)
+                      }
+                      await testSignerInstance.blockUntilReady()
```
riccardobl, line 244: Here, we initiate and test the connection to the signer, providing two benefits:

1. The user immediately knows if the NIP-46 signer is functioning without needing to attempt a crosspost.
2. We complete the challenge and authentication at this stage, preventing interruptions during navigation later.

lib/nostr.js
```diff
@@ -66,16 +67,30 @@ export default class Nostr {
...
+   * @param {string} [args.userPreferences.signerType] - signer type to use for signing (nip07, nip46)
+   * @param {string} [args.userPreferences.signer] - signer config to use for signing
    * @param {string} [args.privKey] - private key to use for signing
    * @param {string} [args.nip46token] - NIP-46 token to use for signing
-   * @param {boolean} [args.supportNip07] - whether to use NIP-07 signer if available
+   * @param {boolean} [args.nip07] - whether to use NIP-07 signer if available
    * @returns {NDKPrivateKeySigner | NDKNip46Signer | NDKNip07Signer | null} - a signer instance
    */
-  getSigner ({ privKey, nip46token, supportNip07 = true } = {}) {
+  getSigner ({ userPreferences: { signerType: userSignerTypePreference, signer: userSignerPreference, signerInstanceKey } = {}, privKey, nip46token, nip07, nip46LocalSigner } = {}) {
```
riccardobl, line 79: `signerInstanceKey` is the [client key pair](https://github.com/nostr-protocol/nips/blob/master/46.md#terminology), which is one of the key parts needed to reestablish the connection for NIP-46 crossposts across user sessions.

lib/nostr.js
```diff
@@ -200,4 +217,65 @@ class NDKNip46SignerURLPatch extends NDKNip46Signer {
...
+        )
+      }, 2100)
+
+      // try to restore the connection
+      // if it succeeds, clear the timeout.
+      //    One of the two will fail, either the new connection or the restore attempt.
+      confirm(resolve)
+        .then(() => clearTimeout(connector))
+        .catch((err) => console.error('Error restoring connection:', err))
+    })
```
riccardobl, line 279: The [spec](https://github.com/nostr-protocol/nips/blob/master/46.md) is not very clear about how to re-establish connections between user sessions. It only mentions that the client key pair is disposable and suggests storing it locally for the user session. It also states that new connections cannot be established with a reused secret.

The key point here is that nsec.app will respond to connections from an authorized client key pair but will reject new connect RPCs from the same secret and key pair. The issue is that the NDK auth flow doesn’t seem to account for this behavior.

To work around this, I implemented a solution where two requests,one for "connect" and one for "get_public_key", are raced against each other. The first request to complete without an error is used to proceed with authentication.

PR #1763, opened 2024-12-25
LNC receiver V2

riccardobl:
## Description

Remate of https://github.com/stackernews/stacker.news/pull/1342 

## Additional Context

## Checklist


api/resolvers/wallet.js
```diff
@@ -779,6 +784,7 @@ async function upsertWallet (
   if (testCreateInvoice) {
     try {
       await testCreateInvoice(data)
+      await validate({ data, settings, skipGenerated: false })
```
riccardobl, line 787: The changes in this file are not related to this pr specifically.
See https://github.com/stackernews/stacker.news/pull/1769

wallets/config.js
```diff
@@ -63,7 +63,7 @@ export function useWalletConfigurator (wallet) {
         throw err
       }
     } else if (canReceive({ def: wallet.def, config: serverConfig })) {
-      const transformedConfig = await validateWallet(wallet.def, serverConfig)
+      const transformedConfig = await validateWallet(wallet.def, serverConfig, { skipGenerated: true })
```
riccardobl, line 66: The changes in this file are not related to this pr specifically.
See https://github.com/stackernews/stacker.news/pull/1769

wallets/validate.js
```diff
@@ -72,7 +72,7 @@ function composeWalletSchema (walletDef, serverSide, skipGenerated) {
 
     if (clientOnly && serverSide) {
       // For server-side validation, accumulate clientOnly fields as vaultEntries
-      vaultEntrySchemas[optional ? 'optional' : 'required'].push(vaultEntrySchema(name))
+      vaultEntrySchemas[(optional || generated) ? 'optional' : 'required'].push(vaultEntrySchema(name))
```
riccardobl, line 75: The changes in this file are not related to this pr specifically.
See https://github.com/stackernews/stacker.news/pull/1769

api/resolvers/wallet.js
ekzyis: So this means this PR is based on #1769?

riccardobl: yes

huumn:
lncd doesn't appear authenticated. Is the expectation that it's run in a VPC or that authentication is handled in reverse proxy?

riccardobl:
> lncd doesn't appear authenticated. Is the expectation that it's run in a VPC or that authentication is handled in a reverse proxy?

it only needs to be accessible by the backend, so the expectation was to run it in a vpc

huumn:
Everything we run in a vpc we also make authenticated, just in case, so we'll want to do that.

riccardobl:
> Everything we run in a vpc we also make authenticated, just in case, so we'll want to do that.

do you have preferences for the authentication method? Does it need tls?

huumn:
Either basic or bearer auth is fine. Builtin TLS would be nice so I don't have to run a reverse proxy.

Maybe we can gut lnd's gprc auth/tls? Or maybe that's more work than rolling our own.

riccardobl:
Done. I've updated the env variable list in the pr description, but tl;dr is that i've added these vars for the lncd daemon
```
LNCD_TLS_CERT_PATH	Path to the TLS certificate file (empty to disable TLS).
LNCD_TLS_KEY_PATH		Path to the TLS key file (empty to disable TLS).
LNCD_AUTH_TOKEN		Bearer token required to access the server (empty to disable authentication).
```

and these for the sn backend:

```
LNCD_AUTH_TOKEN     same as LNCD_AUTH_TOKEN for lncd
LNCD_CERT                    optional cert hex string used only if lncd has a self-signed certificate
```

huumn:
> the health check endpoint is authenticated too:

I'd prefer it not be. It'll be a load balancer checking it without tls.

riccardobl:
> > the health check endpoint is authenticated too:
> 
> I'd prefer it not be. It'll be a load balancer checking it without tls.

Ok, i've added an unauthenticated http-  /health endpoint that is configurable in the daemon (v0.3.2) using the env variables:
```
LNCD_HEALTHCHECK_SERVICE_PORT	
LNCD_HEALTHCHECK_SERVICE_HOST
```

it defaults to `http://127.0.0.1:7168/health`

PR #1767, opened 2024-12-27
Details & summary tag support

### LINKED ISSUE #1422: Add collapsable answers

kravhen:
## Description

Closes #1422

Added support for collapsible content sections using HTML-style `<details>` and `<summary>` tags. 

- Allows users to create expandable/collapsible sections in their posts
- Preserves markdown within the details blocks

Implemented support for HTML-style `<details>` and `<summary>` tags in markdown content, with minimal styling. It's flexible enough in terms structures; users can use the tags in various ways. Single or multiline, with or without a summary tag, blank lines or not (blank line preceding a content block is still required to activate markdown, as is the expected behavior).

## Key steps
1. Detection: Scan for <details> opening and closing tags in the content
2. Node Collection: Gather all nodes between the tags into a details container
3. Summary Extraction: Look for and process optional <summary> tags 
4. Node Transformation: Convert the collected content into proper AST detail node
5. Styling: Minimal CSS for collapsed/expanded states and hover effects

## Screenshots
Here is example code, see how each structure appears in screenshot below.

```html
Structure 1:

<details>lorem ipsum
without summary tag
lorem ipsum
</details>

----------

Structure 1 with blank line:

<details>

lorem ipsum
without summary tag
lorem ipsum
</details>

----------

Structure 2:

<details>
<summary>summary text here</summary>
lorem ipsum
lorem ipsum
</details>

----------

Structure 2 with blank line:

<details>
<summary>summary text here</summary>

lorem ipsum
lorem ipsum
</details>

----------

Structure 3:

<details>
    <summary>indented summary text here</summary>
1. first thing
2. second thing 
3. third thing
</details>

----------

Structure 3 with blank line around the markdown:

<details>
    <summary>indented summary text here</summary>

1. first thing
2. second thing 
3. third thing

</details>

----------

Structure 4:

<details>
  <summary>
    multiline summary text here
  </summary>

  text inside details with inline markdown working properly, such as *italics* and **bold**
</details>

----------

Structure 5:

<details>
<summary>Shopping list</summary>
- Vegetables
- Fruits
- Fish
</details>

----------

Structure 5 with proper blank line for markdown list:

<details>
<summary>Shopping list</summary>

- Vegetables
- Fruits
- Fish
</details>
```
image
image
image

Note: the yellow borders in the screenshots have been reverted to a neutral color because they looked bad on light mode. Left the screenshots because I thought they kinda look cool in dark mode. If you guys like them enough, can add new theme color variables to globas.scss for this.

## Additional Context

Tried to do it a few different ways because it always seemed so verbose, and every try I gave it ended up being roughly as verbose. This is the implementation I was able to do without any extra imports like `rehype-raw`. Previous attempts used other utilities to transform from and to markdown like `mdast-util-from-markdown` and `mdast-util-to-hast` but I realised I was overmanipulating nodes and those were not necessary to get it to work since markdown is already supported.

Details handler code had to appear early in the RehypeSN transformer or other processors would make certain structures break. For example, if code is moved after the Nostr ID handler processor, the details node still render and function but the closing tags appear duplicated outside as text.

Does **not** handle nesting details within details. I tried it in a prior attempt and I think it requires more node manipulation and complexity for the current use case. If users end up using these extensively and there's a real use case for nesting them in posts, maybe we can take another look at it.

## Checklist

Yes - existing content without details/summary tags will render as before. The changes only affect content that explicitly uses these new tags.

Around 8/10. I could do additional testing for lengthy posts with complex markdown within.

Behaves the same on mobile. Reverted some subtle yellow borders to neutral white/greyish colors for expanded details sections because the yellow looked bad on light mode.

No new environment variables were introduced.

### DIFF
--- a/components/boost-button.js
+++ b/components/boost-button.js
@@ -26,7 +26,7 @@ export default function Boost ({ item, className, ...props }) {
       item={item} As={oprops =>
         <div className='upvoteParent'>
           <div
-            className={styles.upvoteWrapper}
+            className={classNames(styles.upvoteWrapper, item.deletedAt && styles.noSelfTips)}
           >
             <BoostIcon
               {...props}
--- a/components/notifications.js
+++ b/components/notifications.js
@@ -344,7 +344,7 @@ function getPayerSig (lud18Data) {
 
 function InvoicePaid ({ n }) {
   const payerSig = getPayerSig(n.invoice.lud18Data)
-  let actionString = 'desposited to your account'
+  let actionString = 'deposited to your account'
   let sats = n.earnedSats
   if (n.invoice.forwardedSats) {
     actionString = 'sent directly to your attached wallet'
--- a/components/pull-to-refresh.js
+++ b/components/pull-to-refresh.js
@@ -2,7 +2,7 @@ import { useRouter } from 'next/router'
 import { useState, useRef, useEffect, useCallback, useMemo } from 'react'
 import styles from './pull-to-refresh.module.css'
 
-const REFRESH_THRESHOLD = 50
+const REFRESH_THRESHOLD = 150
 
 export default function PullToRefresh ({ children, className }) {
   const router = useRouter()
--- a/components/territory-form.js
+++ b/components/territory-form.js
@@ -197,7 +197,7 @@ export default function TerritoryForm ({ sub }) {
             >
               <Checkbox
                 type='radio'
-                label='100k sats/month'
+                label={`${abbrNum(TERRITORY_PERIOD_COST('MONTHLY'))} sats/month`}
                 value='MONTHLY'
                 name='billingType'
                 id='monthly-checkbox'
@@ -206,7 +206,7 @@ export default function TerritoryForm ({ sub }) {
               />
               <Checkbox
                 type='radio'
-                label='1m sats/year'
+                label={`${abbrNum(TERRITORY_PERIOD_COST('YEARLY'))} sats/year`}
                 value='YEARLY'
                 name='billingType'
                 id='yearly-checkbox'
@@ -215,7 +215,7 @@ export default function TerritoryForm ({ sub }) {
               />
               <Checkbox
                 type='radio'
-                label='3m sats once'
+                label={`${abbrNum(TERRITORY_PERIOD_COST('ONCE'))} sats once`}
                 value='ONCE'
                 name='billingType'
                 id='once-checkbox'
--- a/components/text.js
+++ b/components/text.js
@@ -255,7 +255,7 @@ function P ({ children, node, onlyImages, somethingBefore, somethingAfter, ...pr
   )
 }
 
-function Details({ children, ...props }) {
+function Details ({ children, ...props }) {
   return (
     <details className={styles.details} {...props}>
       {children}
@@ -263,7 +263,7 @@ function Details({ children, ...props }) {
   )
 }
 
-function Summary({ children, ...props }) {
+function Summary ({ children, ...props }) {
   return (
     <summary className={styles.summary} {...props}>
       {children}
--- a/components/text.js
+++ b/components/text.js
@@ -125,7 +125,9 @@ export default memo(function Text ({ rel = UNKNOWN_LINK_REL, imgproxyUrls, child
       return <Link id={props.id} target='_blank' rel={rel} href={href}>{children}</Link>
     },
     img: TextMediaOrLink,
-    embed: Embed
+    embed: Embed,
+    details: Details,
+    summary: Summary
   }), [outlawed, rel, TextMediaOrLink, topLevel])
 
   const carousel = useCarousel()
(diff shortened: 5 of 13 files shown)


huumn (commented):
The rehype code looks more complicated than I'd expect, but I haven't done enough work with nested elements to trust my expectations. 

I'll try to get you an actual review next week.

kravhen:
I also had a nagging feeling that it could be done simpler and with less code. I restarted from scratch once or twice due to life getting in the way and every time I sat down, seeing things with fresh eyes, I thought I could keep it succinct that time around.

But every time it ended up more verbose than I'd wanted heh.

I think it could be made simpler, but the tradeoff would be that users would have to use the tags in a more restrictive way,  following a more opinionated structure or else it wouldn't work. e.g. you must place the tags this exact way or bust.

I'd be glad to see how it's done if the amount of code can be shaved off significantly. I honestly learned a lot about how AST, their manipulation and nodes in general, but I am no pro :P. 

Here is what it actually looks like without the yellow so its more light-mode friendly. Screenshot above was if we didn't care about light mode. This is how things actually end up looking like:

sn-details-dark
sn-details-light

ekzyis:
I just stumbled across this PR. Sorry, reviewing this must have went under!

I can't promise anything, but I will try to review this next week.

kravhen:
> I just stumbled across this PR. Sorry, reviewing this must have went under!
> 
> I can't promise anything, but I will try to review this next week.

All good, no stress. I'd almost forgotten about this one hehe. As always, appreciate critique if any. Went back and forth different ways with this one and wasn't able to get to something that worked as expected while being more... lean and clean-looking. Kept having that gnawing feeling I could be overlooking something, but simpler ways fell short of 1:1 github-like behavior in different ways.

kravhen:
Can I convert this back to draft? 

Unless there was already reviewing done and you like how it works as-is, I think I'd try using stack-based parsing instead of a top-level only aggregator. Advantages include allowing nesting and being less brittle when trying to keep formatting. 

Could also possibly move a lot of the code to a separate utility file. This PR would add a big chunk (~100-200 lines) to `rehype-sn.js`  dedicated solely for this one set of tags which doesn't feel proper.

ekzyis:
Yes, no worries, you can convert this back to draft!

kravhen:
### Rewrite: switch to a stack-based pre-pass for details/summary tags

The rewrite replaces the previous accumulator method (in main visit) of doing it with a stack-based pre-pass step that assembles details & summary outside and before the main visitor in `rehype-sn.js`.


--- 
### Main reasons for change:

- **Better formatting preservation**: The old approach felt somewhat fragile and unsatisfactory because of splitting raw segments and strings which often downgraded content to plain text if the format and structure was not properly and specifically handled well enough.

- **Nesting support**: Nesting works now. The old way explicitly avoided nesting and enforced top-level only.

- **Possible attribute support**: Right now only open is whitelisted as an attribute to have the details block start opened, but we could theoretically allow class and id to allow users to do things like style or format their details blocks based on allowed presets, or use IDs for anchors/links. For now I've left this out though because there's no immediate need to open that door imo. 
I can see a future where it becomes a more attractive feature to allow classes (or a limited predefined set of them) for folks who write long detailed guides and content, perhaps evergreen content of sorts, that could use some extra visual spiff to organize things better, make it more visually pleasing, color code, etc.

- **Better isolation, robustness**: Being outside the main visitor avoids interfering with other transforms (embeds, mentions, etc.). The previous way was also somewhat picky about being included as early as possible in the main visitor to avoid issues. Tags that aren't properly balanced/closed fail safely instead of making broken trees.

- **Cleaner rehype-sn file**: Takes the hundred or so lines out of `rehype-sn` to keep it focused and the pre-pass lives in `rehype-sn-prepass` instead. I've named it as such instead of something like rehype-sn-details so that other pre-pass operations could eventually live there as well.

--- 

### New functionality in a nutshell:

1. The pre-pass function `assembleDetails(tree)` tokenizes details and summary tags in raw nodes.
2. Proper element nodes get built with a stack, applying attributes (open only at the moment) and preserves non-raw children inside.
3. If tags balance, then that level is made into a structured node.
4. The normal `visit()` in `rehype-sn.js` runs on the AST as usual. 

---

### Backwards compatibility & Testing

Existing posts using these tags (if they exist for some reason) gain the new functionality.

Markdown still works. 

No new dependencies (still no `rehype-raw`).

Tested all example structures listed above and behavior is as expected. Tested some more structures that make use of nesting and markdown in different ways and behavior is also as expected.

Copy pasted all these test structures into a github comment box and compared resulting preview with one on SN and it's looking identical afaict. 

---

Here's some screenshots of the nesting in action. I left the previous styling as-is. Open to feedback all around.


image

image

huumn:
We're overhauling our editor and most of this will be unusable, but that's our fault and not yours. 

I'm closing but I will pay you the bounty

kravhen:
> We're overhauling our editor and most of this will be unusable, but that's our fault and not yours. 
> 
> I'm closing but I will pay you the bounty

All good, can't wait to see the new editor! I'd ask if the implementation was any good for educational purposes, but since it's moot no need to spend time reviewing it :P

Issue #1186, opened 2024-05-23
Feature Request : Comment Fee Control


Radentor:
Issue : 1186
     Title : Comment Fee Control
     Author : Public_N_M_E
     Status : Proposed
     Type : Enhancement
     Created : 2024-05-23

# Abstract
This issue describes an enhancement proposal to make territory founder able to control the comment/reply fee within their territory.

# Motivation
> Posting content is low in my territory but engagement is high in comments.  If I were able to adjust the comments/reply fees I think I could make my way to sustainability as a territory.  Replies are 1sat per post. But to post a discussion or something is the only adjustable fee, people are chatting away and engaging with each other for 1sat per comment, but very few posts get posted, the ability to make the comment/reply fees adjustable might really help.

Quoted from SN comment [548398](https://stacker.news/items/548398/r/Radentor?commentId=548406)

# Acknowledgements
This idea was first proposed by [@Public_N_M_E](https://stacker.news/Public_N_M_E) around an hour ago. Since [@Public_N_M_E](https://stacker.news/Public_N_M_E) has yet to submit an issue, i thought i'd submit an issue related to this so a good idea like this won't be lost/ignored.

Gudnessuche:
I found tweaking the category to a free one(for example 'dev') or a slightly cheaper one o help

huumn:
@Radentor do you have a lightning address I can send sats to? The issue has been implemented.

Radentor:
revisedbird84@walletofsatoshi.com

Gudnessuche:
***@***.***

That's my address above

PR #1768, opened 2024-12-27
feat: comment fee control

Soxasora:
## Description

Adds the possibility to customize comment/replies fees for territory owners, addressing #1186 

## Screenshots

territory overview
image

territory edit
image

reply
image

## Additional Context

As originally conceived, if you're not the OP, fee is bound to be multiplied x10 to avoid item spam (eg. 150 -> 1500 -> 15000 etc).

`findRootItem` (line 16) needs to query item more than once to obtain the root in case of replying to another reply

## Checklist

8, replied on bio (1 sat), items, multiple comments. Creation of items and bios follow their original behavior.


components/reply.js
```diff
@@ -173,7 +174,7 @@ export default forwardRef(function Reply ({
       {reply &&
         <div className={styles.reply}>
           <FeeButtonProvider
-            baseLineItems={postCommentBaseLineItems({ baseCost: 1, comment: true, me: !!me })}
+            baseLineItems={postCommentBaseLineItems({ baseCost: bio ? 1 : sub.replyCost, comment: true, me: !!me })}
```
ekzyis, line 177: the saloon is a weird item that has no sub, so we need to check for sub here:

```suggestion
            baseLineItems={postCommentBaseLineItems({ baseCost: bio || !sub ? 1 : sub.replyCost, comment: true, me: !!me })}
```

api/paidAction/itemCreate.js
```diff
@@ -12,8 +12,32 @@ export const paymentMethods = [
...
+    if (bio) return BIO_COST
+
+    if (parentId) {
+      const rootItem = await findRootItem(models, parentId)
+      return rootItem.bio ? BIO_COST : satsToMsats(rootItem.sub?.replyCost)
+    }
+
+    const sub = await models.sub.findUnique({ where: { name: subName } })
+    return satsToMsats(sub.baseCost)
+  }
```
ekzyis, line 37: possibly opinionated nitpick: I think it's easier to read if these functions are defined in the top level (with space between them :eyes:). As written, it suggests there is a "bigger reason" to be defined in the scope of the function but the functions take everything they need as arguments anyway (except `BIO_COST` but that can then also be defined at the top level) so there's also some shadowing going on.

api/paidAction/itemCreate.js
```diff
@@ -12,8 +12,32 @@ export const paymentMethods = [
...
+  const baseCost = await getBaseCost({ models, bio, parentId, subName })
+  // const sub = (parentId || bio) ? null : await models.sub.findUnique({ where: { name: subName } })
+  // const baseCost = sub ? satsToMsats(sub.baseCost) : 1000n
```
ekzyis, line 40:
```suggestion
```

ekzyis (commented):
Just noticed a small thing regarding saloon items and one nitpick but other than that, this is fabulous!

But also want to confirm with @huumn the following:

I noticed that we still have `sub.allowFreebies` in the schema with no way for founders to configure it. The option was removed from the frontend in https://github.com/stackernews/stacker.news/pull/1333 but we kept it in the schema for backwards compatibility. The value is not consistent across territories in prod currently but that's okay since it's no longer used anywhere anyway.

However, reading the description of #1333, I think the intention was to only disable freebies for posts. Since the base cost of comments was always 1 sat, we didn't need to consider freebie comments at that time. 

But since this PR changes the base cost for comments: do we still want to allow freebies if `cost <= baseCost`? Or only if `cost <= 1`?

I think we can ship this as-is (with `cost <= baseCost` as part of the freebie condition) and see how stackers respond.[^1]

[^1]: me totally not trying to simulate @huumn's brain but I do think that's the way to go independently of the simulation

edit: I also wonder if we should use the following UI:

2024-12-29-025159_260x195_scrot

instead of 

2024-12-29-025340_258x214_scrot


<details>
<summary>patch</summary>

```diff
diff --git a/components/territory-header.js b/components/territory-header.js
index a53a99e8..8d64f014 100644
--- a/components/territory-header.js
+++ b/components/territory-header.js
@@ -57,13 +57,16 @@ export function TerritoryInfo ({ sub }) {
           <span> on </span>
           <span className='fw-bold'>{new Date(sub.createdAt).toDateString()}</span>
         </div>
-        <div className='text-muted'>
-          <span>post cost </span>
-          <span className='fw-bold'>{numWithUnits(sub.baseCost)}</span>
-        </div>
-        <div className='text-muted'>
-          <span>reply cost </span>
-          <span className='fw-bold'>{numWithUnits(sub.replyCost)}</span>
+        <div className='d-flex'>
+          <div className='text-muted'>
+            <span>post cost </span>
+            <span className='fw-bold'>{numWithUnits(sub.baseCost)}</span>
+          </div>
+          <span className='px-1'> \ </span>
+          <div className='text-muted'>
+            <span>reply cost </span>
+            <span className='fw-bold'>{numWithUnits(sub.replyCost)}</span>
+          </div>
         </div>
         <TerritoryBillingLine sub={sub} />
       </CardFooter>
```

</details>

components/reply.js
Soxasora: ah, I totally forgot saloon, applied!

api/paidAction/itemCreate.js
Soxasora: Opinionated or not, to have these functions as reusable (and readable!) is a good thing!

api/paidAction/itemCreate.js
Soxasora: oops... sorry, cleaned 👍

Soxasora:
Thanks for the exhaustive review ek!
I've applied all the suggestions and updated `territory-header` component to your design because I think it's cleaner and still readable (if you read post cost, you're gonna also read reply cost).

Regarding `cost <= baseCost`, following the query it seems that `cost` will never be less than `baseCost`.

api/paidAction/itemCreate.js
```diff
@@ -11,9 +11,34 @@ export const paymentMethods = [
...
+    include: { sub: true }
+  })
+}
+
+export async function getBaseCost ({ models, bio, parentId, subName }) {
+  if (bio) return BIO_COST
+
+  if (parentId) {
+    const rootItem = await findRootItem(models, parentId)
+    return rootItem.bio ? BIO_COST : satsToMsats(rootItem.sub?.replyCost)
```
ekzyis, line 33: I noticed you need the same check as in components/reply.js:177 here for saloon replies else this will return `null` because `rootItem.sub` is not defined and then `itemCreate` will throw `Cannot convert null to a BigInt`.

```suggestion
    if (rootItem.bio || !rootItem.sub) return BIO_COST
    return satsToMsats(rootItem.sub.replyCost)
```

I would then also rename `BIO_COST` to something else like `DEFAULT_COST` (maybe you have a better idea for name).

ekzyis (changes requested):
Missed in my previous review that this will also need the same check in the backend

api/paidAction/itemCreate.js
Soxasora: After applying today's patch I moved on and missed it 🫠, sorry and thanks ek!

ekzyis (approved):

huumn:
I made a few nitpick changes:

- any root that's a bio doesn't have a sub - so we only need to check for a sub
- I simplified the root finding logic in `ITEM_CREATE` some
- I moved `reply cost` into the `options` accordion
    - As you've mentioned in chat, we don't want to overload people with form fields, especially if they're optional or have reasonable defaults

I still need to do another review pass and QA, but this should merge tomorrow.

Soxasora:
Thanks a lot for the great changes! As my experience with the codebase grows, I see myself miss less and ship more.

> I moved reply cost into the options accordion
Perfect, it's more appropriate here than elsewhere 👀

PR #1769, opened 2024-12-27
support generated fields on serverside wallets

riccardobl:
## Description

Client-side wallets can transform credentials during initial configuration. Previously, this functionality was also available for server-side wallets, but a recent refactoring broke it. This PR restores it, as it is required for [stackernews/stacker.news#1763](https://github.com/stackernews/stacker.news/pull/1763).

## Checklist

yes

tested as part of https://github.com/stackernews/stacker.news/pull/1763

no

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -47,13 +47,18 @@ function injectResolvers (resolvers) {
         })
       }
 
-      const validData = await validateWallet(walletDef,
-        { ...data, ...settings, vaultEntries: vaultEntries ?? existingVaultEntries },
-        { serverSide: true })
-      if (validData) {
-        data && Object.keys(validData).filter(key => key in data).forEach(key => { data[key] = validData[key] })
-        settings && Object.keys(validData).filter(key => key in settings).forEach(key => { settings[key] = validData[key] })
+      const validate = async ({ data, settings, skipGenerated = false }) => {
+        const validData = await validateWallet(walletDef,
+          { ...data, ...settings, vaultEntries: vaultEntries ?? existingVaultEntries },
+          { serverSide: true, skipGenerated })
+        if (validData) {
+          data && Object.keys(validData).filter(key => key in data).forEach(key => { data[key] = validData[key] })
+          settings && Object.keys(validData).filter(key => key in settings).forEach(key => { settings[key] = validData[key] })
+        }
       }
+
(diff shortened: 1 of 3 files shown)


api/resolvers/wallet.js
```diff
@@ -47,14 +47,19 @@ function injectResolvers (resolvers) {
...
-        data && Object.keys(validData).filter(key => key in data).forEach(key => { data[key] = validData[key] })
-        settings && Object.keys(validData).filter(key => key in settings).forEach(key => { settings[key] = validData[key] })
+      const validate = async ({ data, settings, skipGenerated = false }) => {
+        const validData = await validateWallet(walletDef,
+          { ...data, ...settings, vaultEntries: vaultEntries ?? existingVaultEntries },
+          { serverSide: true, skipGenerated })
+        if (validData) {
+          data && Object.keys(validData).filter(key => key in data).forEach(key => { data[key] = validData[key] })
+          settings && Object.keys(validData).filter(key => key in settings).forEach(key => { settings[key] = validData[key] })
+        }
```
riccardobl, line 57: this becomes a function because we need to run it twice.
- Before testCreateInvoice with skipGenerated=true (because we don't have the generated properties yet)
- After testCreateInvoice with skipGenerated=false

api/resolvers/wallet.js
```diff
@@ -779,6 +784,7 @@ async function upsertWallet (
   if (testCreateInvoice) {
     try {
       await testCreateInvoice(data)
+      await validate({ data, settings, skipGenerated: false })
```
riccardobl, line 787: validate again, since now we'll have the generated fields

wallets/config.js
```diff
@@ -63,7 +63,7 @@ export function useWalletConfigurator (wallet) {
         throw err
       }
     } else if (canReceive({ def: wallet.def, config: serverConfig })) {
-      const transformedConfig = await validateWallet(wallet.def, serverConfig)
+      const transformedConfig = await validateWallet(wallet.def, serverConfig, { skipGenerated: true })
```
riccardobl, line 66: the client  needs to skip generated fields for receivers , because they are generated by the testCreateInvoice that runs serverside

huumn:
Does this currently break something in production?

riccardobl:
> Does this fix something broken in production?

no, i doesn't

Issue #1669, opened 2024-12-01
non custodial poll voting appears buggy


huumn:
It might be another cache race condition and hard to reproduce in dev but I just voted in a pool, the gui kind of flashed and appeared as if I didn't vote, then when I attempted to vote again I got a unique constraint error, then a few seconds later the gui updated to show my vote.

huumn:
Still a problem

Issue #1655, opened 2024-11-27
Zaps in UI occasionally appears to fail but soon succeeds


huumn:
This happens to me once or twice per day in prod.

I'll zap something, the bolt will change color and the sat amount with update, then the bolt will unchange color and the sat amount will un-update. If I navigate around and come back to the item, the zap will be and appear successful.

This happens pretty fast so I suspect the cache is getting updated with the pre-zap values for some reason.

huumn:
It looks like `cache.modify` takes an `optimistic` option which makes it update optimistic caches in addition to the "main" cache. I think that might be the fix for this.

PR #1772, opened 2024-12-27
make possibly conflicting optimistic cache updates write to each other

huumn:
My theory as to why #1655 and #1669 occur is that an earlier and conflicting optimistic mutation fails while another optimistic mutation is in progress. When the optimistic mutation fails, it rolls back the cache to its initial state which causes the in-progress optimistic mutation to have its cache updates overwritten.

In the case of #1655, I believe that this occurs when an earlier zap (in a succession of zaps) fails. In the case of #1669, I believe this occurs when a poll is zapped, then voted on.

I've made the relevant cache updates write to each other. It's very hard to reproduce so I'm not sure I got all the relevant `cache.modify` calls, but I'll continue looking for any cache updates on the optimistic response path in the few places we do it (zaps/boosts/polls) ... it's unclear if there are consequences to doing it on every `cache.modify` call. 

--------------

Update: I've made all the `cache.modify` calls on items update optimistic caches. This might be unnecessary overhead but afaict it shouldn't cause anything bad to happen, and should prevent optimistic rollbacks from affecting these cache updates.

### DIFF
--- a/components/bookmark.js
+++ b/components/bookmark.js
@@ -17,7 +17,8 @@ export default function BookmarkDropdownItem ({ item: { id, meBookmark } }) {
           id: `Item:${id}`,
           fields: {
             meBookmark: () => bookmarkItem.meBookmark
-          }
+          },
+          optimistic: true
         })
       }
     }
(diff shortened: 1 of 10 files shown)


huumn:
I'm going to leave those issues open just in case this doesn't fix it.

PR #1775, opened 2024-12-29
Fix muted parents not uncollapsed

### LINKED ISSUE #1706: Scroll to comment with muted parent doesn't work

### Describe the bug

If someone in a thread is muted, the thread is not uncollapsed so the comment we want to scroll to is never in view.

### Screenshots

_No response_

### Steps To Reproduce

1. Create a post
2. Receive a reply from someone muted
3. Someone who you didn't mute replies to muted person so you receive a notification
4. Click on notification

### Expected behavior

Thread is uncollapsed so I can see the comment I wanted to see

### Logs

_No response_

### Device information

_No response_

### Additional context

Currently, I have to click on "parent" in my notifications to make sure I always can see the comment even if there is someone muted in the thread. Then I have to click on "parent" enough times to see the whole thread.

ekzyis:
## Description

This fixes #1706 by populating a new context for query params with the full comment so we can use `comment.path` to uncollapse all parents.

## Video

https://github.com/user-attachments/assets/1097e4e9-729d-4d3b-8abf-0031aa2d1e89

## Additional Context

I named this new context `QueryParamsContext` since I struggled with a better name and we might want to use this for other stuff at some point even though I currently lack the imagination for what. Open for better naming ideas.

## Checklist

yes

`5`. See video. Also made sure in 91645e1e that if one navigates back and then forward again, the thread is still uncollapsed by storing the state in local storage.

n/a

no

### DIFF
--- a/components/comment.js
+++ b/components/comment.js
@@ -26,6 +26,7 @@ import { commentSubTreeRootId } from '@/lib/item'
 import Pin from '@/svgs/pushpin-fill.svg'
 import LinkToContext from './link-to-context'
 import Boost from './boost-button'
+import { gql, useApolloClient } from '@apollo/client'
 
 function Parent ({ item, rootText }) {
   const root = useRoot()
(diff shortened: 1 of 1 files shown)


huumn (changes requested):
This seems kind of crazy tbh.

If the comment is on the page, it's in the apollo cache isn't it, and you can just query the cache for the path?

If that isn't the case, then it might be better to just add more state to the url, e.g. `?commentPath=1.2.3.4`

ekzyis:
Wow, you are right, this was an impressive brain fart.

Changed it to use `cache.readFragment` now.

Issue #1734, opened 2024-12-18
Async invoice creation


ekzyis:
### Describe the problem you're trying to solve

Receiver wallets can take pretty long to create invoices. Since our mutation waits for wallets to create an invoice, our load balancer might timeout before this happens.

### Describe the solution you'd like

Instead of increasing timeouts for wallets and load balancer, we could return an id that the frontend can use to poll for the invoice.

This might not be simply the invoice id since we only store invoices in the database when we actually have something to store.

### Additional context

Doing https://github.com/stackernews/stacker.news/issues/1717 first might be useful.

https://github.com/stackernews/stacker.news/issues/1492 might also be related since this ticket is also about sending invoices to the client. Ideally, the code for #1492 can also be used for async invoice creations.

ekzyis:
Doing https://github.com/stackernews/stacker.news/issues/1717 first might be useful

PR #1776, opened 2024-12-30
Automated retries

### LINKED ISSUE #1492: automatic retries of failed paid actions
currently when a paid action fails, we notify the stacker. It would be better UX to simply retry N times in the background before notifying them.

ekzyis:
## Description

close #1492 based on #1785, #1787

All failed invoices are returned to every client periodically via a new query `failedInvoices`. To make sure an invoice is only retried by one client at a time, we implemented expiring locks: we first try to update a new timestamp column `retryPendingSince` on the invoice. Only one client at a time will be able to successfully update the invoice. This lock expires after a minute after which another retry for the same invoice can be attempted by a client.

To stop after three payment attempts (= two retries), a new integer column `Invoice.paymentAttempt` tracks at which payment attempt we are. This number is increased when we retry an invoice and pass `newAttempt: true` which we do when we retry these fetched failed invoices. When the number is increased, the payment will start from the beginning with all sender and receiver wallets available.

TODO:

- [x] retry returned failed invoices on client
- [x] count retries so we can stop returning them after we hit a limit
- [x] only show failed invoices that have exhausted retries in notifications
- [x] don't retry invoices that have been manually cancelled?

added `userCancel` column, see #1785

- [x] don't return failed invoices to clients that don't have send wallets

client only polls when it has send wallets

- [x] don't return intermediate failed invoices that will be retried due to sender or receiver fallbacks

added `"cancelledAt" < now() - interval '${WALLET_RETRY_AFTER_MS} milliseconds'` filter

- [x] :warning: make sure when this is deployed, this will not start retrying old failed invoices :warning: 

added `WALLET_RETRY_BEFORE_MS` used in this filter:

```sql
AND now()::timestamp <@ tsrange(
  "cancelledAt" + interval '${WALLET_RETRY_AFTER_MS} milliseconds',
  "cancelledAt" + interval '${WALLET_RETRY_BEFORE_MS} milliseconds'
)
```

- [x] use lock in retry mutation

new TODO since feedback:

- [x] also update `hasNewNotes`
- [x] fix no notification if no wallets are attached and QR payment fails since they are never retried

## Additional Context

- this will still retry invoices that have been created before deployment and failed if they aren't older than a day
- if a client detaches all wallets after they tried a payment for the first time, it will never show up in notifications as failed since it wasn't retried often enough. I consider this to be an edge case we don't need to worry about.
- ~see https://github.com/stackernews/stacker.news/pull/1776/files#r1907791409~

## Checklist

yes

`7`. Tested automated retries for posting, replies and zapping with this patch:

```diff
diff --git a/api/resolvers/wallet.js b/api/resolvers/wallet.js
index 8e6f3f0e..fb9edd2f 100644
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -469,7 +469,6 @@ const resolvers = {
         SELECT * FROM "Invoice"
         WHERE "userId" = ${me.id}
         AND "actionState" = 'FAILED'
-        AND "userCancel" = false
         AND "cancelledAt" < now() - ${`${WALLET_RETRY_AFTER_MS} milliseconds`}::interval
         AND "cancelledAt" > now() - ${`${WALLET_RETRY_BEFORE_MS} milliseconds`}::interval
         AND "paymentAttempt" < ${WALLET_MAX_RETRIES}
diff --git a/lib/constants.js b/lib/constants.js
index fe46b71a..09b6981a 100644
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -198,7 +198,7 @@ export const WALLET_CREATE_INVOICE_TIMEOUT_MS = 45_000
 // interval between which failed invoices are returned to a client for automated retries.
 // retry-after must be high enough such that intermediate failed invoices that will already
 // be retried by the client due to sender or receiver fallbacks are not returned to the client.
-export const WALLET_RETRY_AFTER_MS = 60_000 // 1 minute
+export const WALLET_RETRY_AFTER_MS = 5_000 // 1 minute
 export const WALLET_RETRY_BEFORE_MS = 86_400_000 // 24 hours
 // we want to attempt a payment three times so we retry two times
 export const WALLET_MAX_RETRIES = 2
@@ -277,7 +277,7 @@ function RetryHandler ({ children }) {
           console.error('retry poll failed:', err)
         }
         if (!stopped) queuePoll()
-      }, NORMAL_POLL_INTERVAL)
+      }, 1_000)
     }
     const stopPolling = () => {
       stopped = true
```

Simulate multiple clients with this patch:

```diff
diff --git a/wallets/index.js b/wallets/index.js
index 42fb55be..8a9f5f3f 100644
--- a/wallets/index.js
+++ b/wallets/index.js
@@ -255,9 +255,9 @@ function RetryHandler ({ children }) {
         return
       }

-      for (const inv of failedInvoices) {
+      for (const inv of [...failedInvoices, ...failedInvoices]) {
         try {
-          await retry(inv)
+          retry(inv).catch(err => console.error('retry failed:', err))
         } catch (err) {
           // some retries are expected to fail since only one client at a time is allowed to retry
           // these should show up as 'invoice not found' errors
```

Test p2p zaps with this patch that makes forwards fail and disables the fallback to CCs:

```diff
diff --git a/api/paidAction/index.js b/api/paidAction/index.js
index 8feabc30..031e4d7e 100644
--- a/api/paidAction/index.js
+++ b/api/paidAction/index.js
@@ -355,6 +355,7 @@ export async function retryPaidAction (actionType, args, incomingContext) {
       invoiceArgs = { bolt11, wrappedBolt11, wallet, maxFee }
     } catch (err) {
       console.log('failed to retry wrapped invoice, falling back to SN:', err)
+      throw err
     }
   }
diff --git a/worker/paidAction.js b/worker/paidAction.js
index 9b3ecb5a..b9172ebf 100644
--- a/worker/paidAction.js
+++ b/worker/paidAction.js
@@ -268,7 +268,7 @@ export async function paidActionForwarding ({ data: { invoiceId, ...args }, mode
     payViaPaymentRequest({
       lnd,
       request: bolt11,
-      max_fee_mtokens: String(maxFeeMsats),
+      max_fee_mtokens: String(0),
       pathfinding_timeout: LND_PATHFINDING_TIMEOUT_MS,
       confidence: LND_PATHFINDING_TIME_PREF_PPM,
       max_timeout_height: maxTimeoutHeight
```

n/a

no

### DIFF
--- a/api/resolvers/paidAction.js
+++ b/api/resolvers/paidAction.js
@@ -69,12 +69,12 @@ export default {
 
       // a locked invoice means we want to retry a payment from the beginning
       // with all sender and receiver wallets so we need to increment the retry count
-      const retry = invoice.lockedAt ? invoice.retry + 1 : invoice.retry
-      if (retry > WALLET_MAX_RETRIES) {
+      const paymentAttempt = invoice.lockedAt ? invoice.paymentAttempt + 1 : invoice.paymentAttempt
+      if (paymentAttempt > WALLET_MAX_RETRIES) {
         throw new Error('Payment has been retried too many times')
       }
 
-      const result = await retryPaidAction(invoice.actionType, { invoice }, { retry, models, me, lnd })
+      const result = await retryPaidAction(invoice.actionType, { invoice }, { paymentAttempt, models, me, lnd })
 
       return {
         ...result,
--- a/api/resolvers/paidAction.js
+++ b/api/resolvers/paidAction.js
@@ -58,7 +58,7 @@ export default {
       // make sure only one client at a time can retry by acquiring a lock that expires
       const [invoice] = await models.$queryRaw`
           UPDATE "Invoice"
-          SET "retryPendingSince" = CURRENT_TIMESTAMP
+          SET "retryPendingSince" = now()
           WHERE
             id = ${invoiceId} AND
             "userId" = ${me.id} AND
--- a/api/resolvers/paidAction.js
+++ b/api/resolvers/paidAction.js
@@ -67,11 +67,14 @@ export default {
         throw new Error(`Invoice is not in failed state: ${invoice.actionState}`)
       }
 
-      if (invoice.retry >= WALLET_MAX_RETRIES) {
+      // a locked invoice means we want to retry a payment from the beginning
+      // with all sender and receiver wallets so we need to increment the retry count
+      const retry = invoice.lockedAt ? invoice.retry + 1 : invoice.retry
+      if (retry > WALLET_MAX_RETRIES) {
         throw new Error('Payment has been retried too many times')
       }
 
-      const result = await retryPaidAction(invoice.actionType, { invoice }, { models, me, lnd })
+      const result = await retryPaidAction(invoice.actionType, { invoice }, { retry, models, me, lnd })
 
       return {
         ...result,
--- a/api/resolvers/paidAction.js
+++ b/api/resolvers/paidAction.js
@@ -50,35 +50,44 @@ export default {
     }
   },
   Mutation: {
-    retryPaidAction: async (parent, { invoiceId }, { models, me, lnd }) => {
+    retryPaidAction: async (parent, { invoiceId, newAttempt }, { models, me, lnd }) => {
       if (!me) {
         throw new Error('You must be logged in')
       }
 
-      const invoice = await models.invoice.findUnique({ where: { id: invoiceId, userId: me.id } })
+      // make sure only one client at a time can retry by immediately transitioning to an intermediate state
+      const [invoice] = await models.$queryRaw`
+          UPDATE "Invoice"
+          SET "actionState" = 'RETRY_PENDING'
+          WHERE id in (
+            SELECT id FROM "Invoice"
+            WHERE id = ${invoiceId} AND "userId" = ${me.id} AND "actionState" = 'FAILED'
+            FOR UPDATE
+          )
+          RETURNING *`
       if (!invoice) {
         throw new Error('Invoice not found')
       }
 
-      if (invoice.actionState !== 'FAILED') {
-        if (invoice.actionState === 'PAID') {
-          throw new Error('Invoice is already paid')
-        }
-        throw new Error(`Invoice is not in failed state: ${invoice.actionState}`)
-      }
-
-      // a locked invoice means we want to retry a payment from the beginning
-      // with all sender and receiver wallets so we need to increment the retry count
-      const paymentAttempt = invoice.lockedAt ? invoice.paymentAttempt + 1 : invoice.paymentAttempt
+      // do we want to retry a payment from the beginning with all sender and receiver wallets?
+      const paymentAttempt = newAttempt ? invoice.paymentAttempt + 1 : invoice.paymentAttempt
       if (paymentAttempt > WALLET_MAX_RETRIES) {
         throw new Error('Payment has been retried too many times')
       }
 
-      const result = await retryPaidAction(invoice.actionType, { invoice }, { paymentAttempt, models, me, lnd })
-
-      return {
-        ...result,
-        type: paidActionType(invoice.actionType)
+      try {
+        const result = await retryPaidAction(invoice.actionType, { invoice }, { paymentAttempt, models, me, lnd })
+        return {
+          ...result,
+          type: paidActionType(invoice.actionType)
+        }
+      } catch (err) {
+        // revert state transition to allow new retry for this invoice
+        await models.invoice.update({
+          where: { id: invoiceId, actionState: 'RETRY_PENDING' },
+          data: { actionState: 'FAILED' }
+        })
+        throw err
       }
     }
   },
--- a/api/resolvers/paidAction.js
+++ b/api/resolvers/paidAction.js
@@ -1,5 +1,5 @@
 import { retryPaidAction } from '../paidAction'
-import { USER_ID } from '@/lib/constants'
+import { USER_ID, WALLET_MAX_RETRIES } from '@/lib/constants'
 
 function paidActionType (actionType) {
   switch (actionType) {
@@ -65,6 +65,10 @@ export default {
           throw new Error('Invoice is already paid')
         }
         throw new Error(`Invoice is not in failed state: ${invoice.actionState}`)
+      }
+
+      if (invoice.retry >= WALLET_MAX_RETRIES) {
+        throw new Error('Payment has been retried too many times')
       }
 
       const result = await retryPaidAction(invoice.actionType, { invoice }, { models, me, lnd })
--- a/api/resolvers/paidAction.js
+++ b/api/resolvers/paidAction.js
@@ -1,5 +1,5 @@
 import { retryPaidAction } from '../paidAction'
-import { USER_ID, WALLET_MAX_RETRIES } from '@/lib/constants'
+import { USER_ID, WALLET_MAX_RETRIES, WALLET_RETRY_TIMEOUT_MS } from '@/lib/constants'
 
 function paidActionType (actionType) {
   switch (actionType) {
@@ -55,18 +55,18 @@ export default {
         throw new Error('You must be logged in')
       }
 
-      // make sure only one client at a time can retry by immediately transitioning to an intermediate state
+      // make sure only one client at a time can retry by acquiring a lock that expires
       const [invoice] = await models.$queryRaw`
           UPDATE "Invoice"
-          SET "actionState" = 'RETRY_PENDING'
-          WHERE id in (
-            SELECT id FROM "Invoice"
-            WHERE id = ${invoiceId} AND "userId" = ${me.id} AND "actionState" = 'FAILED'
-            FOR UPDATE
-          )
+          SET "retryPendingSince" = CURRENT_TIMESTAMP
+          WHERE
+            id = ${invoiceId} AND
+            "userId" = ${me.id} AND
+            "actionState" = 'FAILED' AND
+            ("retryPendingSince" IS NULL OR "retryPendingSince" < now() - ${`${WALLET_RETRY_TIMEOUT_MS} milliseconds`}::interval)
           RETURNING *`
       if (!invoice) {
-        throw new Error('Invoice not found')
+        throw new Error('Invoice not found or retry pending')
       }
 
       // do we want to retry a payment from the beginning with all sender and receiver wallets?
@@ -75,19 +75,10 @@ export default {
         throw new Error('Payment has been retried too many times')
       }
 
-      try {
-        const result = await retryPaidAction(invoice.actionType, { invoice }, { paymentAttempt, models, me, lnd })
-        return {
-          ...result,
-          type: paidActionType(invoice.actionType)
-        }
-      } catch (err) {
-        // revert state transition to allow new retry for this invoice
-        await models.invoice.update({
-          where: { id: invoiceId, actionState: 'RETRY_PENDING' },
-          data: { actionState: 'FAILED' }
-        })
-        throw err
+      const result = await retryPaidAction(invoice.actionType, { invoice }, { paymentAttempt, models, me, lnd })
+      return {
+        ...result,
+        type: paidActionType(invoice.actionType)
       }
     }
   },
--- a/api/resolvers/paidAction.js
+++ b/api/resolvers/paidAction.js
@@ -76,6 +76,7 @@ export default {
       }
 
       const result = await retryPaidAction(invoice.actionType, { invoice }, { paymentAttempt, models, me, lnd })
+
       return {
         ...result,
         type: paidActionType(invoice.actionType)
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -4,9 +4,9 @@ import { decodeCursor, LIMIT, nextCursorEncoded } from '@/lib/cursor'
 import { msatsToSats } from '@/lib/format'
 import { bioSchema, emailSchema, settingsSchema, validateSchema, userSchema } from '@/lib/validate'
 import { getItem, updateItem, filterClause, createItem, whereClause, muteClause, activeOrMine } from './item'
-import { USER_ID, RESERVED_MAX_USER_ID, SN_NO_REWARDS_IDS, INVOICE_ACTION_NOTIFICATION_TYPES, WALLET_MAX_RETRIES } from '@/lib/constants'
+import { USER_ID, RESERVED_MAX_USER_ID, SN_NO_REWARDS_IDS, INVOICE_ACTION_NOTIFICATION_TYPES, WALLET_MAX_RETRIES, WALLET_RETRY_BEFORE_MS } from '@/lib/constants'
 import { viewGroup } from './growth'
-import { timeUnitForRange, whenRange } from '@/lib/time'
+import { datePivot, timeUnitForRange, whenRange } from '@/lib/time'
 import assertApiKeyNotPermitted from './apiKey'
 import { hashEmail } from '@/lib/crypto'
 import { isMuted } from '@/lib/user'
@@ -557,6 +557,11 @@ export default {
                   },
                   {
                     userCancel: true
+                  },
+                  {
+                    cancelledAt: {
+                      lte: datePivot(new Date(), { milliseconds: -WALLET_RETRY_BEFORE_MS })
+                    }
                   }
                 ]
               }
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -544,29 +544,25 @@ export default {
             in: INVOICE_ACTION_NOTIFICATION_TYPES
           },
           actionState: 'FAILED',
-          ...(user.sendWallets
-            ? {
-                OR: [
-                  {
-                    paymentAttempt: {
-                      gte: WALLET_MAX_RETRIES
-                    }
-                  },
-                  {
-                    actionType: 'ITEM_CREATE'
-                  },
-                  {
-                    userCancel: true
-                  },
-                  {
-                    cancelledAt: {
-                      lte: datePivot(new Date(), { milliseconds: -WALLET_RETRY_BEFORE_MS })
-                    }
-                  }
-                ]
+          OR: [
+            {
+              paymentAttempt: {
+                gte: WALLET_MAX_RETRIES
               }
-            : {})
-
+            },
+            {
+              actionType: 'ITEM_CREATE'
+            },
+            {
+              userCancel: true
+            },
+            {
+              // TODO: test this since invoice is not updated after retry-before
+              cancelledAt: {
+                lte: datePivot(new Date(), { milliseconds: -WALLET_RETRY_BEFORE_MS })
+              }
+            }
+          ]
         }
       })
 
@@ -886,14 +882,6 @@ export default {
 
       await models.user.update({ where: { id: me.id }, data: { hideWelcomeBanner: true } })
       return true
-    },
-    setSendWallets: async (parent, { sendWallets }, { me, models }) => {
-      if (!me) {
-        throw new GqlAuthenticationError()
-      }
-
-      await models.user.update({ where: { id: me.id }, data: { sendWallets } })
-      return sendWallets
     }
   },
 
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -551,9 +551,6 @@ export default {
               }
             },
             {
-              actionType: 'ITEM_CREATE'
-            },
-            {
               userCancel: true
             },
             {
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -4,7 +4,7 @@ import { decodeCursor, LIMIT, nextCursorEncoded } from '@/lib/cursor'
 import { msatsToSats } from '@/lib/format'
 import { bioSchema, emailSchema, settingsSchema, validateSchema, userSchema } from '@/lib/validate'
 import { getItem, updateItem, filterClause, createItem, whereClause, muteClause, activeOrMine } from './item'
-import { USER_ID, RESERVED_MAX_USER_ID, SN_NO_REWARDS_IDS, INVOICE_ACTION_NOTIFICATION_TYPES } from '@/lib/constants'
+import { USER_ID, RESERVED_MAX_USER_ID, SN_NO_REWARDS_IDS, INVOICE_ACTION_NOTIFICATION_TYPES, WALLET_MAX_RETRIES } from '@/lib/constants'
 import { viewGroup } from './growth'
 import { timeUnitForRange, whenRange } from '@/lib/time'
 import assertApiKeyNotPermitted from './apiKey'
@@ -543,7 +543,17 @@ export default {
           actionType: {
             in: INVOICE_ACTION_NOTIFICATION_TYPES
           },
-          actionState: 'FAILED'
+          actionState: 'FAILED',
+          OR: [
+            {
+              paymentAttempt: {
+                gte: WALLET_MAX_RETRIES
+              }
+            },
+            {
+              actionType: 'ITEM_CREATE'
+            }
+          ]
         }
       })
 
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -544,16 +544,21 @@ export default {
             in: INVOICE_ACTION_NOTIFICATION_TYPES
           },
           actionState: 'FAILED',
-          OR: [
-            {
-              paymentAttempt: {
-                gte: WALLET_MAX_RETRIES
+          ...(user.sendWallets
+            ? {
+                OR: [
+                  {
+                    paymentAttempt: {
+                      gte: WALLET_MAX_RETRIES
+                    }
+                  },
+                  {
+                    actionType: 'ITEM_CREATE'
+                  }
+                ]
               }
-            },
-            {
-              actionType: 'ITEM_CREATE'
-            }
-          ]
+            : {})
+
         }
       })
 
@@ -873,6 +878,14 @@ export default {
 
       await models.user.update({ where: { id: me.id }, data: { hideWelcomeBanner: true } })
       return true
+    },
+    setSendWallets: async (parent, { sendWallets }, { me, models }) => {
+      if (!me) {
+        throw new GqlAuthenticationError()
+      }
+
+      await models.user.update({ where: { id: me.id }, data: { sendWallets } })
+      return sendWallets
     }
   },
 
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -557,7 +557,6 @@ export default {
               userCancel: true
             },
             {
-              // TODO: test this since invoice is not updated after retry-before
               cancelledAt: {
                 lte: datePivot(new Date(), { milliseconds: -WALLET_RETRY_BEFORE_MS })
               }
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -554,6 +554,9 @@ export default {
                   },
                   {
                     actionType: 'ITEM_CREATE'
+                  },
+                  {
+                    userCancel: true
                   }
                 ]
               }
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -552,17 +552,37 @@ export default {
             },
             {
               userCancel: true
-            },
-            {
-              cancelledAt: {
-                lte: datePivot(new Date(), { milliseconds: -WALLET_RETRY_BEFORE_MS })
-              }
             }
           ]
         }
       })
 
       if (invoiceActionFailed) {
+        foundNotes()
+        return true
+      }
+
+      const invoiceActionFailed2 = await models.invoice.findFirst({
+        where: {
+          userId: me.id,
+          updatedAt: {
+            gt: datePivot(lastChecked, { milliseconds: -WALLET_RETRY_BEFORE_MS })
+          },
+          actionType: {
+            in: INVOICE_ACTION_NOTIFICATION_TYPES
+          },
+          actionState: 'FAILED',
+          paymentAttempt: {
+            lt: WALLET_MAX_RETRIES
+          },
+          userCancel: false,
+          cancelledAt: {
+            lte: datePivot(new Date(), { milliseconds: -WALLET_RETRY_BEFORE_MS })
+          }
+        }
+      })
+
+      if (invoiceActionFailed2) {
         foundNotes()
         return true
       }
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -11,7 +11,8 @@ import {
   PAID_ACTION_PAYMENT_METHODS,
   WALLET_CREATE_INVOICE_TIMEOUT_MS,
   WALLET_RETRY_AFTER_MS,
-  WALLET_RETRY_BEFORE_MS
+  WALLET_RETRY_BEFORE_MS,
+  WALLET_MAX_RETRIES
 } from '@/lib/constants'
 import { amountSchema, validateSchema, withdrawlSchema, lnAddrSchema } from '@/lib/validate'
 import assertGofacYourself from './ofac'
@@ -480,6 +481,7 @@ const resolvers = {
               "cancelledAt" + $3::interval
             )
             AND "lockedAt" IS NULL
+            AND "retry" < $4
             ORDER BY id DESC
             FOR UPDATE SKIP LOCKED
           )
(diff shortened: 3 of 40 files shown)


api/resolvers/wallet.js
```diff
@@ -456,6 +459,48 @@ const resolvers = {
...
+      return await models.$queryRawUnsafe(`
+        WITH failed AS (
+          UPDATE "Invoice"
+          SET "lockedAt" = now()
+          WHERE id IN (
+            SELECT id FROM "Invoice"
+            WHERE "userId" = $1
+            AND "actionState" = 'FAILED'
+            AND "userCancel" = false
+            AND now()::timestamp <@ tsrange(
```
huumn, line 479: I haven't looked at much else of the PR, but we're going to want to double check that this tsrange operation uses an index.

We also don't need a tsrange, right? If it's just syntactic sugar it might be worth avoiding so we can use `$queryRaw`.

wallets/index.js
```diff
@@ -221,7 +225,38 @@ export function useWallet (name) {
...
+
+function RetryHandler ({ children }) {
+  const failedInvoices = useFailedInvoices()
+  const waitForWalletPayment = useWalletPayment()
+  const invoiceHelper = useInvoice()
+
+  useEffect(() => {
+    (async () => {
+      for (const invoice of failedInvoices) {
+        // TODO: don't retry forever
```
ekzyis, line 241: whoops, forgot to remove TODO

components/use-invoice.js
```diff
@@ -53,5 +53,5 @@ export default function useInvoice () {
     return newInvoice
   }, [retryPaidAction])
 
-  return { cancel, retry, isInvoice }
+  return useMemo(() => ({ cancel, retry, isInvoice }), [cancel, retry, isInvoice])
```
ekzyis, line 56: This `useMemo` and the one in `useSendWallets` is pretty important since this avoid that `waitForWalletPayment` changes between renders which would mean that we run `useEffect` with the same failed invoices multiple times.

In general, this means it is very important that `waitForWalletPayment` is stable across renders.

wallets/index.js
```diff
@@ -221,7 +225,38 @@ export function useWallet (name) {
...
+    })()
+  }, [failedInvoices, invoiceHelper, waitForWalletPayment])
+
+  return children
+}
+
+function useFailedInvoices () {
+  const wallets = useSendWallets()
+
+  // TODO: use longer poll interval in prod?
```
ekzyis, line 254: Polling for failed invoices every second was nice during development, but I think in prod every 30s is enough, so we'd want to use `NORMAL_POLL_INTERVAL`.

wallets/index.js
```diff
@@ -204,7 +206,9 @@ export function WalletsProvider ({ children }) {
         removeLocalWallets
       }}
     >
-      {children}
+      <RetryHandler>
+        {children}
+      </RetryHandler>
```
ekzyis, line 211: I think we should use more context providers inside context providers. There are too many in pages/_app.js.

For example, we could put all providers for the carousel (price, block height, chain fee) into the same component.

api/resolvers/wallet.js
ekzyis: I found `tsrange` a lot easier to reason about but you were right, it wasn't using an index. Range queries like `<@` also require a special index.

As suggested, I changed it to use `$queryRaw` with `<` and `>` which can use a b-tree index in ccbde0d3.

api/resolvers/wallet.js
```diff
@@ -456,6 +459,42 @@ const resolvers = {
...
+          )
+          RETURNING *
+        ),
+        _ AS (
+          INSERT INTO pgboss.job (name, data, startafter, keepuntil)
+          SELECT
+            'unlockInvoice',
+            jsonb_build_object('id', id),
+            now() + interval '10 minutes',
+            now() + interval '15 minutes'
```
huumn, line 494: This is a cool construct. 

1. Can you explain your reasoning behind the times for unlocking?
2. As to why it needs the lock, is this correct: it *only* exists to prevent races creating the invoice (either wrapped or direct to SN) that the race winner will pay? 
3. If I have (2) correct, did you consider skip locking in `retryPaidAction` in a separate tx0, before the invoice was created, rather than when it's fetched? Then, it'd be implicitly unlocked in tx1, where the invoice is set to `RETRYING`? I guess it's possible tx1 fails, and doesn't unlock by setting to `RETRYING`, but maybe that can be fixed by having tx0 set an expiring lock like `lockedAt is null OR lockedAt < now() - interval '2 minutes'`
    - I'm just trying to brainstorm ways to avoid the pgboss job (async stuff like this tends to be the most confusing for me)

ekzyis: > Can you explain your reasoning behind the times for unlocking?

I didn't think too much about when exactly it should unlock. It should just be high enough such that we can reasonably assume that the client who received and locked this invoice is not going to retry this invoice. I think we can probably safely decrease the unlock time 10 minutes to 1 minute.

> As to why it needs the lock, is this correct: it only exists to prevent races creating the invoice (either wrapped or direct to SN) that the race winner will pay?

Mhh, yes, by making sure we only hand out invoices to one client at a time to avoid any such races.

> I guess it's possible tx1 fails, and doesn't unlock by setting to `RETRYING`, but maybe that can be fixed by having tx0 set an expiring lock like `lockedAt is null OR lockedAt < now() - interval '2 minutes'`

Yeah, an expiring lock would be the alternative to the async stuff.

huumn: I just noticed you didn't address (3) exactly. I might have been confusing. 

I don't think we should lock on query - we should lock in `retryPaidAction`. ie multiple clients can call `retryPaidAction`, but only one will successfully get an invoice to retry.

Is that not possible or is the current way better for some reason?

huumn: bump

ekzyis: whoops, forgot to reply

> I just noticed you didn't address (3) exactly. I might have been confusing.

I didn't address (3) exactly, only the async stuff because of

> I'm just trying to brainstorm ways to avoid the pgboss job

so I thought that was the main issue and didn't elaborate on the "lock location"

> I don't think we should lock on query - we should lock in retryPaidAction. ie multiple clients can call retryPaidAction, but only one will successfully get an invoice to retry.
>
> Is that not possible or is the current way better for some reason?

I am also using the lock set on query to know if the payment attempt counter should be increased in the retry mutation.

I could increase the payment attempt counter on the client and the client includes it in the retry mutation but letting the client pick that number makes this a little harder to think about for example wrt trusting inputs :thinking:

huumn: > so I thought that was the main issue and didn't elaborate on the "lock location"

The lock location is only an issue because they way it's written is `request -> query-gets-a-lock -> request -> retry-makes-invoice` vs `request -> query -> request -> retry-gets-a-lock-then-makes-invoice` ... It seems a little odd to grab a lock across requests if we don't have to, doesn't it? Maybe I'm missing something though.

> I could increase the payment attempt counter on the client.

I don't see why that would change with the way I'm thinking it should work. We still acquire the lock - the only difference is that it doesn't happen in a separate graphql query.

pseudocode:
```js
func retryPaidAction(invoiceId) {
   if (tx0.getInvoiceLock(invoiceId)) {
       newInvoice = makeNewBolt11(invoiceId)
       tx1.markInvoiceRetrying(invoiceId)
       return tx1.createDbInvoice(newInvoice)
   }

   throw new Error ("Invoice already retrying")
}
```

ekzyis: > It seems a little odd to grab a lock across requests if we don't have to, doesn't it? Maybe I'm missing something though.

I think you're missing the importance of [this line](https://github.com/stackernews/stacker.news/pull/1776/files#diff-b7c4a08fb27c07ba98911a0feffc9fd5117111337487bc2c6fac63f76f31ec8cR72):

```js
// a locked invoice means we want to retry a payment from the beginning
// with all sender and receiver wallets so we need to increment the retry count
const paymentAttempt = invoice.lockedAt ? invoice.paymentAttempt + 1 : invoice.paymentAttempt
```

As it's currently implemented, the invoice is immediately locked when returned to the client so when the client retries, we know that's a retry that should increment the payment counter because the invoice is locked.

Your pseudo code would always try to acquire a lock for an invoice which prevents multiple clients retrying, yes, but afaict, it doesn't allow to distinguish when the payment counter should be incremented.

Maybe that's weird that the lock is also used to increment the counter but I thought it was a nice side-effect since it's literally just one line of code in the retry mutation (the line above).

huumn: Basically, anytime an invoice is transitioned from `FAILED` to `RETRYING`, we want this incremented, right?

I admit I haven't read the code in detail, so if you're certain that what I'm asking makes that impossible, then I'll accept it is until I've taken a closer look.

ekzyis: > Basically, anytime an invoice is transitioned from FAILED to RETRYING, we want this incremented in the new invoice record, right?

No, during sender and receiver fallbacks, we are also retrying but we want to keep the payment attempt counter the same so we can filter receiver wallets based on that. When we retry a locked invoice, we increment the counter to make all sender/receiver wallets available again and start a "new chain of invoices" where all have the same payment attempt counter.

> Anyway, if you're certain that what I'm asking makes that impossible, then I'll accept it is until I've taken a closer look.

Changing the lock+payment attempt counter increment logic is not impossible but I think it's also good enough as it is now

huumn: I guess I don't see why splitting this up into separate requests makes this possible in way that separate txs in `retryPaidAction` wouldn't. I'm mostly suggesting the logic/counting stays the same, but we can get rid of the lock on query by locking before the retry, and if we can't acquire the lock because another client retried it already, returning null/error.

Anyway, I haven't read the code so I'm probably missing something subtle.

ekzyis: > I'm mostly suggesting the logic/counting stays the same, but we can get rid of the lock on query by locking before the retry

Maybe I am actually missing something, but I can't tell from this description how you want to know during retry if you should increment the counter if you lock every invoice before retry. If the logic stays the same, every retry would then increment the counter which would break receiver fallbacks.

So basically, the lock on query allows us to distinguish "normal retries" (= same payment counter) from "full retries" (= increment payment counter).

huumn: Without solving this myself, assuming it can be solved, the easiest way for you to understand might be to ask a riddle: how would you do exactly what you do now without acquiring the lock in the query by acquiring the lock only during the mutation? `hint: you can run multiple txs, and logic in between them, in the same mutation.`

ekzyis: I created an intermediate retry state `RETRY_PENDING` in a604f0b3 to lock during retry:

```js
// make sure only one client at a time can retry by immediately transitioning to an intermediate state
const [invoice] = await models.$queryRaw`
    UPDATE "Invoice"
    SET "actionState" = 'RETRY_PENDING'
    WHERE id in (
      SELECT id FROM "Invoice"
      WHERE id = ${invoiceId} AND "userId" = ${me.id} AND "actionState" = 'FAILED'
      FOR UPDATE
    )
    RETURNING *`
if (!invoice) {
  throw new Error('Invoice not found')
}
```

This intermediate state avoids that concurrent retries request invoices from attached wallets even though their state transition from `FAILED` to  `RETRYING` at the end will fail.

I also updated the documentation that `FAILED` now always transitions to `RETRY_PENDING` before `RETRYING`.

I still have to change the foreign key from `InvoiceForward` to `Invoice`. Will do this now.

api/resolvers/paidAction.js
```diff
@@ -50,35 +50,44 @@ export default {
...
+          RETURNING *`
       if (!invoice) {
         throw new Error('Invoice not found')
       }
 
-      if (invoice.actionState !== 'FAILED') {
-        if (invoice.actionState === 'PAID') {
-          throw new Error('Invoice is already paid')
-        }
-        throw new Error(`Invoice is not in failed state: ${invoice.actionState}`)
```
ekzyis, line 67: removing these error messages probably makes error messages during bugs worse, for example if `retry` via notifications fail

components/use-invoice.js
ekzyis: no longer the case since I am using  `useLazyQuery` + `setTimeout` now instead of relying on `useEffect`

api/resolvers/paidAction.js
huumn: I think the thing to do in this case is return the pending/paid invoice rather than an error.

We should do the same thing in #1669 - if an attempt is made to vote again, just return the original vote.

It's kind of painting over the bug - which is some kind of clientside cache issue - but it's way better UX.

ekzyis: wouldn't this cause the client to retry a paid invoice if the retry mutation doesn't throw?

huumn: Yeah, we'd have to modify the client logic when retrying, or ....

we could just throw a more specific error, catch it, then update the cache appropriately.

ekzyis: I have put this PR back into review after some Q&A of 89a6b42. I think what we discuss here should be done in a separate PR.

huumn: Oh for sure. I was just riffing on your comment. I keep encountering that poll bug.

api/resolvers/paidAction.js
```diff
@@ -50,28 +50,44 @@ export default {
...
-      return {
-        ...result,
-        type: paidActionType(invoice.actionType)
+      try {
+        const result = await retryPaidAction(invoice.actionType, { invoice }, { paymentAttempt, models, me, lnd })
+        return {
+          ...result,
+          type: paidActionType(invoice.actionType)
+        }
+      } catch (err) {
```
huumn, line 84: What happens if `retryPaidAction` causes the server to crash and the catch block isn't hit?

Won't the invoice be kept in `RETRY_PENDING` limbo?

Wouldn't the expiring lock you had previously still work with this retry flow? Or no?

api/resolvers/paidAction.js
```diff
@@ -50,28 +50,44 @@ export default {
...
-      const invoice = await models.invoice.findUnique({ where: { id: invoiceId, userId: me.id } })
+      // make sure only one client at a time can retry by immediately transitioning to an intermediate state
+      const [invoice] = await models.$queryRaw`
+          UPDATE "Invoice"
+          SET "actionState" = 'RETRY_PENDING'
+          WHERE id in (
+            SELECT id FROM "Invoice"
+            WHERE id = ${invoiceId} AND "userId" = ${me.id} AND "actionState" = 'FAILED'
+            FOR UPDATE
+          )
```
huumn, line 66: Why use a set query here? Wouldn't this suffice:

```sql
          UPDATE "Invoice"
          SET "actionState" = 'RETRY_PENDING'
          WHERE id = ${invoiceId} AND "userId" = ${me.id} AND "actionState" = 'FAILED'
          RETURNING *
```

`id` is a unique column so there'd only ever be one.

ekzyis: Mhh, this `SELECT FOR UPDATE` was probably a remainder of the previous query which locked multiple invoices at once and I used `SELECT FOR UPDATE` because I wanted to use `SKIP LOCKED`.

I don't use `SKIP LOCKED` anymore since none of the locks are acquired in long-running transactions. So the update queries never try long to acquire the lock. There would only ever be a delay while another update is trying to immediately acquire the lock or fail but that should be insignificant.

api/resolvers/paidAction.js
ekzyis: > What happens if `retryPaidAction` causes the server to crash and the catch block isn't hit?
>
> Won't the invoice be kept in RETRY_PENDING limbo?

Yes, the retry would be pending forever :eyes: This is definitely the achilles' heel of me trying to integrate the lock into our state machine via a new state. However, this state is irregular because it should only be a temporary state.

I could add an async job again just to make sure invoices are transitioned back to `FAILED` if they have been in `RETRY_PENDING` for too long but we already discussed that we don't like more async stuff.

> Wouldn't the expiring lock you had previously still work with this retry flow? Or no?

You are right, I changed it back to use a timestamp column in 95e4000d. I renamed it from `lockedAt` to `retryPendingSince` since I think that's a more descriptive name for what it is used.

ekzyis:
I have tested 95e4000d again. I used this patch to test the locking and make sure only one client at a time will be able to retry. Essentially, it allows retries of QR codes, makes retrying faster and attempts to retry every invoice twice:

```diff
diff --git a/api/resolvers/wallet.js b/api/resolvers/wallet.js
index 8e6f3f0e..fb9edd2f 100644
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -469,7 +469,6 @@ const resolvers = {
         SELECT * FROM "Invoice"
         WHERE "userId" = ${me.id}
         AND "actionState" = 'FAILED'
-        AND "userCancel" = false
         AND "cancelledAt" < now() - ${`${WALLET_RETRY_AFTER_MS} milliseconds`}::interval
         AND "cancelledAt" > now() - ${`${WALLET_RETRY_BEFORE_MS} milliseconds`}::interval
         AND "paymentAttempt" < ${WALLET_MAX_RETRIES}
diff --git a/lib/constants.js b/lib/constants.js
index 9ec20245..de55ac05 100644
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -201,7 +201,7 @@ export const WALLET_CREATE_INVOICE_TIMEOUT_MS = 45_000
 // interval between which failed invoices are returned to a client for automated retries.
 // retry-after must be high enough such that intermediate failed invoices that will already
 // be retried by the client due to sender or receiver fallbacks are not returned to the client.
-export const WALLET_RETRY_AFTER_MS = 60_000 // 1 minute
+export const WALLET_RETRY_AFTER_MS = 5_000 // 1 minute
 export const WALLET_RETRY_BEFORE_MS = 86_400_000 // 24 hours
 // we want to attempt a payment three times so we retry two times
 export const WALLET_MAX_RETRIES = 2
diff --git a/wallets/index.js b/wallets/index.js
index 42fb55be..ab2580c3 100644
--- a/wallets/index.js
+++ b/wallets/index.js
@@ -255,7 +255,9 @@ function RetryHandler ({ children }) {
         return
       }

-      for (const inv of failedInvoices) {
+      console.log('XXX failedInvoices', failedInvoices)
+
+      for (const inv of [...failedInvoices, ...failedInvoices]) {
         try {
           await retry(inv)
         } catch (err) {
@@ -277,7 +279,7 @@ function RetryHandler ({ children }) {
           console.error('retry poll failed:', err)
         }
         if (!stopped) queuePoll()
-      }, NORMAL_POLL_INTERVAL)
+      }, 1_000)
     }
     const stopPolling = () => {
       stopped = true
```

https://github.com/user-attachments/assets/84c3097b-8145-4c60-b250-f239530a701c

huumn:
#1776 is the year civilization was automatically retied in america

lib/constants.js
```diff
@@ -198,4 +198,14 @@ export const ZAP_UNDO_DELAY_MS = 5_000
...
 
+// interval between which failed invoices are returned to a client for automated retries.
+// retry-after must be high enough such that intermediate failed invoices that will already
+// be retried by the client due to sender or receiver fallbacks are not returned to the client.
+export const WALLET_RETRY_AFTER_MS = 60_000 // 1 minute
+export const WALLET_RETRY_BEFORE_MS = 86_400_000 // 24 hours
+// we want to attempt a payment three times so we retry two times
+export const WALLET_MAX_RETRIES = 2
+// when a pending retry for an invoice should be considered expired and can be attempted again
+export const WALLET_RETRY_TIMEOUT_MS = 60_000 // 1 minute
```
huumn, line 209: We can't be subject to multiple `WALLET_CREATE_INVOICE_TIMEOUT_MS` during a retry right? I remember you mentioning something like that in a previous pr.

huumn (approved):
This looks good to me. I'll QA fully tomorrow and fix any nitpicks (if I have any) tomorrow.

lib/constants.js
ekzyis: Ohhh, good catch!

We actually can be subject to multiple `WALLET_CREATE_INVOICE_TIMEOUT_MS` during a retry because the timeout is applied in `walletCreateInvoice` which is called in the receiver wallet loop until we found a wallet that returned an invoice:

https://github.com/stackernews/stacker.news/blob/95e98501ec3e000c6b3c710d5809c5d04839d2b4/wallets/server.js#L33-L51

**This also affects `WALLET_RETRY_AFTER_MS` afaict.** It exists to avoid returning failed invoices which will already get retried by a client due to sender or receiver fallbacks (the time between `FAILED` and `RETRYING` for an invoice). But since we transition an invoice to `RETRYING` at the end of the mutation and therefore after we called `walletCreateInvoice`, this means an invoice can also stay failed for `n` * `WALLET_CREATE_INVOICE_TIMEOUT_MS` which is currently larger than `WALLET_RETRY_AFTER_MS` for n=2 even though we can assume it will get retried.

I think the solution is to also put a timeout on the full loop instead of only on each external wallet call. To give each wallet independent of order the same amount of time, we could divide the loop timeout by the amount of receiving wallets available. This means that the more receiving wallets you have, the less time is allocated to each wallet. This might be confusing since adding a second receiving wallet can make the first receiving wallet more likely to fail since it effectively has now only half the amount of time to respond as before. But I think having a timeout on the loop that doesn't depend on how many wallets someone has is more important and we can of course tweak the timeouts as we like.

For example, we could use a 1 minute timeout on the full loop so if you have two receiving wallets, each wallet gets 30 seconds to return an invoice.

huumn: This is at least less of an issue than I thought initially. Worst case here is that some unused invoices get created. Initially I thought that multiple payments might be made, but only one call to `retryPaidAction` will succeed because the invoice is transitioned (and returned) conditional on it being in a `FAILED` state.

tl;dr we can keep this as is.

huumn:
If there are no send wallets attached, payment failures never show up in notifications because they are never retried. 

I think we'll probably want another conditional on the notification that says: if it's been 10 minutes (or whatever) and 0 retries have been attempted, display notification. Perhaps we can reuse `WALLET_RETRY_BEFORE_MS` and make it narrower.

lib/constants.js
ekzyis: I think doing https://github.com/stackernews/stacker.news/issues/1734 would be a better idea than any more messing with timeouts anyway

huumn:
Also as noted in chat, the `hasNewNotes` resolver is not updated to reflect the additional conditions on paid action notifications.

huumn:
From chat:

I think the move for no sending wallet is: 

1. just retry (without a qr fallback) whether there's a sending wallet or not. The attempts counter will increase and all will work as intended with just a one line change.
2. Make a special case when the paid action is an `ITEM_CREATE` ... don't auto-retry on `userCancel` but show the notification.

ekzyis:
2c14275 now immediately shows failed payments for posts or comments in /notifications and also allows auto-retries for them:

https://github.com/user-attachments/assets/6f9259b4-0932-40c3-8938-1624f543e048

I also made sure that auto-retries work even if you already retried manually once. It works because we only increment the payment counter if it was automatically retried. This is also why we can retry manually forever.

However, this commit does implement auto-retries even if `userCancel` is set which is against what was mentioned here:

> don't auto-retry on `userCancel` but still show the retry notification.

I think it's better to have automated retries for everything. It doesn't mess with showing them in the notifications. If you're currently looking at your notifications and an invoice is automatically retried, it will update the notification you're looking at.

Or am I missing something why we shouldn't allow automated retries for `userCancel` in this specific case?

We don't do zaps because there we don't know if they wanted to abort the zap by closing the QR code.

But if you posted something and already tried to pay once, I think we can assume they want to continue trying to pay for it.

huumn:
I'll fully QA tomorrow morning, but `ITEM_CREATE`-cancels without a send wallet don't trigger `hasNewNotes`.

ekzyis:
> `ITEM_CREATE`-cancels without a send wallet don't trigger `hasNewNotes`.

Did you wait at least 30 seconds? That's the poll interval:

https://github.com/user-attachments/assets/b92cedc4-650b-4998-b0f5-59805cc0a743

huumn:
I did wait. Mine was a comment if that makes a difference. I might've also navigated and reloaded in between there.

huumn:
It might be that I was running a fresh cluster and hadn't checked notifications yet. I'll continue to QA. If you say it works I believe you.

huumn:
Is this intended: I don't get to retry if my zap fails to forward when I don't have a sending wallet?

Repro: Make `paidActionForwarding` throw an error. p2p zap someone while logged in without sending wallets attached.

huumn:
Because we have so many possible combinations of things, I find myself making matrices in my mind to test.

Zapping retry notifications (how it currently works):

| sndr/recvr | send wallet | no wallet | anon |
| - | - | - | - |
| recv wallet | x |  | |
| no wallet | x | | |

Zapping retry notifications (how it should work):

| | send wallet | no wallet | anon |
| - | - | - | - |
| recv wallet | x | x | |
| no wallet | x | x | |

Item retry notifications (how it does work after latest fixes):

| | send wallet | no wallet | anon |
| - | - | - | - |
| post | x | x | |
| comment | x | x | |

Item retry notifications (how it used to work before latest fixes):

| | send wallet | no wallet | anon |
| - | - | - | - |
| post | x |  | |
| comment | x |  | |

I think is what I originally meant with "just retry (without a qr fallback) whether there's a sending wallet or not" ... the no sending wallets special case means you can't rely on the payment attempt counter for showing notifications.

huumn (changes requested):
For people without sending wallets, they ***never*** get retry notifications for: zapping, poll votes, boost, down zaps.

See my other comments. You made a special case for `ITEM_CREATE` but the problem remains for other paid action types.

ekzyis:
> Is this intended: I don't get to retry if my zap fails to forward when I don't have a sending wallet?
>
> Repro: Make paidActionForwarding throw an error. p2p zap someone while logged in without sending wallets attached.

No, I assumed you would either get a QR code which succeeds or you cancel (= no notification) but I forgot that we have optimistic QR codes. I made a special case for `ITEM_CREATE` because there, we want to notify them always, even if they closed the QR code.

> I suspect for people without a sending wallet, even when the failure is not their fault, they no longer get retry notifications for: zapping, down zaps, poll votes, boost.

I haven't tested myself but I think you are right as mentioned above. I only briefly considered the case without sending wallets. I relied on `userCancel` always being set when a user would not have wallets attached :eyes: 

> just retry (without a qr fallback) whether there's a sending wallet or not

I think this isn't going to be reliable since a client without sending wallets could consume all payment retries: they would receive failed invoices, retry them (= increase the payment counter) but then fail to pay them and waste retries.

ekzyis:
Mhh, maybe it's enough to send `hasSendWallet` to `hasNewNotes` and the notifications query

huumn:
How do you think about `setHasSendWallets` when there are multiple devices without device sync? I'm having trouble reasoning through it.

ekzyis:
Notifications are now aware of if a stacker has send wallets via a new column `users.sendWallets`. I didn't simply pass this information as a GraphQL variable because notifications are rendered on the server and I didn't want to mess with that. Searching for `getGetServerSideProps` looked like we never pass GraphQL variables to it.[^1]

[^1]: Additionally, `user.sendWallets` is nice information for us :eyes:

With this information, I simplified the notification logic such that filter for failed payment notifications is _exactly_ the inverse of the filter for when automated retries should be attempted. To achieve this, I also reverted my decision to not show failed payments with `userCancel` set. They will not get automatically retried and thus should show up in notifications.

This means that if someone enables wallet, some notifications might disappear if they will not get automatically retried and vice versa. I think this is not bad UX but actually good.

huumn:
> Additionally, user.sendWallets is nice information for us. 

I agree.

Anyway, to add some brainstorming. In a situation where we're worried about invoice "waste," don't want to retry without send wallets, and I'm thinking about the refactor, I think we should just send `hasSendWallet` with paid actions. We already do it with zaps.

We can probably write a graphql interface for paid action input types that requires `hasSendWallet`. `retryPaidAction` can also take it so that retries can be auto-retried if a wallet is added/found later.

ekzyis:
dea2e7a now always retries, even without send wallets, to make sure that failed payments eventually show up in notifications.

However, there was an issue if a stacker goes offline longer than we would try automated retries. In that case, the payment counter will not reach the maximum and we need to check `cancelledAt` to show the notification.

This works to show the notification in /notifications, but not for the indicator, since the indicator assumes that `updatedAt` will get updated but it doesn't since the invoice isn't updated in a query.

To fix this, I added a `retryTimeout` job in 804f58c9. I will sleep over this to make sure updating the invoice like this doesn't mess with anything.

worker/wallet.js
```diff
@@ -161,6 +161,10 @@ export async function checkInvoice ({ data: { hash, invoice }, boss, models, lnd
   }
 }
 
+export async function retryTimeout ({ data: { hash }, models, lnd, boss }) {
+  await models.invoice.update({ where: { hash }, data: { updatedAt: new Date() } })
```
ekzyis, line 165: This might need to set a date slightly in the future to avoid race conditions with the notification indicator query but not sure, need to test

ekzyis: This fixes the notification indicator, but it doesn't seem to work as expected: It does not seem to use the date passed to `updatedAt`. I suspect the trigger for updated_at overrides any value passed to `updatedAt`, however, the value set does not seem to be the actual update date :thinking: 

Or maybe this does not run when I think it does run.

ekzyis: Mhh, it actually works as expected. It DOES use the `updatedAt` I passed in. Not sure why I was confused before. Updated to use a date 5 seconds in the future just to be sure.

huumn:
> This works to show the notification in /notifications, but not for the indicator, since the indicator assumes that updatedAt will get updated but it doesn't since the invoice isn't updated in a query.
>
> To fix this, I added a retryTimeout job in https://github.com/stackernews/stacker.news/commit/ddc115ef5072ff171fffad48718001b77a901931. I will sleep over this to make sure updating the invoice like this doesn't mess with anything.

I'm surprised the indicator query couldn't be changed instead.

ekzyis:
Since possibly receiving a notification about a failed payment after 1 day is bad UX, I changed it to stop automated retries after one hour.

> I'm surprised the indicator query couldn't be changed instead.

I think this is the first case where we want to show something in notifications after a timeout. This means it already existed in the database and wasn't updated between last poll and when we want to show it.

I considered to calculate `updated_at` on the fly using a `CASE` statement but I suspect that would mess with index usage.

And we can't create an index with an expression that depends on the current time afaik :thinking: 

So not sure how to change it. Using a job to update the timestamp makes everything work as expected.

However, we could add a new final `FAILED` state. That would also be an async job though.

huumn (commented):
Okay I understand why you added the job. I think it can be fixed by making the notification and indicator queries more complicated - which sucks. We'd basically need a separate queries for `userCancel` or not and have to fuss with `lastChecked` and `sortTime` using `WALLET_RETRY_BEFORE_MS` if `userCancel = false`. Probably not much better than your job but it'll at least be in two places (indicator/notifications) rather than four (paidAction,indicator,notifications,worker)? I think without a refactor this is kind of where we are going to end up with this stuff. 

I don't know if it's worth making those changes. Up to you. Just let me know about the `ITEM_CREATE` thing and I'll give it a final QA.

api/resolvers/wallet.js
```diff
@@ -456,6 +459,23 @@ const resolvers = {
...
+        SELECT * FROM "Invoice"
+        WHERE "userId" = ${me.id}
+        AND "actionState" = 'FAILED'
+        -- never retry if user has cancelled the invoice manually
+        AND "userCancel" = false
+        AND "cancelledAt" < now() - ${`${WALLET_RETRY_AFTER_MS} milliseconds`}::interval
+        AND "cancelledAt" > now() - ${`${WALLET_RETRY_BEFORE_MS} milliseconds`}::interval
+        AND "paymentAttempt" < ${WALLET_MAX_RETRIES}
+        -- never retry failed posts because we always immediately show them in notifications
+        AND "actionType" <> 'ITEM_CREATE'
```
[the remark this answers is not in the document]
ekzyis: Mhh, I can't find it in chat or here but I thought we not only want to receive a notification for failed posts (not just `retry payment` next to post), but also receive it as soon as possible.

But since I can't find us discussing it, I think my brain made this requirement up. I am happy with removing this special case.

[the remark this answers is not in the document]
ekzyis: Done in d971229

api/resolvers/user.js
```diff
@@ -567,6 +562,31 @@ export default {
         return true
       }
 
+      const invoiceActionFailed2 = await models.invoice.findFirst({
```
ekzyis, line 565: I struggled with writing a good comment since I already struggled with grasping this query, lol.

I think this didn't break anything but I didn't fully QA everything again.

ekzyis:
Oh, I am still missing the fix for `sortTime`

huumn:
QA without a wallet is flawless. I'll QA with a wallet after SNL.

PR #1780, opened 2024-12-30
filter out POST /api/graphql spam from sndev logs

riccardobl:
## Description

Since a Next.js update, the dev environment has started logging HTTP requests, which has caused the logs to become cluttered with logs of POST /api/graphql requests. This is an actual issue of  Next.js, see the related discussion here: https://github.com/vercel/next.js/discussions/65992.

This PR adds a workaround to filter out these logs by using grep in the sndev logs command.

## Screenshots
before / after

image

### DIFF
--- a/sndev
+++ b/sndev
@@ -7,6 +7,10 @@ if [ -f .env.local ]; then
   . ./.env.local
 fi
 
+logFilter() {
+  grep -v --line-buffered --color=never -P 'POST /api/graphql \x1b\[32m200\x1b\[39m in [0-9]+ms'
+}
+
 docker__compose() {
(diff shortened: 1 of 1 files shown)


huumn (changes requested):
`-P` isn't universal I guess: `grep: invalid option -- P`

Also this still logs GET requests: `GET / 200 in 156ms`. Is that as intended?

riccardobl:
Yes, that's intentional. I don't want to make the filter too broad to avoid the risk of filtering out legitimate logs. I'm focusing on filtering the graphql calls, as they are the primary source of pollution.

I've changed the command to not use -P, can you try it now?

Issue #1779, opened 2024-12-30
Adding a new Switch Account does not switch to new account


huumn:
When I add a new account to switch to, it adds, but I remain logged in as the original account until I switch to the new account again.

ekzyis:
Mhh, thought I fixed any such case in https://github.com/stackernews/stacker.news/pull/1618

riccardobl:
i can reproduce, i'll look into this

PR #1781, opened 2024-12-30
Fix adding an account doesn't automatically switch to it  #1779

riccardobl:
Fixes https://github.com/stackernews/stacker.news/issues/1779 

The issue was that the code responsible for setting the default multi-auth cookie for the user on their first login was being called on every authentication, including when adding a new account. This caused the authentication cookies to be set but then immediately overwritten by the default cookie, reverting to the previous user.

## Screenshots

before the patch
image

after the patch
image

## Checklist

yes

Tested with sndev login and nostr nip05
- [x] the default multi_auth cookie is set on first login
- [x] switching happens automatically when a new account is added
- [x] switching to anon works
- [x] switching to another account works


ekzyis (commented):
I've noticed that this breaks adding an account when you're switched to anon, see #1783 for details:

https://github.com/user-attachments/assets/6659741b-0a21-48b0-8e2f-282bf9379e51

PR #1783, opened 2024-12-31
Always return new user token for JWT callback to use

### LINKED ISSUE #1779: Adding a new Switch Account does not switch to new account

ekzyis:
## Description

As @riccardobl mentioned in #1781, #1779 is caused by the cookie pointer update in `pubkeyAuth` getting overridden back to the initial cookie pointer in the JWT callback.

As I mentioned in https://github.com/stackernews/stacker.news/issues/1779#issuecomment-2565585659, I thought I fixed  this in #1618 but it appears that I didn't really test all cases, but only the case when we switched to anon and then try to add a new account. Then we automatically get switched because the JWT callback will run with the token set to the new user since we return the new user token (the variable that is called `user`) in `pubkeyAuth`, not the old user token (the variable that is called `token`).

However, as noticed in #1779, this does not work if you're logged in as some account and then add an account because we then return `token` (the old user). This is what I must have broken in #1618 and @riccardobl fixed in #1781. However, #1781 then breaks switching from anon to a new account because `setMultiAuthCookies` will not be called in `pubkeyAuth` and in the JWT callback.

Since the token that the JWT callback receives is the token we returned in `pubkeyAuth`, returning the token of the user we want to switch to in `pubkeyAuth` makes sure that the JWT callback will call `setMultiAuthCookies` with the token we want. So afaict, we don't actually need to set the cookie pointer in `pubkeyAuth` but can rely on the JWT callback to do this for us. This is what this PR does.

fix #1779 supersedes #1781

## Videos

switching around to new accounts:

https://github.com/user-attachments/assets/9b7fb5d1-44fa-4322-8389-2fbac3fbc43d

switching around to existing accounts:

https://github.com/user-attachments/assets/833b1154-e3dc-44ae-88bd-b3571ce225c1

unlinking + linking new nostr key:

https://github.com/user-attachments/assets/1e67f9b4-65f7-421b-8408-f097d4761fc8

## Checklist

yes

`7`. It was a little tricky to consider all cases and make sense of them but I've tested all cases afaict now as can be seen in the videos and all make sense to me now. Simply return token we want to switch to and let JWT callback set the cookies. We only need to consider account switching in `pubkeyAuth` to not accidentally update pubkeys when we instead wanted to switch accounts.

n/a

no

### DIFF
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -91,15 +91,14 @@ function getCallbacks (req, res) {
         token.sub = Number(token.id)
       }
 
-      // response is only defined during signup/login
+      // this only runs during a signup/login because response is only defined during signup/login
+      // and will add the multi_auth cookies for the user we just logged in as
       if (req && res) {
         req = new NodeNextRequest(req)
         res = new NodeNextResponse(res)
         const secret = process.env.NEXTAUTH_SECRET
         const jwt = await encodeJWT({ token, secret })
         const me = await prisma.user.findUnique({ where: { id: token.id } })
-        // we set multi_auth cookies on login/signup with only one user so the rest of the code doesn't
-        // have to consider the case where they aren't set yet because account switching wasn't used yet
         setMultiAuthCookies(req, res, { ...me, jwt })
       }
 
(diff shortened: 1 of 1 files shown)


ekzyis:
> We only need to consider account switching in `pubkeyAuth` to not accidentally update pubkeys when we instead wanted to switch accounts.

btw, this is the reason why the OAuth providers don't work because we don't have this control. They always think we want to update an authentication method instead of logging in as a new account.

huumn:
> btw, this is the reason why the OAuth providers don't work because we don't have this control. They always think we want to update an authentication method instead of logging in as a new account.

Whether an account is linked or creates a new one is affected by these lines: https://github.com/stackernews/stacker.news/blob/ba2cdc2275c7a1b63da4b3926bc7893d19dab17e/pages/api/auth/%5B...nextauth%5D.js#L87-L92

Before this branch, accounts would not link. Maybe there's a way to conditionally enter that branch?

PR #1785, opened 2025-01-01
Distinguish invoices cancelled by user

ekzyis:
## Description

Requirement for #1776 

This adds a boolean column `Invoice.userCancel`. It is set if the invoice was cancelled by the user, i.e. the invoice was shown to the user via a QR code and they closed it. In any other case, it is not set.

I also added a constraint so cancelled invoices cannot be in a weird state where it's not clear if it has been cancelled and how.

## Video

https://github.com/user-attachments/assets/37de1f5b-db26-4dd0-9225-f281f9ad2932

## Additional Context

I decided to use a trigger to set `userCancel` to false by default if it was not specified in the same call as `cancelled` and `cancelledAt` are set.

## Checklist

Yes, old rows are migrated by setting `userCancel` to false. I also checked that there are no existing rows that would fail the new constraint:

```
> SELECT COUNT(1) FROM "Invoice" WHERE cancelled AND "cancelledAt" IS NULL;
0
> SELECT COUNT(1) FROM "Invoice" WHERE NOT cancelled AND "cancelledAt" IS NOT NULL;
0
```

`7`. See video

n/a

no

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -489,7 +489,7 @@ const resolvers = {
     },
     createWithdrawl: createWithdrawal,
     sendToLnAddr,
-    cancelInvoice: async (parent, { hash, hmac }, { me, models, lnd, boss }) => {
+    cancelInvoice: async (parent, { hash, hmac, userCancel }, { me, models, lnd, boss }) => {
       // stackers can cancel their own invoices without hmac
       if (me && !hmac) {
         const inv = await models.invoice.findUnique({ where: { hash } })
@@ -499,7 +499,7 @@ const resolvers = {
         verifyHmac(hash, hmac)
       }
       await finalizeHodlInvoice({ data: { hash }, lnd, models, boss })
-      return await models.invoice.findFirst({ where: { hash } })
+      return await models.invoice.update({ where: { hash }, data: { userCancel: !!userCancel } })
     },
     dropBolt11: async (parent, { hash }, { me, models, lnd }) => {
       if (!me) {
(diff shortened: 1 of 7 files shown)


api/resolvers/wallet.js
```diff
@@ -487,10 +487,10 @@ const resolvers = {
     },
     createWithdrawl: createWithdrawal,
     sendToLnAddr,
-    cancelInvoice: async (parent, { hash, hmac }, { models, lnd, boss }) => {
+    cancelInvoice: async (parent, { hash, hmac, userCancel }, { models, lnd, boss }) => {
       verifyHmac(hash, hmac)
       await finalizeHodlInvoice({ data: { hash }, lnd, models, boss })
-      return await models.invoice.findFirst({ where: { hash } })
+      return await models.invoice.update({ where: { hash }, data: { userCancel: !!userCancel } })
```
ekzyis, line 493: we usually use `Number(x)` or `String(x)` so maybe I should have also used `Boolean(x)` here instead of `!!`.

huumn:
Is there a way we can reduce these columns? Once I got a code review from an OG at my old company. I was passing 5 booleans to a function. He's like, why not use a bitmask? Is there some equivalent here? 

We don't need to clean this up asap, but it's worth forecasting a bit.

What do all of these do? AFAICT

- `cancelledAt` tells us when
- `cancelled` tells us "yes it's cancelled"
- `userCancel` tells us who

Seems like we can eventually do away with `cancelled` because `cancelledAt` gives us the same info, right?

ekzyis:
> Seems like we can eventually do away with cancelled because cancelledAt gives us the same info, right?

Definitely, forgot to mention. I'll create a PR that does which can be merged first.

I think without the extra `cancelled` column, `userCancel` + `cancelledAt` is fine but I haven't though much about an alternative solution yet.

huumn:
> I think without the extra cancelled column, userCancel + cancelledAt is fine but I haven't though much about an alternative solution yet.

Another solution (to reduce the model further) is: 

1. status column (`confirmed`, `cancelled`, `user_cancelled`)
2. statusUpdatedAt column

Anyway ... all just for forecasting. Refactoring now is a bad idea.

PR #1788, opened 2025-01-02
Fix expirein used instead of keepuntil

ekzyis:
## Description

We've been using `expirein` for reminder, delete and boost jobs far into the future. We thought `expirein > startafter` would make sure they don't disappear before they get run. However, `expirein` is for how long a job can stay in the **active state**:

> **Expiration options**
>
> - expireInSeconds, number
>  How many seconds a job may be in active state before it is failed because of expiration. Must be >=1

-- https://timgit.github.io/pg-boss/#/./api/jobs

The docs aren't clear, but I assume `keepuntil` is called "Retention options" in the docs and they state this:

> **Retention options**
>
> - retentionSeconds, number
>  How many seconds a job may be in created or retry state before it's archived. Must be >=1

So we want to use `keepuntil` instead of `expirein` for reminder, delete and boost jobs.

Our defaults via `\d pgboss.job`:

- `expirein`: 15 minutes
- `keepuntil`: 14 days

TODO:

- [ ] ~reschedule reminders~ 

since the jobs are just for the push notifications, I decided to not reschedule them to avoid duplicate push notifications for past or future reminders

- [ ] ~reschedule deletes~ 

won't do in this PR because I need an up-to-date db dump from prod after this was deployed

- [x] fix existing active boost jobs by updating `keepuntil`

added migration

## Additional Context

We could add a constraint on `pgboss.job` to make sure that `keepuntil` > `startafter` since I think any such case is a bug.

## Checklist

- some delete jobs have been lost (see TODO)
- some reminders might not send push notifications but that's not a big deal
- fixed existing `expireBoost` via update query in migrations

`5`. tested that reminders, deletes, expire boost, `checkInvoice` jobs are inserted without error.

n/a

no

### DIFF
--- a/api/paidAction/boost.js
+++ b/api/paidAction/boost.js
@@ -67,9 +67,9 @@ export async function onPaid ({ invoice, actId }, { tx }) {
   })
 
   await tx.$executeRaw`
-    INSERT INTO pgboss.job (name, data, retrylimit, retrybackoff, startafter, expirein)
+    INSERT INTO pgboss.job (name, data, retrylimit, retrybackoff, startafter, keepuntil)
     VALUES ('expireBoost', jsonb_build_object('id', ${itemAct.itemId}::INTEGER), 21, true,
-              now() + interval '30 days', interval '40 days')`
+              now() + interval '30 days', now() + interval '40 days')`
 }
 
 export async function onFail ({ invoice }, { tx }) {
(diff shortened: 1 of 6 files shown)


huumn:
The migration here doesn't work. The shadow db doesn't know about `pgboss`. I've removed it in the cowboy credit branch.

ekzyis:
Oh, then I must have forgotten to run it with Prisma

Issue #1517, opened 2024-10-25
Founders of archived territories are get post notifications when unarchived by new founders 


huumn:
Had this reported today.

huumn:
On unarchive or transfer, the original territory founder should be unsubscribed.

PR #1791, opened 2025-01-03
Unsubscribe original territory owner on transfer or unarchive

riccardobl:
## Description

Fixes https://github.com/stackernews/stacker.news/issues/1517

## Checklist

yes

- [x] transfer
- [x] transfer when source is not subscribed
- [x] transfer when destination is already subscribed
- [x] unarchive
- [x] unarchive when source is not subscribed
- [x] unarchive when destination is already subscribed
- [x] self unarchive

no

### DIFF
--- a/api/paidAction/territoryUnarchive.js
+++ b/api/paidAction/territoryUnarchive.js
@@ -42,13 +42,6 @@ export async function perform ({ name, invoiceId, ...data }, { me, cost, tx }) {
     // unsubscribe the old user
     const oldSubscription = await tx.subSubscription.findUnique({ where: { userId_subName: { userId: oldUserId, subName: name } } })
     if (oldSubscription) await tx.subSubscription.delete({ where: { userId_subName: { subName: name, userId: oldUserId } } })
-
-    // subscribe the new user if they aren't already
-    // await tx.subSubscription.upsert({
-    //   where: { userId_subName: { userId: newUserId, subName: name } },
-    //   update: {},
-    //   create: { userId: newUserId, subName: name }
-    // })
   }
 
   await tx.subAct.create({
--- a/api/resolvers/sub.js
+++ b/api/resolvers/sub.js
@@ -255,14 +255,6 @@ export default {
         // unsubscribe the old user
         const oldSubscription = await tx.subSubscription.findUnique({ where: { userId_subName: { userId: oldUserId, subName } } })
         if (oldSubscription) await tx.subSubscription.delete({ where: { userId_subName: { subName, userId: oldUserId } } })
-
-        // subscribe the new user if they aren't already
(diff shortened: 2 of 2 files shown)


api/paidAction/territoryUnarchive.js
```diff
@@ -34,7 +34,14 @@ export async function perform ({ name, invoiceId, ...data }, { me, cost, tx }) {
...
   if (sub.userId !== me.id) {
-    await tx.territoryTransfer.create({ data: { subName: name, oldUserId: sub.userId, newUserId: me.id } })
+    const oldUserId = sub.userId
+    const newUserId = me.id
+
+    await tx.territoryTransfer.create({ data: { subName: name, oldUserId, newUserId } })
+
+    // unsubscribe the old user
+    const oldSubscription = await tx.subSubscription.findUnique({ where: { userId_subName: { userId: oldUserId, subName: name } } })
+    if (oldSubscription) await tx.subSubscription.delete({ where: { userId_subName: { subName: name, userId: oldUserId } } })
```
riccardobl, line 44: do you prefer this or deleteMany?

api/resolvers/sub.js
```diff
@@ -240,14 +240,23 @@ export default {
...
-        models.sub.update({ where: { name: subName }, data: { userId: user.id, billingAutoRenew: false } })
-      ])
+      const updatedSub = await models.$transaction(async tx => {
+        await tx.territoryTransfer.create({ data: { subName, oldUserId, newUserId } })
+        const updatedSub = await tx.sub.update({ where: { name: subName }, data: { userId: newUserId, billingAutoRenew: false } })
+
+        // unsubscribe the old user
+        const oldSubscription = await tx.subSubscription.findUnique({ where: { userId_subName: { userId: oldUserId, subName } } })
+        if (oldSubscription) await tx.subSubscription.delete({ where: { userId_subName: { subName, userId: oldUserId } } })
+        return updatedSub
```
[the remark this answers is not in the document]
ekzyis: Nah, I'd say let's not automatically subscribe them for the reason you mentioned

api/paidAction/territoryUnarchive.js
ekzyis: I think `delete` should be fine. But do you need to check first if it exists?

riccardobl: yes, delete will error if the record does not exist (eg. if the user unsubscribed manually before transferring the territory) [the workaround is to use deleteMany that will not error](https://github.com/prisma/prisma/discussions/21682).
But in this case, since it is already inside a transaction, maybe this is better?

ekzyis: I would use `deleteMany` then

PR #1792, opened 2025-01-03
Fix UPDATE before adding invoice cancel constraint

ekzyis:
## Description

Fix migration that was added in #1785. I assume changing the existing migration is okay because it didn't run in prod.

## Checklist

yes, they will fix that #1785 was not backwards compatible with existing data in prod

`5`. Run migration locally but after `sndev delete` so I did not have any invoices anyway. But I am pretty sure this will fix the constraint failure in prod.

n/a

no

### DIFF
--- a/prisma/migrations/20241231223214_invoice_user_cancel/migration.sql
+++ b/prisma/migrations/20241231223214_invoice_user_cancel/migration.sql
@@ -2,7 +2,7 @@
 ALTER TABLE "Invoice" ADD COLUMN "userCancel" BOOLEAN;
 
 -- Migrate existing rows
-UPDATE "Invoice" SET "userCancel" = false;
+UPDATE "Invoice" SET "userCancel" = false WHERE "cancelled";
 
 -- Add constraint to ensure consistent cancel state
 ALTER TABLE "Invoice" ADD CONSTRAINT "Invoice_cancel" CHECK (


ekzyis:
I have verified locally that updating the old checksum `6c57e2e384fefdfa2d68c17fc49747c5e8d75d1628b31460cdd07e7747c5d841` to the new checksum `cafea2458befe90010c028bc2d9ea361f1d7d1b4e1fc3f0921136c3cdaf6f4db` makes `npx prisma migrate dev` run without errors:

https://github.com/user-attachments/assets/d703b802-1e8b-4e10-8cab-a95f405d186d

So I think we should merge this and then run

```sql
UPDATE _prisma_migrations
SET checksum = 'cafea2458befe90010c028bc2d9ea361f1d7d1b4e1fc3f0921136c3cdaf6f4db'
WHERE checksum = '6c57e2e384fefdfa2d68c17fc49747c5e8d75d1628b31460cdd07e7747c5d841';
```

before the next deployment.

But even without this update, migration in prod should not even show a warning anymore since the warning was removed in v3.12.0, see [Prisma docs](https://www.prisma.io/docs/orm/prisma-migrate/workflows/patching-and-hotfixing#migration-history-conflicts). It will only show up in `prisma migrate status`, see https://github.com/prisma/prisma/issues/9829 and https://github.com/prisma/docs/issues/5244.

So this should be safe to merge so local dev is not broken for seeds with invoices.

PR #1799, opened 2025-01-05
Add FAQ to version control

ekzyis:
## Description

I think it's a good idea to add our [special items](https://github.com/stackernews/stacker.news/blob/529b5d1fef47333482d6c7ec29040ed099659545/lib/constants.js#L180-L183) to version control so we have a history of changes and a familiar process to discuss, review and apply changes to them. It's (unfortunately) also easier to edit these markdown documents in a proper text editor instead of the SN editor.

I've added a script that can "deploy" the markdown content via an API key from an account included in `SN_USER_IDS`.

For now, I only did this to the [FAQ](https://stacker.news/faq) in case @huumn does not like this idea.

I tested this locally by adding my test user to `SN_USER_IDS` and using an API key from that test account.


huumn (changes requested):
I'm generally in support of this - mostly for version tracking. 

My main complaint is that `markdown` is a ambiguous toplevel directory for us to have.

I kind of like`docfood`, but that's confusing in a "wtf" way. 

Given this is run as a script remotely, it might make sense to put the script in `scripts`?

I don't know about all the other stuff, but it shouldn't be named `markdown`.

ekzyis:
> My main complaint is that `markdown` is a ambiguous toplevel directory for us to have.

I agree, I moved it into docs/user now

> Given this is run as a script remotely, it might make sense to put the script in `scripts`?

:+1:

PR #1794, opened 2025-01-04
fix: iOS PWA push notifications

Soxasora:
## Description

This PR tracks the progress on fixing issues related to iOS PWA push notifications (see #756)

## Screenshots

Mention counts:
22817271-66E0-4594-82A7-789C85070A5B_1_201_a

## Additional Context

_Was anything unclear during your work on this PR? Anything we should definitely take a closer look at?_

## Checklist

**Are your changes backwards compatible? Please answer below:**

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:**

**Did you introduce any new environment variables? If so, call them out explicitly here:**

## Progress

WIP:
- [ ] notification reply counts and mention counts and subscribe to stacker counts are wacky

Issues:
- [ ] push notifications continue to be lost
- [ ] when I click on notification replies I get sent to really old threads
- [ ] sat stacking notifications aren't replaced

Issue #756, opened 2024-01-16
iOS pwa push notification issues


huumn:
Figured we should just create persistent issue for this given they're numerous and hard to pin down:

1. push notifications continue to be lost #411 
2. notification reply counts and mention counts and subscribe to stacker counts are wacky 
    - https://github.com/stackernews/stacker.news/assets/34140557/d321e3ba-fe1d-48b3-8354-ac78ff870adb
3. when I click on notification replies I get sent to really old threads
4. sat stacking notifications aren't replaced

ekzyis:
My deposit notifications are also weird now on Android. I got a "4 sats were deposited in your account" notification but only 1 sat was deposited.

I looked at the logs and they didn't fully explain what was going on. I noticed the bug that is fixed in #751 though.

I didn't change the Android code though. So I am confused. I also wasn't able to reproduce the issue locally. Seems to only affect deposit notifications.

ekzyis:
Apple is pulling PWAs for their European users in iOS 17.4:

> To comply with the Digital Markets Act, Apple has done an enormous amount of engineering work to add new functionality and capabilities for developers and users in the European Union — including more than 600 new APIs and a wide range of developer tools.
>
> The iOS system has traditionally provided support for Home Screen web apps by building directly on WebKit and its security architecture. That integration means Home Screen web apps are managed to align with the security and privacy model for native apps on iOS, including isolation of storage and enforcement of system prompts to access privacy impacting capabilities on a per-site basis.
>
> Without this type of isolation and enforcement, malicious web apps could read data from other web apps and recapture their permissions to gain access to a user’s camera, microphone or location without a user’s consent. Browsers also could install web apps on the system without a user’s awareness and consent. Addressing the complex security and privacy concerns associated with web apps using alternative browser engines would require building an entirely new integration architecture that does not currently exist in iOS and was not practical to undertake given the other demands of the DMA and the very low user adoption of Home Screen web apps. And so, to comply with the DMA’s requirements, we had to remove the Home Screen web apps feature in the EU.
>
> EU users will be able to continue accessing websites directly from their Home Screen through a bookmark with minimal impact to their functionality. We expect this change to affect a small number of users. Still, we regret any impact this change — that was made as part of the work to comply with the DMA — may have on developers of Home Screen web apps and our users.

Even though this only should affect EU users ("a small number of users"), I guess this makes their lack of priority regarding [this](https://bugs.webkit.org/show_bug.cgi?id=258922) more clear.

[...]

PR #1802, opened 2025-01-05
fix: iOS push subscription gets lost

Soxasora:
## Description

Partial (final) fix of #756, 1: (push notifications continue to be lost #411)

Sometimes an `await showNotification` can be considered as an **hidden push notification** (like those spam websites), when 3 of those gets recognized as 'hidden' consecutively, Apple revokes the subscription.

We need to tell the service worker that the work is finished (sending a notification) and we could do this by using `event.waitUntil`[^1] instead of `await`

This PR aligns showNotification calls with Apple's recommended way, which is `event.waitUntil(showNotification)`.

## Screenshots
n/a

## Additional Context

This measure aligns our service worker with Apple's vision of how push notifications should work and it should be cross-platform. Not having Android I would need some test on it to see if notifications get sent correctly (Chrome didn't have problems)

This is part of a series of fixes tracked on #1794 

Edit: I just found this [comment](https://github.com/stackernews/stacker.news/issues/411#issuecomment-1871399648), which is correct but weirdly enough `return await showNotification` sometimes creates a notification without telling the service worker that it happened, causing a 'silent notification'. 
`event.waitUntil` tells the service worker that work is happening and doesn't return a Promise[^1]
I can be incorrect in evaluating what is happening here

Passing `event` from `onPush` to `mergeAndShowNotification` should ensure that `showNotification` gets done within the same context not violating Apple's requirement

## Checklist

Yesnt, I need to test with Android

6, tested only on iOS, so far no lost push subscription in a 2-day test environment.

n/a

No

[^1]: The ExtendableEvent.waitUntil() method tells the event dispatcher that work is ongoing. It can also be used to detect whether that work was successful. In service workers, waitUntil() tells the browser that work is ongoing until the promise settles, and it shouldn't terminate the service worker if it wants that work to complete.
[waitUntil specification](https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil)

### DIFF
--- a/sw/eventListener.js
+++ b/sw/eventListener.js
@@ -67,7 +67,7 @@ export function onPush (sw) {
           notifications.filter(({ tag: nTag }) => nTag === tag).forEach(n => n.close())
         }
         log(`[sw:push] ${nid} - show notification with title "${payload.title}"`)
-        return await sw.registration.showNotification(payload.title, payload.options)
+        return event.waitUntil(sw.registration.showNotification(payload.title, payload.options))
       }
 
       // according to the spec, there should only be zero or one notification since we used a tag filter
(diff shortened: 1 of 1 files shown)


Soxasora:
Sending too many notifications one after another causes push sub to be lost, investigating

Soxasora:
Merging to #1794

PR #1805, opened 2025-01-05
Use sats / ccs format in item details

ekzyis:
I think this:

2025-01-05-175920_515x328_scrot

looks better than this:

2025-01-05-175939_513x401_scrot

On mobile:

localhost_3000_(iPhone SE) (1)

stacker news_(iPhone SE) (1)

### DIFF
--- a/components/item-info.js
+++ b/components/item-info.js
@@ -252,20 +252,14 @@ function InfoDropdownItem ({ item }) {
           <div>{item.createdAt}</div>
           <div>cost</div>
           <div>{item.cost}</div>
-          <div>sats</div>
-          <div>{item.sats - item.credits}</div>
-          <div>CCs</div>
-          <div>{item.credits}</div>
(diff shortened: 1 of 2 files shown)


riccardobl:
+1 much easier to compare stacked sats vs cc on desktop, but i would reduce line-height for mobile so that it doesn't look like `stacked` and `(comments)` are two different entries.
Or maybe it could be reworked like this

```
title
         value
title 2
        value2
````
for mobile

huumn:
We have a css class `.line-height-md` and `.line-height-sm` I use in modals most of the time.

ekzyis:
I used `row-gap: 0.5rem` + `line-height: 1.25` now. `line-height` alone didn't help because it increases the height of every line, so it does not separate rows visually from each other. `row-gap` does that.

It now looks like this:

2025-01-06-004220_368x395_scrot

PR #1808, opened 2025-01-06
lint & fix

riccardobl:
Proposal to add a permissive eslint config that catches some issues while also making the sndev coding style pass.
All deps are devDependencies. Usage is `npm run eslint`.

I fixed all the issues  unrelated to coding style, that were just dead code, unused imports, wrong react pops and a potential global shadowing that did not cause any issue yet.


api/paidAction/zap.js
```diff
@@ -102,7 +102,6 @@ export async function onPaid ({ invoice, actIds }, { tx }) {
       }
     })
     acts = await tx.itemAct.findMany({ where: { invoiceId: invoice.id }, include: { item: true } })
-    actIds = acts.map(act => act.id)
```
riccardobl, line 105: dead code?

pages/_error.js
```diff
@@ -62,7 +62,7 @@ function ErrorImage ({ statusCode }) {
   return <Image className='rounded-1 shadow-sm' width='500' height='376' src={`${process.env.NEXT_PUBLIC_ASSET_PREFIX}/falling.gif`} fluid />
 }
 
-export default function Error ({ statusCode }) {
+export default function ErrorPage ({ statusCode }) {
```
riccardobl, line 65: potential global shadowing inside this file

wallets/server.js
```diff
@@ -82,19 +82,18 @@ export async function createInvoice (userId, { msats, description, descriptionHa
...
+    msats: toPositiveBigInt(msats) * (100n - feePercent) / 100n,
+    description,
+    descriptionHash,
+    expiry
+  }, { predecessorId, models })
 
-    logger = walletLogger({ wallet, models })
-    bolt11 = invoice
+  const logger = walletLogger({ wallet, models })
+  const bolt11 = invoice
```
riccardobl, line 94: if the only thing this catch does is logging the error and rethrow, there is no point in wrapping all the rest of the code in the try-catch, also eslint didn't like this.

huumn (commented):
This looks like it caught some stuff (largely inconsequential from the looks of it, but they'd confuse a newcomer, so nice to be ride of it). 

We already have [a linter](https://standardjs.com/) (and we add a nextjs plugin) so I guess I'd be curious:

1. which rules this is adding that we don't already have with the existing linters (the delta - do we know?)
2. why we chose the rules that were chosen and not all the rules those plugins provide
3. why we should implement these plugins into a separate linter rather than plugging into to the existing one

Maybe those things are obvious, but I'm not super familiar with lint rules and managing them.

riccardobl:
> why we should implement these plugins into a separate linter rather than plugging into to the existing one

I don't know much about standardjs, maybe it is possible to achieve the same goal in a different way, but the current setup seems to be targeted mostly at detecting formatting deviations, while the eslint config i've added here is mostly aimed at detecting oversights and bad patterns that make the code fragile or confusing. 

> why we chose the rules that were chosen and not all the rules those plugins provide

The suggested rules for the plugins are too strict for sn (they were finding 300+ issues) so i've disabled those that I thought were considered undesired, just by my experience with the project (no scientific method here, and might need more tweaks). I've disabled also the rules related to improving regexes, that i think are not very important, and would need some special attention.

> which rules this is adding that we don't already have with the existing linters (the delta - do we know?)

the rule set is [here](https://github.com/SonarSource/SonarJS/blob/master/packages/jsts/src/rules/README.md#rules) and [here](https://github.com/eslint-community/eslint-plugin-promise?tab=readme-ov-file#rules), minus the ones disabled.



Ultimately my goal here was to add something to make the review process lighter by having some form of guidance to catch stylistic issues and avoid nitpicking in prs for new contributors (make the unwritten rules written), and to catch oversights from veterans. 
I don't have strong opinions on whether to use eslint or other linters.

PR #1812, opened 2025-01-06
fix: zapping too fast causes duplicate notifications

### LINKED ISSUE #1809: Zapping too fast causes duplicate notifications

### Describe the bug

notifyZapped will get triggered the amount of times the user has zapped but will take Item.msats as the value to be shown.
If a user zaps an Item too fast, it will cause the same notification to be shown n times

### Screenshots

image

### Steps To Reproduce

1. Zap an Item n times in a row, really fast

### Expected behavior

For each TIP, a notification with the correct sat value should be shown

### Logs

_No response_

### Device information

- OS: iOS
- Client: PWA

### Additional context

Should be OS agnostic

Soxasora:
## Description

Fixes #1809
Checks if there are new ItemActs before triggering notifyZapped, avoiding duplicate notifications

## Screenshots
IMG_69E9C8279350-1

## Additional Context

n/a

## Checklist

Yes

7, sent a lot of zaps really really fast!

n/a

No

### DIFF
--- a/api/paidAction/zap.js
+++ b/api/paidAction/zap.js
@@ -206,6 +206,8 @@ export async function nonCriticalSideEffects ({ invoice, actIds }, { models }) {
     where: invoice ? { invoiceId: invoice.id } : { id: { in: actIds } },
     include: { item: true }
   })
+  // avoid duplicate notifications with the same zap amount
+  // by checking if there are any other pending acts on the item
   const pendingActs = await models.itemAct.count({
     where: {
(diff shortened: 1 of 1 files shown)


ekzyis (approved):
did not test but looks good to me!

huumn:
This risks missing a zap notification if the future zap is p2p and fails, but missing a zap is less buggy UX than duplicates. So we'll ship it.

Issue #1736, opened 2024-12-18
globally pinned items rank in global


huumn:
![](https://i.imgur.com/AUwoFHI.png)

Soxasora:
Can't seem to find the duplicate anymore, before I could find it after a couple of 'fetch more' which could maybe indicate some trouble in fetching and ssr?

I'll continue to monitor

ekzyis:
> which could maybe indicate some trouble in fetching and ssr?

Do you mean in production? I think it just means that the item dropped off in ranking because it got old. Items in `hot` have a time decay.

Soxasora:
Makes much more sense, I was indeed exploring how ranking works to try and find the root cause of this bug (or completely exclude global pins from ranking, they have position for it)

PR #1814, opened 2025-01-12
fix: globally pinned items rank in global

Soxasora:
## Description

Excludes items that have a `pinId` but no `subName` (global items) from the items query if it's ran in home, fixing #1736

## Screenshots
before:
image

after:
image

## Additional Context

n/a

## Checklist

Yes

10, array from query doesn't contain items that have a pinId and no subName, pins are later injected correctly by `itemsWithPins`

n/a

No

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -536,8 +536,8 @@ export default {
                     LEFT JOIN "Sub" ON "Sub"."name" = "Item"."subName"
                     ${joinZapRankPersonalView(me, models)}
                     ${whereClause(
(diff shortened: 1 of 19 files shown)


huumn (commented):
As is, this will prevent an item that's pinned in another territory from ranking in hot on the homepage won't it? People got really upset when we did that before.

I suspect we'll want this to be done in the SQL query. We probably already have the correct behavior for pins in territories, but it doesn't work when `sub = null`, ie when it's the homepage.

Soxasora:
> I suspect we'll want this to be done in the SQL query.

Way better! Didn't think of going deeper than `itemsWithPins`

PR #1818, opened 2025-01-14
Login with magic code

### LINKED ISSUE #727: Cannot login PWA (iOS) with e-mail or Nostr

Soxasora:
## Description

Partially fixes #727 
Introduces a 6-character bech32 magic code (token) that expires in 5 minutes and can be used to login with email on PWA, especially iOS since atm there's no way to open a PWA from a link.

On `signIn`, the email and the callbackUrl will be temporarily saved on sessionStorage

Upon submitting the magic code, it will build and redirect to a magic link (as we are used to) with the user inserted token and `email, callbackUrl` from sessionStorage

After 3 failed attempts, a custom `useVerificationToken` will delete email and token from the database, requiring the user to get a new one.

Introduces **`MultiInput`**, a customizable component that allows for n inputs and supports keyboard traveling, can prepend a sequence (1-input, 2-input, etc.)

## Screenshots
code input page:
image

invalid magic code:
image

if no callback url or email
image

email template:
image

error page:
image

## Additional Context

afaik in this version of NextAuth we can't send parameters from signIn to /verify-request to /email without risking to be disruptive, thus I had to resort to sessionStorage

Also we can't send the default token AND the new custom token

Changed from 'login as xx@xx.com' to 'login with xx@xx.com' to make it independent from context (if login or setting an email); removed any reference to links

InputInner component has been modified to check for `hideError`, if true error messages will be showed by MultiInput instead of InputInner

## Checklist

Changes token generation

7: login, signup, link email, emails
7: token invalidation occurs on success and on 4th failed attempt

Yes

No

## Progress
Removed checkboxes, was lagging too much

- fix flex rules
- disable magic code input if callback is not available
- magic code input via MultiInput
  - change layout method, inputs are too big on desktop
  - fix mobile layout
  - cleanup
  - comments
  - single error message instead of per input
  - autofocus
- magic code for non-PWA users
- check if we can avoid sessionStorage
- better token invalidation tests
- avoid isPWA duplicate (`pull-to-refresh.js`; `email.js`)
- expire token/session after 3 failed tries
- use bech32 as generator
- Auth, revise token generation
- UX, better email
- UX, adjust remaining email templates
- cleanup


Soxasora:
We could also enable magic code for non-PWA users, imho it would be handy

ekzyis (commented):
_Turns out I didn't submit this review two days ago :eyes:_

I didn't test yet, only looked at the code. Left some questions.

> We could also enable magic code for non-PWA users, imho it would be handy, token is 6-digits anyway and you can still click the link

I think this is the way to go. Less complexity in our code, less confusing for users and would allow to use emails on mobile to login on desktop and vice versa.

pages/api/auth/[...nextauth].js
```diff
@@ -321,6 +324,40 @@ export const getAuthOptions = (req, res) => ({
...
+            })
+          } else { // otherwise, just increment the failed attempts
+            await prisma.verificationToken.update({
+              where: { id: verificationRequest.id },
+              data: { attempts: newAttempts }
+            })
+          }
+        }
+      }
+      return null
```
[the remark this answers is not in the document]
Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:360]: It looks cooler 👀, I'm moving towards a less verbose way of writing so I fully agree!

pages/api/auth/[...nextauth].js
```diff
@@ -366,9 +403,15 @@ export default async (req, res) => {
   await NextAuth(req, res, getAuthOptions(req, res))
 }
 
+function randomizeToken () {
+  const words = bech32.toWords(Buffer.from(randomBytes(3)))
+  return bech32.encode('token', words).slice(6, 12)
+}
```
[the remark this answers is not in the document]
Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:409]: This is more a product of `better safe than sorry`. Discussing with @huumn we agreed that bech32 encoding would produce a safer token than a standard 6-digit or even an 8-digit one.

The real reason is the resulting entropy, there are 32 possible characters per every 'digit', 32^6 = 1.073.741.824 (auto calc came in clutch lol).
24-bit or 32-bit entropy will result anyway in 32^6 after slicing![^1]

[^1]: if I'm not mistaken

pages/auth/error.js
```diff
@@ -27,8 +27,8 @@ export default function AuthError ({ error }) {
     return (
       <StaticLayout>
         <Image className='rounded-1 shadow-sm' width='500' height='375' src={`${process.env.NEXT_PUBLIC_ASSET_PREFIX}/double.gif`} fluid />
-        <h2 className='pt-4'>This magic link has expired.</h2>
-        <h4 className='text-muted pt-2'>Get another by logging in.</h4>
+        <h2 className='pt-4'>Where did the magic go?</h2>
```
Soxasora, line 30: this is a placeholder for eventual stackernews-themed sentences 👀

components/form.js
```diff
@@ -1241,5 +1241,102 @@ export function PasswordInput ({ newPass, qr, copy, readOnly, append, value: ini
...
+            name={name}
+            key={index}
+            type={inputType}
+            value={value}
+            innerRef={(el) => { inputRefs.current[index] = el }}
+            onChange={(formik, e) => handleChange(formik, e, index)}
+            onKeyDown={e => handleKeyDown(e, index)}
+            style={{
+              textAlign: 'center',
+              maxWidth: `${charLength * 44}px` // adjusts the max width of the input based on the charLength
```
Soxasora, line 1322: I found that a character sits comfortably in 44px given InputInner padding etc.

pages/api/auth/[...nextauth].js
```diff
@@ -366,9 +403,15 @@ export default async (req, res) => {
   await NextAuth(req, res, getAuthOptions(req, res))
 }
 
+function randomizeToken () {
+  const words = bech32.toWords(Buffer.from(randomBytes(3)))
+  return bech32.encode('token', words).slice(6, 12)
```
Comment (huumn) [inline pages/api/auth/[...nextauth].js:408]: I don't think we want to use the bech32 encoding lib:

1. I don't understand it
2. it has features that we don't need like prefix and checksum

I suggested bech32 for the character set because in lower case, none of the characters can be confused for other numbers/letters. For upper case, there might be a better character set, but basically we want something like the following to generate the string:

```
function generateRandomString(length = 6, charset) {
  // Default charset if none provided
  charset = charset || 'qpzry9x8gf2tvdw0s3jn54khce6mua7l';
  
  const bytes = crypto.randomBytes(length);
  let result = '';
  
  // Map each byte to a character in the charset
  for (let i = 0; i < length; i++) {
    result += charset[bytes[i] % charset.length];
  }
  
  return result;
}

**Review by huumn — CHANGES_REQUESTED** [review]
I'll QA and stuff first thing tomorrow, but we'll want to change the usage of the `bech32` lib.

File: pages/api/auth/[...nextauth].js (hunk 5 above)
Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:408]: Okay! I juggled around the bech32 library because we were already using it (appropriately) for other stuff. I'll gladly switch to manual generation and avoid shenanigans.
Didn't know about avoiding confusion with uppercase characters but I see it know, switching asap!

File: pages/api/auth/[...nextauth].js [hunk 6]
```diff
@@ -403,9 +403,16 @@ export default async (req, res) => {
...
+  const bytes = randomBytes(length)
+  const result = new Array(length)
+
+  // Map each byte to a character in the charset
+  for (let i = 0; i < length; i++) {
+    result[i] = charset[bytes[i] % charset.length]
+  }
+
+  return result.join('')
 }
```
Comment (Soxasora) [inline pages/api/auth/[...nextauth].js:416]: Just added the bech32 charset to constants, made it default and switched to array building instead of string concatenation (it would allocate a string per addition, actually not a problem with 6 characters but it's better to be safe than sorry)

pages/api/auth/[...nextauth].js
Comment (huumn) [reply pages/api/auth/[...nextauth].js]: FYI Probably not going to get to QA until later today. Troubleshooting weird lightning stuff.

Comment (huumn) [reply pages/api/auth/[...nextauth].js]: I didn't get to it at all today - add to #1853 that the lnd cert expired, paginated comments deployment caused a meltdown and had weird ux issues. 

The fixes you made look good though. I'll QA first thing tomorrow hopefully.

Comment (Soxasora) [reply pages/api/auth/[...nextauth].js]: A signal from the universe that comments are 🔥🔥🔥, jokes aside don't worry, this PR ain't going somewhere else while we sleep ^^

huumn (commented):
I haven't done a line by line review yet but this works well in my QA so far.

Is it intentional that the code must be typed manually? The email says "copy it" but each input only takes one character so I can't copy then paste.

Also it might be nice if the input autofocused to assist copy and pasting.

I'll continue QAing and let you know if anything else comes up

Soxasora:
> Is it intentional that the code must be typed manually?

Oh, didn't think about it! I'm going to push it right now.

pages/api/auth/[...nextauth].js
```diff
@@ -329,14 +329,17 @@ export const getAuthOptions = (req, res) => ({
...
+          attempts: {
+            lt: 2 // count starts at 0
+          }
         },
         orderBy: {
           createdAt: 'desc'
         }
       })
 
-      if (!verificationRequest) return null
```
Comment (huumn) [inline pages/api/auth/[...nextauth].js]: This forces a redirect to the login page if all 3 attempts have been tried which lets us avoid a the "go back or login again" on the error page. Instead, the button on the error page, just goes back in the browser history.

pages/api/auth/[...nextauth].js
```diff
@@ -345,17 +348,14 @@ export const getAuthOptions = (req, res) => ({
...
-      } else { // otherwise, just increment the failed attempts
-        await prisma.verificationToken.update({
-          where: { id: verificationRequest.id },
-          data: { attempts: newAttempts }
-        })
-      }
+      await prisma.verificationToken.update({
+        where: { id: verificationRequest.id },
+        data: { attempts: { increment: 1 } }
+      })
```
Comment (huumn) [inline pages/api/auth/[...nextauth].js:354]: Your changes were perfectly fine, but the increment wasn't an atomic read-write; it read from the db in one tx, then updated it in another. While it's still possible with my fixes for someone to get more than 3 attempts in quick succession, this makes it a little less likely.

Ideally, this code would not have any races at all, but it's a bit tricky because `identifier` isn't unique. We could wrap it all in a transaction and do an optimistic lock, or make it seriallizable, but our attempt limit is so low they won't get many more attempts anyway.

huumn:
Great work! Couldn't fault except for relative nitpicks.

Sebastix:
Woa, this is very neat!! Nice solution for fixing the loggin in with the PWA 👍🏻

pages/api/auth/[...nextauth].js
Comment (Soxasora) [reply pages/api/auth/[...nextauth].js]: This is a way better UX, thanks!

PR #1820, opened 2025-01-17
fix: duplicate push notification on subscribed user and territory

### LINKED ISSUE #1819: Duplicate(ish) push notifications for subscribed user and territory

### Describe the bug

When stacker to whom you are subscribed posts in a territory to which you’re subscribed, you receive a push notification for each, e.g.:
1. User posted foo
2. user posted foo in bar territory 

These are deduped in the notifications page. They should also be deduped in push notifications. 

### Screenshots

I forgot to take one before clearing the notifications. 

### Steps To Reproduce

1. Enable push notifications
2. Subscribe to a user
3. subscribe to a territory
4. wait for the user to post in the territory 
5. see 2 notifications

### Expected behavior

Only 1 push notification should come through

### Logs

_No response_

### Device information

iOS PWA

### Additional context

_No response_

Soxasora:
## Description

Fixes #1819 
Adds a condition to the `notifyUserSubscribers`' query, ensuring that a push notification is not sent if the subscribed user has posted in a territory the recipient is subscribed to, avoiding duplicate notifications.

## Screenshots
before:
Image

after:
Image

## Additional Context
n/a

## Checklist

Yes

9, tested with userA and userB, the latter creating an Item in a territory user A is subscribed to

n/a

No

### DIFF
--- a/lib/webPush.js
+++ b/lib/webPush.js
@@ -4,6 +4,7 @@ import { COMMENT_DEPTH_LIMIT, FOUND_BLURBS, LOST_BLURBS } from './constants'
 import { msatsToSats, numWithUnits } from './format'
 import models from '@/api/models'
 import { isMuted } from '@/lib/user'
+import { Prisma } from '@prisma/client'
 
 const webPushEnabled = process.env.NODE_ENV === 'production' ||
   (process.env.VAPID_MAILTO && process.env.NEXT_PUBLIC_VAPID_PUBKEY && process.env.VAPID_PRIVKEY)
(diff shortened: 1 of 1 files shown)


lib/webPush.js
```diff
@@ -137,6 +137,13 @@ export const notifyUserSubscribers = async ({ models, item }) => {
       AND "Reply"."ancestorUserId" = follower.id
       AND follower."noteAllDescendants"
     )
+    -- ignore subscription if user has posted to a territory the recipient is subscribed to
+    AND NOT EXISTS (
+      SELECT 1 FROM "SubSubcription"
```
ekzyis, line 142: :eyes: 

```suggestion
      SELECT 1 FROM "SubSubscription"
```

lib/webPush.js
```diff
@@ -137,6 +137,13 @@ export const notifyUserSubscribers = async ({ models, item }) => {
...
+      INNER JOIN "Item" ON "Item"."subName" = "SubSubscription"."subName"
+      WHERE "SubSubscription"."userId" = "UserSubscription"."followerId"
+      AND "Item"."userId" = $1
+    )
     `, Number(item.userId), Number(item.id))
```
ekzyis, line 147: I don't think you need this join, you can use `item.subName` as `$3` to get the territory and then use this filter:

```suggestion
      WHERE "SubSubscription"."userId" = "UserSubscription"."followerId"
      AND "SubSubscription"."subName" = $3
    )
    `, Number(item.userId), Number(item.id), item.subName)
```

ekzyis (changes requested):
Perfect location for the fix, but I noticed three things:

1. Syntax error that I assume went unnoticed because a syntax error also results in no push notification as expected :eyes:
2. unnecessary join?
3. this also prevents the push notification for comments but the conflicting subscription for territories only handles posts

For 3., you could add the following to the subquery:

```sql
AND ${isPost ? 'true' : 'false'}
```

but maybe you have a better idea

lib/webPush.js
Soxasora: embarrassed to say the least 🫠

lib/webPush.js
Soxasora: Added an isPost to $3 because comments don't have a subName

Soxasora:
Thank you ek, your review is spot on and my mind wasn't at the time, sorry about that.
Comments now show up and duplicate notifications are gone!

huumn:
I modified the query to use [prisma's tagged template helpers](https://www.prisma.io/docs/orm/prisma-client/using-raw-sql/raw-queries#tagged-template-helpers). We wrote a lot of these queries before prisma supported these tagged template fragments. In general, when we're modifying queries these days, I like to move away from `queryRawUnsafe` when I can - harder to mess up.

PR #1826, opened 2025-01-18
normalized wallet logs

ekzyis:
## Description

We want to normalize the wallet logs. They currently don't link to invoices or withdrawals via foreign keys but store all the data in the JSON `context` column.

~However, afaict, we can't get rid of `context` entirely since either a) the data we want to show is unrelated to an invoice or b) the invoice was not created yet.~

~**update:** Decided to drop the `context` column anyway to keep it simple. If we really need some information, we can add it back or save it in a relational way.~

**update:** Realized we would lose `created invoice for X sats` log messages. We log them when we just fetched a user invoice before they get wrapped. Since wrapping can fail, this would mean we would show an error without having shown that we created an invoice.

There are and will most likely be more cases like this so I decided to keep the `context` column.

So all this PR does is to add the link to the invoice/withdrawal if we already have an invoice/withdrawal id and resolve the context from this link instead of from the JSON column.

TODO:

- [x] resolve context fields required by frontend
- [x] ~refactor context in frontend?~ _will completely remove client logs in the future and move them to server, too_

Regression tests:

- [x] test wallet status derived from logs
- [x] test log messages still show same information in frontend

## Additional Context

To not use the JSON column if not necessary, I now also resolve `context.status` and `context.recv` (needed for wallet status) from the existing information instead of saving it.

In a future PR, I will replace the logic to derive status with two status column on the `Wallet` table anyway.  

## Checklist

maybe, but I didn't test and we don't care so we just truncate the table

`8`. Tested successful/failed autowithdrawals, successful/failed p2p zaps (forward error, wrap error) and sending payments (even though sender logs were not changed at all) 

n/a

no

### DIFF
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -767,9 +767,8 @@ const resolvers = {
         }
       }
 
-      if (invoice) {
-        return await logContextFromBolt11(invoice.bolt11)
-      }
+      // XXX never return invoice as context because it might leak sensitive sender details
+      // if (invoice) { ... }
 
       return context
     }
--- a/api/resolvers/wallet.js
+++ b/api/resolvers/wallet.js
@@ -745,10 +745,44 @@ const resolvers = {
       return item
     },
     sats: fact => msatsToSatsDecimal(fact.msats)
+  },
+
+  WalletLogEntry: {
+    context: async ({ level, context, invoiceId, withdrawalId }, args, { models }) => {
+      const isError = ['error', 'warn'].includes(level.toLowerCase())
+
+      if (withdrawalId) {
+        const wdrwl = await models.withdrawl.findUnique({ where: { id: withdrawalId } })
+        return {
+          ...await logContextFromBolt11(wdrwl.bolt11),
+          ...(wdrwl.preimage ? { preimage: wdrwl.preimage } : {}),
+          ...(isError ? { max_fee: formatMsats(wdrwl.msatsFeePaying) } : {})
+        }
+      }
+
+      if (invoiceId) {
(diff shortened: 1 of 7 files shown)


huumn (approved):
QA for this went fine. As discussed in chat, I'd recommend refactoring this to not require 2N+1 DB dips where N = # log lines. 

I'd also recommend thinking about preventing context from (in the future ... perhaps by accident) leaking sender details to receivers and receiver details to senders. Perhaps wallet logs for receiving should never have an invoiceId and for sending should never have a withdrawalId.

ekzyis:
> I'd recommend refactoring this to not require 2N+1 DB dips where N = # log lines.

Done in 8be91e80

> leaking sender details to receivers and receiver details to senders.

I realized I was actually indeed leaking a sender's invoice to the receiver. :eyes: Fixed that in 0d5c0351. I now never return an invoice as the context for a receiver's log.

 > Perhaps wallet logs for receiving should never have an invoiceId and for sending should never have a withdrawalId.

Mhh, wouldn't we miss out on interesting relationships that way again, too? But if we want to do this, we would just need to remove [this](https://github.com/stackernews/stacker.news/pull/1826/files#diff-522f172044005415493e60122e9bda8e0c714355defb8c796655fc64026422ffR325) and [this line](https://github.com/stackernews/stacker.news/pull/1826/files#diff-522f172044005415493e60122e9bda8e0c714355defb8c796655fc64026422ffR447).

We could also completely drop the `invoiceId` column and add it back when we actually have sender logs on the server.

huumn:
> Mhh, wouldn't we miss out on interesting relationships that way again?

No, they'd always be available via join. walletLogEntry -> withdrawal -> invoiceForward -> invoice

PR #1829, opened 2025-01-20
Fix recent sort order for retried items

ekzyis:
## Description

I created [this item](https://stacker.news/items/857641) but my payment failed (different story).

I then paid manually after an hour and noticed that my item didn't show up at the top of the recent feed but was inserted where it would have been if I had immediately paid for it.

I think this is a bug: recent sort should sort by `Item.invoicePaidAt` if available else fallback to `Item.created_at`

TODO:

- [x] fix timestamp shown in feed
- [x] fix index not used
- [x] fix for comments

Postgres function diff:

```diff
diff --git a/prisma/migrations/20250122114228_recent_sort_by_invoice_paid_at/diff.sql b/prisma/migrations/20250122114228_recent_sort_by_invoice_paid_at/diff.sql
index e568839b..ee12b551 100644
--- a/prisma/migrations/20250122114228_recent_sort_by_invoice_paid_at/diff.sql
+++ b/prisma/migrations/20250122114228_recent_sort_by_invoice_paid_at/diff.sql
@@ -10,7 +10,7 @@ BEGIN
     END IF;

     EXECUTE 'CREATE TEMP TABLE IF NOT EXISTS t_item ON COMMIT DROP AS'
-        || '    SELECT "Item".*, "Item".created_at at time zone ''UTC'' AS "createdAt", "Item".updated_at at time zone ''UTC'' AS "updatedAt", '
+        || '    SELECT "Item".*, COALESCE("Item"."invoicePaidAt", "Item".created_at) at time zone ''UTC'' AS "createdAt", "Item".updated_at at time zone ''UTC'' AS "updatedAt", '
         || '    "Item"."invoicePaidAt" at time zone ''UTC'' AS "invoicePaidAtUTC", to_jsonb(users.*) || jsonb_build_object(''meMute'', "Mute"."mutedId" IS NOT NULL) AS user, '
         || '    COALESCE("ItemAct"."meMsats", 0) AS "meMsats", COALESCE("ItemAct"."mePendingMsats", 0) as "mePendingMsats", COALESCE("ItemAct"."meDontLikeMsats", 0) AS "meDontLikeMsats", '
         || '    COALESCE("ItemAct"."meMcredits", 0) AS "meMcredits", COALESCE("ItemAct"."mePendingMcredits", 0) as "mePendingMcredits", '

```

## Checklist

yes

`3`. Tested this by first creating item 1 and not paying for it, then creating item 2 and paying for it, then paying for item 1. Item 1 now shows up before item 2 as expected because it "went live" by virtue of payment

 Did this for posts and comments.

n/a

no


huumn (changes requested):
1. `invoicePaidAt` doesn't have an index. This will require sorting all qualifying items, >100k probably, for each `recent` page load.
2. `ORDER BY` expressions probably need to be indexed in addition to the columns used in the expression, ie `COALESCE(column1, column2)` probably means "evaluate expression and sort" and not "use index on `column1` unless its null, then use index on `column2`)

huumn:
fyi This bug will also be a bug for comments.

ekzyis:
Fixed missing index use in 343b90d2 and fixed it for comments in 66ee938a

api/resolvers/item.js
```diff
@@ -415,7 +415,7 @@ export default {
               ORDER BY "Item".created_at DESC
```
huumn, line 415: You'll want to order the inner query too so that the LIMIT doesn't cut out the item. I'll make the change myself on review but the way this query monster works is we have LIMIT'd subquery so we can efficiently join the result0 with a bunch things in result1. After the join result0 can get out of order so we have to reorder it.

huumn:
In addition to the fix needed in the above comment, I made it so that we don't hide the value of `createdAt` from the frontend/anyone calling `items` or querying comments. That just seemed like a bad idea to avoid having to modify the frontend.

PR #1840, opened 2025-01-22
fix: Images on iOS are cropped weird

### LINKED ISSUE #1793: Images on iOS are cropped weird
I'll look for an example soon. I just keep forgetting to fix this.

Soxasora:
## Description

Fixes #1793

My hypothesis is that iOS gives up on decoding and thus loading the image if during asynchronous decoding it runs out of resources, indeed on reload the problem might be resolved (essentially retrying decoding).

If we force synchronous decoding, iOS gives this job more priority by using the main thread, potentially fixing weird artifacts and image misplacement.

## Screenshots
image

## Additional Context

In a no-cache dev scenario this problem really doesn't exist because after the page is fully loaded, images start loading one by one.

Another possible fix, if this fails, could be to update CSS after all the images are loaded, much like we already do for videos as this fixes it in prod.

There's also another fix that implies reducing the thumbnail resolution to avoid running out of iOS resources

Previously drafted to explore new solutions

## Checklist

Yes

4, issue doesn't present itself in dev

n/a

No


huumn:
Did you explore [`decode()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decode) at all? That came up when I was researching the `decode` attribute. According to MDN, which is occasionally wrong, it appears that [safari decodes sync by default (most of the time)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding#value).

Soxasora:
Ah, yes I drafted the PR earlier to explore better solutions but I didn't deep-dive into decode(), thank you. 

> it appears that safari decodes sync by default (most of the time)

I was trying with decoding attribute mainly for that, like MDN stated Safari can [choose a decoding method](https://x.com/smfr/status/1511370699885068290) opaquely. 
But we don't live by hypotheses, I'll redraft and get to it, thanks k00b!

components/media-or-link.js
```diff
@@ -133,9 +133,12 @@ export const useMediaHelper = ({ src, srcSet: srcSetIntital, topLevel, tab }) =>
       // hack
       // if it's not a video it will throw an error, so we can assume it's an image
       const img = new window.Image()
-      img.onload = () => setIsImage(true)
-      img.decoding = 'sync' // iOS could give up on decoding if async
       img.src = src
+      img.decode().then(() => {
```
Soxasora, line 137: It is quite interesting to load an image this way, following the decoding pipeline it should theoretically wait for the image to be ready before showing

components/media-or-link.js
```diff
@@ -133,9 +133,12 @@ export const useMediaHelper = ({ src, srcSet: srcSetIntital, topLevel, tab }) =>
...
+        setIsImage(true)
+      }).catch((e) => {
+        console.error('Cannot decode image', e)
+      })
```
Soxasora, line 141: It is quite interesting to load an image this way, following the decoding pipeline it should theoretically wait for the image to be ready before showing

Soxasora:
Loading the image after successful decoding should potentially fix this issue. We're still in a blind scenario but it's still good practice to do so anyway. Before claiming it as ready for review I'll try reproducing with caching

huumn:
Yep, I've confirmed there are no regressions. Let's give a shot.

PR #1843, opened 2025-01-25
Fix missing push notifications for thread subscriptions

ekzyis:
## Description

I subscribed to a post and I expected to receive push notifications but I didn't because we don't send any.

Turns out we likely never had push notifications for thread subscriptions afaict. At least I couldn't find a commit where we had them using `git log -S "ThreadSubscription"`.

This PR adds push notifications for these. It sends the same push notification as if you subscribed to the user that replied to the post you subscribed to because that's the same notification format I was going for.

It also makes sure that you only receive push notifications for stuff you would also find in your notifications by considering `users."noteAllDescendants"`:

push notifications:

```sql
-- send notifications for all levels?
AND CASE WHEN users."noteAllDescendants" THEN TRUE ELSE r.level = 1 END
```

notifications page:

https://github.com/stackernews/stacker.news/blob/b28407ee99fc68db91abe2cf99983a073ccece1d/api/resolvers/notifications.js#L88

TODO:

- [x] fix mutes not considered
- [x] fix duplicate push notification (thread subscription + reply)

## Additional Context

- Seems like `Reply.ancestorId` should have been called `Reply.rootId`
- https://github.com/stackernews/stacker.news/issues/1844

## Checklist

yes

`7`. Tested by subscribing to a post, and then replying to it with anon on different levels, muting anon and disabling `noteAllDescendants` setting

n/a

no


huumn:
I've noticed we're missing a bunch of push notification equivalents for stuff. Referral rewards is another one I think about. I think also territory revenue doesn't have one.

PR #1857, opened 2025-02-02
Update price button accessibility in header

jason-me:
## Description

Accessibility changes to price button to ensure that the `button` role is being used and the `aria-describedby` announces the correct hint for when the user activates the button.

## Screenshots
No changes to visual UI

## Additional Context

This is my first commit. I plan on doing more accessibility work for stacker.news in the future but wanted to start small to ensure I am in sync with your contribution process. 

This PR simply changes a few lines on prices.js to ensure the price element is announced as a `button` to screen reader users with the proper `aria-describedby` which describes the next action the button will perform and that it is focusable and actionable by the keyboard. 

## Checklist

I believe so. These are very small changes that should not impact backwards compatibility.

8 - I tested my changes on a local dev environment running only the minimal containers to view real-time changes to the UI. Some blockheight and other bitcoin data was not present in the dev environment however. 

Yes, tested in an emulated mobile viewport on Chrome, also tested light mode in addition to dark

No, only used standard `html` elements, `aria` attributes, and added a visually-hidden `css` class.

### DIFF
--- a/components/header.module.css
+++ b/components/header.module.css
@@ -110,3 +110,14 @@
     background-color: var(--bs-body-bg);
     z-index: 1000;
 }
+
+.visually-hidden {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    padding: 0;
+    margin: -1px;
+    overflow: hidden;
+    clip: rect(0, 0, 0, 0);
+    border: 0;
+}
--- a/components/header.module.css
+++ b/components/header.module.css
@@ -110,14 +110,3 @@
     background-color: var(--bs-body-bg);
     z-index: 1000;
 }
-
-.visually-hidden {
-    position: absolute;
-    width: 1px;
-    height: 1px;
-    padding: 0;
-    margin: -1px;
-    overflow: hidden;
-    clip: rect(0, 0, 0, 0);
-    border: 0;
-}
--- a/components/price.js
+++ b/components/price.js
@@ -44,6 +44,15 @@ export function PriceProvider ({ price, children }) {
   )
 }
 
+function AccessibleButton ({ id, description, children, ...props }) {
+  return (
+    <div>
+      <button {...props} aria-describedby={id}>{children}</button>
+      <div id={id} className='visually-hidden'>{description}</div>
+    </div>
+  )
+}
+
 export default function Price ({ className }) {
   const [selection, handleClick] = usePriceCarousel()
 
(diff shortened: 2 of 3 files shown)


huumn:
Awesome Jason!

huumn:
Works! Thanks. Do you have a lightning address I can send sats to?

A few changes I made:

- I moved `.visually-hidden` to `./styles/globals.scss`
    - we use something called [css modules](https://nextjs.org/docs/14/pages/building-your-application/styling/css-modules) which means classes aren't accessible by their name alone unless they're in `./styles/globals.scss`
- In React, we can't use `class=` and instead have to use `className=` because React is creating the html programmatically
    - see: https://developer.mozilla.org/en-US/docs/Web/API/Element/className 
- I made an `AccessibleButton` component to make it a little more reusable. It's not necessary but it'll make changing it easier in the future.

jason-me:
Thanks @huumn ! I appreciate the css module reference... should get a bit more familiar with how React handles this :)

You can send sats to jasonbits@stacker.news 

Look forward to becoming more familiar with your codebase and contributing a bit more soon!

huumn:
> jsonbits@stacker.news

Have you attached a receiving wallet on SN? We transitioned to non-custodial, so without one you'll get gift credits that can only be spent on SN.

I can also send the sats to an invoice if that's easier.

jason-me:
I also just setup auto withdrawal to my voltage node so either way should work now.

huumn:
@jason-me I originally had trouble paying your invoice due to not being able to find a route. It looks like you may have hooked up the same node to SN.

Can you double check your lightning channels and make sure there's enough liquidity?

jason-me:
Hmnn... The invoice was from breez and the node I connected to my stacker.news lightning address was a separate voltage node. That node has plenty of liquidity so not sure why it failed.

You can try sending directly to the voltage address which is rrbtc@vlt.ge or you can try the invoice below from that same node generated via ZeusLN:

lnbc1m1pn606v4pp5dfjrytdn344uwpu5uxadsfegmqf25s3jvrty4nayu04ajumkja6qdqqcqzzsxqrrsssp5p85rdhsql68kmy8ggnee6xulft8pvrs9ul5w3e49vas8hk8alssq9p4gqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqpqysgqqkysxjaurjssc0gvx0hjrm20fuptlsw3q06zr7wgfpjcvgapfp5kd8arzwcflxt6hne6zp603l8g8zem25utnf24fc67sjt6e7s2elqpy74h8e

If that doesnt work. Dont worry about it you can get me on the next one. :)

J

jason-me:
Got it! Thanks, Keyan

PR #1858, opened 2025-02-03
fix: WebLN QR fallback for anon users

### LINKED ISSUE #1758: QR code fallbacks don't work for anon if WebLN enabled

### Describe the bug

If the WebLN wallet was enabled and we switch to anon, anon will also use the WebLN wallet because it only looks at `wallet:webln` (not scoped to user id).

However, if the WebLN prompt was canceled, we won't show a QR code because anon can't retry invoices.

This means that following attempts will simply do nothing and anon looks broken because WebLN prompts don't open again once they were canceled. Only a page reload brings back the WebLN prompt but still not possible to fallback to a QR code.

I think anons should be able to use WebLN but that we only try it if it was enabled while signed in is weird.

### Screenshots

_No response_

### Steps To Reproduce

1. Enable WebLN wallet
2. Switch to anon
3. Try to post something
4. Cancel WebLN prompt

### Expected behavior

Anon can fallback to QR code if WebLN enabled or gets immediately shown a QR code

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

Soxasora:
## Description

Fixes #1758 
When a WebLN wallet request fails (eg. cancelling a transaction) it can't be re-requested until next reload, throwing an error and prompting an invoice retry with QR that fails if the user is not logged in.

This fix checks for 'me' on `useWalletPayment`, throwing a new `AnonWalletError` that can be caught during QR fallback to show both the WebLN client and the QR payment

## Screenshots
https://github.com/user-attachments/assets/7bd6ee12-2089-44dd-8c0f-96a4075ac2aa

## Additional Context

Fix specifically applied for WebLN

## Checklist

Yes, only affects WebLN wallet

7, more testing

n/a

No


components/use-paid-mutation.js
```diff
@@ -61,10 +61,15 @@ export function usePaidMutation (mutation,
...
     const paymentAttempted = walletError instanceof WalletPaymentError
     if (paymentAttempted) {
-      walletInvoice = await invoiceHelper.retry(walletInvoice, { update: updateOnFallback })
+      try {
+        walletInvoice = await invoiceHelper.retry(walletInvoice, { update: updateOnFallback })
+      } catch (err) {
+        if (walletError.wallet === 'webln') { // show QR code for WebLN errors
+          throw err
+        }
+      }
```
Soxasora, line 72: While thinking about it, the major errors that this would ignore are already handled: Invoice cancel, Invoice expiration. I think that in this context, errors like Invoice already paid or Invoice failed shouldn't really happen... if I'm not mistaken

ekzyis (changes requested):
Mhh, it's a bit hard to understand the implications of these changes. For example, I am worried if the priority of the WebLN wallet makes a difference for these changes here.

I think a better way to fix #1758 is to make `useWalletPayment` aware of if we're logged in or not.

If we're not logged in, try to pay with WebLN if available (`typeof window.webln !== 'undefined'`) and ignore the rest of the code. That it currently might not ignore it for anons because WebLN might have been enabled (`wallets.length > 0)`) is the bug we should fix imo. 99% of our code for payments assumes you're logged in and thus that anons will never hit that code.

What anons can do in comparison to stackers is very limited (no sender and receiver fallbacks, no retries etc.) so I suspect if we (continue to) handle anons during payments immediately, it will be much easier to implement, maintain, read etc.

Soxasora:
Thanks a lot for the correct redirection, now we call WebLN's `sendPayment` when QR shows up! But only if we reached QR stage because of `AnonWalletError` that gets thrown by `useWalletPayment` if the user attempting to pay is Anon.

I think it's the correct behavior and gives a choice, if anything we could 'await' for sendPayment to throw an error to show the QR if we ever change idea.
Thanks again ^^

wallets/payment.js
```diff
@@ -26,6 +28,11 @@ export function useWalletPayment () {
       throw new WalletsNotAvailableError()
     }
 
+    // anon user cannot pay with wallets
+    if (!me) {
+      throw new AnonWalletError()
+    }
+
```
ekzyis, line 35: This depends on WebLN being enabled while logged in; otherwise, the check above (`wallets.length === 0`) will trigger first. If we place this before `wallets.length === 0`, anons can always use WebLN to pay if they have it.

ekzyis (changes requested):
You fixed the main issue in #1758:

> However, if the WebLN prompt was canceled, we won't show a QR code because anon can't retry invoices.

But we can do better and always allow WebLN payments for anon, even if the wallet wasn't enabled while logged in:

> I think anons should be able to use WebLN but that we only try it if it was enabled while signed in is weird.

So I suggested to change the order in which we check for errors.

Other than that, this is approved!

wallets/payment.js
Soxasora: I made the equivalent of typo in ordering lol, that was what was intended, fixed it ^^'

ekzyis (approved):
Approving this because this is more than enough to fix #1758, nice work!

But for the sake of a complete review, I want to mention that I noticed it would be nice if WebLN errors showed up in the QR code. I did not test, but I think that is not the case currently. I also noticed that the other `throw` statements should probably also use `WalletError` now. Since it's a small innocent change afaict, I went ahead and did that in ba93d347.

But these were nitpicks, mostly unrelated to the bug described in #1758 and not really important whatsoever so don't worry about not noticing them; just mentioning them to "open your third eye" regarding details even more (lol). We can do the WebLN<>QR error thing when we actually care about it if ever.

Issue #1861, opened 2025-02-04
Referral notifications should indicate the source content


Soxasora:
### Describe the problem you're trying to solve

Referral notifications don't clearly indicate the source of the referral.

### Describe the solution you'd like

For instance, when you refer someone, you’ll receive a notification indicating that they’ve been referred. You should be able to see the motivation behind the referral, such as the content you shared.

### Additional context

See [this comment](https://stacker.news/items/874374?commentId=875204)

Soxasora:
Implemented via #1862

PR #1862, opened 2025-02-04
enhance: referral notifications with source

Soxasora:
## Description

Addresses #1861 

Includes the source of referral on referral notification, showing the related Item, profile or territory.
Instead of getting data from the first entry of OneDayReferral, we push a row containing the referee's landing page, marked with a boolean.

## Screenshots
General and Post
image

Comment
image

Profile
image

Territory
image

## Additional Context
If there's nothing to show or the referral has type REFERRAL, it shows "because of you", as usual.
Push notification that shows "because of you" can probably stay the same, notifications page would provide more details

## Checklist

Yes

7, tested with multiple referral types

Yes

No


fragments/notifications.js
```diff
@@ -112,6 +112,17 @@ export const NOTIFICATIONS = gql`
...
+            item {
+              ...ItemFullFields
+            }
+            sub {
+              ...SubFields
+            }
+            profile {
+              name
+            }
+          }
```
huumn, line 125: This is a nitpick. But because these are mutually exclusive, it'd be more idiomatic to use a [graphql union](https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces#union-type).

huumn (changes requested):
When someone joins (and is a referree), we use cookies to assign their referrer. One day referrers only apply to referring other logged in stackers (stackers that have already joined).

The relevant code is: 
- where one day referrers are recorded: https://github.com/stackernews/stacker.news/blob/95e98501ec3e000c6b3c710d5809c5d04839d2b4/api/ssrApollo.js#L59
- this is where `r/<referrer>` is handled and passed along: https://github.com/stackernews/stacker.news/blob/95e98501ec3e000c6b3c710d5809c5d04839d2b4/middleware.js#L15
- when someone signs up, we grab the referral from cookies and attribute the referrer here: https://github.com/stackernews/stacker.news/blob/95e98501ec3e000c6b3c710d5809c5d04839d2b4/pages/api/auth/%5B...nextauth%5D.js#L79

I'm not sure what exactly needs to change to know which referral someone joined through, but using one day referrers won't do it.

fragments/notifications.js
Soxasora: Yes, absolutely!

Soxasora:
> One day referrers only apply to referring other logged in stackers

Oh, now I understand correctly the behavior, what seemed to confuse me was the fact that after signup the user would be redirected to the page they were before, giving me a correct referral source. But this wouldn't work if they sign up from a different page or get redirected somewhere else.
Mmh, I think that recording the first referred page they visit can suffice here

huumn:
Perhaps the move on sign up is to insert a row into the one day referrer table, and mark it with a boolean.

pages/api/auth/[...nextauth].js
```diff
@@ -79,6 +100,10 @@ function getCallbacks (req, res) {
         if (req.cookies.sn_referrer && user?.id) {
           const referrerId = await getReferrerId(req.cookies.sn_referrer)
           if (referrerId && referrerId !== parseInt(user?.id)) {
+            if (req.cookies.sn_referee_landing) { // if we have a landing referrer, record it
+              const refereeLanding = await getRefereeLanding(req.cookies.sn_referee_landing, user.id)
+              if (refereeLanding) await prisma.oneDayReferral.create({ data: refereeLanding })
+            }
             const { count } = await prisma.user.updateMany({ where: { id: user.id, referrerId: null }, data: { referrerId } })
             if (count > 0) notifyReferral(referrerId)
           }
```
Comment (Soxasora) [inline pages/api/auth/[...nextauth].js:109]: Here I kept the same logic and re-used some stuff that was already there

pages/api/auth/[...nextauth].js
```diff
@@ -56,6 +56,27 @@ async function getReferrerId (referrer) {
...
+    type = 'PROFILE'
+    typeId = refereeLanding.slice(8)
+  } else if (refereeLanding.startsWith('territory-')) {
+    type = 'TERRITORY'
+    typeId = refereeLanding.slice(10)
+  }
+
+  return { referrerId, refereeId, type, typeId, landing: true }
+}
+
```
Comment (Soxasora) [inline pages/api/auth/[...nextauth].js:79]: It will exit the moment it doesn't find an ID following the contentReferrer criteria, causing the referee to not have a 'refereeLanding'. The referrer will have this generic notification
image

middleware.js
```diff
@@ -25,6 +45,14 @@ function referrerMiddleware (request) {
...
     // explicit referees might click around before signing up.
     response.cookies.set(SN_REFERRER, referrer, { maxAge: 60 * 60 * 24 })
+
+    // we record the first page the user lands on and keep it for 24 hours
+    // in addition to the explicit referrer, this allows us to tell the referrer
+    // which share link the user clicked on
+    const contentReferrer = getContentReferrer(request, url)
+    if (contentReferrer) {
+      response.cookies.set(SN_REFEREE_LANDING, contentReferrer, { maxAge: 60 * 60 * 24 })
+    }
```
huumn, line 55: I moved this into the explicit referrer block. The way it was before, if I shared `https://stacker.news/r/k00b` then the referree visited an item, the notification would say "you shared the post/comment" when I didn't.

pages/api/auth/[...nextauth].js
```diff
@@ -77,10 +103,14 @@ function getCallbacks (req, res) {
...
-            const { count } = await prisma.user.updateMany({ where: { id: user.id, referrerId: null }, data: { referrerId } })
-            if (count > 0) notifyReferral(referrerId)
+          const referrerData = await getReferrerData(req.cookies.sn_referrer, req.cookies.sn_referee_landing)
+          if (referrerData?.referrerId && referrerData.referrerId !== parseInt(user?.id)) {
+            // if we have recorded a referee landing, record it in the db
+            if (referrerData.type && referrerData.typeId) {
+              await prisma.oneDayReferral.create({ data: { ...referrerData, refereeId: user.id, landing: true } })
+            }
+            const { count } = await prisma.user.updateMany({ where: { id: user.id, referrerId: null }, data: { referrerId: referrerData.referrerId } })
+            if (count > 0) notifyReferral(referrerData.referrerId)
```
Comment (huumn) [inline pages/api/auth/[...nextauth].js:113]: All the changes I made here relate to the changes I made in the middleware. ie the landing cookie is ONLY set if this is via an explicit (`/r/k00b`) referral because otherwise it's redundant.

worker/earn.js
```diff
@@ -74,6 +74,7 @@ export async function earn ({ name }) {
       FROM earners
       LEFT JOIN "OneDayReferral" ON "OneDayReferral"."refereeId" = earners."userId"
       WHERE "OneDayReferral".created_at >= date_trunc('day', now() AT TIME ZONE 'America/Chicago' - interval '1 day')
+      AND "OneDayReferral".landing IS NOT TRUE
```
huumn, line 77: Nice catch with this change! I would've forgot about that.

components/notifications.js
```diff
@@ -528,11 +529,27 @@ function WithdrawlPaid ({ n }) {
...
+    case 'Item':
+      referralSource = (Number(me?.id) === Number(n.source.user?.id) ? 'of your' : 'you shared this') + ' ' + (n.source.title ? 'post' : 'comment')
+      break
+    case 'Sub':
+      referralSource = (Number(me?.id) === Number(n.source.userId) ? 'of your' : 'you shared the') + ' ~' + n.source.name + ' territory'
+      break
+    case 'User':
+      referralSource = (me?.name === n.source.name ? 'of your profile' : `you shared ${n.source.name}'s profile`)
+      break
+  }
```
huumn, line 544: I mostly just made the copy a little more natural sounding if the source was your own item/territory/profile.

pages/api/auth/[...nextauth].js
Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:113]: These changes feel amazing! I was conflicted on referral handling here so I just went with adapting to the already present one.

middleware.js
Soxasora: Oh, what a fumble, I didn't realize that! Thanks a lot k00b ^^

PR #1894, opened 2025-02-10
THREAD notifications to distinguish direct replies from follow-ups

### LINKED ISSUE #1844: noteAllDescendants also disables push notifications for direct replies

### Describe the bug

We currently only send out push notifications for replies if `noteAllDescendants` is set:

https://github.com/stackernews/stacker.news/blob/b28407ee99fc68db91abe2cf99983a073ccece1d/lib/webPush.js#L37-L54

However, direct replies always show up in /notifications so push notifications are inconsistent with the notifications we show there.

To fix this, we would need to always send a push notification for every reply. 

### Screenshots

_No response_

### Steps To Reproduce

1. Enable push notifications
2. Disable `someone replies to someone who replied to me` in settings
3. Reply to yourself with anon
4. No push notification received but notification in /notifications

### Expected behavior

Push notification received

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

Soxasora:
## Description
Fixes #1844
Introduces THREAD notification type, it gets sent instead of REPLY whenever the item's parentId doesn't match the recipient's userId

## Additional Context

n/a

## Checklist

it checks for isDirect which can only be true if there are follow-up replies, otherwise it will send the usual REPLY notification

7, will edit once I tested on more scenarios

n/a

n/a

### DIFF
--- a/sw/eventListener.js
+++ b/sw/eventListener.js
@@ -91,7 +91,7 @@ const mergeNotification = (event, sw, payload, currentNotifications, tag, nid) =
   // merge notifications into single notification payload
   // ---
   // tags that need to know the amount of notifications with same tag for merging
-  const AMOUNT_TAGS = ['REPLY', 'MENTION', 'ITEM_MENTION', 'REFERRAL', 'INVITE', 'FOLLOW', 'TERRITORY_POST']
+  const AMOUNT_TAGS = ['REPLY', 'THREAD', 'MENTION', 'ITEM_MENTION', 'REFERRAL', 'INVITE', 'FOLLOW', 'TERRITORY_POST']
   // tags that need to know the sum of sats of notifications with same tag for merging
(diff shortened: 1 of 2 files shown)


sw/eventListener.js
```diff
@@ -116,6 +116,8 @@ const mergeNotification = (event, sw, payload, currentNotifications, tag, nid) =
   if (AMOUNT_TAGS.includes(compareTag)) {
     if (compareTag === 'REPLY') {
       title = `you have ${amount} new replies`
+    } else if (compareTag === 'THREAD') {
+      title = `you have ${amount} new follow-up replies`
```
ekzyis, line 120: I was too slow to review / forgot to review after I requested a review from myself but I suspect we will want to merge this with `REPLY` later. I don't really need to know in my push notification at which level I received a reply. We also don't distinguish them in /notifications.

But we can keep it as-is for now and see. Just wanted to mention this as a UX thing to consider in the future :)

PR #1899, opened 2025-02-12
Don't parse content in code blocks

### LINKED ISSUE #1898: Text after @ is swallowed in code block
See https://stacker.news/items/881667 and compare with https://stacker.news/items/881667/edit

ekzyis:
## Description

Fix #1898

Currently, SN will render this:

```
# test

asdf@foobar

[](url)

![](image)

```

like this:

```
# test 

asdf
```

This is fixed by preventing any parsing inside code blocks.

## Additional Context

If you run `console.log(tree)` before `visit` is called, you still get the processed tree printed to console. I think that's because the tree is processed inline and the console will show the latest state of an object.

Therefore, I used `console.log(JSON.stringify(tree, null, 2))` to see the unprocessed tree in the console.

## Checklist

If you mean there's no regression, I think so, yes.

~`3`. Only tested the code block in the description but not processing anything inside a code block sounds like what we want.~

Tested some more, I give this a `7` now.

n/a

no

### DIFF
--- a/lib/rehype-sn.js
+++ b/lib/rehype-sn.js
@@ -16,6 +16,11 @@ export default function rehypeSN (options = {}) {
   return function transformer (tree) {
     try {
       visit(tree, (node, index, parent) => {
+        if (parent?.tagName === 'code') {
+          // don't process code blocks
+          return
+        }
+
         // Handle inline code property
         if (node.tagName === 'code') {
           node.properties.inline = !(parent && parent.tagName === 'pre')


huumn (commented):
Seems logical to me. I'm no wiz with this stuff. I went deep on it one week and forgot most of it.

I'd be happy to ship if you QA enough and are confident in it.

ekzyis:
> I'm no wiz with this stuff. I went deep on it one week and forgot most of it.

But I thought you love trees @huumn :eyes: haha

> I'd be happy to ship if you QA enough and are confident in it.

Tested some more formatted text with and without code blocks, looks good to me!

Issue #1889, opened 2025-02-09
Territory ranking improvements


huumn:
Currently we have global trust scores and personalized trust scores for stackers, but neither of these consider the topic/domain/territory.

Additionally, founders of territories are conflicted - they'll often zap stuff in their territory to support people posting in their territory and not necessarily because they like the post. This can negatively affect their trust scores and ideally it wouldn't.

To improve this, we should

- modify trust computation to give conflicted and unconfliced global trust scores to territory founders
    - that is a global trust score when they are and aren't zapping things in their territory, so that their effect on homepage ranking is more accurate
- create independent trust graph for each territory where founder is the seed, and rank posts in the territory based on this graph and not global
    - this will involve adding a `subName` to the `Arc` table (where null is for global)
    - this will require giving each `Item` a `weightedVotes` for each territory it belongs to (when crossposting is live) and for the frontpage/global (`subName = null`)

Darth-Coin:
Please consider including in this issue the previous request: https://github.com/stackernews/stacker.news/issues/1665

PR #1900, opened 2025-02-13
WIP trust without conflicts

huumn:
This aims to do the first bullet of #1889 (which should be its own subissue)

As an interim solution to improve quality of life ASAP, and avoid boiling the ocean, this computes two trust graphs - one without conflicts and with and records them independently.

TODO

- [ ] rewrite `zap_rank_personal_view` to take into account conflicts
- [ ] make required changes to `zap`
- [ ] testing
- [ ] make backwards compatible


huumn:
Deleting in favor of another approach.

Issue #1596, opened 2024-11-15
Refresh JWTs for account switching


ekzyis:
### Describe the problem you're trying to solve

The JWT stored in the `multi_auth` cookies are never refreshed. This means they expire in 30 days.

This could be the reason for #1573 and why account switching stops working after some time. It would at least make sense.

Currently, account switching probably just completely stops working and you need to clear cookies to recover. This is pretty bad and something I didn't really make sure fails gracefully in #489.

### Describe the solution you'd like

Refresh all JWTs stored `multi_auth` cookies on every request—or at least the one that is currently used since we always use the `multi_auth` cookie if the pointer cookie is set.

### Describe alternatives you've considered

Fail gracefully if JWTs are expired

ekzyis:
Maybe we can refresh all JWTs at the same time we're checking them which user we should use.

We do this here:

https://github.com/stackernews/stacker.news/blob/8c43caed80b38b7450796df5a35bea196ed1845a/pages/api/graphql.js#L85-L123

AndreaDiazCorreia:
Can I solve this issue?

ekzyis:
Sure, I'll assign you

riccardobl:
are you still working on this @AndreaDiazCorreia ?

AndreaDiazCorreia:
Yes, sorry for the delay

riccardobl:
That's alright, take your time, I just wanted to check in.

ekzyis:
This was implemented in https://github.com/stackernews/stacker.news/pull/1957. Closing.

PR #1904, opened 2025-02-13
Refresh active JWT tokens

Soxasora:
## Description

Addresses #1596
Refreshes the logged-in multi-auth account JWT token together with session

## Additional Context

By passing response we trigger setMultiAuthCookies that we used to trigger only on login/sign-up.

We could also specialize the token Set-Cookie to avoid accessing db for user data if we **just** want the token to be refreshed:
```javascript
res.appendHeader('Set-Cookie', cookie.serialize(`multi_auth.${id}`, jwt, cookieOptions))
```
But by refreshing both the multi-auth current user cookie and the user list, we'll probably have less points of failure

## Checklist

todo

**5**, normal multi-auth between 3 users, refreshes the multi-auth token together with session, no hiccups

n/a

todo


ekzyis (changes requested):
This looks very close to what we want!

However, two things:

1. As you mentioned in chat, this does a db call to fetch the current user which we don't want since this runs on every request
2. This only updates the `multi_auth.<id>` token of the current user, but not the other ones

I imagine we can solve both via a new function `refreshMultiAuthCookies` that looks at the user list stored in the `multi_auth` cookie to loop over every cookie with `multi_auth.<id>` and then issues a new JWT given the existing JWT (`decodeJWT` + `encodeJWT`).

Since we now always need `req` and `res` in the jwt callback, I noticed that instead of checking only for `req && res` to distinguish new logins and conditionally run `setMultiAuthCookies`, we can additionally check for `user`.

So this means if `user && req && res` is true, we run `setMultiAuthCookies` else we run `refreshMultiAuthCookies`.

Does this make sense?

Soxasora:
I tested with 5 accounts and got a flawless experience in account switching for the scenarios I tested!
Cookies are always refreshed for every logged-in account and I got the `multi_auth` user list cookie to update too to maintain refresh functionality.

ekzyis (commented):
This does refresh all JWTs of other users but I noticed following things:

1. it does not refresh `multi_auth.user-id` which might be important to do since else, the cookie pointer might not get sent to the server anymore :thinking: 

2. when I switch between users, I get `ERR_HTTP_HEADERS_SENT`  in the server logs but the switch seems to work as expected, see video:

https://github.com/user-attachments/assets/351bbd2b-809c-406f-a3b8-d260cb5add6c

3. this also happens randomly while idle:

https://github.com/user-attachments/assets/5a61abd0-9a71-4079-95bf-8df2b1251098

4. I noticed that a signout can lead to `multi_auth.<id>` and `multi_auth.user-id` getting deleted but not `multi_auth` so the code in `refreshMultiAuthCookies` would try to decode `undefined`, see screenshot:

2025-03-05-145229_1914x1049_scrot

This isn't caused by your code, but we need to handle it until it's fixed. I think we can just check if the token exists before trying to decode it.

I'll debug pages/api/signout.js today. I also noticed that once you switch between users, `multi_auth.user-id` expires after the session unlike the other cookies for authentication, including `next-auth.session-token`, which expire after 30 days. Maybe that's related to the issues we're seeing with account switching. :pray:

Soxasora:
Awesome! Thanks for catching all the stuff I didn't see!

> `ERR_HTTP_HEADERS_SENT`

I realized that headers were being sent prematurely which meant that a practical way to solve this would be to group all the cookies to be set and then push all of them together, I didn't get any error after doing this and switching works for me ^^

> 4. I noticed that a signout can lead to `multi_auth.<id>` and `multi_auth.user-id` getting deleted but not `multi_auth` so the code in `refreshMultiAuthCookies` would try to decode `undefined`

This is something I didn't encounter though, in my case `signout` do edit `multi_auth` cookie
Before signout:
```
[
  { id: 21858, name: 'soxa_dev', photoId: null },
  { id: 21859, name: 'soxa_dev2', photoId: null }
]
```

After signout:
```
[ { id: 21859, name: 'soxa_dev2', photoId: null } ]
```

Nonetheless to avoid every kind of weird scenarios in which an incorrect multi_auth edit happens, now there's a condition to refresh the token if it's actually present.

> `multi_auth.user-id` expires after the session unlike the other cookies for authentication

Another great catch that makes a lot of sense, now it gets refreshed!

Super grateful of the work you do with me ^^
I'll continue monitoring the code and its effects to iron out other kinks that may appear.

pages/api/auth/[...nextauth].js
```diff
@@ -182,26 +183,35 @@ function setMultiAuthCookies (req, res, { id, jwt, name, photoId }) {
...
+      cookies.push(cookie.serialize(`multi_auth.${user.id}`, jwt, cookieOptions))
+    }
   }
 
   // keep multi_auth user list cookie up to date
-  res.appendHeader('Set-Cookie', cookie.serialize('multi_auth', b64Encode(multiAuth), { ...cookieOptions, httpOnly: false }))
+  cookies.push(cookie.serialize('multi_auth', b64Encode(multiAuth), { ...cookieOptions, httpOnly: false }))
+
+  // set all the cookies
+  res.setHeader('Set-Cookie', cookies)
```
Comment (ekzyis) [inline pages/api/auth/[...nextauth].js:214]: I think we need to use `appendHeader` to not accidentally overwrite other calls :thinking:

ekzyis (commented):
I will rebase #1957 on this PR so all the changes to account switching can be reviewed at once.

Normally, it's nice to review them separately but I think since there's some interaction between the code in #1957 and the code in this PR, it's probably actually easier to implement, test and review them together and not separate.

ekzyis:
> I realized that headers were being sent prematurely which meant that a practical way to solve this would be to group all the cookies to be set and then push all of them together, I didn't get any error after doing this and switching works for me ^^

I realized that the issue was most likely a missing `await`. I encountered the same issue when I was implementing the refresh in #1957 because `encodeJWT` and `decodeJWT` are async, I was missing an `await` on `refreshMultiAuthCookies`. That fixed it.

Soxasora:
> was most likely a missing `await`

A facepalm moment for sure thanks for spotting it, I thought it was because cookies were getting lost... in some way?

huumn:
Should this be closed in favor of https://github.com/stackernews/stacker.news/pull/1957?

PR #1907, opened 2025-02-14
Fix receiver fallback on caller error

ekzyis:
## Description

Our loop over receiver wallets is setup in a way that if anything fails after we created the invoice, we will not try the next wallet since we already exited the loop.

This affects p2p zaps, direct payments and autowithdrawals.

This PR fixes this by making the caller aware of the loop via an [async generator function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator). This means the caller can continue the loop on any errors.

This is the first time we use generators in our codebase, but I think this is a very good use case for generators.

## Videos

bug + fix for p2p zaps:

https://github.com/user-attachments/assets/bbe40c6e-6032-4fbb-b95b-300fdeb6f9e9

bug + fix for direct payments:

https://github.com/user-attachments/assets/2aa644f7-cce3-4c2d-87d9-5b8e11492900

bug + fix for autowithdrawals:

https://github.com/user-attachments/assets/71fb26b4-0e20-49ae-aac0-b2511e643174

## Additional context

In the video for autowithdrawals, you can see that the logs are in a wrong order. I suspect that's because we don't await the logger calls. I will fix this in a different PR or while normalizing the wallet logs.

## Checklist

yes

`7`. See videos.

n/a

no

### DIFF
--- a/worker/autowithdraw.js
+++ b/worker/autowithdraw.js
@@ -1,6 +1,6 @@
 import { msatsSatsFloor, msatsToSats, satsToMsats } from '@/lib/format'
 import { createWithdrawal } from '@/api/resolvers/wallet'
-import { createInvoice } from '@/wallets/server'
+import { createUserInvoice } from '@/wallets/server'
 
 export async function autoWithdraw ({ data: { id }, models, lnd }) {
   const user = await models.user.findUnique({ where: { id } })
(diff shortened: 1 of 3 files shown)


worker/autowithdraw.js
```diff
@@ -42,14 +42,18 @@ export async function autoWithdraw ({ data: { id }, models, lnd }) {
...
+    expiry: 360
+  }, { models })) {
+    try {
+      return await createWithdrawal(null,
+        { invoice, maxFee: msatsToSats(maxFeeMsats) },
+        { me: { id }, models, lnd, wallet, logger })
+    } catch (err) {
+      console.error('failed to create autowithdrawal:', err)
+      logger?.error(`incoming payment failed: ${err}`, { bolt11: invoice })
+    }
```
ekzyis, line 57: oh noes, forgot to rethrow the error here

PR #1909, opened 2025-02-15
fix: Dynamically import React Syntax Highlighter

### LINKED ISSUE #1242: Syntax highlighting

Soxasora:
## Description
Fixes #1242 
Dynamically imports React Syntax Highlighter and uses the correct theme for the highlight.js based highlighter

## Screenshots
image

## Additional Context

This moves a fair bit of `react-syntax-highlighter` stuff that was present inside the `_app` bundle and by dynamically importing the highlight.js based highlighter we can highlight stuff again while preventing over 0.7~1.5 MB of base bundle growth.[^1]

To avoid disrupting the already established Markdown parser, `Code` dynamically imports libraries when called instead of dynamically importing `Code` itself

Noting that #1606 seems to have an alternative package that is much smaller than `r-s-h` that imports 1.5 MB of highlight.js

## Checklist

Yes

7, tested with javascript, java, python

## Bundle Analysis

First load JS
Before
image

After
image

New NextJS chunk, loaded when there's a code block (see additional context)
image

[^1]: this 1.5 MB of growth is not present in the actual code because syntax highlighting isn't enabled

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -1,8 +1,7 @@
 import styles from './text.module.css'
 import ReactMarkdown from 'react-markdown'
 import gfm from 'remark-gfm'
-import { LightAsync as SyntaxHighlighter } from 'react-syntax-highlighter'
-import atomDark from 'react-syntax-highlighter/dist/cjs/styles/prism/atom-dark'
+import dynamic from 'next/dynamic'
 import React, { useState, memo, useRef, useCallback, useMemo, useEffect } from 'react'
 import MediaOrLink from './media-or-link'
 import { IMGPROXY_URL_REGEXP, decodeProxyUrl } from '@/lib/url'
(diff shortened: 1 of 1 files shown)


huumn:
1.5 MB for first load is crazy. It used be to 500kB. I wonder what bloated that.

Soxasora:
Exactly, 1.5MB is the parsed value of the chunk that gets now lazy loaded, it contains all the languages supported by highlight.js and other RSH dependencies.

Before we would have had ~700 kB of RSH sitting in _app and highlight.js too if working, overall I'm glad the first load benefits from this ^^

Issue #1832, opened 2025-01-21
Reduce bundle size by 10-50%


huumn:
Each additional 10% reduction will double the reward. 20% gets 200k, 30% 300k, 40% 500k, 50% 1m.

joelklabo:
I've submitted a PR targeting bundle size reduction: https://github.com/stackernews/stacker.news/pull/2817

Changes include:
- Enabled gzip compression (was disabled)
- Enabled SWC minification
- Added optimizePackageImports for large libraries (Apollo, Lexical, Nostr, charts)
- Added modularizeImports for react-bootstrap

These optimizations should reduce bundle size through better tree-shaking and smaller transferred files.

KingParmenides:
Submitted PR: https://github.com/stackernews/stacker.news/pull/2964

This targets the 50% bundle-size bounty tier by moving the rich Lexical editor stack out of the app-wide initial `_app` bundle. Measured on current `master` (`64afb581`):

- `_app` JS chunk: 12,381,692 B -> 2,756,405 B (-77.7%)
- `_app` entrypoint incl. webpack/framework/main: 12,722,902 B -> 3,099,098 B (-75.6%)

Checks run locally: `npm run build`, `npm run lint`, `git diff --check`.

BTC payout address if needed: 39Q34P8A7g375yqEr8buvNJkUbRgKfbKQZ

skotchdopolepet:
Submitted PR #2965 for this bounty: lazy-loads the rich text editor so the shared _app chunk drops from ~12,096 KB raw to ~2,696 KB raw in local production builds (~77.7% reduction).

KingParmenides:
Submitted PR #2966 for this bounty: https://github.com/stackernews/stacker.news/pull/2966

Measured production build result from .next/static/chunks: total emitted JS 13.84 MiB -> 5.50 MiB (60.3% reduction), _app JS 11.81 MiB -> 2.77 MiB (76.5% reduction). PR includes verification commands and payout address.

skotchdopolepet:
Update on PR #2965: added a second runtime split beyond the editor lazy-load. Wallet send protocol modules and Nostr runtime now load on demand instead of shipping in the shared _app chunk. Final local production build measurement: _app 2,209,246 B raw / 679,373 B gzip; source map no longer includes @lexical/code-shiki, @nostr-dev-kit/ndk, @nostr-dev-kit/ndk-wallet, @shocknet/clink-sdk, or @cashu/cashu-ts.

[...]

PR #1910, opened 2025-02-16
Dynamically import MathJax

Soxasora:
## Description

Addresses #1832 
Dynamically imports MathJax whenever is needed. Checks for `$$some math$$` inside the text to be processed and loads MathJax when positive.

## Screenshots
image

## Additional Context

It created two chunks containing MathJax-full and MathJax's LaTeX font, configuring a 1.5MB download client-side when is needed.

I noticed that on Math blocks we load Code. I added a condition for React Syntax Highlighter that catches 'math' language and avoids loading the library since MathJax should handle math.

## Checklist

Yes

7, loads MathJax only on "$$math$$" and displays correctly, conditional rendering is also correct. RSH avoidance works correctly.

n/a

No

## Bundle Analysis
First Load JS 
~35% reduction of _app First Load JS
Before
image

After
image

Bundle:
~35% reduction of the _app bundle reduction, from 4.59MB down to 3.12MB.
Before
image
After
image

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -57,7 +57,7 @@ export default memo(function Text ({ rel = UNKNOWN_LINK_REL, imgproxyUrls, child
 
   // we only need mathjax if there's math content between $$ tags
   useEffect(() => {
-    if (/\$\$(.*?)\$\$/g.test(children)) {
+    if (/\$\$(.|\n)+\$\$/g.test(children)) {
       import('rehype-mathjax').then(mod => {
         setMathJaxPlugin(() => mod.default)
       }).catch(err => {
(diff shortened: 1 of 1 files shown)


components/text.js
```diff
@@ -227,43 +238,15 @@ function Table ({ node, ...props }) {
...
+const Code = dynamic(() => import('./code-block'), {
+  ssr: false,
+  loading: ({ inline, className, children, ...props }) => (
+    <code className={className} {...props}>
       {children}
-    </ReactSyntaxHighlighter>
+    </code>
   )
-}
+})
```
Soxasora, line 249: I realized that Code gets called every time Markdown gets processed, it's better to dynamically load Code only when actually needed.
Before, a MathJax rendering would also download RSH's library

huumn:
I noticed it didn't support multiline so I updated the regex. Back to vacation lol

```
$$
L = \frac{1}{2} \rho v^2 S C_L
$$ 
```

Issue #1905, opened 2025-02-14
Item popover does not consider commentId param


ekzyis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

If you paste a SN link like https://stacker.news/items/885043?commentId=885108, we render the comment id since that's where you will get scrolled to. However, the popover does not show the comment.

### Steps To Reproduce

1. Paste https://stacker.news/items/885043?commentId=885108 into a comment
2. Hover over `#885108`
3. Popover shows item 885043

### Expected behavior

Popover shows 885108

### Additional context

See https://stacker.news/items/885398

huumn:
+1 I've noticed this too

PR #1911, opened 2025-02-16
Fix issue #1905, item popover with commentId

Scroogey-SN:
This makes the popover show the comment instead of the post.

To test this locally, I had to modify .env.local with
```
PUBLIC_URL=https://localhost:3000
SELF_URL=https://localhost:3000
```
screenshot

Tested that
- item links still popover as before
- item and comment links, when clicked, still result as before

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -20,7 +20,6 @@ import TextareaAutosize from 'react-textarea-autosize'
 import { useToast } from './toast'
 import { numWithUnits } from '@/lib/format'
 import textAreaCaret from 'textarea-caret'
-import ReactDatePicker from 'react-datepicker'
 import 'react-datepicker/dist/react-datepicker.css'
 import useDebounceCallback, { debounce } from './use-debounce-callback'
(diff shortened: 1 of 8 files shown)


huumn:
Sorry about the wait. I'm on vacation until thursday but I'll get to it first thing.

huumn:
I reverted the unrelated change you made to lib/url.js - I'm not sure why that was there but if it's a fix for something else please open another pr.

I also ended up reverting the fix here - which to put it bluntly was a hack and broke other links like `http://localhost:3000/items/458289/related` because it was using the link text to determine the id. Instead, I made the change in the plugin that parses and passes `id` such that it passes the comment id if it exists. 

-------------

Welcome to the team! Please send a long a lightning address when you get a chance so I can send you some sats.

Scroogey-SN:
Thanks! Scroogey@coinos.io

The lib/url.js change is for issue #1924. I should make branches for individual push requests.

Issue #1049, opened 2024-04-09
Feature Request : Perpetual Edit


Radentor:
Title : Perpetual Edit
     Author : Radentor
     Status : Proposed
     Type : Feature
     Created : 2024-04-09

# Abstract
This issue describes the proposal of granting author with ability to perpetually edit his/her SN post.

# Motivation
Certain type of SN post (such as directory & statistic report) requires constant update to avoid possible misinformation. Even regular post could slightly benefited from this feature in case the author forget to include certain detail after 10 minute time window has passed.

# Consideration
- Alternatives :
a. Reposting with update
b. Delete older post & reposting with update
c. Invent new type of post with perpetual update

- Cost :
a. 100% original cost (infinitely free editable post)
b. 100% original cost with cooldown (cost-limited editable post)
c. Increased (150% or 200% of original cost per edit, with cooldown)

- Moderation :
a. Total control to author
b. Moderated by territory founder

- Timestamping (if edit history is implemented) :
a. Regular SN numbering
b. Timestamped with Bitcoin block 

- Transparency :
a. Edit history toggle-able by author
b. Edit history always visible to reader
c. No edit history at all

# Pros & Cons
- Pros :
1. Easier navigation (bookmarking once is enough)
2. Better UX (it's awkward to publish new post just to add "small update")
3. Decluttering (single maintained post > multiple post on every update)

- Cons :
1. Degradation of post quality (why would i thoroughly craft my post if i can edit it later on?)

# Earlier Mentions
- [Books And Articles Newsletter, Issue 10](https://stacker.news/items/496231)
- [Perpetual Edits?](https://stacker.news/items/492553)
- [Feature idea: turn off exponential pricing to encourage long-form writing](https://stacker.news/items/312673)
- [Feature idea: perpetually-editable SN posts](https://stacker.news/items/306665)

huumn:
I discuss design goal [here](https://stacker.news/items/492553/r/k00b?commentId=492677) and I think we want to use the [history table pattern](https://stacker.news/items/492553/r/k00b?commentId=492722).

Radentor:
> I discuss design goal [here](https://stacker.news/items/492553/r/k00b?commentId=492677) and I think we want to use the [history table pattern](https://stacker.news/items/492553/r/k00b?commentId=492722).

Looking forward to it

ekzyis:
[...]

PR #1915, opened 2025-02-18
wip Store Item before editing in OldItem

Soxasora:
## Description
Starts the work on #1049 
What we want is to guarantee an Item's immutability by storing the original Item before editing. Allowing us to see and compare the editing history.

We do this by using a `BEFORE UPDATE` trigger function on Item update, recording only if there's edited content.

## Additional Context

The goal was to achieve this via Postgres' polymorphism but Prisma doesn't support inheritance and at the moment we store what really can be changed via editing, which is the content (title, text, url).

It doesn't record an Item if edited in the 10 minutes timeframe.

## Checklist

tbd

tbd

tbd

tbd

## Progress

- [x] Store original Item in OldItem -- wip, working
- [ ] Editing history UI
- [ ] Edited Item UI
- [ ] Allow edits after 10 minutes
- [ ] Pay-per-edit predisposition


api/paidAction/itemUpdate.js
```diff
@@ -65,6 +65,28 @@ export async function perform (args, context) {
...
+        cloneBornAt: old.cloneBornAt,
+        cloneDiedAt: new Date(),
+        originalItemId: parseInt(id)
+      }
+    })
+
+    data.cloneBornAt = new Date() // we can use this to determine if the item has been edited
+    data.cloneDiedAt = null
+  }
+
```
Soxasora, line 89: Since we really just want to track Item content editing, I switched to a Prisma-based version of the trigger that I think it's more appropriate. Before, every update to Item could fire the `procedure` and confront the conditions.

Not having Item inheritance made me think that having just enough columns can be okay, but also having the whole item is okay to me

huumn:
The non-trigger-based versioning, and of only select fields, makes me a little nervous tbh. I haven't looked at it very carefully at it though. I'll try to take a closer look when I'm less tired.

Soxasora:
Reverting to trigger is something I was debating myself some days ago, I just want to do it better than what it was before switching to Prisma.

I think we can do a 1:1 copy of Item for OldItem! I was selecting just the bare minimum for it to be displayed.

Thanks k00b, redrafting! ^^

PR #1919, opened 2025-02-20
enhance: Allow switching to a more expensive territory

### LINKED ISSUE #1918: Allow switch to expensive territories by paying the difference

You can only change the territory of a post if the new territory has equal or less post cost.

Allow switching to an expensive territory by paying the difference between your old territory and your new territory.

If for example you paid 100 sats to post on ~Design but you actually wanted to post on ~Photography where a post costs 250 sats, pay only 150 sats.

Deleting a post and re-creating it in the correct territory, paying full price.

### Additional context

[#890604](https://stacker.news/items/890416/?commentId=890604)

Soxasora:
## Description

Fixes #1918 
Allows switching to a more expensive territories during post editing by paying just the difference between the old territory post cost and the new one. 
A downgrade will have 0 cost.

## Screenshots
I'm moving from B2B which costs 1 sat to everythinghere that costs 2780 sats.
The edit cost will be 2779 then
image

## Additional Context

Perpetual Edits will allow this change to be done indefinitely

## Checklist

Yes

7, tested before on perpetual edits, tested on multiple territories, downgrade has 0 cost

n/a

No

### DIFF
--- a/api/paidAction/itemUpdate.js
+++ b/api/paidAction/itemUpdate.js
@@ -12,12 +12,19 @@ export const paymentMethods = [
   PAID_ACTION_PAYMENT_METHODS.PESSIMISTIC
 ]
 
-export async function getCost ({ id, boost = 0, uploadIds, bio }, { me, models }) {
+export async function getCost ({ subName, id, boost = 0, uploadIds, bio }, { me, models }) {
   // the only reason updating items costs anything is when it has new uploads
-  // or more boost
+  // or more boost or is switch to a more expensive sub
   const old = await models.item.findUnique({ where: { id: parseInt(id) } })
   const { totalFeesMsats } = await uploadFees(uploadIds, { models, me })
-  const cost = BigInt(totalFeesMsats) + satsToMsats(boost - old.boost)
+  let cost = BigInt(totalFeesMsats) + satsToMsats(boost - old.boost)
+  if (old.subName !== subName) {
+    const oldSub = await models.sub.findUnique({ where: { name: old.subName } })
+    const newSub = await models.sub.findUnique({ where: { name: subName } })
+    const oldCost = oldSub?.baseCost ?? 0
+    const newCost = newSub?.baseCost ?? 0
+    cost += satsToMsats(Math.max(0, newCost - oldCost)) // user will pay just the difference
+  }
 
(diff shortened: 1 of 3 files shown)


huumn (commented):
The main issue with this is that it neglects the spam multiplier.

This is hard to do because without recording the spam multiplier, all we're able to do is deduce that there was a spam multiplier. 

This creates a small spam loophole. For example:

00:00: Post once in ~bitcoin for base cost of 100 sats
01:00: Post something else in ~bitdevs for base cost of 10 sats * 10 = 100 sats
01:01: Edit post to be in ~bitcoin for 90 sats more, effectively saving 810 sats

----------

As part of my paid action refactor plans, I intend to create a `Receipt` table which will have the line items in it and this kind of thing should be easier to do. Not sure what to do in the meantime.

Soxasora:
Oh, you're right! I didn't foresee this loophole... and I think we can adapt the spam multiplier from itemCreate to itemUpdate for it (just for sub switch[^1]), I'm going to take a look.

Thanks a lot!

[^1]: I also thought about extending this to perpetual editing but there may be a lot of typo editing, at least in the first 10 minutes

api/paidAction/itemUpdate.js
```diff
@@ -12,20 +12,34 @@ export const paymentMethods = [
...
+  if (subDifferenceCost === 0) return 0
+
+  // calculate spam cost exclusively for sub switch
+  const [{ totalCost }] = await models.$queryRaw`
+  SELECT ${satsToMsats(subDifferenceCost)}::INTEGER
+    * POWER(10, item_spam(NULL, ${me?.id ?? USER_ID.anon}::INTEGER,
+        ${me?.id ? ITEM_SPAM_INTERVAL : ANON_ITEM_SPAM_INTERVAL}::INTERVAL))
+    * ${me ? 1 : 100}::INTEGER as "totalCost"`
+
+  return totalCost
```
Soxasora, line 33: It's mostly the same as itemCreate's spam multiplier, another change is parallel sub base cost querying.
I did some tests, following also the loophole scenario you mentioned and it's smooth!

huumn:
It was initially blocked by complications of the spam multiplier IIRC. Given this isn't pressing, it'll probably be worth waiting for the refactor.

PR #1930, opened 2025-02-27
fix #1167: allow pointer events in linkBoxParent pre for scrollbar

### LINKED ISSUE #1167: It's not possible to scroll left-right in code blocks on the `notifications` page
**Description**

It's not possible to scroll left-right in code blocks on the `notifications` page.

This issue is only noticed on the `notifications` page so far, not in comments || replies

**Steps to Reproduce**
A clear and concise way we might be able to reproduce the bug.

- Create a reply to some stacker's post with content like 
    ```
     \``` <-- ignore that slash... i had to put it there for formatting this report
     very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line very long text line
     \```
    ```
- get one of your awesome friends to zap the reply
- go to your `stacker.news/notifications` to see that your reply can not be scrolled left-to-right
    

**Expected behavior**

you should be able to scroll left-to-right on the long code block

**Environment:**
 - Browser [chrome & safari, at least]

Scroogey-SN:
This affects also other item lists, like search results (e.g. recent comments).

The scrollbar isn't visible because of CSS pointer-events: none.
Adding an exemption for pre fixes that.

This also makes it so the code block can be marked and copied to clipboard, which I think is good.

However, it makes it somewhat harder to click the item to open it, especially if the code block takes up most of the area.


huumn:
I wonder if it might be better to force code blocks to wrap. idk though. I'll review soon and come to a conclusion.

Issue #1849, opened 2025-01-28
Create awards.csv entries on PR merge


huumn:
When I merge a PR, I want a github action that automatically adds rows and columns to https://github.com/stackernews/stacker.news/blob/master/awards.csv. It can either update the any PRs pending review with an additional edit to `awards.csv`, or open another PR automatically with the changes, or whatever. It doesn't matter so long as it's automatic.

e.g. when #1659 was merged, it would've appended the following lines to `awards.csv`

```csv
Soxasora,pr,#1659,#1657,easy,,,,100k,soxasora@blink.sv,???
sig47,issue,#1659,#1657,easy,,,,10k,siggy47@stacker.news,???
```

If we don't have an existing entry with an lnaddr for a github handle, it'd put `???` in the `receive method` column.

It should also take into account all rules from the OSCARS (see readme), ie difficulty levels, priority multiplier, penalties, etc.

Scroogey-SN:
Is there a formal link between the merged PR (1659) and the issue it solves (1657), i.e. is Github really aware that the former solves the latter?

I see nothing obvious in the API https://api.github.com/repos/stackernews/stacker.news/pulls/1659 and AFAIK the action interface doesn't have more, either.

Or do you have to unreliably search through the PR title and body for numbers that might be issues?

Scroogey-SN:
Are the following lines legitimately missing?

```
Soxasora,pr,#1925,#1832,easy,,,100k,soxasora@blink.sv,???
Soxasora,pr,#1910,#1832,easy,,,100k,soxasora@blink.sv,???
Soxasora,pr,#1909,#1242,medium,,,250k,soxasora@blink.sv,???
Soxasora,pr,#1862,#1861,medium,,1,225k,soxasora@blink.sv,???
Soxasora,issue,#1862,#1861,medium,,,22500,soxasora@blink.sv,???
```

The first two merged PRs also reference the same issue, what are the rules in this case?

huumn:
I can't give detailed requirements. I of course don't mean do the impossible or do what's unreasonable. 

Part of what I'm paying for is the implementer thinking and reasoning through things like this.

m0wer:
Should this issue be closed? Looks like it was implemented already.

huumn:
Yep, thanks!

PR #1937, opened 2025-03-02
Extend awards action

Scroogey-SN:
## Description

Implements #1849

Pass ${{ github }} through env var GITHUB_CONTEXT to the script instead of passing it on the command line.

Tested with various PR body texts (backticks, single quotes, double quote, new lines, unicode, etc.) and multiple PR merge methods (squash, merge, etc.).

### DIFF
--- a/.github/workflows/extend-awards.yml
+++ b/.github/workflows/extend-awards.yml
@@ -7,7 +7,11 @@ on:
       - master
 jobs:
   if_merged:
-    if: github.event_name == 'pull_request' && github.event.action == 'closed' && github.event.pull_request.merged == true
+    if: |
+      github.event_name == 'pull_request' &&
+      github.event.action == 'closed' &&
+      github.event.pull_request.merged == true &&
+      github.event.pull_request.head.ref != 'extend-awards/patch'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


Scroogey-SN:
[Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable) covers this in detail.

huumn:
It failed again but it looks related to repo permissions. I'm not sure what needs to change.

Scroogey-SN:
I'll check.

Scroogey-SN:
It seems the permission can be set on enterprise, organization, and repository level:

https://stackoverflow.com/questions/72376229/github-actions-is-not-permitted-to-create-or-approve-pull-requests-createpullre

But you previously posted a screenshot showing it enabled in the repo, so it might not be this.

Scroogey-SN:
As per [Modifying the permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) you can see the permissions of the token in the Actions log:

screenshot

You have 'read' for everything, while my private test repo shows 'write' there.

The link above explains how this can be changed in enterprise/organization levels.

We can try to add a permissions: line in the .yml.

huumn:
Both the repo and the org have read/write enabled. It was the first thing I checked. I guess we need the yaml thing.

PR #1939, opened 2025-03-03
Extend awards action

Scroogey-SN:
Implements #1849 

add permissions: line


huumn:
I modified to only need relevant write permissions. First I only added `pull-requests: write` but then `contents: write`. I think that should do it but we'll need to wait for another PR to test I guess.

PR #1941, opened 2025-03-03
Also check for `user` before setting multi auth cookies

ekzyis:
While reviewing #1904, I noticed checking `req && res` to detect if it's a login is weird. Not sure why I did that since `next-auth` already gives us a way to detect logins:

> The arguments _user_, _account_, _profile_ and _isNewUser_ are only passed the first time this callback is called on a new session, after the user signs in. In subsequent calls, only `token` will be available.

-- https://next-auth.js.org/configuration/callbacks#jwt-callback

So I moved the code into the existing `if (user)` block and tested that we can still:

1. login
2. add a new account
3. logout (until full logout)

Authentication is critical piece of code so to keep the `if (req && res)` check might make sense but `req` is definitely set (we also access it above) so I only added `if (res)`. `res` should also always be set ... but only afaict.

**Update**

Changed my mind regarding this PR, see https://github.com/stackernews/stacker.news/pull/1941#discussion_r1978479895

### DIFF
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -116,14 +116,6 @@ function getCallbacks (req, res) {
             }
           }
         }
-
-        // add multi_auth cookie for user that just logged in
-        if (res) {
-          const secret = process.env.NEXTAUTH_SECRET
-          const jwt = await encodeJWT({ token, secret })
-          const me = await prisma.user.findUnique({ where: { id: token.id } })
-          setMultiAuthCookies(new NodeNextRequest(req), new NodeNextResponse(res), { ...me, jwt })
-        }
       }
 
       if (token?.id) {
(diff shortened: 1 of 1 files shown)


Soxasora (approved):

pages/api/auth/[...nextauth].js
```diff
@@ -116,6 +116,14 @@ function getCallbacks (req, res) {
...
           }
         }
+
+        // add multi_auth cookie for user that just logged in
+        if (res) {
+          const secret = process.env.NEXTAUTH_SECRET
+          const jwt = await encodeJWT({ token, secret })
+          const me = await prisma.user.findUnique({ where: { id: token.id } })
+          setMultiAuthCookies(new NodeNextRequest(req), new NodeNextResponse(res), { ...me, jwt })
+        }
```
Comment (huumn) [inline pages/api/auth/[...nextauth].js:126]: Did you check that multiple auth method linking still works? `token.sub = Number(token.id)` is needed to make that work, and it came before this previously and now it comes after.

Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:126]: Oh I didn't catch that during tests, the move here would be to do what @ekzyis first suggested and do `if (user && req && res) else if (req && res)`

Comment (ekzyis) [reply pages/api/auth/[...nextauth].js:126]: Ohh, good catch! I did not test before but did now with nostr and it successfully linked it. But I think the real test is with Github or Twitter. Will test Github.

Comment (ekzyis) [reply pages/api/auth/[...nextauth].js:126]: Tested linking Github now, still works!

Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:126]: Ah, the multi auth cookie we save on login won't have the sub property with token.id but we do indeed return a token with the property later, that would explain why it works!

Comment (ekzyis) [reply pages/api/auth/[...nextauth].js:126]: > the multi auth cookie we save on login won't have the sub property with token.id

Oh, you're right!

I think it's better to store the same token as we return. I'll change the code back.

PR #1953, opened 2025-03-07
Fix no pw manager autofill for device sync password

### LINKED ISSUE #1950: Device Sync does not use password text bot on mobile
On mobile it brings up my normal keyboard instead of the system keyboard when typing in device sync passphrase. This is annoying because then I can't open my password manager easily, and I don't trust my normal keyboard with passwords.

ekzyis:
## Description

fix #1950

I think that we use `<textarea>` instead of `<input>` is related, see video.

I tried yesterday to get my password manager autofill to work with `<textarea>`, which is what we currently use for the device sync password input, but I am not sure it's possible even though it has `autocomplete="current-password"` and `type="password"` set. So I think we need to switch to `<input>`.

## Video

https://github.com/user-attachments/assets/c291b846-7b1d-4eaf-aed1-d5a270d215d1

## Additional Context

Password manager autofill is also related to us wanting to switch to a dedicated input component for bip39 with separate inputs for each word: I guess with separate inputs you can't use password manager autofill 🤔

edit: or maybe it could work by using autofill on the first word input and then our code puts each word into its own input by splitting on whitespace :thinking:  

## Checklist

yes

`7`. See video.

n/a

no

### DIFF
--- a/pages/settings/passphrase/index.js
+++ b/pages/settings/passphrase/index.js
@@ -131,7 +131,6 @@ function Enabled ({ setVaultKey, clearVault }) {
           placeholder=''
           required
           autoFocus
-          rows={3}
           qr
         />
         <div className='mt-3'>
--- a/pages/settings/passphrase/index.js
+++ b/pages/settings/passphrase/index.js
@@ -131,7 +131,6 @@ function Enabled ({ setVaultKey, clearVault }) {
           placeholder=''
           required
           autoFocus
-          as='textarea'
           rows={3}
           qr
         />


Soxasora:
> Password manager autofill is also related to us wanting to switch to a dedicated input component for bip39 with separate inputs for each word: I guess with separate inputs you can't use password manager autofill 🤔
edit: or maybe it could work by using autofill on the first word input and then our code puts each word into its own input by splitting on whitespace 🤔

For quite a bit of time I wanted to adjust MultiInput to effortlessly support also device sync, atm it supports autofill (with proper adjustments?) the way you envisioned it as it switches inputs once it reaches eg. 6 characters ^^

ekzyis:
I'll check if we can keep the textarea functionality where it wraps to the next line instead of horizontal scroll

PR #1957, opened 2025-03-09
Reset multi_auth cookies on error

### LINKED ISSUE #1821: logged out randomly can't switch back to account
https://github.com/user-attachments/assets/c2595494-a3ff-4be2-8d6a-c807707385f6

I forgot to look at my cookies before I logged in.

I was able to get the behavior to come back by switching to another account. I tried switching to `@sn` and it switch me to anon. Then I tried switching back to `@k00b` which I was logged back into and now it won't switch.

Image

ekzyis:
## Description

fix #1821 based on #1956 #1974 supersedes #1904

I found out that we have a race condition on logout that exists without any account switching code: if a request starts before we receive the signout response (that deletes `next-auth.session-token`) and finishes after it, we're logged back in because it includes `next-auth.session-token` in the request headers and therefore also in the response headers thanks to the implicit token refresh on all requests.

However, I have also seen in the timings of the network tab that it's possible that a request still includes the cookie _after_ it was deleted AND _after_ we reloaded the page. This is somehow related to slow requests since this is reliably reproducible if we poll for `me` every second but sleep for 5 seconds on the server before replying. I don't know how it is possible to include a cookie in a request header that we previously deleted. It must be something the browser is doing like a request retry, but I don't think it makes sense for a browser to retry a request after a page was reloaded and any client-side code would be running in JS and JS can't set `httpOnly` cookies.

Since I wasn't able to prevent this from happening and I am not 100% sure that this is what causes #1821 or other issues with account switching that appear non-deterministic, I decided to show an error in the account switch dialog when we noticed that something is off with the cookies on the server.

This PR also includes the changes of #1904 so it also refreshes the cookies if there was no error.

## Screenshots

2025-03-17-171124_513x393_scrot

## Additional context

1. `next-auth.session-token` returned to refresh while switched to anon

This also refreshes the session token while we're switched to anon. However, this means that we now return a response header that includes the session token even when we're switched to anon.

This might be unexpected but since we still delete the session token on the server (just for the request lifecycle) when switched to anon and thus the server still returns responses as if we're logged out, I don't think this breaks anything.

But I wanted to mention this since it can be weird to return a session token even though we're not considered logged in.

This only happens if the multi auth check is successful. If we're completely logged out, there are no multi auth cookies and thus this does not happen.

2. no checks for invalid tokens

> Decided to not check for invalid tokens because it's not just a try/catch but also requires logic to update the list and switch to the next account which might not exist if the last token ended up becoming invalid. However, I think tokens can only become invalid if they expired so I think this complexity isn't worth it. I am pretty confident the issues we're seeing with account switching aren't related to invalid tokens that didn't expire yet.

-- https://github.com/stackernews/stacker.news/pull/1957#issuecomment-2726165243

## Checklist

yes

`7`. Tested switching between users and anon; switching to next account on logout; deleting any `multi_auth` cookie deletes all `multi_auth` cookies; token refresh

yes, tested mobile, light and dark mode

yes

### DIFF
--- a/components/account.js
+++ b/components/account.js
@@ -8,39 +8,30 @@ import { useQuery } from '@apollo/client'
 import { UserListRow } from '@/components/user-list'
 import Link from 'next/link'
 import AddIcon from '@/svgs/add-fill.svg'
+import { MultiAuthErrorBanner } from '@/components/banners'
 
 const AccountContext = createContext()
 
+const CHECK_ERRORS_INTERVAL_MS = 5_000
+
 const b64Decode = str => Buffer.from(str, 'base64').toString('utf-8')
-const b64Encode = obj => Buffer.from(JSON.stringify(obj)).toString('base64')
 
 const maybeSecureCookie = cookie => {
   return window.location.protocol === 'https:' ? cookie + '; Secure' : cookie
 }
 
 export const AccountProvider = ({ children }) => {
-  const { me } = useMe()
   const [accounts, setAccounts] = useState([])
   const [meAnon, setMeAnon] = useState(true)
+  const [errors, setErrors] = useState([])
 
   const updateAccountsFromCookie = useCallback(() => {
-    try {
-      const { multi_auth: multiAuthCookie } = cookie.parse(document.cookie)
-      const accounts = multiAuthCookie
-        ? JSON.parse(b64Decode(multiAuthCookie))
-        : me ? [{ id: Number(me.id), name: me.name, photoId: me.photoId }] : []
-      setAccounts(accounts)
-      // required for backwards compatibility: sync cookie with accounts if no multi auth cookie exists
-      // this is the case for sessions that existed before we deployed account switching
-      if (!multiAuthCookie && !!me) {
-        document.cookie = maybeSecureCookie(`multi_auth=${b64Encode(accounts)}; Path=/`)
-      }
-    } catch (err) {
-      console.error('error parsing cookies:', err)
-    }
+    const { multi_auth: multiAuthCookie } = cookie.parse(document.cookie)
+    const accounts = multiAuthCookie
+      ? JSON.parse(b64Decode(multiAuthCookie))
+      : []
+    setAccounts(accounts)
   }, [])
-
-  useEffect(updateAccountsFromCookie, [])
 
   const addAccount = useCallback(user => {
     setAccounts(accounts => [...accounts, user])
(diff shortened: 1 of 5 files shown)


huumn:
So in my example issue, the improvement this offers is telling me about the cookies (which if I'm not a SN dev I won't understand)? Do I need to logout of every account? Just one? 

Is there a way to just delete cookies/logout of accounts that can't be switched to? If we can't do that on the client, could we maybe loop over the accounts and make requests to the server to see if switching is possible, and if not, remove those accounts from the list?

ekzyis:
> So in my example issue, the improvement this offers is telling me about the cookies (which if I'm not a SN dev I won't understand)? 

I would say the improvement is that this tries to avoid any state where we're stuck in a weird state where we would need to clear the cookies manually. Instead, with this PR, this happens automatically and shows an error message that tells the stacker what to do to get account switching back: sign out and sign in again.

The error message about the cookies was meant for us devs and "To resolve these issues, please sign in and out again." was meant for stackers.

> Do I need to logout of every account? Just one?

When you see this error message, there is only one account you can logout of which is the one you're currently logged in as

> Is there a way to just delete cookies/logout of accounts that can't be switched to? If we can't do that on the client, could we maybe loop over the accounts and make requests to the server to see if switching is possible, and if not, remove those accounts from the list (auto-logout of them)?

Mhh, I think we could do that :eyes: Didn't think of that tbh. It could be a simple API route that tries to decode each token and returns which one failed if any.

But I am not sure if that would catch all issues we're seeing with account switching and solve #1821.

However, since if tokens become invalid for some reason[^1] the changes in this PR would NOT detect them so I think we should do both: check if any token fails to decode and then remove that one AND check if all cookies exist as they should (this PR). I noticed that when something broke regarding account switching, the user JWT was missing because that's set on the server but because of the logout race condition, we might be logged back in with broken account switching since there was some (broken) code for backwards compatibility to set cookies on the client, too.

[^1]: why would a valid token become invalid if it didn't expire yet?

Soxasora (commented):
This was undoubtedly necessary, a really great step towards better communication between us and the users! 

I tested the errors (only one cookie and both) and also cookie removal, everything works as intended. 

Now, I have just one suggestion on the message: I think it will be somewhat better, even if opaque, to either redirect to logout (or with a button) *or* hide the problematic cookies under a collapsed section
eg.

**We lost track of some of your accounts**
To resolve this issue, please sign out and sign in again.

 <details>

<summary>Technical details</summary>
multi_auth not found
multi_auth.user-id not found
</details>

I second removing only the affected cookies by checking if the session on server is valid.

> why would a valid token become invalid if it didn't expire yet?

I think it can’t become invalid on its own unless a [Ionizing particle from a cosmic ray](https://www.thegamer.com/how-ionizing-particle-outer-space-helped-super-mario-64-speedrunner-save-time/) hits the user’s device. It should require tampering *or* expiration date inconsistency (which also would be weird)

> check if any token fails to decode and then remove that one AND check if all cookies exist as they should (this PR).

To be safer than sorry then this is the way to go!

btw, this is an extreme scenario, ideally we shouldn’t *_ever_* have the need to show this but not having it it’s worse.

Refreshing tokens will minimize the risks of encountering this message anyway, I hope.

But other than these considerations, this was really needed for UX. And for us too in regards to troubleshooting.

I'm going to stress it a bit more and wait for any change you consider ^^

huumn:
> When you see this error message, there is only one account you can logout of which is the one you're currently logged in as

What if I'm anon as in my example and I can't switch to any account to logout?

ekzyis:
> > When you see this error message, there is only one account you can logout of which is the one you're currently logged in as
> 
> What if I'm anon as in my example and I can't switch to any account to logout?

You can't see the error message as anon because either

a) you're completely logged out because there's no `next-auth.session-token` and no multi_auth cookies anymore so there also won't be a `switch accounts` option in the dropdown
b) you're back to your original account because there still is `next-auth.session-token` but no more multi_auth cookies

I think it's okay that if there's an issue with account switching while you switched to anon that you will get switched back to your original account. Do you agree?

huumn:
I'm confused in my issue video. I'm anon and I'm shown accounts I can switch to but I can't switch to them.

Are you saying that's okay?

ekzyis:
Oh, I see, seems like I never tested removing the cookies while switched to anon, only while switched to another user. :sweat_smile: 

You're right, when we're anon, it will NOT reset the multi auth cookies (and thus switch you back to the account you originally logged in as) if one of them is missing because the jwt callback is never called.

Will see if I can move the logic to the `me` resolver, then it will always get called (logged in + logged out). Will also add the decode check to make sure we only keep valid tokens in the cookies as you suggested in https://github.com/stackernews/stacker.news/pull/1957#issuecomment-2714157209.

huumn (commented):
I'm not familiar with this code and functionality. I'd appreciate some help answering the following questions:

1. How do you recommend I QA this? 
2. What might've potentially broken?
3. Which parts of this are you not very confident in?

ekzyis:
> 1. How do you recommend I QA this?

I tested this by using all functionality of account switching (add account, switch between users and anon, logout to next account). After that was confirmed to not have regressed, I verified that if a check in `checkMultiAuthCookies` fails that all `multi_auth` cookies are removed[^1]. You can make a check fail by manually removing a cookie. Finally, all `multi_auth` tokens and `next-auth.session-token` should get refreshed on every graphql request (even if you switched to anon).

> What might've potentially broken?

To be honest, nothing really. :eyes: 

I heavily tested that account switching didn't regress and putting everything into lib/auth.js also made the multi auth flow on the server easier to understand. The server and client code is also 90% the same as it was before and the new code has a limited blast radius afaict.

I think this is a strict improvement with no regressions.

> Which parts of this are you not very confident in?

The logout to switch to the next account might also suffer from what I've seen with a logout with no accounts (see initial two paragraphs in the PR description) to switch to but if that happens, the check on the server + reset should handle exactly.

So I am only not confident in knowing _how_ something can break, but I am pretty confident that this PR will allow us to recover from any such scenario.

[^1]: I could have done what you suggested and try to remove as little as possible (i.e. only update the list if a token has gone missing), but since I don't think this is related to the issues we're seeing, I didn't, see also crossed out checkbox for "check invalid tokens" [here](https://github.com/stackernews/stacker.news/pull/1957#issuecomment-2726165243).

PR #1961, opened 2025-03-10
Add welcome banner back

ekzyis:
## Description

It was removed in https://github.com/stackernews/stacker.news/commit/f2ba61e64bcc0a7ff06a3aa8eec6df3fd8cf7ac9 as an accident I suppose.

## Checklist

yes

`7`. Hiding it still works.

works like before

no


huumn:
Not an accident. It caused layout shift iirc and was ugly. 

I'd prefer a welcome notification.

Issue #1343, opened 2024-08-30
URL search not showing up when sorted by recent


huumn:
https://stacker.news/items/667080

aegroto:
Tried it today, if i switch to "recent" then I get only one result with the exact post discussing the bug, linked in the original message. Are you sure this is not intentional? It looks fair to exclude older posts with a "recent" search criteria.

aegroto:
> sorting by "recent" makes the search more strict because the most recent thing is likely to be highly irrelevant.
> 
> We just need to made it less strict.

Ok, now I have a clear idea of the problem, thanks and sorry for the initial misunderstanding. I have been playing with the search code for a while today so I will give a deeper look into this issue if I've got some more spare time.

Issue #866, opened 2024-02-20
Search returns fuzzy matches above exact matches


felipebueno:
I specifically searched for "zig" but only the first three results on the list contain the word 'zig'. I clicked on other results to check for any references to 'zig,' but none of them had any.

![image](https://github.com/stackernews/stacker.news/assets/241271/652356f3-b6a1-4892-a071-ecea0c1c64ac)

huumn:
Looks like you're right. It's returning fuzzy matches higher than exact matches. Here's all the exact matches: https://stacker.news/search?q=%22zig%22

ed-kung:
I believe this could be due to Zaprank being the default sort field.  Not only does it sort by zaprank, it also boosts the relevance score of higher zaprank posts, so it's retrieving items that are semantically similar to "zig" but have a high zap rank, and prioritizing those items over items with the actual word "zig".

This behavior stops when including zig in quotation marks because a clause is added to the search query requiring that the exact phrase be found in the document.

Proposed solution: there should be a "relevance" sort field that doesn't boost by zaprank *or* recency.  My guess is that this might be the most common type of search people do, if they're looking for an old post.

huumn:
I think that's right. The only thing I'd add is that among items that have similar relevance, we do want zap rank to sort those. 

In SQL, we'd express this like `ORDER BY relevance DESC, zaprank DESC` where zaprank is a tiebreaker. I don't recall if we can do that with open search's query language.

Issue #1217, opened 2024-06-03
Search anomaly, is ":" in the title throwing something off?


cointastical:
criteria: tech crippled
https://stacker.news/search?q=tech+crippled&sort=recent

doesn't show any results

But there is the following item which should have been a match:
Bitcoin Tech Talk #399: Shrinking "Home," Calories, Crippled Miners, Clams++
https://stacker.news/items/560579

It seems like if one of the search criteria is on one side of the ":" in the title, and a second criteria is on the other side, that no match occurs.

criteria: crippled miners 
Results include the item above.

criterial: tech talk
Results include the item above.

ed-kung:
Is this still an issue? I can't reproduce it.

![Image](https://github.com/user-attachments/assets/0f1269d8-43b6-4182-b0e8-d4626c0ee8ca)

ekzyis:
If you sort by recent, you can see that it doesn't find anything: https://stacker.news/search?q=tech+crippled&sort=recent

I think the issue is that search by recent is broken, not necessarily about `:` in titles.

ed-kung:
This issue should be closed by https://github.com/stackernews/stacker.news/pull/1962

PR #1962, opened 2025-03-10
Search improvements: Add relevance search and make recent searches less strict

ed-kung:
## Description

Resolves https://github.com/stackernews/stacker.news/issues/1343 and https://github.com/stackernews/stacker.news/issues/1217 by making "recent" searches less strict.

Resolves https://github.com/stackernews/stacker.news/issues/866 by adding "relevance" search that prioritizes search term relevance only.

Changes the recency modifier from square of `createdAt` to Gaussian decay.

Refactors some variable names and restructures `api/resolvers/search.js` to (in my opinion) enhance logical flow and readability.

## Additional Context

You're gonna want to play around with different choices for the parameters, such as the relative important to give title vs. text, the decay rate and decay bandwidth, and potential boost modifiers.

## Checklist

yes

3. I tested that it works, and that the different types of search are generally doing what they're supposed to do. I don't have good confidence that the searches are returning quality matches, since I don't have a realistic test database to work with. 

n/a

no


api/resolvers/search.js
```diff
@@ -279,20 +292,51 @@ export default {
...
+          {
+            multi_match: {
+              query,
+              type: 'phrase',
+              fields: ['title^10', 'text'],
+              boost: 1000
+            }
           }
-        }
+        ]
```
huumn, line 339: This is the main thing that I added, which just enhances relevance for zaprank and other sorts by adding boosted exact matches.

ed-kung:
Thanks for merging! How did the testing go, since I wasn't sure at all how it'd perform on real data? I'd still be interested in having access to a data dump for further testing on search/WoT :)

huumn:
Testing revealed that we can't nest hybrid queries like we were. Newer versions of opensearch error, while the older version just silently ranked stuff poorly.

> I'd still be interested in having access to a data dump for further testing on search/WoT :)

I'll plan on producing it.

PR #1969, opened 2025-03-12
Fix saloon replies

ekzyis:
## Description

The query will return `{ replyCost: null }` if no sub was found because of the left join.

Instead of changing the query to use a inner join for the `Sub` table, I decided to check `sub?.replyCost` instead. Using a different join to keep `if (!sub)` felt wrong but not really sure why.

## Checklist

yes

`8`. Tested replies to items with no sub (like the saloon would be in prod) and with subs.

n/a

no

### DIFF
--- a/api/paidAction/itemCreate.js
+++ b/api/paidAction/itemCreate.js
@@ -27,8 +27,8 @@ export async function getBaseCost ({ models, bio, parentId, subName }) {
       LEFT JOIN "Sub" s ON s.name = COALESCE(r."subName", i."subName")
       WHERE i.id = ${Number(parentId)}`
 
-    if (!sub) return DEFAULT_ITEM_COST
-    return satsToMsats(sub.replyCost)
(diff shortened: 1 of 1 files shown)


api/paidAction/itemCreate.js
```diff
@@ -27,8 +27,8 @@ export async function getBaseCost ({ models, bio, parentId, subName }) {
       LEFT JOIN "Sub" s ON s.name = COALESCE(r."subName", i."subName")
       WHERE i.id = ${Number(parentId)}`
 
-    if (!sub) return DEFAULT_ITEM_COST
-    return satsToMsats(sub.replyCost)
+    if (sub?.replyCost) return satsToMsats(sub.replyCost)
```
ekzyis, line 30: this will be false for 0 but `replyCost` can't be lower than 1 according to our formik validation

huumn:
Nice catch!

ekzyis:
Yeah, fortunately the first item on the front page in our seed db does not have a territory

Issue #1254, opened 2024-07-04
Top Cowboys page has territory selector which doesn't actually do anything


SatsAllDay:
**Description**
If you go to https://stacker.news/top/cowboys, you'll see that it has the territory selector. By default, it's on `home`. If you change it to a new territory e.g. `~AGORA`, you'll be taken to https://stacker.news/~AGORA/top/cowboys, but the results are unchanged, meaning you still see the top cowboys across the entire site.

I don't think the territory selector should be included on the top cowboys page.

**Steps to Reproduce**
1. Go to https://stacker.news/top/cowboys
2. Click on a territory besides `home` in the territory selector e.g. `~AGORA`
3. Notice you're taken to https://stacker.news/~AGORA/top/cowboys, but the results are unchanged from before.

**Expected behavior**
I don't think the territory selector should be included on the top cowboys page.

The URL would still be navigable, but at that point it's undefined behavior.

**Screenshots**
<img width="1058" alt="image" src="https://github.com/stackernews/stacker.news/assets/128755788/b86114aa-4063-4b6c-97f9-bca2251be198">

**Logs**
N/A

**Environment:**
N/A

**Additional context**
N/A

ed-kung:
Selecting the territory while in /top/cowboys actually leads to weird behavior. It routes to /~territory/top/cowboys but still shows "home" in the territory selector, while the type selector is still limited to ["posts", "comments"] as if we were actually in a territory. See video.

https://github.com/user-attachments/assets/b6e7cdde-6f7f-461f-8554-5f39945073f6

Instead of getting rid of the territory selector, which would create inconsistent formatting for the top page and I think be a heavier lift, my suggestion would be to fix it so that if you select a territory while in /top/cowboys, it routes you to /~territory/top/posts/day, the default landing page for /top

PR #1972, opened 2025-03-13
Top cowboys territory selector fix

ed-kung:
## Description

Closes https://github.com/stackernews/stacker.news/issues/1254

Selecting a territory while on `/top/cowboys` now sends you to `/~:sub/top/posts/day` (or stay in `top/cowboys` if home is selected). 

Redirects `/~:sub/top/cowboys` to `/top/cowboys`.

## Checklist

yes

9

n/a

no

### DIFF
--- a/components/sub-select.js
+++ b/components/sub-select.js
@@ -94,8 +94,13 @@ export default function SubSelect ({ prependSubs, sub, onChange, size, appendSub
           }
         } else {
           // we're currently on the home sub
-          // are we in a sub aware route?
-          if (router.pathname.startsWith('/~')) {
+          // if we're in top/cowboys and a territory is selected,
+          // go to ~sub/top/posts/day
+          if (router.pathname.endsWith('top/cowboys')) {
+            router.push(sub ? `/~${sub}/top/posts/day` : '/top/cowboys')
+            return
+          } else if (router.pathname.startsWith('/~')) {
+            // are we in a sub aware route?
             // if we are, go to the same path but in the sub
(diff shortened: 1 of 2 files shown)


components/sub-select.js
```diff
@@ -94,8 +94,13 @@ export default function SubSelect ({ prependSubs, sub, onChange, size, appendSub
           }
         } else {
           // we're currently on the home sub
-          // are we in a sub aware route?
-          if (router.pathname.startsWith('/~')) {
+          // if we're in top/cowboys and a territory is selected,
+          // go to ~sub/top/posts/day
+          if (router.pathname.endsWith('top/cowboys')) {
```
ekzyis, line 99: I was only wondering if this would make this a little more reliable but I am about to leave for the day so I am just leaving this suggestion that we can probably ignore:

```suggestion
          if (router.pathname.endsWith('/top/cowboys')) {
```

ekzyis (approved):
Looks good to me! Just one nitpick comment.

ekzyis (commented):
This is good, but can you also do this for the other filters? :innocent: 

Territories only have 'posts' and 'comments' as a filter, but on 'home', there is also 'stackers' and 'territories' that have the same issue as 'cowboys'.

components/sub-select.js
ed-kung, line 1: * Changed routing logic from `endsWith` to `startsWith` since stackers and territories can have :when
* Verbose, but keeping the three cases separate would make it easier to make future changes like allowing for top stackers by territory.

PR #1975, opened 2025-03-15
Support .onion for phoenixd

### LINKED ISSUE #1964: Support .onion for phoenixd

Some stackers run phoenixd as a hidden service. We don't support that for phoenixd so they can't connect.

We should be able to use the same code we used for LNbits for phoenixd.

_No response_

### Additional context

https://stacker.news/items/909755

Scroogey-SN:
Fix #1964

Support .onion urls like lnbits does.

### DIFF
--- a/wallets/phoenixd/server.js
+++ b/wallets/phoenixd/server.js
@@ -28,7 +28,7 @@ export async function createInvoice (
   body.append('description', description)
   body.append('amountSat', msatsToSats(msats))
 
-  let hostname = url.replace(/^https?:\/\//, '').replace(/\/+$/, '')
+  const hostname = url.replace(/^https?:\/\//, '').replace(/\/+$/, '')
   const agent = getAgent({ hostname })
 
   const res = await fetchWithTimeout(`${agent.protocol}//${hostname}${path}`, {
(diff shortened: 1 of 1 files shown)


ekzyis:
Can you please not ignore the template we have setup for PRs?

I would like to know if you have tested this. It's okay if you didn't, since it's reasonably to assume that it will work if it works with LNbits, but I want to know if that's what you were thinking including everything else you thought about this PR and might be relevant.

Scroogey-SN:
Sure, will fill out the template for future PRs.

No, I didn't test it.

Setting up my own phoenixd on an .onion address is something I have never done before. I would have asked @itsMoro to test once it's life.

Same for PR #1973

ekzyis (changes requested):
This breaks attaching phoenixd over HTTPS:


https://github.com/user-attachments/assets/64c6247d-3d68-4089-8838-d4696dfcfdf7

You can test with these (mainnet) credentials:

- url: https://phoenixd.ekzy.is
- secondary password: `346e413c3d1eb73d271d02d824bbed459d9a561770cf411df86c3a8a62b45002`

ekzyis:
Oh sorry, this is only an issue during local dev since we force HTTP in that case, see line 140 here:

https://github.com/stackernews/stacker.news/blob/344c23ed5c4609f6061139c15c88bc487491da68/lib/proxy.js#L123-L146

ekzyis (approved):
Compared with how Tor support is implemented with LNbits. LGTM! This should work but didn't test since I don't have access to a phoenixd instance running as a hidden service.

Issue #841, opened 2024-02-16
Account creation via social media logins could be potentially improved


kristapsk:
Restarted Google Chrome (memory leaks, you know) and somehow forgot which social media account I used to log in to SN. Turns out it was Twitter. But I first tried GitHub. Without asking any questions it created new account kristapsk2 and now I can't even link GitHub to my original kristapsk account ("This auth method is linked to another account. To link to this account first unlink the other account"). Don't have concrete suggestions here, but UX in this regard could be improved.

ekzyis:
Yes, you are not the first one who is confused by new accounts getting automatically created. Login and signup are currently doing basically the same. I think the solution would be to only create accounts when signup is used. If the auth method was already used, it should throw an error.[^1] Vice versa, when login is used, it should throw an error when the auth method is not used already.

> it created new account kristapsk2 and now I can't even link GitHub to my original kristapsk account

I think you need to login with your Github account again to access the wrong account and then unlink the auth method as mentioned in the error message with "first unlink the other account". Then you should be able to link Github to your actual account.

Did you try that?

[^1]: in a way that preserves privacy as much as possible but a leak like "account already exists" is probably not preventable to make this change worthwhile

huumn:
@kristapsk I owe you some sats. Do you have a lightning address I can send them to?

PR #1976, opened 2025-03-16
Prevent new account creation on login

### LINKED ISSUE #1239: Prevent new account creation from /login
If someone tries to login with the wrong auth method, we create a new account for them. This is annoying and generates a bunch of customer service requests. New accounts should only get created from the `signup` page.

Soxasora:
## Description

Fixes #1239 and #841 
Applies a cookie to remember the user’s choice between logging in or signing up.
NextAuth picks up this cookie for the general purpose `createUser` function used by Email and Providers (social media) authentication methods, as well as for `pubkeyAuth`, which is used by Lightning and Nostr Credentials Providers.

## Screenshots
image

## Additional Context

I first tried by passing `signup` to the signIn method, this works but not for OAuth since there are a lot of steps before landing back to the callback and every additional option is lost.
Cookies are a reliable method of persistence in this case.

Email and OAuth throws `Callback` and CredentialsProvider throws `CredentialsSignIn`, as Next Auth doesn't support custom error messages, a not invasive way to handle this is to show a vague generic message that applies to every scenario. Implementing a custom error system for Next Auth is something that can be done but it would be too invasive imo.

## Checklist

Doesn't create a user anymore during login process, only during signup.

6, Nostr is not yet tested; Twitter uses a paid model.

n/a

n/a

### DIFF
--- a/components/login.js
+++ b/components/login.js
@@ -53,14 +53,14 @@ export function authErrorMessage (error) {
   return error && (authErrorMessages[error] ?? authErrorMessages.default)
 }
 
-export default function Login ({ providers, callbackUrl, multiAuth, error, text, Header, Footer, signup }) {
+export default function Login ({ providers, callbackUrl, multiAuth, error, text, Header, Footer, signin }) {
   const [errorMessage, setErrorMessage] = useState(authErrorMessage(error))
   const router = useRouter()
 
   // signup/signin awareness cookie
   useEffect(() => {
     const cookieOptions = [
-      `signup=${!!signup}`,
+      `signin=${!!signin}`,
       'path=/',
       'max-age=' + 60 * 60 * 24, // 24 hours
       'SameSite=Lax',
@@ -68,7 +68,7 @@ export default function Login ({ providers, callbackUrl, multiAuth, error, text,
     ].filter(Boolean).join(';')
 
     document.cookie = cookieOptions
-  }, [signup])
+  }, [signin])
 
   if (router.query.type === 'lightning') {
     return <LightningAuthWithExplainer callbackUrl={callbackUrl} text={text} multiAuth={multiAuth} />
--- a/components/login.js
+++ b/components/login.js
@@ -1,7 +1,7 @@
 import { signIn } from 'next-auth/react'
 import styles from './login.module.css'
 import { Form, Input, SubmitButton } from '@/components/form'
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
 import Alert from 'react-bootstrap/Alert'
 import { useRouter } from 'next/router'
 import { LightningAuthWithExplainer } from './lightning-auth'
@@ -53,9 +53,25 @@ export function authErrorMessage (error) {
   return error && (authErrorMessages[error] ?? authErrorMessages.default)
 }
 
-export default function Login ({ providers, callbackUrl, multiAuth, error, text, Header, Footer }) {
+export default function Login ({ providers, callbackUrl, multiAuth, error, text, Header, Footer, signup }) {
   const [errorMessage, setErrorMessage] = useState(authErrorMessage(error))
   const router = useRouter()
+
+  const setSignupCookie = (isSignup) => {
+    const cookieOptions = [
+      `signup=${isSignup}`,
+      'path=/',
+      'max-age=300',
+      'SameSite=Lax',
+      process.env.NODE_ENV === 'production' ? 'Secure' : ''
+    ].filter(Boolean).join(';')
+
+    document.cookie = cookieOptions
+  }
+
+  useEffect(() => {
+    setSignupCookie(!!signup)
+  }, [signup])
 
   if (router.query.type === 'lightning') {
     return <LightningAuthWithExplainer callbackUrl={callbackUrl} text={text} multiAuth={multiAuth} />
--- a/components/login.js
+++ b/components/login.js
@@ -42,10 +42,10 @@ const authErrorMessages = {
   OAuthCallback: 'Error handling OAuth response. Try again or choose a different method.',
   OAuthCreateAccount: 'Could not create OAuth account. Try again or choose a different method.',
   EmailCreateAccount: 'Could not create Email account. Try again or choose a different method.',
-  Callback: 'Try again or choose a different method.',
+  Callback: 'Could not authenticate. Try again or choose a different method.',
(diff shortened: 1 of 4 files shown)


pages/api/auth/[...nextauth].js
```diff
@@ -314,6 +318,7 @@ export const getAuthOptions = (req, res) => ({
   adapter: {
     ...PrismaAdapter(prisma),
     createUser: data => {
+      if (req.cookies.signin === 'true') return null
```
Comment (ekzyis) [inline pages/api/auth/[...nextauth].js:321]: Is it enough to just check if a cookie with that name exists to eliminate a class of potential future bugs?

components/login.js
```diff
@@ -42,21 +42,34 @@ const authErrorMessages = {
   OAuthCallback: 'Error handling OAuth response. Try again or choose a different method.',
   OAuthCreateAccount: 'Could not create OAuth account. Try again or choose a different method.',
   EmailCreateAccount: 'Could not create Email account. Try again or choose a different method.',
-  Callback: 'Try again or choose a different method.',
+  Callback: 'Could not authenticate. Try again or choose a different method.',
   OAuthAccountNotLinked: 'This auth method is linked to another account. To link to this account first unlink the other account.',
   EmailSignin: 'Failed to send email. Make sure you entered your email address correctly.',
-  CredentialsSignin: 'Auth failed. Try again or choose a different method.',
+  CredentialsSignin: 'Could not authenticate. Try again or choose a different method.',
```
ekzyis, line 48: Is it possible to return a special error message if the user does not exist?

ekzyis (commented):
This does fix #1976 with very few changes which is cool, nice!

Left some comments to explore how we can make this even better.

components/login.js
Soxasora: Thaaaat's something I researched quite a bit, it can be done but it requires extra-nextauth error handling. I think that if strictly needed we can do it ^^ but I also think that Auth.js v5 might have this as a native feature

pages/api/auth/[...nextauth].js
Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:321]: Yes it should be enough. That was the original way indeed:

https://github.com/stackernews/stacker.news/commit/8904e76eaf5b1b4e0b8c07296cd49e320366f807#diff-e0861d63eed5eacaa4b7c51e620e1da44a7579bccf244d96b26efb5d93c6b8a3L324

But I amended it because during tests I was still able to create a new account from signin. Which honestly I can't reproduce anymore so I'll put it back

pages/api/auth/[...nextauth].js
```diff
@@ -94,6 +94,8 @@ function getCallbacks (req, res) {
      */
     async jwt ({ token, user, account, profile, isNewUser }) {
       if (user) {
+        // reset signup cookie if any
+        res.setHeader('Set-Cookie', cookie.serialize('signup', '', { path: '/', expires: 0, maxAge: 0 }))
```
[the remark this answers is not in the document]
Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:98]: Ah, yes, I thought that by being in the callback space we would have headroom for `setHeader`. Yes appendHeader is a better choice!

pages/api/auth/[...nextauth].js
```diff
@@ -177,6 +179,7 @@ function setMultiAuthCookies (req, res, { id, jwt, name, photoId }) {
 
 async function pubkeyAuth (credentials, req, res, pubkeyColumnName) {
   const { k1, pubkey } = credentials
+  const signin = req.headers.cookie?.includes('signin=true')
```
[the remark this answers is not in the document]
Comment (Soxasora) [reply pages/api/auth/[...nextauth].js:182]: Mhh, I think we can do this
```javascript
if (new NodeNextRequest(req).cookies.signin) return null
```

components/login.js
ekzyis: Mhh, if waiting for the upgrade for a custom error message makes sense, then I think we should ship this without the upgrade to not be blocked by it.

You can resolve this conversation if that's also what you're thinking.

components/login.js
```diff
@@ -42,21 +42,34 @@ const authErrorMessages = {
...
+    const cookieOptions = [
+      `signin=${!!signin}`,
+      'path=/',
+      'max-age=' + 60 * 60 * 24, // 24 hours
+      'SameSite=Lax',
+      process.env.NODE_ENV === 'production' ? 'Secure' : ''
+    ].filter(Boolean).join(';')
+
+    document.cookie = cookieOptions
+  }, [signin])
```
huumn, line 71: Won't this prevent account creation for 24 hours if they visit the `login` page before the `sign up` page?

Soxasora: It gets overwritten by `signin=false` in that case because `signup.js` don't have the `signin` prop

ekzyis: Oh, but we stopped checking for the value on the server, only that the cookie exists, so I think @huumn is right.

I think it's better to remove this cookie in a `useEffect` cleanup function.

I also wanted to test this cookie with account switching more but during my testing yesterday, it seemed to worked as expected except that it just silently fails if the account we want to add doesn't exist but that's related to error handling and we had no error handling for account switching before, too.

Soxasora: Ah yes, you're right mb I thought of that case previously with the check for value, we can go with the useEffect. Pushing it in a bit ^^

components/login.js
```diff
@@ -42,21 +42,34 @@ const authErrorMessages = {
...
+export default function Login ({ providers, callbackUrl, multiAuth, error, text, Header, Footer, signin }) {
   const [errorMessage, setErrorMessage] = useState(authErrorMessage(error))
   const router = useRouter()
 
+  // signup/signin awareness cookie
+  useEffect(() => {
+    const cookieOptions = [
+      `signin=${!!signin}`,
+      'path=/',
+      'max-age=' + (signin ? 60 * 60 * 24 : 0), // 24 hours if signin is true, expire the cookie otherwise
```
Soxasora, line 65: _reduce, reuse, recycle?_

huumn:
I wonder if we should prevent sending the email if the account doesn't exist. That's one weirdness with this.

Soxasora:
Oh, you're right! I previously thought that not signaling that it's the wrong email would make it a bit more vague, kind of a natural deterrent against email fishing.
It's better to not send the email at all, but could still be good to redirect to the code page, maybe saying that they will receive a code if we found a match. What do you think?

huumn:
Yes, I think that's the approach - don't send the email if an account does not exist. AFAICT that's what every other site does.

PR #1341, opened 2024-08-28
Test payments with HODL invoices

ekzyis:
## Description

This PR replaces the `testSendPayment` function by wrapping `sendPayment` with code that actually attempts to pay an invoice.

That invoice is a HODL invoice and when it is held, we cancel it and return success. If the payment is not received within a timeout, or if the wallet throws an error first, the test fails and the wallet is not attached.

**However, I realized that payments might also fail because it has insufficient balance. The wallet should still get attached in that case, see TODO.**

TODO:

- [ ] how to attach wallet if payment fails due to insufficient balance? can we assume and parse the error message?
- [ ] make sure `logger` in `sendPayment` is optional in every wallet
- [ ] make sure we use any config returned by `sendPayment` (or find an alternative solution for LNC)

Based on #1340 

## Additional Context

Wallets that can't pay regtest invoices like phoenixd or Blink will not be attachable during development.

## Checklist

**Are your changes backwards compatible? Please answer below:**

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**

**For frontend changes: Tested on mobile? Please answer below:**

**Did you introduce any new environment variables? If so, call them out explicitly here:**

PR #1981, opened 2025-03-18
Add CLN send via CLNRest

### LINKED ISSUE #1967: Add CLN send via CLNRest

Looks like the CLNRest plugin we're already using to receive also supports spending:

> Mhh, we might be able to pay using the CLNRest plugin we're already using. When you click on 'Try it!' [here](https://docs.corelightning.org/reference/get_list_methods_resource), it lists all available methods and `pay`, `renepay` and `xpay` are among them. 👀

-- https://stacker.news/items/910893

Use CLNRest to spend from a CLN node.

_No response_

### Additional context

_No response_

Scroogey-SN:
## Description

Fix #1967

Adds client.js to wallets/cln

## Additional Context

I tested the /v1/pay call against the demo server cln-regtest-demo.blockstream.com with their demo rune.
First created an invoice, then paid that invoice.
The demo calls showed the syntax for parameters and result.

## Checklist

If the new field runePay isn't filled, there should be no regressions. Couldn't test configuring a server-only CLN wallet, though.

1

Ran sndev with COMPOSE_PROFILES=minimal,payments, configured cln wallet for sending, zapped a post to make a payment. It failed due to the demo credentials, of course.

So this will definitely require testing with a real CLNRest server.

Desktop only.

No.

### DIFF
--- a/wallets/cln/client.js
+++ b/wallets/cln/client.js
@@ -1,1 +1,39 @@
+import { fetchWithTimeout } from '@/lib/fetch'
+import { assertContentTypeJson, assertResponseOk } from '@/lib/url'
+
 export * from '@/wallets/cln'
+
+export async function testSendPayment ({ socket, runePay }, { logger, signal }) {
+  // TODO: check rune
+}
+
+export async function sendPayment (bolt11, { socket, runePay }, { logger, signal }) {
+  const url = `https://${socket}/v1/pay`
+
+  const headers = new Headers()
+  headers.set('Content-Type', 'application/json')
+  headers.set('Rune', runePay)
+  // see nodeId in lib/cln.js
+  headers.set('nodeId', '02cb2e2d5a6c5b17fa67b1a883e2973c82e328fb9bd08b2b156a9e23820c87a490')
+
+  const body = new URLSearchParams()
+  body.append('bolt11', bolt11)
+
+  const res = await fetchWithTimeout(url, {
+    method: 'POST',
+    headers,
+    body,
+    signal
+  })
+
+  assertResponseOk(res)
+  assertContentTypeJson(res)
+
+  const payment = await res.json()
+  const preimage = payment.data.payment_preimage
+  if (!preimage) {
+    throw new Error(payment.reason)
+  }
+
+  return preimage
+}
--- a/wallets/cln/index.js
+++ b/wallets/cln/index.js
@@ -14,7 +14,6 @@ export const fields = [
     placeholder: '55.5.555.55:3010',
     hint: 'tor or clearnet',
     clear: true,
-    serverOnly: true,
     validate: string().socket()
   },
   {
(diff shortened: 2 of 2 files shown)


wallets/cln/client.js
```diff
@@ -1 +1,39 @@
+import { fetchWithTimeout } from '@/lib/fetch'
+import { assertContentTypeJson, assertResponseOk } from '@/lib/url'
+
 export * from '@/wallets/cln'
+
+export async function testSendPayment ({ socket, runePay }, { logger, signal }) {
+  // TODO: check rune
```
ekzyis, line 7: Missing TODO.

We usually do "integration tests" here that involve the network unlike the formik/yup validation that triggers no network requests.

So for example, you could try to use the rune in some way that wouldn't work if the rune didn't exist but since we don't actually want to pay something (this was planned in #1341 but I got blocked by an issue), you can't actually use the `pay` method. Maybe it helps to look at `testSendPayment` from existing wallets.

wallets/cln/index.js
```diff
@@ -52,6 +54,27 @@ export const fields = [
...
+      .test({
+        name: 'runePay',
+        test: (v, context) => {
+          if (!v) return true
+          const decoded = decodeRune(v)
+          if (!decoded) return context.createError({ message: 'invalid rune' })
+          return true
+        }
+      })
+  },
```
ekzyis, line 77: Please add `help` like we have for the invoice only rune.

wallets/cln/index.js
```diff
@@ -52,6 +54,27 @@ export const fields = [
...
+    requiredWithout: 'rune',
+    validate: string().matches(B64_URL_REGEX, { message: 'invalid rune' })
+      .test({
+        name: 'runePay',
+        test: (v, context) => {
+          if (!v) return true
+          const decoded = decodeRune(v)
+          if (!decoded) return context.createError({ message: 'invalid rune' })
+          return true
+        }
```
ekzyis, line 75: this doesn't test that the rune can actually pay.

wallets/cln/client.js
```diff
@@ -1 +1,39 @@
...
+}
+
+export async function sendPayment (bolt11, { socket, runePay }, { logger, signal }) {
+  const url = `https://${socket}/v1/pay`
```
ekzyis, line 11: You can't hardcode the protocol since we allow HTTP over Tor.

This also breaks testing in local dev so I couldn't test this.

> So this will definitely require testing with a real CLNRest server.

We have a CLN node in our docker-compose setup that you should be able to test with.

ekzyis (changes requested):

huumn:
Moving to draft because author is unresponsive.

ekzyis:
closing because superseded by https://github.com/stackernews/stacker.news/pull/2475

PR #1998, opened 2025-03-20
Use __Secure- cookie prefix

ekzyis:
## Description

Based on #1997 #2005 #2006

I realized that in prod, the next-auth cookies use the `__Secure-` prefix, see [_Cookie prefixes_](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#cookie_prefixes) on MDN. We should use this prefix also for our `multi_auth` cookies.

TODO:

- [x] test this with HTTPS

## Additional context

~**Account switching will be unavailable for everyone when we deploy** until they logged back in.~ _This is fixed in #2007._

## Checklist

No, but #2007 makes it backwards compatible

`8`. Tested account switching functionality with https://sn.ekzy.is/.

n/a

no

### DIFF
--- a/api/ssrApollo.js
+++ b/api/ssrApollo.js
@@ -15,6 +15,7 @@ import { getServerSession } from 'next-auth/next'
 import { getAuthOptions } from '@/pages/api/auth/[...nextauth]'
 import { NOFOLLOW_LIMIT } from '@/lib/constants'
 import { satsToMsats } from '@/lib/format'
+import { MULTI_AUTH_ANON, MULTI_AUTH_LIST } from '@/lib/auth'
 
 export default async function getSSRApolloClient ({ req, res, me = null }) {
   const session = req && await getServerSession(req, res, getAuthOptions(req))
@@ -155,7 +156,7 @@ export function getGetServerSideProps (
 
     // required to redirect to /signup on page reload
     // if we switched to anon and authentication is required
-    if (req.cookies['multi_auth.user-id'] === 'anonymous') {
+    if (req.cookies[MULTI_AUTH_LIST] === MULTI_AUTH_ANON) {
       me = null
     }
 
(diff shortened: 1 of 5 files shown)


huumn:
Asking everyone to log in and out is pretty traumatic. Is there a way to do the reset without requiring logging in and out?

ekzyis:
> Maybe just reinitialize the multiauth cookies without requiring a fresh login? This would also fix the ugliness with the existing multiauth reset when an issue is detected.

Mhh, yes. I think my main reason to just reset it is to get additional data points to when it breaks; maybe that helps to find the underlying issue. But that might again be me who wants to "expose all the things" as a debugging dev :eyes:

I'll look into resetting to the initial state instead of to a degraded experience tomorrow!

ekzyis:
66bf70b3 is a WIP comment but it doesn't work yet because I get this error:

```
[next-auth][error][JWT_SESSION_ERROR]
https://next-auth.js.org/errors#jwt_session_error Invalid Compact JWE {
  message: 'Invalid Compact JWE',
  stack: 'JWEInvalid: Invalid Compact JWE\n' +
    '    at compactDecrypt (/app/node_modules/jose/dist/node/cjs/jwe/compact/decrypt.js:16:15)\n' +
    '    at jwtDecrypt (/app/node_modules/jose/dist/node/cjs/jwt/decrypt.js:8:61)\n' +
    '    at Object.decode (/app/node_modules/next-auth/jwt/index.js:56:34)\n' +
    '    at async Object.session (/app/node_modules/next-auth/core/routes/session.js:39:28)\n' +
    '    at async AuthHandler (/app/node_modules/next-auth/core/index.js:136:27)\n' +
    '    at async getServerSession (/app/node_modules/next-auth/next/index.js:129:19)\n' +
    '    at async context (webpack-internal:///(api)/./pages/api/graphql.js:99:23)\n' +
    '    at async ApolloServer.executeHTTPGraphQLRequest (file:///app/node_modules/@apollo/server/dist/esm/ApolloServer.js:511:32)',
  name: 'JWEInvalid'
}
```

PR #1999, opened 2025-03-20
don't send a verification email during sign in if no match

Soxasora:
## Description

Updates #1976 
If we're in `signin` mode (cookie) and can't find a user with the inserted email, we resolve earlier and don't send a verification email. This works by passing `req` to `sendVerificationRequest`, this way we're able to access cookies and check for `signin`

## Screenshots
n/a

## Additional Context
I added some _mystery_ to the email.js message, in line with major websites behavior

## Checklist

Yes, we're just passing req to check the `signin` cookie

7, I can receive an email if I'm registered and I don't receive an email if I'm not registered.

n/a

n/a

### DIFF
--- a/pages/email.js
+++ b/pages/email.js
@@ -12,8 +12,10 @@ export const getServerSideProps = getGetServerSideProps({ query: null })
 export default function Email () {
   const router = useRouter()
   const [callback, setCallback] = useState(null) // callback.email, callback.callbackUrl
+  const [signin, setSignin] = useState(false)
 
   useEffect(() => {
+    setSignin(document.cookie.includes('signin='))
     setCallback(JSON.parse(window.sessionStorage.getItem('callback')))
   }, [])
 
@@ -27,6 +29,13 @@ export default function Email () {
     router.push(url)
   }, [callback, router])
 
+  const buildMessage = useCallback(() => {
+    const email = callback?.email || 'your email address'
+    return signin
+      ? `if there's a match, a magic code will be sent to ${email}`
+      : `a magic code has been sent to ${email}`
+  }, [callback, signin])
+
   return (
     <StaticLayout>
       <div className='p-4 text-center'>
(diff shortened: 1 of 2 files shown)


huumn (changes requested):

pages/email.js
```diff
@@ -35,7 +35,7 @@ export default function Email () {
           <Image className='rounded-1 shadow-sm' width='640' height='302' src={`${process.env.NEXT_PUBLIC_ASSET_PREFIX}/cowboy-saloon.gif`} fluid />
         </video>
         <h2 className='pt-4'>Check your email</h2>
-        <h4 className='text-muted pt-2 pb-4'>a magic code has been sent to {callback ? callback.email : 'your email address'}</h4>
+        <h4 className='text-muted pt-2 pb-4'>if there's a match, a magic code will be sent to {callback ? callback.email : 'your email address'}</h4>
```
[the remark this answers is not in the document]
Soxasora: Oh you're right! Since we attempt login only after attempting to use the code, the signin cookie is still there and can be used. Pushing a conditional message 🫡 thanks!

pages/email.js
```diff
@@ -12,8 +12,10 @@ export const getServerSideProps = getGetServerSideProps({ query: null })
 export default function Email () {
   const router = useRouter()
   const [callback, setCallback] = useState(null) // callback.email, callback.callbackUrl
+  const [signin, setSignin] = useState(false)
 
   useEffect(() => {
+    setSignin(document.cookie.includes('signin='))
```
Soxasora, line 18: I didn't use a cookie parser here to avoid bloating, I tested this for quite a bit and it didn't fail on me!

ekzyis: but can you test the future? :eyes: 

to me, this is like a bomb that is ticking but has no screen with a countdown so you never know if it's just ticking or it's actually going to blow up at some point

a lot of stuff outside our control can set cookies, for example LNbits requests will return cookies iirc

Soxasora: I didn't get that power, _yet!_ You're totally right, there are advantages in using the lib that I didn't actually value like exact match, this case being raised. To be safer than sorry, I'll push it back with the cookie lib, thanks ^^

PR #2002, opened 2025-03-20
Wallet indicators

ekzyis:
## Description

~based on #2046~ _(merged)_

Show an indicator if no wallet has been setup yet like we do for bios.

TODO:

- [x] show orange button similar to the 'create bio' button

## Additional context

1. This will always show an indicator if no wallet is configured, not only the first time.
2. On back navigation, the button should maybe not show up again? It's at least annoying during development.

## Screenshots

localhost_3000_wallets(iPhone SE) (2)

localhost_3000_wallets(iPhone SE) (1)

## Checklist

yes

`7`. Tested that wallet indicator disappers if recv or send wallet is attached. Indicator appears again when all wallets are detached.

yes

no

### DIFF
--- a/components/nav/common.js
+++ b/components/nav/common.js
@@ -22,6 +22,7 @@ import classNames from 'classnames'
 import SnIcon from '@/svgs/sn.svg'
 import { useHasNewNotes } from '../use-has-new-notes'
 import { useWallets } from '@/wallets/index'
+import { useWalletIndicator } from '@/components/wallet-indicator'
 import SwitchAccountList, { nextAccount, useAccounts } from '@/components/account'
 import { useShowModal } from '@/components/modal'
 import { numWithUnits } from '@/lib/format'
@@ -190,6 +191,8 @@ export function MeDropdown ({ me, dropNavKey }) {
   if (!me) return null
 
   const profileIndicator = !me.bioId
+  const walletIndicator = useWalletIndicator()
+  const indicator = profileIndicator || walletIndicator
 
   return (
     <div className=''>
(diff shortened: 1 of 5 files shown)


components/nav/common.js
```diff
@@ -164,21 +164,49 @@ export function NavWalletSummary ({ className }) {
...
+  )
+}
+
 export function MeDropdown ({ me, dropNavKey }) {
   if (!me) return null
+
+  const wallets = useConfiguredWallets()
+
+  const profileIndicator = !me.bioId
+  const walletIndicator = wallets.length === 0
```
ekzyis, line 199: If one removes all wallets, the indicator will show up again. Not sure if that's good.

Soxasora: Could be an encouragement to attach a wallet ⚡️

ekzyis: > do you see it getting old over time?

Yes. Some people don't want to attach a wallet so they should have the option to remove this indicator / keep it removed in some way.

ekzyis: I guess we can ignore this for now and consider that case in the next iteration

components/nav/mobile/offcanvas.js
```diff
@@ -25,6 +26,11 @@ export default function OffCanvas ({ me, dropNavKey }) {
       )
     : <span className='text-muted pointer'><AnonIcon onClick={onClick} width='22' height='22' /></span>
 
+  const wallets = useConfiguredWallets()
+
+  const profileIndicator = me && !me.bioId
+  const walletIndicator = wallets.length === 0
```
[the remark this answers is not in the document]
ekzyis: see https://github.com/stackernews/stacker.news/pull/2002#discussion_r2020419076

huumn:
If webln is present it won't show for me. Is that intended?

ekzyis:
> If webln is present it won't show for me. Is that intended?

Mhh, it should show up when you enabled it in /wallets/webln but I didn't test with WebLN so maybe it doesn't actually

PR #2003, opened 2025-03-20
Refactor login cookie with cookieOptions function

ekzyis:
## Description

I think it's better to remove the cookie when we left the login page instead of when we visit /signup.

However, if we navigate away by entering a URL, the `useEffect` cleanup will not run.

We could fix this with a global `useEffect` that listens for path changes and sets/unsets the cookie if we're on /login or not.

TODO:

- [x] ~should we use `path: /login`?~ _no, see https://github.com/stackernews/stacker.news/pull/2003#pullrequestreview-2704354436_
- [x] ~use global `useEffect` to always set/unset signin cookie if we're on /login or not?~ _no, that would be too much_

## Video

https://github.com/user-attachments/assets/349d8f22-1569-4464-b7b1-b5c7bdc00fef

## Checklist

yes, especially because #1976 was not deployed yet

`5`. Tested that the cookie is removed when I leave /login.

n/a

no

### DIFF
--- a/components/login.js
+++ b/components/login.js
@@ -9,6 +9,9 @@ import { NostrAuthWithExplainer } from './nostr-auth'
 import LoginButton from './login-button'
 import { emailSchema } from '@/lib/validate'
 import { OverlayTrigger, Tooltip } from 'react-bootstrap'
+import { datePivot } from '@/lib/time'
+import * as cookie from 'cookie'
+import { cookieOptions } from '@/lib/auth'
 
 export function EmailLoginForm ({ text, callbackUrl, multiAuth }) {
   const disabled = multiAuth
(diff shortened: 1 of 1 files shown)


Soxasora (commented):
I don't personally think that there's a real problem with having the `signin` cookie just lying around since we always update its state when actually needed and it will get consumed eventually when the anon decides to login/signup.

Although... I admit it's not that clean and I would want to remove it too on leave. My only fear with going with the global route listener is that we might be adding complexity to a cookie that has a fairly innocent purpose.

tldr
It would be okay, just not sure if it deserves being checked at every route change

huumn:
I think because this does not work in corner cases, while the other solution does, we should probably stick to the other solution. I agree that this seems better - but because it's unreliable, I think it isn't.

ekzyis:
Oh, you are right, the other solution was more reliable. 0a9e499 now also expires the cookie on /signup, but I kept the `useEffect` cleanup function.

huumn:
The redirect to email doesn't cause this cookie to clear?

ekzyis:
As discussed in chat, I removed the cleanup function now.

All this PR does now is to use `cookieOptions` instead of manual joining of cookie options.

I considered to also set `expires` and `maxAge` to `undefined` so we'd have a session cookie but I don't want to over-engineer this / mess with this any further without a good reason.

PR #2007, opened 2025-03-22
Reset multi_auth to initial state on error

ekzyis:
## Description

Based on ~#1998~ ~#2018~ ~#2019~

Instead of just removing all `multi_auth` cookies on error, we now reset to the initial state.

## Checklist

yes

`7`. Tested by removing each `multi_auth` cookie and seeing it getting set again.

https://github.com/user-attachments/assets/c65f7a53-ef14-4940-8bf5-006a32299021

n/a

no

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -92,7 +92,7 @@ function switchSessionCookie (request) {
 }
 
 async function checkMultiAuthCookies (req, res) {
-  if (!req.cookies[MULTI_AUTH_LIST] || !req.cookies[MULTI_AUTH_POINTER]) {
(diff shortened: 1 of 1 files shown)


huumn:
Works well. Only, I see weird behavior when reseting cookies as anon. See video


https://github.com/user-attachments/assets/b2745650-0ea1-407f-b31d-b8fb45d3f6cf

ekzyis:
Ah, this happens because my code which account is currently selected sucks.

I think I'll create a `useCookie` function which detects cookie changes and returns updated cookies in a React-y way instead of what I am currently doing.

PR #2010, opened 2025-03-23
Simplify state of selected account

ekzyis:
## Description

~Based on #2009~ _merged_

Since we always call `router.reload()` after selecting an account, most of my code regarding which account is selected was not necessary.

The only difference is that we now no longer show that the selection changed even before the page reloaded but I think removing this code is worth it.

## Additional Context

Cleaning up the foundation to add a `useCookie` function to fix https://github.com/stackernews/stacker.news/pull/2007#issuecomment-2745895997.

## Checklist

yes

`8`. I tested switching to anon and to a different user.

n/a

no

### DIFF
--- a/components/account.js
+++ b/components/account.js
@@ -1,7 +1,6 @@
 import { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react'
 import { useRouter } from 'next/router'
 import * as cookie from 'cookie'
-import { useMe } from '@/components/me'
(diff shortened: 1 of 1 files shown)


ekzyis:
Will test this more tomorrow together with a new `useCookie` hook and when https://github.com/stackernews/stacker.news/pull/2007#issuecomment-2745895997 is fixed.

ekzyis:
This PR is useless, all changes are overriden by #2019. Closing.

PR #2015, opened 2025-03-24
Remove OpenSearch API Password

secp512k2:
## Description

Removed hardcoded OpenSearch admin password from docker-compose.yml and replaced it with an environment variable reference. This change improves security by ensuring sensitive credentials are not stored in version control and instead are properly managed through environment variables.

The change specifically:
- Replaces `OPENSEARCH_INITIAL_ADMIN_PASSWORD="password"` with `OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD}`
- Utilizes the existing `OPENSEARCH_PASSWORD` variable from .env.development
- Maintains functionality while improving security practices

## Screenshots

N/A - Configuration change only

## Additional Context

The change was straightforward as the environment variable was already defined in .env.development. This suggests there might be other instances of hardcoded credentials in the codebase that could benefit from similar treatment.

## Checklist

Yes, fully backwards compatible. The change only affects how the password is sourced, not the functionality. Existing deployments will continue to work as long as they have the OPENSEARCH_PASSWORD environment variable set.

8/10. Tested that:
- OpenSearch container starts successfully
- Health checks pass
- Authentication works with the environment variable
- Verified the password is correctly passed through to the container

N/A - This is a backend configuration change only

No new environment variables were introduced. The change utilizes the existing `OPENSEARCH_PASSWORD` variable that was already present in .env.development.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,7 +28,11 @@ services:
     restart: unless-stopped
     healthcheck:
       <<: *healthcheck
-      test: ["CMD-SHELL", "PGPASSWORD=${POSTGRES_PASSWORD} pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB} -h 127.0.0.1 && psql -U ${POSTGRES_USER} ${POSTGRES_DB} -c 'SELECT 1 FROM users LIMIT 1'"]
+      test:
+        [
+          "CMD-SHELL",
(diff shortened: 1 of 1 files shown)


ekzyis (changes requested):
This PR appears to be AI-generated or AI-assisted to me.

Regardless, please remove any unrelated changes. Then I will review the relevant change.

PR #2016, opened 2025-03-24
Removed OpenSearch API Password

secp512k2:
Removed OpenSearch API Password on Line 176

## Description

Removed hardcoded OpenSearch admin password from docker-compose.yml and replaced it with an environment variable reference. 

The change specifically:
- Replaces `OPENSEARCH_INITIAL_ADMIN_PASSWORD=password` with `OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD}`
- Utilizes the existing `OPENSEARCH_PASSWORD` variable from .env.development
- Maintains functionality while improving security practices

## Screenshots

N/A - Configuration change only

## Additional Context

The change was straightforward as the environment variable was already defined in .env.development. This suggests there might be other instances of hardcoded credentials in the codebase that could benefit from similar treatment.

## Checklist

Yes, fully backwards compatible. The change only affects how the password is sourced, not the functionality. Existing deployments will continue to work as long as they have the OPENSEARCH_PASSWORD environment variable set.

8/10. Tested that:
- OpenSearch container starts successfully
- Health checks pass
- Authentication works with the environment variable
- Verified the password is correctly passed through to the container

N/A - This is a backend configuration change only

No new environment variables were introduced. The change utilizes the existing `OPENSEARCH_PASSWORD` variable that was already present in .env.development.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -173,7 +173,7 @@ services:
     restart: unless-stopped
     env_file: *env_file
     environment:
-      - OPENSEARCH_INITIAL_ADMIN_PASSWORD=mVchg1T5oA9wudUh
+      - OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD}
       - plugins.security.disabled=true
       - discovery.type=single-node
     ports:


ekzyis:
Duplicate of https://github.com/stackernews/stacker.news/pull/2015

secp512k2:
This was the one I resubmitted with only the password fix, and not the formatting issues.

ekzyis (approved):
This does not improve security since .env.development is also checked into version control, but it's only for local dev and it's indeed better to read it from env instead of hardcoding it to avoid confusion so this gets my approval.

huumn (changes requested):
the password needs to be added to `.env.development` and checked in

ekzyis:
> the password needs to be added to `.env.development` and checked in

Is it not already?

```
$ git ls-files .env.development
.env.development
$ cat .env.development | grep OPENSEARCH
OPENSEARCH_URL=http://opensearch:9200
OPENSEARCH_USERNAME=admin
OPENSEARCH_PASSWORD=mVchg1T5oA9wudUh
OPENSEARCH_INDEX=item
OPENSEARCH_MODEL_ID=
OPENSEARCH_INITIAL_ADMIN_PASSWORD=mVchg1T5oA9wudUh
```

huumn:
> Is it not already?

Ah, I'm just wondering why it's there to begin with then. It might've been that the env vars weren't exported to the container before. Did you run this?

ekzyis:
> Did you run this?

I did not but I did now and `opensearch` starts successfully.

When I remove `OPENSEARCH_PASSWORD` from .env.development, it does not with this error:

> No custom admin password found. Please provide a password via the environment variable OPENSEARCH_INITIAL_ADMIN_PASSWORD.

Anyway, this PR should either use the environment variable `OPENSEARCH_INITIAL_ADMIN_PASSWORD` (that has the same value as `OPENSEARCH_PASSWORD`) or remove that variable from .env.development.

ekzyis (changes requested):
see previous comment

PR #2019, opened 2025-03-24
Refactor multi auth client-side code with new useCookie hook

ekzyis:
## Description

Based on ~#2018~ _(merged)_

I was able to remove almost all React state and only rely on a new `useCookie` hook to tell us what the current state of account switching is (which accounts are available + which account is selected). This looks much simpler now.

Unfortunately, I had to use polling to keep track of cookie changes because the [`CookieStore` API](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore/change_event) is not supported in all browsers (Safari :eyes:).

## Checklist

yes

`8`. Tested switching between anon and user, user and user, logout until fully logged out.

n/a

no

### DIFF
--- a/components/account.js
+++ b/components/account.js
@@ -1,114 +1,44 @@
-import { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react'
 import { useRouter } from 'next/router'
-import * as cookie from 'cookie'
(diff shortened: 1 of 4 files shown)


huumn:
Can you resolve the conflicts in these. I don't want to accidentally break them.

PR #2021, opened 2025-03-25
Reset if pointer is not a number or JWT cannot be decoded

ekzyis:
## Description

This can handle more potential issues.

## Checklist

yes

`7` . Cookies are reset if pointer is not a number or if we fail to decode the JWT.

n/a

no

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -91,14 +91,24 @@ function switchSessionCookie (request) {
   return request
 }
 
-export function checkMultiAuthCookies (req, res) {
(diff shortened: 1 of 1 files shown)


lib/auth.js
```diff
@@ -91,14 +91,24 @@ function switchSessionCookie (request) {
   return request
 }
 
-export function checkMultiAuthCookies (req, res) {
+async function checkMultiAuthCookies (req, res) {
```
ekzyis, line 94: `export` wasn't needed and was accidentally added back in #1998

PR #2027, opened 2025-03-26
Merge follow-up replies with normal replies

ekzyis:
## Description

This merges `THREAD` push notifications with `REPLY` push notifications but still respects the user setting for `THREAD` push notifications.

## Video

https://github.com/user-attachments/assets/0ab80a13-8cc5-4113-b64b-9b82d895cbea

## Checklist

yes

`8`. See video.

n/a

no

### DIFF
--- a/lib/webPush.js
+++ b/lib/webPush.js
@@ -99,6 +99,11 @@ async function sendUserNotification (userId, notification) {
       delete notification.item
     }
     const userFilter = createUserFilter(notification.tag)
+
+    // XXX we only want to use the tag to filter follow-up replies by user settings
+    // but still merge them with normal replies
+    if (notification.tag === 'THREAD') notification.tag = 'REPLY'
+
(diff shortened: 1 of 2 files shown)


lib/webPush.js
```diff
@@ -99,6 +99,11 @@ async function sendUserNotification (userId, notification) {
       delete notification.item
     }
     const userFilter = createUserFilter(notification.tag)
+
+    // XXX we only want to use the tag to filter follow-up replies by user settings
+    // but still merge them with normal replies
+    if (notification.tag === 'THREAD') notification.tag = 'REPLY'
+
```
Soxasora, line 106: Such a cool way to execute this. Just by looking at the code this gets my approval :)

Soxasora (approved):
Tested with iOS and it works correctly, also it can't possibly cause a delay (that would revoke a push sub) but I still tested for it and everything okay.

PR #2030, opened 2025-03-26
Add robots.txt +  sitemap.xml

SuperAtic:
## Description
Follow up from @Design_r's post about SN SEO https://stacker.news/items/925433

- add `robots.txt` to allow indexing
- add `sitemap.xml` to let crawlers know which contents to index

## Additional Context
Not sure how the `sitemap.xml` could be generated, I'll let this to you to figure out

PR #2031, opened 2025-03-26
Create robots.txt

SuperAtic:
As per #2030

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -1,0 +1,4 @@
+User-agent: *
+Allow: /
+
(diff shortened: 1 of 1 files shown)


huumn:
I don't think this is necessary. These things are primarily used for blocking search not "enabling" it.

https://www.reddit.com/r/SEO/comments/x4t78p/is_robotstxt_necessity_or_nicetohave/

ekzyis:
Closing since we don't plan to add this.

PR #2035, opened 2025-03-27
System logger for users

ekzyis:
## Description

We currently can only create wallet logs linked to a specific wallet. This means that if we have a user id but no wallet yet, we can't create a log message.

This PR fixes that by dropping the `NOT NULL` constraint on the `WalletLog.wallet` column and by defaulting to "system" for the wallet name in /wallet/logs if we didn't pass in a wallet to the logger call.

## Checklist

yes, old logs will continue to work.

`7`. Tested successful p2p zaps with anon as sender, receiver fallbacks, fallback from proxy to direct and ofc generating and paying invoices via lightning address.

n/a

no


pages/api/lnurlp/[username]/pay.js
```diff
@@ -4,17 +4,21 @@ import { lnurlPayDescriptionHashForUser, lnurlPayMetadataString, lnurlPayDescrip
...
+import { walletLogger } from '@/api/resolvers/wallet'
 
 export default async ({ query: { username, amount, nostr, comment, payerdata: payerData }, headers }, res) => {
   const user = await models.user.findUnique({ where: { name: username } })
   if (!user) {
     return res.status(400).json({ status: 'ERROR', reason: `user @${username} does not exist` })
   }
 
+  const logger = walletLogger({ models, me: user })
+  logger.info('LNURL-pay callback', { amount: formatMsats(amount), nostr, comment })
```
Comment (ekzyis) [inline pages/api/lnurlp/[username]/pay.js:20]: "LNURL-pay" might still be confusing but it's better than "lnurlp" imo.

Most stackers are probably the most familiar with the concept of a lightning address, but not sure how to mention that here.

PR #2040, opened 2025-03-28
Wallet badges

ekzyis:
## Description

This changes the reason when the gun or horse badge are shown to indicate if the stacker has a send/receive wallet attached and enabled.

Since it would require an extra mutation on save or enable/disable of send credentials only to detect if the sender has a sending wallet enabled, I decided to check if there's a vault entry for an enabled wallet and it's an incentive to enable device sync.

The badges can all be hidden using the same setting (`hide my cowboy essentials (e.g. cowboy hat)`).

TODO:

- [x] use columns (via triggers like we did for streaks) in resolvers
- [x] continue to send notifications on attach send or recv
  - [x] /notifications
  - [x] push notifications
  - [x] update indicator query | _ok lol, this time I remembered to update it, but there was no update needed ..._
- [x] fix backwards compatibility

## Checklist

yes, didn't drop any database columns ... ~but have to check if dropping graphql fields is bad~ 

~I think renaming graphql fields is not a problem during deployment since the client will deploy after~

~Turns out I didn't consider client/server desync, see https://github.com/stackernews/stacker.news/pull/2040#discussion_r2025584496~

Fixed that in https://github.com/stackernews/stacker.news/pull/2040/commits/40983ce47fa81b2031d6dfa43bb04b606ab4335f.

`8` (a2f61cb9)

Tests:

- [x] horse appears when first receive wallet is attached (and enabled)
- [x] gun appears when first send wallet is attached and device sync is enabled
- [x] horse disappears when last receive wallet is disabled (or detached)
- [x] gun disappears when last send wallet is disabled (or detached)
- [x] cowboy hat streaks
  - [x] `startedAt` is saved as midnight
  - [x] when I run `INSERT INTO pgboss.job(name) VALUES('streak')`, the streak job runs successfully ...
  - [x] ... and when I gained a cowboy hat by zapping 100 sats and then running the job again with `COWBOY_HAT_STREAK` set to 1000, I lose my streak

```
stackernews=# select * from "Streak" WHERE type = 'COWBOY_HAT' ORDER BY created_at DESC;
  id   |       created_at        |       updated_at        |      startedAt      |       endedAt       | userId |    type
-------+-------------------------+-------------------------+---------------------+---------------------+--------+------------
 13466 | 2025-04-07 21:10:45.713 | 2025-04-07 21:10:45.713 | 2025-04-06 00:00:00 |                     |    876 | COWBOY_HAT
 13467 | 2025-04-07 21:10:45.713 | 2025-04-07 21:10:45.713 | 2025-04-06 00:00:00 |                     |   8945 | COWBOY_HAT
 13465 | 2025-04-07 20:57:27.578 | 2025-04-07 21:10:45.713 | 2025-04-07 00:00:00 | 2025-04-06 00:00:00 |  21858 | COWBOY_HAT
(3 rows)
```

_(but the notification shows -1 days, shouldn't it show 0 days? See https://github.com/stackernews/stacker.news/pull/2040#discussion_r2032024140 :thinking:)_

- [x] (push) notifications for when a badge appears or disappears
- [x] privacy setting
- [x] test with multiple wallets

https://github.com/user-attachments/assets/00ea85c6-6fd3-498b-a96e-75c7536ba837

icons didn't change

no


prisma/schema.prisma
```diff
@@ -77,6 +77,8 @@ model User {
   streak                    Int?
   gunStreak                 Int?
   horseStreak               Int?
+  hasSendWallet             Boolean              @default(false)
+  hasRecvWallet             Boolean              @default(false)
```
huumn, line 81: what's the point of these columns if the resolvers don't use them?

api/typeDefs/user.js
```diff
@@ -209,8 +209,8 @@ export default gql`
     spent(when: String, from: String, to: String): Int
     referrals(when: String, from: String, to: String): Int
     streak: Int
-    gunStreak: Int
-    horseStreak: Int
```
huumn, line 213: I believe removing these fields from the typeDef and resolvers will cause a brief outage on deployment.

api/resolvers/user.js
```diff
@@ -1082,19 +1083,37 @@ export default {
...
+        where: {
+          userId: user.id,
+          enabled: true
+        }
+      })
+
+      return wallets.some(({ type, wallet: config }) => {
+        const def = getWalletByType(type)
+        return canReceive({ def, config })
+      })
```
huumn, line 1116: This adds two new db queries to each page load for logged in users? did you consider mitigating this?

The db columns suggested to me that you were going to update these when wallets were set and unset ... but the resolvers don't use them.

huumn (changes requested):
It works fine. I left comments about the issues that I have with it.

prisma/schema.prisma
ekzyis: Oh, that was my initial idea but something must have changed my mind and then I forgot to clean this up :grimacing: 

I think I realized that `hasSendWallet` won't work without device sync but it also doesn't work without with the resolver approach, so not sure what I was thinking.

I'll try to set these columns and then use them in the resolver.

api/resolvers/user.js
ekzyis: You are right, I should use the columns, see https://github.com/stackernews/stacker.news/pull/2040#discussion_r2025612148

api/typeDefs/user.js
ekzyis: I forgot that the client/server can also be out of sync, not just server and db :grimacing:

api/resolvers/notifications.js
```diff
@@ -320,9 +320,31 @@ export default {
           FROM "Streak"
           WHERE "userId" = $1
           AND updated_at < $2
+          AND type = 'COWBOY_HAT'
           ORDER BY "sortTime" DESC
           LIMIT ${LIMIT})`
         )
+        for (const type of ['HORSE', 'GUN']) {
+          queries.push(
+            `(SELECT (id::text || '-${type}-FOUND')::text AS id, "startedAt" AS "sortTime", 0 as "earnedSats", 'Streak' AS type
```
ekzyis, line 329: Since the same row is used to detect found and lost, I am encoding in the id if it was found or lost.

I then use that information when resolving `days` to return a number > 0 (lost) or not (found).

components/notifications.js
```diff
@@ -176,22 +176,30 @@ const defaultOnClick = n => {
 
 function Streak ({ n }) {
   function blurb (n) {
-    const type = n.type ?? 'COWBOY_HAT'
-    const index = Number(n.id) % Math.min(FOUND_BLURBS[type].length, LOST_BLURBS[type].length)
+    const id = Number(n.id.split('-')[0])
+    const type = n.id.includes('-HORSE')
+      ? 'HORSE'
+      : (n.id.includes('-GUN') ? 'GUN' : 'COWBOY_HAT')
```
ekzyis, line 182: Same code as in `days` resolver ...

prisma/migrations/20250328203730_wallet_badges/migration.sql
```diff
@@ -0,0 +1,95 @@
...
+    IF old_has_recv_wallet = false AND new_has_recv_wallet = true THEN
+        INSERT INTO "Streak" ("userId", "startedAt", "type", created_at, updated_at)
+        VALUES (user_id, now_utc(), 'HORSE', now_utc(), now_utc());
+    ELSIF old_has_recv_wallet = true AND new_has_recv_wallet = false THEN
+        UPDATE "Streak"
+        SET "endedAt" = now_utc(), updated_at = now_utc()
+        WHERE "userId" = user_id
+        AND "type" = 'HORSE'
+        AND "endedAt" IS NULL;
+    END IF;
```
ekzyis, line 44: @huumn is this what you meant with reusing the `Streak` table for notifications?

This will now show a notification every time you enable the first send/recv wallet or disable the last send/recv wallet. This is consistent with when the badges show up / disappear.

prisma/migrations/20250328203730_wallet_badges/migration.sql
```diff
@@ -0,0 +1,95 @@
+ALTER TABLE "users"
+    ADD COLUMN     "hasRecvWallet" BOOLEAN NOT NULL DEFAULT false,
+    ADD COLUMN     "hasSendWallet" BOOLEAN NOT NULL DEFAULT false;
+
+ALTER TABLE "Streak"
+    ALTER COLUMN "startedAt" SET DATA TYPE TIMESTAMP(3),
+    ALTER COLUMN "endedAt" SET DATA TYPE TIMESTAMP(3);
```
ekzyis, line 7: I had to change the types to support sending found/lost notifications for guns/horses at the same day.

I need to test that this didn't break cowboy hat streaks in some time-trippy way and if existing data is properly migrated to a timestamp at midnight in using the America/Chicago time zone.

prisma/schema.prisma
ekzyis: These columns are now set via triggers, see 019cdace.

ekzyis (commented):
This seems to work as intended but I need to add push notifications and Q&A more.

worker/streak.js
```diff
@@ -2,59 +2,56 @@ import { notifyNewStreak, notifyStreakLost } from '@/lib/webPush'
...
+      UPDATE users SET "streak" = 1 FROM new_streaks WHERE new_streaks.id = users.id
+    ), user_update_end_streaks AS (
+      UPDATE users SET "streak" = NULL FROM ending_streaks WHERE ending_streaks.id = users.id
+    ), user_update_extend_streaks AS (
+      UPDATE users
+      SET "streak" = (now() AT TIME ZONE 'America/Chicago')::date - extending_streaks.started_at::date
+      FROM extending_streaks WHERE extending_streaks.id = users.id
+    )
+    UPDATE "Streak"
+    SET "endedAt" = (now() AT TIME ZONE 'America/Chicago' - interval '1 day')::date, updated_at = now_utc()
```
ekzyis, line 49: @huumn why is this using `- interval '1 day'`?

Shouldn't we start to count from 1 and not from 0 for streaks? Like if you got a cowboy hat and then lose it, you at least kept it for one day since we run this at midnight (= next day)?

Asking because I ran this on the same day with `COWBOY_HAT_THRESHOLD` set to 1000 to make me lose my hat and then it shows up as -1 days. I expected it to say 0 days.

prisma/migrations/20250328203730_wallet_badges/migration.sql
ekzyis: Tested, cowboy hat streaks continue to work as before

prisma/schema.prisma
ekzyis: As discussed, I can simply create a job that then checks if receive or send is set and then update `hasSendWallet`/`hasRecvWallet` from there.

worker/streak.js
huumn: That statement in total means "yesterday's date" because this is run shortly after midnight.

prisma/schema.prisma
ekzyis: Done in a2f61cb9

worker/streak.js
```diff
@@ -85,39 +82,14 @@ export async function checkStreak ({ data: { id, type = 'COWBOY_HAT' }, models }
   if (!streak) return
 
   // new streak started for user
-  notifyNewStreak(id, streak)
+  notifyNewStreak(id, streak.type)
```
ekzyis, line 85: whoops, didn't mean to commit this

worker/wallet.js
```diff
@@ -284,3 +286,73 @@ export async function checkPendingWithdrawals (args) {
...
+  const endStreak = async (type) => {
+    const [streak] = await models.$queryRaw`
+      UPDATE "Streak"
+      SET "endedAt" = now(), updated_at = now()
+      WHERE "userId" = ${userId}
+      AND "type" = ${type}::"StreakType"
+      AND "endedAt" IS NULL
+      RETURNING "id"
+    `
+    return streak?.id
```
ekzyis, line 332: in case we ever start to delete streaks, this optional chaining plus `if (streakId)` will handle that

we don't do that currently in prod afaik, but during development, I tended to delete existing streaks for testing purposes so this was useful

api/resolvers/notifications.js
ekzyis: Because the same row is used to generate two notifications, so their id needs to be different:

> Since the same row is used to detect found and lost, I am encoding in the id if it was found or lost.

ekzyis: Refactored this in c15c317a. Thanks for the feedback, looks much better now!

api/typeDefs/notifications.js
```diff
@@ -156,11 +149,37 @@ export default gql`
...
+
+  type NewGun {
+    id: ID!
+    sortTime: Date!
+  }
+
+  type LostGun {
+    id: ID!
+    sortTime: Date!
+  }
```
ekzyis, line 176: Instead of `type Horse` with a `found: Boolean` field, I decided to create dedicated types so I don't have to write code just to distinguish the types when I already can distinguish them I query for the rows.

I tried it first with `found: Boolean` but it sucked iirc.

ekzyis (commented):
This is ready for review again. I'll work on the migration for existing users now.

I also think we can keep the blurbs for now, even though they aren't ideal since they sound like the horse + gun are still easter eggs.

api/resolvers/notifications.js
```diff
@@ -316,13 +316,36 @@ export default {
...
+          `(SELECT id::text, updated_at AS "sortTime", 0 as "earnedSats", 'CowboyHat' AS type
           FROM "Streak"
           WHERE "userId" = $1
           AND updated_at < $2
+          AND type = 'COWBOY_HAT'
           ORDER BY "sortTime" DESC
           LIMIT ${LIMIT})`
         )
+        for (const type of ['HORSE', 'GUN']) {
+          const gqlType = type.charAt(0) + type.slice(1).toLowerCase()
```
ekzyis, line 328: Will refactor this into a function in a follow-up PR since we use code like this at several places already and I don't want to refactor all these places in this PR.

huumn: Sounds good. Reviewing this now.

worker/wallet.js
```diff
@@ -284,3 +286,71 @@ export async function checkPendingWithdrawals (args) {
...
+  const newHasRecvWallet = wallets.some(({ type, wallet }) => canReceive({ def: getWalletByType(type), config: wallet }))
+  const newHasSendWallet = wallets.some(({ vaultEntries }) => vaultEntries.length > 0)
+
+  await models.user.update({
+    where: { id: userId },
+    data: {
+      hasRecvWallet: newHasRecvWallet,
+      hasSendWallet: newHasSendWallet
+    }
+  })
```
huumn, line 312: This should happen atomically with streak creation/update.

huumn (changes requested):
I'm still doing QA and review but I noticed that streak checking is not atomic which could lead to data consistency bugs.

worker/wallet.js
ekzyis: Oh, right, will wrap it in a transaction

huumn (changes requested):
I'm getting 404 when I try to visit profiles.

worker/wallet.js
ekzyis: Fixed in f5da4d52

ekzyis:
> I'm getting 404 when I try to visit profiles.

Fixed in 669c1522!

PR #2048, opened 2025-03-31
Move wallet components to wallets/

ekzyis:
## Description

based on #2002 

Wallet components were spread around. Some were in components/wallet-\*.js and some were in wallets/\*.js.

This PR moves all components/wallet-*.js files to wallets/.

## Checklist

yes

`7.5`. This is a little sketchy because of the amount of changes and that a bug would crash the site. But I tested sending 
and receiving with LNbits (p2p), device sync, wallet logs and clicked around. I think I hit all the wallet code and nothing broke.

I also searched for any occurrences of `components/wallet` that I forgot to replace but didn't find any.

n/a

no


ekzyis:
will resolve conflicts

ekzyis:
want to test this more since latest rebase

Issue #2047, opened 2025-03-31
Automate nlp setup for local dev search


huumn:
via https://github.com/stackernews/stacker.news/issues/2042#issuecomment-2764935974

ed-kung:
Yikes, after banging my head on this for a few days, I'm convinced this should be upgraded to difficulty:hard, at least for a fully automated "one-press" setup from fresh containers.  The opensearch container just seems really finicky and prone to failure, even after increasing resource requirements.

I'm going to take the approach of a setup script that you manually run *after* `sndev start`.  It would require you to restart your containers afterwards, but the approach seems more robust and less prone to failure.

PR #2049, opened 2025-04-01
automate opensearch nlp setup

ed-kung:
## Description

Closes https://github.com/stackernews/stacker.news/issues/2047

Closes https://github.com/stackernews/stacker.news/issues/2061

Closes https://github.com/stackernews/stacker.news/issues/2058

Automates setup of OpenSearch, including setting up semantic search if `OPENSEARCH_INDEX=item-nlp`. Any other value for `OPENSEARCH_INDEX` will not setup vector search and default to keyword search.

To do:

- [x] Create an OpenSearch setup script which is called from docker-compose. 
- [x] Setup script can replicate current functionality of setting up a default non-semantic index 
- [ ] If `OPENSEARCH_INDEX=item-nlp`, setup script will set up semantic index
- [ ] Setup script can determine if a re-indexing of all items needs to be done
- [ ] Configure settings to avoid circuit_breaking_exception errors

## Checklist

n/a

The script automatically adds:

```
OPENSEARCH_MODEL_ID=<model_id>
```

to `.env.local` if `OPENSEARCH_INDEX=item-nlp`


ed-kung:
Gonna close this PR to switch to a different approach, while leaving this branch open.  Where I left off with this one is that the NLP index and model would get successfully set up, the model_id would be dynamically generated and pushed to the app and worker containers, but the indexing process during `indexAllItems` was unstable

PR #2059, opened 2025-04-01
Wallet prompt

ekzyis:
## Description

based on ~#2040~ ~#2002~ ~#2089~

When a stacker does not have a receive wallet attached, we will prompt them to attach one when they post something.

The prompt includes three options:

1. immediately save a lightning address
2. go to /wallets
3. skip (and opt-in to not show again)

TODO:

- [x] save address on click
- [x] don't toast massive error if address validation failed (create test invoice)
- [ ] let user know that their post is saved as a draft so they won't lose it when they go to /wallets? but how?
- [x] skip
- [x] respect 'don't show again' if it was selected on skip
- [x] ~use `user.privates.hideWalletPrompt` to also hide wallet indicator (#2002)~ _actually, let's do this when I added a wallet send prompt on zap and this one is skipped, too_
- [x] toast success on lightning address save

## Additional context

There's currently no way to reset `hideWalletRecvPrompt` via the frontend for a stacker but I think that's okay.

## Screenshots

localhost_3000_(iPhone SE)

## Checklist

yes

`8`. Tested successful and failed lightning address save, tested skip with and without `don't show again` checked.

yes, see screenshot for mobile

no


ekzyis:
Note for myself: use the mutation for `don't show again` to also hide the wallet indicator for people who don't want to attach wallets

wallets/prompt.js
```diff
@@ -0,0 +1,151 @@
...
+        )
+      }, { keepOpen: true, onClose: () => reject(new WalletPromptClosed()) })
+    })
+  }, [!!me, me?.optional?.hasRecvWallet, me?.privates?.hideWalletRecvPrompt, showModal, onAttach, onSkip])
+}
+
+const Header = () => (
+  <div className='fw-bold text-center mb-3'>
+    You need to attach a<br />
+    <span className='fw-bold text-primary fs-1' style={{ fontFamily: 'lightning' }}>lightning wallet</span>
```
ekzyis, line 66: Btw, I would really appreciate it if we can keep this. :eyes:

wallets/prompt.js
```diff
@@ -0,0 +1,151 @@
...
+
+  const onSkip = useCallback(({ onClose, resolve }) =>
+    () => {
+      resolve()
+      onClose()
+    }, [])
+
+  return useCallback((e) => {
+    return new Promise((resolve, reject) => {
+      // TODO: check if user told us to not show again
```
huumn, line 43: old comment? looks done

huumn:
This is really well made. Very thoughtful all around. Nice work. 

After thinking about our discussion more this morning, I think we need to be pushy like this. Being gentle but ineffective is fool's errand.

wallets/prompt.js
ekzyis: oh, yes

Issue #2064, opened 2025-04-03
Incorrect Dupe Results for gitlab.io links


kepford:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When I post a link URL that is on the domain gitlab I get a list of dupes that are not duplicates. Wondering it if it is a regex bug. Not sure how the dupe check works.

### Screenshots

![Image](https://github.com/user-attachments/assets/9d56a55c-aae3-4cb4-a08d-9159554109bd)

### Steps To Reproduce

1. Grab a URL from gitlab.io or gitlab.com like https://about.gitlab.com/
2. Create a link post and paste the url in the url field
3. See that you have dupes that are false

### Expected behavior

When creating a link post only actual duplicate urls should show up as dupes.

ekzyis:
Thanks for the ticket!

We ignore subdomains in our dupes check, see here:

https://github.com/stackernews/stacker.news/blob/78f7e006d5b95e58787bf9871b528bf8b206460a/api/resolvers/item.js#L616-L620

Since you provided no path in your URL, it then matches all URLs from `gitlab.io`. Thatis not very intelligent for non-whitelisted domains, I agree.

SouthKoreaLN:
Will do this one to test if `sndev` works on this machine. Guess it's just a matter of commenting lines 617 to 620.

PR #2068, opened 2025-04-05
Fix: Treat subdomains as distinct in dupe URL detection

SouthKoreaLN:
Related issue: https://github.com/stackernews/stacker.news/issues/2064

## Description

This PR fixes the logic in the duplicate URL check to **preserve subdomains**. Previously, URLs like `about.gitlab.com` and `docs.gitlab.com` were treated as duplicates of any URL from `gitlab.com` due to subdomain stripping. This caused false positives when posting link URLs. Now, the logic compares full hostnames including subdomains, making the dupe detection more accurate.

## Screenshots

_No visual changes — backend logic only._

## Additional Context

The issue was discovered when trying to post a link from `about.gitlab.com`, which was flagged as a duplicate of an unrelated GitLab post. The root cause was this block:

\`\`\`js
// remove subdomain from hostname
if (parseResult?.subdomain?.length > 0) {
  hostname = hostname.replace(\`\${parseResult.subdomain}.\`, '')
}
\`\`\`

This PR removes that block to preserve the full hostname in the dupe comparison.

## Checklist

Yes — only affects dupe logic in link post creation.

**On a scale of 1–10 how well and how have you QA'd this change and any features it might affect? Please answer below:**  
Could not test it. The `sndev` environment only contains `example.com` as link examples, so one cannot check the subdomain issue.

N/A — backend-only change.

No.

Resolves https://github.com/stackernews/stacker.news/issues/2064


huumn:
The main issue here was that we didn't enforce the path to match, so we were getting prefix matches instead.

e.g. `about.gitlab.io` would match `about.gitlab.io/dafdsa/fdsadfas` and because we removed the subdomain `about.gitlab.io` would also match `something.gitlab.io/fdsafdas/fdsafdas`

For most domains, we do want to remove the subdomain so that `www.google.com` matches `google.com`. This breaks in certain cases, but the issue specifically points to an issue with domains that do not have paths more than needing subdomain matching.

I'll still award the full bounty.

SouthKoreaLN:
OK, my bad. I'll try to better recognize the issue at hand next time. Thanks for the bounty.

Issue #1990, opened 2025-03-19
Allow all custom zap amounts to be customized


huumn:
Currently we support a few custom zap amounts (on long press of the zap bolt), but keep the higher powers of 10 there. In effect this means you only get 3 custom zap amounts. I think we should allow replacing those powers of 10 too.

SouthKoreaLN:
I suggest capping the number of slots to 7 (or any other value you deem appropriate).
We have the 4 default values. Once someone starts adding custom tips, it adds them to the list, until we reach 7 entries.
Once 7 slots are occupied, it drops the oldest custom tip. If that works for you, I have just implemented it and I can create a PR with the changes.

huumn:
Yes, that's exactly what I wanted!

PR #2069, opened 2025-04-05
Gradually replace default tips with user-defined ones on long press (up to 7)

SouthKoreaLN:
## Description

Cfr https://github.com/stackernews/stacker.news/issues/1990

This PR improves the custom zap tip experience by allowing users to **gradually replace the default zap amounts** (`defaultTips`) with their own custom amounts (`customTips`). 

Previously, the app always displayed the default powers of 10 and allowed only 3 custom tips. Now, users can store up to 7 custom tips, and these will progressively take the place of the defaults in the UI.

Changes include:
- Allowing up to 7 custom zap tips instead of 3
- Removing the restriction that blocked default values from being added as custom
- Updating the display logic to prioritize custom tips and only fall back to non-duplicate default tips if needed

## Screenshots

_N/A – no visual changes to the UI layout, only logic affecting what zap amounts are shown on long-press._

## Additional Context

The main behavior change is that default tips will now disappear as the user adds their own preferred amounts, providing a more personalized and cleaner long-press menu. 

## Checklist

Yes, fully backwards compatible. Users with no custom tips will still see the default ones.

8 – tested long press interaction and tip rendering logic with various custom/default tip combinations.

No

No new environment variables.

Resolves https://github.com/stackernews/stacker.news/issues/1990


huumn:
This worked well. While my changes are semantically identical, I made the logic a little more concise while I tested it.

SouthKoreaLN:
Ok, thanks. Will aim for conciseness next time.

components/item-act.js
```diff
@@ -18,18 +18,10 @@ import { useSendWallets } from '@/wallets/index'
...
-    // enough custom tips to fully replace
-    tips = customTips.slice(0, 7)
-  } else {
-    // fill the rest with defaults (but avoid duplicates)
-    const remainingDefaults = defaultTips.filter(d => !customTips.includes(d))
-    tips = [...customTips, ...remainingDefaults].slice(0, 7)
-  }
-  tips.sort((a, b) => a - b)
+  const defaultNoCustom = defaultTips.filter(d => !customTips.includes(d))
+  const tips = [...customTips, ...defaultNoCustom].slice(0, 7).sort((a, b) => a - b)
```
ekzyis, line 23: Shouldn't this sort first and then slice?

PR #2071, opened 2025-04-06
show sats on full view pinned comments

### LINKED ISSUE #1475: Comment has been zapped but does not show number of sats have been zapped
This item (comment) shows a "green" Lightning bolt emoji, indicating it has been zapped, but there is nothing that shows the quantity of sats that it has been zapped:

https://stacker.news/items/707679

Requesting some indicator showing the quantity of sats, just like any other item would. 

This item was Pinned to the parent, so I suspect that could be a contributing factor.

Soxasora:
## Description

Fixes #1475 
Shows sats value for pinned comments in full view (reply on another page).
If your own Item is pinned, it shows Pin instead of Boost icon on comment view

## Screenshots
Comment view
image

Full view
image

## Additional Context

Simplified condition readability
This is a simple fix but for UX reasons we might want to evaluate the possibility of showing the Zap button for pinned comments along with a pinned icon, like this:
image

## Checklist

Yes

6, tested on pinned comments that belongs to an Item that belongs to a territory; also tested on pinned comments on pinned posts without a territory.

n/a

n/a


huumn:
I'm sure it's just some small issue but I don't see the pin if I pin my comment on my own item.

Screenshot 2025-04-14 at 7 46 22 PM

Then if I visit the item directly (ie "full") then the lightning bolt is hidden.

Screenshot 2025-04-14 at 7 46 34 PM

It might've been buggy like this before. idk. But we probably want to fix that.

Soxasora:
> I don't see the pin if I pin my comment on my own item.

Ah yes, that's the standard behavior, I'll push something for that ^^

> Then if I visit the item directly (ie "full") then the lightning bolt is hidden.

Weird, this doesn't happen to me
image

I think that the only case in which we don't show something there is when we want to hide the boost icon for biographies 
```javascript
: item.mine || item.meForward
    ? <Boost item={item} className={classNames(styles.upvote, item.bio && 'invisible')} />
```
And the boost icon can also only happen on own items (but we're on anon)
Hmm, I'll investigate

PR #2075, opened 2025-04-07
Recognize media type via Content-Type

### LINKED ISSUE #1433: AVIF images are rendered as video
https://stacker.news/items/230608

Soxasora:
## Description

Fixes #1433 
Instead of loading media beforehand to recognize if it's an image or a video, this method fetches Content-Type header to recognize it.
This fixes AVIF images that loads as video by following its MIME type rather than trying to load it as video/image.

## Screenshots
This is an AVIF image downloaded from stacker.news after Imgproxy processing, it gets rendered as an image, before it would've rendered as video on browsers like Chrome
image

## Additional Context

Fetching headers should be faster than loading media beforehand as it doesn't fetch the whole file.

## Checklist

Yes, it falls back to the original method if HEAD fetch fails

7, tested on Chrome (it happens way more here)

n/a

n/a

### DIFF
--- a/components/media-or-link.js
+++ b/components/media-or-link.js
@@ -123,29 +123,57 @@ export const useMediaHelper = ({ src, srcSet: srcSetIntital, topLevel, tab }) =>
     // don't load the video at all if user doesn't want these
     if (!showMedia || isVideo || isImage) return
 
-    // check if it's a video by trying to load it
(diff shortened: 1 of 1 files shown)


huumn:
Closing because we do server side byte sniffing now.

Issue #2051, opened 2025-04-01
option to randomize ordering of poll choices on page load


huumn:
https://stacker.news/items/931219/r/k00b?commentId=931273

huumn:
Yep, sounds spot on!

PR #2082, opened 2025-04-08
feat(Poll): add option to randomize poll choices

brymut:
## Description

Added a checkbox to the poll creation form to allow poll creators to choose whether or not the poll choices will be rendered in a random order or not. As discussed here: https://github.com/stackernews/stacker.news/issues/2051#issuecomment-2786565851

fixes https://github.com/stackernews/stacker.news/issues/2051

## Screenshots
Screen recording where I create poll with randomisation activated and I refresh a number of times and then I deactivate and try the same to demonstrate functionality.

https://github.com/user-attachments/assets/ac18230f-4fc7-48b2-8c0a-36a1c6fc0d2b

## Additional Context
- Changes to DB required adding a new boolean column to the ITEM table to enable this feature.

## Checklist

10

Checked mobile & light & dark mode.

No


brymut:
Hi @huumn, ready for review.

huumn:
At first glance this looks pretty solid. Nice work! I'll try to QA and do a more in depth review by early next week.

brymut:
Ok thanks, would it be ok if I got started on another while I wait @huumn

brymut:
Ok great, thanks

components/poll.js
```diff
@@ -14,6 +14,13 @@ export default function Poll ({ item }) {
   const { me } = useMe()
   const pollVote = usePollVote({ query: POLL_VOTE, itemId: item.id })
   const toaster = useToast()
+  const shuffleOptions = (options) => {
+    return options
+      .map(value => ({ value, sort: Math.random() }))
+      .sort((a, b) => a.sort - b.sort)
+      .map(({ value }) => value)
+  }
```
ekzyis, line 22:
```suggestion
  const shuffleOptions = (options) => {
    // shuffle copy because item.poll.options is read-only
    return [...options].sort(() => Math.random() - 0.5)
  }
```

ekzyis (changes requested):
Welcome to contributing to SN!

I found three issues with your PR, and I am requesting changes because of the first two, but I can see how it's easy to miss them, so no worries.

The first two issues are related to shuffling the options on the render path of the component.

1. The options will reshuffle on every render, see video:

https://github.com/user-attachments/assets/5f76ab47-7544-4de5-8ded-a998b97c295e

2. Hydration errors on full page load because the HTML rendered on the server will be different:

https://github.com/user-attachments/assets/3537dc48-61e1-4d48-ac4c-209c9e872024

I think both issues can be fixed with [`useEffect`](https://react.dev/reference/react/useEffect#displaying-different-content-on-the-server-and-the-client).[^1]

The other issue is a nitpick about code style: I think you're using `map` to create a new array and shuffle that one since the `item.poll.options` is read-only.

Since I think that's not clear from just reading the code—the 2x `map` looked unnecessary so I checked why you're doing that—I would recommend to simply create a copy of the `options` input and then return a shuffled version of that with one `sort` call, see suggestion. A comment in cases like this where the most obvious solution doesn't work is also nice.

[^1]: Maybe we even want to use [`useLayoutEffect`](https://react.dev/reference/react/useLayoutEffect) to not show the sorted options in the first render but that would be a nitpick and I wouldn't request changes for it—I would just change it myself if I think it will improve UX, so don't worry about this detail too much.

components/poll.js
brymut: ok, I'll give this a go

brymut: I think I've addressed everything you've brought up here. @ekzyis

ekzyis (changes requested):
The hydration error is gone, but on page load, it now reshuffles the options a few times:

https://github.com/user-attachments/assets/34fb6f09-c089-4d26-8d71-8d3995a94713

This might be a feature to signal that the poll options are randomized, but the code isn't written in a way to make this intentional so I consider this bug.

So I recommend to just fix it instead of making a feature out of it.

brymut:
Apologies for the back and forth, I've noticed that we don't need anything in the dependency list for the `useLayoutEffect` since the options wouldn't be changing while the poll is in use. Also added a `useRef` to keep track of the first shuffle and prevent any more. It should be better now. @ekzyis

components/poll.js
```diff
@@ -14,6 +15,17 @@ export default function Poll ({ item }) {
   const { me } = useMe()
   const pollVote = usePollVote({ query: POLL_VOTE, itemId: item.id })
   const toaster = useToast()
+  const [pollOptions, setPollOptions] = useState(item.poll.options)
+  const hasShuffled = useRef(false)
```
ekzyis, line 19: I don't think we need `useRef`.

I assume you added this because you weren't aware of [React Strict mode](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-effects-in-development): It runs all effects twice in development mode but this won't happen in production:

> When Strict Mode is on, React will also run one extra setup+cleanup cycle in development for every Effect. This may feel surprising, but it helps reveal subtle bugs that are hard to catch manually.

If that is why you added it, please remove `useRef`. If not, please explain why it's needed.

Next time, please mention the context of the decisions you made like this one. I don't want to have to guess why you did what you did when it's not obvious. Also make sure you understand why something is happening before you try to fix it.

components/poll.js
```diff
@@ -14,6 +15,17 @@ export default function Poll ({ item }) {
...
+  const shuffleOptions = (options) => {
+    return [...options].sort(() => Math.random() - 0.5)
+  }
```
ekzyis, line 22: While we're at it, please also move this function outside the component since it doesn't require any context of it.

(I would have done this myself at the end but since I am requesting changes anyway, I decided to mention this here)

ekzyis (changes requested):
Sorry to request changes again, but `useRef` is not meant to be used like this, see comment.

components/poll.js
brymut: Removed. It was a remnant of my debugging process before I thought of clearing the dependency list in the useLayouteffect hook. Also I thought it would be a cool way to control how many shuffles had been run. But I guess that has been rendered irrelevant.

ekzyis (approved):

brymut:
Any chance you've had a look? @huumn

huumn:
I'll look at it first thing tomorrow. Thank you for your patience.

brymut:
Thanks @huumn would be nice to get that first PR merge in.

api/resolvers/item.js
```diff
@@ -1150,15 +1151,9 @@ export default {
...
+      poll.randPollOptions = item?.randPollOptions
+      poll.options = poll.randPollOptions ? shuffleArray(options) : options
       poll.count = options.reduce((t, o) => t + o.count, 0)
-      const pollSettings = await models.item.findFirst({
-        where: {
-          id: item.id
-        },
-        select: { randPollOptions: true }
-      })
-      poll.randPollOptions = pollSettings?.randPollOptions || false
```
huumn: The argument `item` had all the `item` members including `randPollOptions`.

I also decided it would be better to randomize on the server so that we don't have to rely on the client doing it. It's not necessarily better, but in general I prefer the server do work so clients are "thinner."

components/adv-post-form.js
```diff
@@ -311,16 +311,6 @@ export default function AdvPostForm ({ children, item, sub, storageKeyPrefix })
...
-          {me && itemType === 'poll' &&
-            <Checkbox
-              label={
-                <div className='d-flex align-items-center'>randomize order of poll choices
-                </div>
-            }
-              name='randPollOptions'
-              checked={formik?.values.randPollOptions}
-              onChange={formik.handleChange}
-            />}
```
huumn: `AdvPostForm` is used by every `Item` form. We have an "escape hatch" for members to be added to the `AdvPostForm` accordian via the `children` prop.

components/poll-form.js
```diff
@@ -69,6 +69,11 @@ export function PollForm ({ item, sub, editThreshold, children }) {
           label='poll expiration'
           name='pollExpiresAt'
           className='pr-4'
+          groupClassName='mb-0'
+        />
+        <Checkbox
+          label={<div className='d-flex align-items-center'>randomize order of poll choices</div>}
+          name='randPollOptions'
```
huumn, line 76: checkbox as `children` like poll expiration.

components/adv-post-form.js
huumn: Also the formik stuff was unecessary afaict

huumn:
overall, nice work stitching this all together! welcome to the team

brymut:
Thanks @huumn, enjoyed contributing too. May I ask how the payment of sats is settled? I'm on stacker as brymut@stacker.news

huumn:
Payment will be sent to your lightning address.

huumn:
I plan on it.

Issue #1656, opened 2024-11-27
payment QR popup despite wallets being attached and enabled


huumn:
This happened once today. I opened the PWA and every zap I made, prompted with the QR immediately without the wallet payment failing (meaning the client thought there wasn't a wallet available). This continued even after I visited the attached wallets page and verified my nwc creds were synced.

I couldn't check the pwa logs, but I suspect the `useWallet` hook was returning a stale value for some reason.

huumn:
This happened to me again today and possibly here: https://stacker.news/items/851008/r/k00b?commentId=851861

ekzyis:
I mentioned in #2028 the following:

> I believe using `wallets.length` in effect dependencies might be the issue:
>
> https://github.com/stackernews/stacker.news/blob/ef8c738582d02ae128738046dd655cac4c93be91/components/item-act.js#L102-L147
>
> https://github.com/stackernews/stacker.news/blob/ef8c738582d02ae128738046dd655cac4c93be91/components/item-act.js#L301-L332
>
> This would lead to the bug if the wallets are initially "partially loaded" and on the next render, the length of the array didn't change so the callbacks aren't updated with the fully loaded wallets.

I thought since this happened to me while I had device sync enabled, it might be related to it since then we fetch wallets from the server. But I couldn't reproduce on local dev today. Even if I make the `wallets` resolver very slow (`await sleep(5000)`), I have seen no case where it might first partially load and only then load the full wallet to trigger the bug as I thought it might get triggered.

ekzyis:
Closing this until this happens again after #2088

ekzyis:
This did happen again:

https://github.com/user-attachments/assets/1c66ea06-9842-4dbb-bdb4-dd05a1dbedaa

m0wer:
This happened to me today. I had a bunch of pending stacker.news tabs on Brave (Android) and one of them didn't let me zap.

ekzyis:
Hopefully this will no longer happen after #2169

m0wer:
I'm not sure. I thing there's something wrong on `item-act.js` that hasn't been fixed there. I'm working on a MR.

ekzyis:
Showing a QR code even though you have sending wallets attached can certainly still happen if you try to zap before the wallets were fetched.

However, before #2169, this also happened even after the wallets were already fetched. You could go to /wallets, see your wallets being there, and still get a QR code. Only a page reload did seem to fix that.

[...]

PR #2088, opened 2025-04-09
Use wallets instead of wallets.length as useZap dependency

ekzyis:
## Description

Potential fix for #1656

## Additional Context

Could not reproduce bug but using `wallets` instead of `wallets.length` should be fine sine `useSendWallets` returns a memoized value.

## Checklist

yes

`5`. Tested zapping with NWC. Still works.

n/a

no

### DIFF
--- a/wallets/index.js
+++ b/wallets/index.js
@@ -196,17 +196,16 @@ export function WalletsProvider ({ children }) {
 
   // provides priority sorted wallets to children, a function to reload local wallets,
   // and a function to set priorities
+  const value = useMemo(() => ({
+    wallets,
+    reloadLocalWallets,
+    setPriorities,
+    onVaultKeySet: syncLocalWallets,
+    beforeDisconnectVault: unsyncLocalWallets,
+    removeLocalWallets
+  }), [wallets, reloadLocalWallets, setPriorities, syncLocalWallets, unsyncLocalWallets, removeLocalWallets])
   return (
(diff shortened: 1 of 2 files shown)


ekzyis:
`useSendWallets` uses `useMemo`, but `useWallets` returns the full context value which is recreated on each render afaict:

https://github.com/stackernews/stacker.news/blob/9df5a52bd3d19491657fedd395f9dae4fc0f39d4/wallets/index.js#L197-L215

wallets/index.js
```diff
@@ -196,17 +196,16 @@ export function WalletsProvider ({ children }) {
...
   // provides priority sorted wallets to children, a function to reload local wallets,
   // and a function to set priorities
+  const value = useMemo(() => ({
+    wallets,
+    reloadLocalWallets,
+    setPriorities,
+    onVaultKeySet: syncLocalWallets,
+    beforeDisconnectVault: unsyncLocalWallets,
+    removeLocalWallets
+  }), [wallets, reloadLocalWallets, setPriorities, syncLocalWallets, unsyncLocalWallets, removeLocalWallets])
```
huumn, line 206: Did this `useMemo` make a difference in the render count?

ekzyis: I think so:

https://github.com/user-attachments/assets/d03e7dd6-d289-46cb-82da-dff978532760

PR #2091, opened 2025-04-10
Fix wallet indicator blink via wallet loading state

ekzyis:
## Description

The wallet indicator currently shows up for everyone on first render because the wallets aren't loaded yet. Fixed that by exposing a loading state via `useWallets`.

However, this means that instead of going from `true` to `false` if there actually is a wallet attached, it now goes from `false` to `true` if there actually isn't a wallet attached, see video.

I think the way it is now (`false`to `true`) is better UX in most cases since most stackers will probably attach a wallet. 

Still something to think about.

<sub>I'm having a déjà vu :eyes:</sub>

## Video

https://github.com/user-attachments/assets/767a7263-1f25-4057-b584-f61911048334

## Additional Context

- https://stacker.news/items/939641?commentId=939788
- this new loading state for the wallets might have been very useful a long time ago already for other things :eyes: Will look into if we can use it to improve other code, UX etc.

## Checklist

yes

`8`. Tested with and without wallets.

n/a

no

### DIFF
--- a/wallets/index.js
+++ b/wallets/index.js
@@ -67,6 +67,7 @@ export function WalletsProvider ({ children }) {
   const [setWalletPriority] = useMutation(SET_WALLET_PRIORITY)
   const [serverWallets, setServerWallets] = useState([])
   const client = useApolloClient()
+  const [loading, setLoading] = useState(true)
 
   const { data, refetch } = useQuery(WALLETS,
     SSR ? {} : { nextFetchPolicy: 'cache-and-network' })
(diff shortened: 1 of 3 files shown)


wallets/index.js
```diff
@@ -223,8 +226,11 @@ export function useWallet (name) {
...
 
 export function useConfiguredWallets () {
-  const { wallets } = useWallets()
-  return useMemo(() => wallets.filter(w => isConfigured(w)), [wallets])
+  const { wallets, loading } = useWallets()
+  return useMemo(() => ({
+    wallets: wallets.filter(w => isConfigured(w)),
+    loading
+  }), [wallets, loading])
 }
```
ekzyis, line 234: Oh, putting this back into draft because I forgot to check if we use `useConfiguredWallets` in other places, too

ekzyis: Ok, we don't.

PR #2092, opened 2025-04-11
Migrate vault entries to new schema

ekzyis:
## Description

This PR replaces the existing `VaultEntry` table which links to a row in the generic `Wallet` table (and to a row in the `users` table) with a new table `Vault` where columns in the specific wallet tables like `WalletNWC` link to it.

TODO:

- [x] migrate existing vault entries to new schema
- `upsertWalletXXX` mutation
  - [x] write to new `Vault` table
  - [x] stop writing to old `VaultEntry` table
- `wallets` query
  - [x] read from new `Vault` table instead of from old `VaultEntry` table
- remove `models.vaultEntry` calls
  - [x] `injectResolvers` (#2114)
  - [x] `getVaultEntries` query
  - [x] `updateVaultKey` mutation
  - [x] `clearVault` mutation
  - [x] `checkWallet` job
- [x] remove `TODO_newName` 
- [x] delete `VaultEntry` table
- [x] Q&A
- [x] review code and refactor

## Checklist

Yes. All rows from `VaultEntry` are migrated to `Vault` with a link to them from their respective wallet tables and all LNC and WebLN rows from `Wallet` are now also inserted into their new wallet tables. 

`8` (7a4832a7). Tested saving recv-only/send-only/recv+send of following wallets with/without device sync:

- CLN
- NWC
- LNbits
- LND
- lightning address
- LNC
- WebLN
- phoenixd

(= every wallet except Blink)
since my internet is kind of bad here (attaching NWC times out Phoenixd 
Tried to test migration with production data but metabase CSV export is broken. It returns NULL for empty strings so it broke on `users."vaultKeyHash"` being NULL.[^1] Didn't bother further if the import already doesn't work because of something like this.

---

Okay, I haven't tested 727d9317 as much as I did 7a4832a7 but it should still be a solid 7 at least!

I tested WebLN and various combinations of LNbits with device sync and not etc. 

n/a

**Did you introduce any new environment variables? If so call them out explicitly here:**

no

[^1]: not sure why this isn't a nullable column but uses an empty string as the default but I'll look into this in a future PR ...


lib/object.js
ekzyis, line 1: I ended up only using `get` to access properties via a nested path like `a.b.c` (see tests).

I'll remove the other object helper functions if they don't turn out to be helpful in other places even though they have already been tested.

prisma/migrations/20250411012334_vault_refactor/migration.sql
ekzyis, line 1: I didn't test the migration with Blink and Phoenixd, only made sure that it looks good so far.

I want to test them with wallets from production since attaching Blink or Phoenixd with only mainnet support is hard in regtest mode.

prisma/schema.prisma
```diff
@@ -322,25 +341,33 @@ model WalletLNbits {
...
   id           Int      @id @default(autoincrement())
   walletId     Int      @unique
   wallet       Wallet   @relation(fields: [walletId], references: [id], onDelete: Cascade)
   createdAt    DateTime @default(now()) @map("created_at")
   updatedAt    DateTime @default(now()) @updatedAt @map("updated_at")
-  apiKeyRecv   String?
-  currencyRecv String?
+  nwcUrlRecv   String?
+  nwcUrlSendId Int?     @unique
+  nwcUrlSend   Vault?   @relation(fields: [nwcUrlSendId], references: [id])
```
ekzyis, line 356: Oh, instead of `TODO_newName` in the wallet def, I could just use `nwcUrl` instead of `nwcUrlSend`. Then the new schema would match the existing typedef. Same for `WalletBlink`.

wallets/blink/index.js
```diff
@@ -16,7 +16,8 @@ export const fields = [
       .matches(/^blink_[A-Za-z0-9]+$/, { message: 'must match pattern blink_A-Za-z0-9' }),
     help: `you can get an API key from [Blink Dashboard](${galoyBlinkDashboardUrl}).\nPlease make sure to select ONLY the 'Read' and 'Write' scopes when generating this API key.`,
     requiredWithout: 'apiKeyRecv',
-    optional: 'for sending'
+    optional: 'for sending',
+    TODO_newName: 'apiKeySend'
```
ekzyis, line 20: Dear future @ekzyis, see other comment regarding `TODO_newName`

prisma/migrations/20250411012334_vault_refactor/migration.sql
ekzyis: Actually, I could attach mainnet Phoenixd just fine. Same is probably true for Blink.

wallets/vault.js
```diff
@@ -0,0 +1,134 @@
+import { getWalletByType } from '@/wallets/common'
+import walletDefs from '@/wallets/client'
+import { get } from '@/lib/object'
+
+// returns fragments for the Prisma Client API
+// input is a wallet row from the db with the new schema
+export function vaultPrismaFragments (wallet) {
```
ekzyis, line 7: It might make sense to refactor this into individual, independent parts.

ekzyis: I would have loved to write a unit tests for this so that I don't accidentally change the output between now and after my refactor but I can't because of NDK's dependency on Cashu ...

2025-04-20-235434_1914x1163_scrot

lib/object.js
ekzyis: mhh, let's just keep them?

api/resolvers/vault.js
```diff
@@ -29,12 +41,21 @@ export default {
         }))
       }
 
-      for (const entry of entries) {
-        txs.push(models.vaultEntry.update({
-          where: { userId_key: { userId: me.id, key: entry.key } },
-          data: { value: entry.value, iv: entry.iv }
-        }))
+      const wallets = await models.wallet.findMany({ where: { userId: me.id } })
```
ekzyis, line 44: I think fetching the wallets outside of the transaction is fine since it's a read-only operation.

Else this needs to become an interactive transaction :eyes:

huumn: Regardless of whether this is interactive, this code is/was vulnerable to race conditions if a wallet is upserted while the vault key is updated.

ekzyis: Okay, I assume this is blocking this PR

ekzyis: Fixed in 72ac2225

ekzyis:
Oh, forgot to put this out of draft yesterday

api/resolvers/vault.js
huumn: This wasn't blocking because even after your changes the code is vulnerable to a race "if a wallet is upserted while the vault key is updated." Your change *only* protects against hash change races.

e.g. 

tx1: read wallets (wallets with ids [1,2])
tx2: insert wallet (wallet [3])
tx2: commit
tx1: update hash
tx1: commit

Wallet [3] is broken because it's encrypted with a different key.

huumn: The above race is very unlikely to occur, but that's why it wasn't blocking. The fix is nontrivial too: the only way to afaict is to give each vault entry a foreign key constraint on the correct key hash. That way, if it's updated any concurrent inserts would fail.

ekzyis: Both transactions read the old vault key hash first, so one of them should fail when they try to update it at the end because they won't find the row anymore, no?

wallets/vault.js
```diff
@@ -0,0 +1,115 @@
...
+  return models[def.walletField].update({
+    where: { walletId: wallet.id },
+    data: vaultPrismaFragments.deleteAll(wallet)
+  })
+}
+
+export function hasVault (wallet) {
+  const def = getWalletByType(wallet.type)
+  const vaultNames = vaultFieldNames(def)
+  return vaultNames.some(name => get(wallet, `wallet.${name}Id`))
```
huumn, line 114: Couldn't this just be written as 

```js
wallet.wallet[`${name}Id`]
```

The only other usage is on 93 which can be written as:

```js
wallet[def.walletField][name]
```

Where else do you imagine using this? And why?

ekzyis: Mhh, I don't really remember.

Might had something to do with me wanting to have a `move` function in `vaultNewSchematoTypedef` since it essentially requires "moving" object values around but I didn't end up using the `move` I wrote after I wrote some primitives (`get`, `set`, `remove`) for such a object helper function.

I'll remove it.

wallets/vault.js
```diff
@@ -0,0 +1,115 @@
...
+
+  if (wallet) {
+    const def = getWalletByType(wallet.type)
+    const names = vaultFieldNames(def)
+    if (names.length === 0) return {}
+    return { [def.walletField]: include[def.walletField] }
+  }
+
+  return include
+}
```
huumn, line 75: The reason we do the JSON trigger is to avoid doing a join with every wallet table like this.

If we're not going to use the JSON trigger, we should remove it. Alternatively, we should remove this N-table join and refactor the JSON triggers.

The con of the N-table join is that we pay the variable (and expected to grow) cost of reading all the wallets very frequently (even your update code does the joins). The pro is that we don't have to maintain triggers.

The con of the trigger is that maintaining the triggers can't be done in javascript and will require DB introspection to work with the changes you are trying to make in this PR (i.e. how do we know the wallet column references a vault entry? Probably by checking for a foreign key relationship.). The pro is that wallets are trivial to read from the DB both in terms of performance and the DB query to retrieve them.

tldr your changes can probably be made to make only reads expensive and writes efficient - relatively. The trigger can be made to make reads efficient and writes expensive - relatively.

huumn (changes requested):
My main gripe is the JSON trigger stuff. We'll need to figure out what to do there. Regardless of that, I'll need to test the migration before merging this.

You've described this PR as an intermediate step to something else so I understand this isn't "the end." I think it might be helpful for me (or any other reviewer) for you to describe "the end" that you're after so you can get feedback on it and the steps (like this) that you'll be taking to get there. Like, what's next after this PR? What will this PR allow you to objectively do that you weren't able to before? And, what will that allow you to do? And so on.

huumn:
How does this relate to #2146? Should this be in draft or will it need to be reviewed?

ekzyis:
> How does this relate to #2146? 

I want to merge #2146 into here. #2146 will contain the wallet schema changes, so this PR here will then have the schema changes for the vault and the wallet.

> Should this be in draft or will it need to be reviewed?

Yes, this should have been in draft since it's waiting for #2146 :eyes:

ekzyis:
I decided to merge this into a new branch `wallet-v2` because I don't see a way to get this PR and #2146 merged into `master` without breaking `master` or too much code for backwards-compatibility that is basically dead on arrival.

Therefore, `wallet-v2` will contain all changes to resolve #1495.

Ideally, the PR to merge `wallet-v2` into `master` will simplify a lot. I expect it to remove more code than its add, so the review of the remaining code should be easy enough.

PR #2093, opened 2025-04-11
fix(extend-awards): add check for existing branch with pending awards

### LINKED ISSUE #1991: Github action for extending awards.csv force pushes over other pending awards.csv changes
If I merge a PR, then another, the github action overwrites the prior changes.

brymut:
## Description
Currently as described in #1991, when a PR is merged while a PR previously opened by the `extend-awards` workflow is still opened & unmerged into master, the workflow runs again and overwrites the prior changes. 

This PR addresses this issue by adding steps in the workflow to first check if the branch that the workflow creates is available, signalling that there is still a PR open trying to extend the awards.csv file. If there is a branch, the workflow instead pushes a commit to that branch rather than creating a new PR and overwriting changes. 

I checked on alternative approaches. Based on the docs for the [create-pull-request](https://github.com/peter-evans/create-pull-request) there is no obvious application/example that addresses this issue. Also, creating differently named branches for each github workflow run would create conflicts as the master branch would change and the master that earlier PR's would be made with would be outdated with new extensions. So I think this is the simplest, most straight forward approach.

Closes #1991 

## Screenshots
My testing approach was to run this workflow on my fork of this repo and created a [test repo](https://github.com/brymut/test-repo/issues)with the labels used for `extend-awards` replicated there. Here is a screen recording demonstrating fixed functionality using the Github web UI:

https://github.com/user-attachments/assets/6c5e8f78-a214-4f7a-9f1c-f854bac75cc0

## Additional Context

**Branch deletion:** Since this approach relies on checking if the `extend-awards/patch` exists to detect if there is an open PR since this workflow creates that branch, we need to delete that branch when we have merged back into master so I have added the "delete-branch" option to the `create-pull-request` step to do so. As explained in the [docs for the Github action](https://github.com/peter-evans/create-pull-request/tree/main?tab=readme-ov-file#delete-branch), this doesn't delete the branch immediately, so I would recommend that the person merging deletes the branch when they are done. I have added a reminder to the PR message with the same. Alternatively, as [mentioned in the docs](https://github.com/peter-evans/create-pull-request/tree/main?tab=readme-ov-file#delete-branch), the repo could use GitHub's `Automatically delete head branches` feature in the repository settings.

## Checklist

Yes, AFAIK

10

### DIFF
--- a/.github/workflows/extend-awards.yml
+++ b/.github/workflows/extend-awards.yml
@@ -22,14 +22,36 @@ jobs:
         with:
           python-version: '3.13' 
       - run: pip install requests
+      - name: Check if branch exists
+        id: check_branch
+        run: |
+          git fetch origin extend-awards/patch || echo "Branch does not exist"
+          if git show-ref --verify --quiet refs/remotes/origin/extend-awards/patch; then
+            echo "exists=true" >> $GITHUB_ENV
+          else
+            echo "exists=false" >> $GITHUB_ENV
+          fi
+      - name: Checkout to existing branch
+        if: env.exists == 'true'
+        run: |
+          git checkout extend-awards/patch
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
       - run: python extend-awards.py
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


huumn:
I'll merge to test it and we'll revert if need be.

PR #2094, opened 2025-04-11
enhance: helpful error message on login

### LINKED ISSUE #2067: Helpful authentication error messages
If you use /login with new credentials, login will fail with this error message:

> Could not authenticate. Try again or choose a different method.

This is very confusing and doesn't help the user at all. They might think that something is wrong on their or our end, while actually nothing is wrong, they just need to use /signup instead.

Showing "user not found" _could_ be fine but it might depend on the authentication method used to not accidentally leak any private information.

We might want to wait with this until we [migrated to next-auth v5](https://authjs.dev/getting-started/migrating-to-v5).

Soxasora:
## Description
Fix #2067
Adds a message that asks the user if they wanted to sign up instead.

## Video
https://github.com/user-attachments/assets/ae6593e9-41b1-455f-a381-e1e1feb21b61

## Additional Context

This implementation is a workaround to next auth's missing custom error support.

## Checklist

Yes, it falls back to the original errors if signin prop is not present

6, the error shows only on sign in and only on errors that gets thrown by next auth in case of failure in returning the user object.

Yes

n/a

### DIFF
--- a/components/login.js
+++ b/components/login.js
@@ -12,6 +12,7 @@ import { OverlayTrigger, Tooltip } from 'react-bootstrap'
 import { datePivot } from '@/lib/time'
 import * as cookie from 'cookie'
 import { cookieOptions } from '@/lib/auth'
+import Link from 'next/link'
 
 export function EmailLoginForm ({ text, callbackUrl, multiAuth }) {
   const disabled = multiAuth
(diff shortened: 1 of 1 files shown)


components/login.js
```diff
@@ -52,12 +53,26 @@ const authErrorMessages = {
...
-  return error && (authErrorMessages[error] ?? authErrorMessages.default)
+export function authErrorMessage (error, signin) {
+  if (!error) return null
+
+  // workaround for signin/signup awareness due to missing support from next-auth
+  const message = error && (authErrorMessages[error] ?? authErrorMessages.default)
+  if (signin && (error === 'Callback' || error === 'CredentialsSignin')) {
+    return (
+      <>
+        We couldn't find an account with these credentials.
```
Soxasora, line 64: Could this message be too revealing? 🤔

ekzyis: I think `CredentialsSignin` error can also happen in other cases.

That's good, since now we can say that either there was no account _or_ authentication failed.

ekzyis: Example:

> Could not authenticate. Try again or choose a different method.
>
> If you are new to Stacker News, please sign up first.

Soxasora: Oh that's way more straightforward, I'll adjust the message. Thanks!!

ekzyis (approved):

PR #2096, opened 2025-04-11
Non-word boundary Regex on user Mentions

### LINKED ISSUE #2076: Mentions with leading ( are not detected

### Describe the bug

Mentions like `(@ek)` are no longer detected since https://github.com/stackernews/stacker.news/pull/2050 because now only mentions that start at a new line or with leading whitespace are detected. 

### Screenshots

_No response_

### Steps To Reproduce

1. Write `(@ek)` into a text box
2. Click preview
3. Mention is not detected

### Expected behavior

Mention is detected

### Logs

_No response_

### Device information

_No response_

### Additional context

See https://stacker.news/items/935195?commentId=935277

This needs more changes than just adding `(` to the non-matching group because this will then include the `(` in the full match which we use for the mention link _text_ but the link itself is correct:

```diff
diff --git a/lib/rehype-sn.js b/lib/rehype-sn.js
index 98a17dbd..35cb2258 100644
--- a/lib/rehype-sn.js
+++ b/lib/rehype-sn.js
@@ -6,7 +6,7 @@ import { toString } from 'mdast-util-to-string'
 const userGroup = '[\\w_]+'
 const subGroup = '[A-Za-z][\\w_]+'

-const mentionRegex = new RegExp('(?:^|\\s)@(' + userGroup + '(?:\\/' + userGroup + ')?)', 'gi')
+const mentionRegex = new RegExp('(?:^|\\s|\\()@(' + userGroup + '(?:\\/' + userGroup + ')?)', 'gi')
 const subRegex = new RegExp('~(' + subGroup + '(?:\\/' + subGroup + ')?)', 'gi')
 const nostrIdRegex = /\b((npub1|nevent1|nprofile1|note1|naddr1)[02-9ac-hj-np-z]+)\b/g
```

Image

Soxasora:
## Description

Fixes #2076
This regex matches user mentions (also with paths) starting with `@`, like @sox or @sox/all.
It ensures that the `@` symbol is not preceded by a word character, that it has the `@` symbol and that we have a username that contains word characters or underscores. Optionally, it could have a single forward slash to indicate paths.

As a bonus, it fixes popovers for mentions with slashes by serving only the username part to the `Mention` component.

## Video
https://github.com/user-attachments/assets/da283057-dada-4103-942d-a36aeeb828a2

## Additional Context

I exponentially tested the possibility of abuse and the ms it took for the regex to conclude didn't grow.
Initially this regex contained * for more than one slash, it happens that user paths never have more than a single slash so it was removed.

## Checklist

Only the client side regex pattern has changed, server-side we only need the username

6, tested for vulnerabilities and functionality. Tested also with 5 common cases and edge cases like `.@mention.`

n/a

n/a

### DIFF
--- a/lib/rehype-sn.js
+++ b/lib/rehype-sn.js
@@ -6,7 +6,7 @@ import { toString } from 'mdast-util-to-string'
 const userGroup = '[\\w_]+'
 const subGroup = '[A-Za-z][\\w_]+'
 
-const mentionRegex = new RegExp('(?:^|\\s)@(' + userGroup + '(?:\\/' + userGroup + ')?)', 'gi')
+const mentionRegex = new RegExp('\\B@(' + userGroup + '(?:\\/' + userGroup + ')*)', 'gi')
 const subRegex = new RegExp('~(' + subGroup + '(?:\\/' + subGroup + ')?)', 'gi')
 const nostrIdRegex = /\b((npub1|nevent1|nprofile1|note1|naddr1)[02-9ac-hj-np-z]+)\b/g
 
@@ -116,8 +116,8 @@ export default function rehypeSN (options = {}) {
             }
 
             const [fullMatch, mentionMatch, subMatch] = match
-            const replacement = mentionMatch ? replaceMention(fullMatch, mentionMatch) : replaceSub(fullMatch, subMatch)
 
+            const replacement = mentionMatch ? replaceMention(fullMatch, mentionMatch) : replaceSub(fullMatch, subMatch)
             if (replacement) {
               newChildren.push(replacement)
             } else {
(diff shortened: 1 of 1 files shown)


lib/rehype-sn.js
```diff
@@ -6,7 +6,7 @@ import { toString } from 'mdast-util-to-string'
 const userGroup = '[\\w_]+'
 const subGroup = '[A-Za-z][\\w_]+'
 
-const mentionRegex = new RegExp('(?:^|\\s)@(' + userGroup + '(?:\\/' + userGroup + ')?)', 'gi')
+const mentionRegex = new RegExp('\\B@(' + userGroup + '(?:\\/' + userGroup + ')?)', 'gi')
```
Soxasora, line 9: With this we also allow mentions presented like this `.@mention.` or `/@mention`. If we don't want that we could just allow `()` parenthesis as a special case.

ekzyis (approved):

lib/rehype-sn.js
ekzyis: I think that's okay

PR #2097, opened 2025-04-11
Fix item deletion hits constraint

### LINKED ISSUE #2095: Can't delete item 934739

### Describe the bug

For some reason, the constraint I added in https://github.com/stackernews/stacker.news/pull/1747 prevents deletion of [this item](https://stacker.news/items/934739).

### Screenshots

Image

### Steps To Reproduce

1. Try to delete [item 934739](https://stacker.news/items/934739)
2. Error and item is not deleted

### Expected behavior

No error and item is deleted

### Logs

_No response_

### Device information

_No response_

### Additional context

https://stacker.news/items/940469?commentId=940954

ekzyis:
## Description

Fix #2095 

This excludes the text `*deleted by author*` from hitting the constraint that I added in https://github.com/stackernews/stacker.news/pull/1747.

## Checklist

yes, it should be because this is excluding more from the constraint so anything that exists should already be valid according to it

`7`. Made sure this fixes the issue.

n/a

no

### DIFF
--- a/prisma/migrations/20250411220921_fix_item_delete/migration.sql
+++ b/prisma/migrations/20250411220921_fix_item_delete/migration.sql
@@ -9,4 +9,4 @@ ALTER TABLE "Item" ADD CONSTRAINT "Item_unique_time_constraint"
     md5(COALESCE("text", '')) WITH =,
     tsrange(created_at, created_at + INTERVAL '10 minutes') WITH &&
   )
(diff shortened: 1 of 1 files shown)


huumn:
Shouldn't we use the `deletedAt` column instead of the contents of the text column?

ekzyis:
> Shouldn't we use the `deletedAt` column instead of the contents of the text column?

You are right, we should

PR #2114, opened 2025-04-18
Remove unused vaultEntries during upsertWallet

ekzyis:
## Description

I _think_ this isn't needed anymore but I still need to verify.

## Checklist

yes

tbd

n/a

no


ekzyis:
Ok this is still needed. It's used here:

https://github.com/stackernews/stacker.news/blob/e94a2311924cc349c32fe6ee0989c28df1c5b20f/wallets/validate.js#L81-L99

_sooo looking forward to refactor all of this code_

PR #2115, opened 2025-04-18
Live updates to comment threads

### LINKED ISSUE #791: Live updates to comment threads

Soxasora:
## Description

Closes #791 
Polls every 10 seconds a new query called `GetNewComments`, returning an array of new comments based on the parent Item and appropriately pushes them to the related `Item` or `Comment` Apollo Cache fragment.

## Screenshots
tbd

## Additional Context
tbd

## Checklist

Yes

5, iterative testing, tbd

Yes

n/a

## TODOs
_Subject to updates_

- [ ] Count also new comments of a new comment
- [ ] Revisit Apollo Cache logic
- [ ] Test
- [ ] Cleanup

### DIFF
--- a/components/use-live-comments.js
+++ b/components/use-live-comments.js
@@ -1,0 +1,91 @@
+import { useQuery, useApolloClient } from '@apollo/client'
+import { SSR } from '../lib/constants'
+import { GET_NEW_COMMENTS, COMMENT_WITH_NEW } from '../fragments/comments'
+import { ITEM_FULL } from '../fragments/items'
+import { useState } from 'react'
+
+export function useLiveComments (rootId, after) {
+  const client = useApolloClient()
+  const [lastChecked, setLastChecked] = useState(after)
+  const { data, error } = useQuery(GET_NEW_COMMENTS, SSR
+    ? {}
+    : {
+        pollInterval: 10000,
+        variables: { rootId, after: lastChecked }
+      })
+
+  console.log('error', error)
+
+  if (data && data.newComments) {
+    saveNewComments(client, rootId, data.newComments.comments)
+    const latestCommentCreatedAt = getLastCommentCreatedAt(data.newComments.comments)
+    if (latestCommentCreatedAt) {
+      setLastChecked(latestCommentCreatedAt)
+    }
+  }
+
+  return null
+}
+
+export function saveNewComments (client, rootId, newComments) {
+  console.log('newComments', newComments)
+  for (const comment of newComments) {
+    console.log('comment', comment)
+    const parentId = comment.parentId
+    if (Number(parentId) === Number(rootId)) {
+      console.log('parentId', parentId)
+      client.cache.updateQuery({
+        query: ITEM_FULL,
+        variables: { id: rootId }
+      }, (data) => {
+        console.log('data', data)
+        if (!data) return data
+        console.log('dataTopLevel', data)
+
+        const { item } = data
+
+        return { item: dedupeComment(item, comment) }
+      })
+    } else {
+      console.log('not top level', parentId)
+      client.cache.updateFragment({
+        id: `Item:${parentId}`,
+        fragment: COMMENT_WITH_NEW,
+        fragmentName: 'CommentWithNew'
+      }, (data) => {
+        if (!data) return data
+
+        console.log('data', data)
+
+        return dedupeComment(data, comment)
+      })
+      console.log('fragment', client.cache.readFragment({
+        id: `Item:${parentId}`,
+        fragment: COMMENT_WITH_NEW,
+        fragmentName: 'CommentWithNew'
+      }))
+    }
+  }
+}
+
+function dedupeComment (item, newComment) {
+  const existingNewComments = item.newComments || []
+  const alreadyInNewComments = existingNewComments.some(c => c.id === newComment.id)
+  const updatedNewComments = alreadyInNewComments ? existingNewComments : [...existingNewComments, newComment]
+  console.log(item)
+  const filteredComments = updatedNewComments.filter((comment) => !item.comments?.comments?.some(c => c.id === comment.id))
+  const final = { ...item, newComments: filteredComments }
+  console.log('final', final)
+  return final
+}
+
+function getLastCommentCreatedAt (comments) {
+  if (comments.length === 0) return null
+  let latest = comments[0].createdAt
+  for (const comment of comments) {
+    if (comment.createdAt > latest) {
+      latest = comment.createdAt
+    }
+  }
+  return latest
+}
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -727,6 +727,32 @@ export default {
         homeMaxBoost: homeAgg._max.boost || 0,
         subMaxBoost: subAgg?._max.boost || 0
       }
+    },
+    newComments: async (parent, { rootId, after }, { models, me }) => {
+      console.log('rootId', rootId)
+      console.log('after', after)
+      const item = await models.item.findUnique({ where: { id: Number(rootId) } })
+      if (!item) {
+        throw new GqlInputError('item not found')
+      }
+
+      const comments = await itemQueryWithMeta({
+        me,
+        models,
+        query: `
+          ${SELECT}
+          FROM "Item"
+          -- comments can be nested, so we need to get all comments that are descendants of the root
+          WHERE "Item".path <@ (SELECT path FROM "Item" WHERE id = $1)
+          AND "Item"."created_at" > $2
+          ORDER BY "Item"."created_at" ASC`
+      }, Number(rootId), after)
+
+      console.log('comments', comments)
+
+      return {
+        comments
+      }
     }
   },
 
--- a/api/typeDefs/item.js
+++ b/api/typeDefs/item.js
@@ -11,6 +11,7 @@ export default gql`
     auctionPosition(sub: String, id: ID, boost: Int): Int!
     boostPosition(sub: String, id: ID, boost: Int): BoostPositions!
     itemRepetition(parentId: ID): Int!
+    newComments(rootId: ID, after: Date): NewComments
   }
 
   type BoostPositions {
@@ -96,6 +97,10 @@ export default gql`
     comments: [Item!]!
   }
 
+  type NewComments {
+    comments: [Item]
+  }
+
   enum InvoiceActionState {
     PENDING
     PENDING_HELD
@@ -148,6 +153,7 @@ export default gql`
     ncomments: Int!
     nDirectComments: Int!
     comments(sort: String, cursor: String): Comments!
+    newComments(rootId: ID, after: Date): NewComments
     path: String
     position: Int
     prior: Int
--- a/components/comment.js
+++ b/components/comment.js
@@ -28,6 +28,8 @@ import LinkToContext from './link-to-context'
 import Boost from './boost-button'
 import { gql, useApolloClient } from '@apollo/client'
 import classNames from 'classnames'
+import { ITEM_FULL } from '@/fragments/items'
+import { COMMENT_WITH_NEW } from '@/fragments/comments'
 
 function Parent ({ item, rootText }) {
   const root = useRoot()
@@ -111,8 +113,11 @@ export default function Comment ({
   const router = useRouter()
   const root = useRoot()
   const { ref: textRef, quote, quoteReply, cancelQuote } = useQuoteReply({ text: item.text })
-
   const { cache } = useApolloClient()
+
+  useEffect(() => {
+    console.log('item', item)
+  }, [item])
 
   useEffect(() => {
     const comment = cache.readFragment({
@@ -275,6 +280,9 @@ export default function Comment ({
                   : null}
                 {/* TODO: add link to more comments if they're limited */}
               </div>
+              {item.newComments?.length > 0 && (
+                <ShowNewComments newComments={item.newComments} itemId={item.id} />
+              )}
             </div>
             )
       )}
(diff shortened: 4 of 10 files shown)


components/show-new-comments.js
```diff
@@ -0,0 +1,80 @@
...
+
+    // count the total number of comments including nested comments
+    let ncomments = data.ncomments + freshNewComments.length
+    for (const comment of freshNewComments) {
+      ncomments += (comment.ncomments || 0)
+    }
+
+    // update all ancestors
+    const ancestors = data.path.split('.')
+    updateAncestorsCommentCount(client, ancestors, ncomments)
```
Soxasora: > The updateAncestorsCommentCount function is incorrectly passed the total comment count (ncomments) instead of the actual increment (the count of new comments and their nested comments), causing ancestor comment counts to be inflated.

> **instead of the actual increment (the count of new comments and their nested comments)**

It is quite literally what I'm passing, maybe the naming isn't the best.

edit: and also I'm stupid because _I was_ passing the parent ncomments to its ancestors instead of only the new comments

Soxasora:
> Double-counting in `totalNComments` calculation

The `prepareComments` function counts of a parent (`data`):
1. Its direct new children (`freshNewComments.length`)
2. All `item.comments` comments nested under those new children (sum of each child's `ncomments`), not the comments that live in `newComments`.

This count is then propagated to the parent and its ancestors to update their `ncomments` field.

> Repeated updates from multiple components: When ShowNewComments components are rendered at multiple levels of a comment tree, the ancestor comment count update logic is triggered repeatedly for the same ancestors, causing new comments to be counted multiple times.

Items/comments that resides in `newComments` are not counted by the `ncomments` field yet, as it is computed for comments inside item.comments.*

> Additionally, the `data.path.split('.')` call in `prepareComments` lacks a null or undefined check for `data.path`, which can cause a runtime error if the path is missing or invalid.

A successful fragment that triggered the `prepareComments` function must have the `path`, this can't happen.

huumn:
This looks good at a glance. One thing I'd be curious to see is how it looks on mobile with deeply nested comments.

huumn:
If you're at the stage where you're open to enhancement suggestions. I think it'd be cool to:

1. change the favicon when a tab I'm on has new comments
2. give blue outlines of these freshly loaded comments (like we do when you visit a thread after a first visit and there are new comments)
3. put this `show new comment` thing also at the top of the thread to load all new comments

Soxasora:
Of course! These are pretty cool, going to implement them now ^^
I'll also make a mobile video example right after the new enhancements.

There's only a bug left: the deepest comment can't be updated because it lacks `item.comments` per `item_comments_zaprank_with_me/_limited`.

It's not a serious problem because you need to go to another page to see its nested comments, but the new comment indicator on **`view x replies on another page`** is a nice touch.
A workaround would be to just update the `newComments` field, but I'm comparing options to avoid bloat on the `updateCommentFragment` function... if it makes sense.

huumn:
I recommend starting these enhancements in a new branch. I think this branch can live on its own.

huumn:
It looks pretty good in my testing. It's easy to imagine how useful it will be. 

A few notes on first pass:
- I found the comment count to be unreliable if multiple come in
    - it seems to underestimate the number, or it only counts direct descendants in some cases. I can't tell. Recreate by commenting multiple times, one after the other, at a variety of levels.
- I found where the 'show new comment' appearing in parents and children simultaneously somewhat unexpected.

I'll look more deeply at the code and see what I can find. Please resist force pushing from now on, so that I can patch things as I notice them.

Soxasora:
> It's easy to imagine how useful it will be.

I'm actually pretty excited! Imagine arguing with someone, _live!_

> it seems to underestimate the number, or it only counts direct descendants in some cases. I can't tell.

It is a logic error, `collectAllNewComments` only counts the `newComments` field of each comment level, instead of also counting the already present children. This means that if you receive a new comment that already have 5 comments and also 1 new comment, the counter will only show 2.
Because of memoization, it can't show subsequent new comments of a new comment.

I think it's best to revise memoization on this

Being pretty close to `showAllNewCommentsRecursively` I think I got distracted, preparing a fix!

> I found where the 'show new comment' appearing in parents and children simultaneously somewhat unexpected.

It is a bit ugly, it shows the button whenever an item has `newComments`, which can be overwhelming 🤔

> Please resist force pushing from now on

I have this tendency of committing everything so sometimes I squash when things start to become too much, will commit consciously and patch without force pushing 🫡

components/show-new-comments.js
```diff
@@ -0,0 +1,140 @@
...
+  if (item.newComments && item.newComments.length > 0) {
+    const dedupedNewComments = dedupeNewComments(item.newComments, item.comments?.comments)
+    onLevel(dedupedNewComments, item)
+
+    for (const newCommentId of dedupedNewComments) {
+      const newComment = readNestedCommentsFragment(client, newCommentId)
+      traverseNewComments(client, newComment, onLevel, currentDepth + 1)
+    }
+  }
+}
```
Soxasora: By borrowing the recursive code of the new PR, I actually forgot that I was also counting/showing `newComments` of ALL the children. This caused some confusion, we just need to deal with direct new comments and their nested comments/new comments.

Since injection and counting shares the same logic, this is a general function that passes deduped new comments and the comment object to a `injectNewComments` or `countAllNewComments`.

huumn:
Seems to be working better since yesterday. 

Some QA feedback before I begin looking at the actual code

1. there's vertical layout shift when a top level comment arrives. I'm not sure how to avoid this without fixed positioning, but we definitely do not want vertical layout shift.

Screenshot 2025-07-17 at 3 04 16 PM

2. Also, I think, we probably want that top level `show new comments` from (1), wherever we place it, to reveal all new comments in the comment tree when pressed.
3. The outline of new comments is unreliable in my testing. The outlines showed initially on the first few new comments, but then didn't appear again.

huumn:
> there's vertical layout shift when a top level comment arrives. I'm not sure how to avoid this without fixed positioning, but we definitely do not want vertical layout shift.

Twitter uses fixed positioning to present any new posts:

Screenshot 2025-07-17 at 3 19 51 PM

It'd be nice to not need to do that at the top level but, again, I'm not sure where else to put it.

Soxasora:
> there's vertical layout shift when a top level comment arrives. ... we definitely do not want vertical layout shift.

I agree, I was even actually thinking of showing it in the comments header to avoid this, hmm

> Also, I think, we probably want that top level show new comments from (1), wherever we place it, to reveal all new comments in the comment tree when pressed.

That was something I debated with myself, we have the logic already, might as well use it. Mainly thought that it could be more comfortable to read at the top level comment first, and then load its replies. But it's actually way more comfortable to load it all.

> The outline of new comments is unreliable in my testing. The outlines showed initially on the first few new comments, but then didn't appear again.

Outline is only present in the _enhancements_ PR. Since we highlight new comments based on `router.query.commentsViewedAt` that we get from local storage, and this might be unavailable if the item has not been reached with client-side navigation, it required some highlight specialization for live comments.

Essentially outline works only if the commentsViewedAt query param is present, and if the new comment is part of a thread that had the outline unset, it stops highlighting

https://github.com/stackernews/stacker.news/blob/d5c9ffbddf953d0a9433d3d7e3b5f9780bbd7b3d/components/comment.js#L142-L148

https://github.com/stackernews/stacker.news/blob/d5c9ffbddf953d0a9433d3d7e3b5f9780bbd7b3d/styles/globals.scss#L911-L917

In the new PR, highlighting for live comments is managed via a client-side `injected` field, which may be overkill so I still want to try with the `commentsViewedAt` method to be less invasive.

components/use-live-comments.js
```diff
@@ -0,0 +1,85 @@
...
+  useEffect(() => {
+    if (!data?.newComments?.comments?.length) return
+
+    // sometimes new comments can arrive as orphans because their parent might not be in the cache yet
+    // queue them up, retry until the parent shows up.
+    const newComments = [...data.newComments.comments, ...queue.current]
+    const { queuedComments } = cacheNewComments(client, rootId, newComments, sort)
+
+    // keep the queued comments for the next poll
+    queue.current = queuedComments
```
huumn: why not just discard orphans? won't them and their parents get found in subsequent queries?

shouldn't the parent always be present in the response or in the cache if the child is in the response?

lib/comments.js
```diff
@@ -0,0 +1,83 @@
...
+// updates a comment fragment in the cache, with a fallback for comments lacking CommentsRecursive
+export function commentUpdateFragment (client, id, fn) {
+  let result = client.cache.updateFragment({
+    id: `Item:${id}`,
+    fragment: COMMENT_WITH_NEW_RECURSIVE,
+    fragmentName: 'CommentWithNewRecursive'
+  }, (data) => {
+    if (!data) return data
+    return fn(data)
+  })
```
huumn: nice reuse of these via `fn`

huumn (commented):
I can't fault it. Nice work!

I guess all it needs is removing the layout shift.

Oh, and did you get it to handle limited comment threads?

components/use-live-comments.js
Soxasora: Oh, you're right! _Many_ iterations ago, the queue was needed because of bad fragment structures. If you commented on a parent that wasn't injected yet, it wouldn't find the correct fragment in the cache and just forget about it forever. I can't see how/why it would happen now.

Soxasora:
> I can't fault it. Nice work!

Thanks! Iterative QA made the development **_slower than it needed to be_**, but I think that the result is much more sophisticated than what it would've been.

> I guess all it needs is removing the layout shift.

This, and deepest comments support:

> Oh, and did you get it to handle limited comment threads?

Kind of. The fallback mechanism on `commentUpdateFragment`/`readCommentFragment` (wow ok maybe the naming is all over the place) actually fixes this, by providing a fallback fragment, but not for the deepest comment, which lacks the `comments` field altogether.

Not a problem if you're loading the deepest levels yourself, since they're injected with the comment structure we need, but it is a problem if they're already there.

I didn't mess with `item_comments_zaprank_with_me/_limited` because it would inevitably impact the already established query performance and behavior, so I think I'll rather use another fallback fragment for this, albeit kinda ugly.

components/use-live-comments.js
```diff
@@ -0,0 +1,75 @@
...
+    // merge and cache new comments in their parent comment/post
+    cacheNewComments(client, rootId, data.newComments.comments, sort)
+
+    // update latest timestamp to the latest comment created at
+    // save it to session storage, to persist between client-side navigations
+    const newLatest = getLatestCommentCreatedAt(data.newComments.comments, latest)
+    setLatest(newLatest)
+    if (typeof window !== 'undefined') {
+      window.sessionStorage.setItem(latestKey, newLatest)
+    }
```
Soxasora, line 73: Client-side navigations would make the hook refetch (from cache) from the first timestamp it received.
Other than being a performance waste, it tried to inject new comments on comments that already have them, triggering deduplication every time.

Riding this bug, deduplication doesn't even work on comments at the deepest level as they lack the `comments` field, so it would perpetually inject new comments, even if they're there.

Now the latest new comment timestamp persists via `sessionStorage`, win win

huumn: Another option might be to use the logic we already have in `lib/new-comments.js` which handles the new comment outlining, and could allow you to use this value to solve outlining issues. I don't know though.

Soxasora: The thing is, outlining works somewhat correctly if you visit the post with client-side navigation, but if you visit it by its url, it fails because there isn't `commentsViewedAt` in the query params. I think I can work around this and make it work without the local `injected` field!

Here I update `commentsViewedAt` so that on the next visit, you don't get new comments you may have injected as unread:

```javascript
    const latestCommentCreatedAt = getLatestCommentCreatedAt(freshNewComments, data.createdAt)
    const rootId = data.path.split('.')[0]
    commentsViewedAfterComment(rootId, latestCommentCreatedAt)
```

But, wait, I just discovered a bug: I'm calling `commentsViewedAfterComment` after injecting a batch of comments, ignoring that the function was made only for single replies, since it updates `commentsViewedNum` by 1

https://github.com/stackernews/stacker.news/blob/56a185d477a61ae71aacd378b03e6658101d71a0/lib/new-comments.js#L11-L15

Pushing a patch to also add `totalNComments`.

---

Sadly I can't use `commentsViewedAt` for live comments, as it gets updated only when new comments are showed (new comments that have not been shown will be outlined this way on next visit).

Soxasora:
> This causes existingRootComments + ncomments to perform string concatenation (e.g., "5" + 1 = "51") instead of numeric addition, resulting in incorrect comment counts.

oh well, that was actually happening before too, maybe we never reached a point in which `viewNum` would be used.

```javascript
    if (viewedAt && viewNum) {
      return viewedAt < new Date(item.lastCommentAt).getTime() || viewNum < item.ncomments
    }
```

edit: just realized that if you visit a post for the first time and comment in dev, the new comment dot appears even if it's yours. Hmm

edit2: Ah, it's based off `item.lastCommentAt` which may be null in **dev**.

https://github.com/stackernews/stacker.news/blob/56a185d477a61ae71aacd378b03e6658101d71a0/components/item-full.js#L164-L166

huumn:
Top level comments are messed up on mobile.

Screenshot 2025-07-20 at 3 22 32 PM

I probably wasn't explicit enough, but this is why I thought we had no choice but to do fixed positioning.

Another solution to this kind of thing is to reserve empty space for the `show new comments` component, originally rendering it as `.invisible`.

Soxasora:
> top level comments are messed up on mobile

Ah! Sorry about it, that wasn't supposed to happen. 

We can make space for it in the `CommentsHeader` but I knew this positioning was anyway rushed, overall not so good. I think we can reserve empty space like you mentioned and avoid fixed positioning!

I'll look into this right now.

Soxasora:
I tried some positions yesterday

A smaller padding, feels overwhelming, looks a bit ugly
IMG_1604

Under `CommentsHeader`, too much space, still a bit ugly

https://github.com/user-attachments/assets/58fca15f-3ba9-4011-aa20-e2c8c80d4509



Above `CommentsHeader`, now it feels seamless, but still something feels off (bigger than what's underneath it)

https://github.com/user-attachments/assets/e64222ca-bd6b-4e28-9c20-e0d3f2034e7d



I think the last one is the most comfortable to look at and interact with, but it highlights the need for a scroll button or, yes, a fixed button that floats and then brings you to the comments

PR #2135, opened 2025-04-25
Exact search

### LINKED ISSUE #1391: For exact search matches, standard analyzer should be used
https://stacker.news/items/681047

m0wer:
## Description

Added exact search functionality for quoted phrases and words. Modified the search resolver to use wildcard queries on keyword fields instead of multi_match phrase queries, which improves the accuracy of quoted searches by making them truly case-sensitive and exact.

This change ensures that when users search for content in quotes, the search will find exact matches rather than partial or fuzzy matches, providing more precise search results.

Closes #1391.

### Before the changes

before

#### First commit only (no highlight)

after

#### Second commit (with existing highlight)

after2

## Additional Context

The existing implementation has issues with syntax highlighting for exact matches, matching the individual words instead of the exact input.

This implementation has two commits: one with no highlighting for exact words and a second one that enables something similar to the existing highlighting. I see no clean/simple way of implementing highlighting for exact matches without refactoring or significant additional logic.

## Checklist

Yes, this change is backwards compatible. It modifies the search behavior for quoted text to be more precise but doesn't break existing functionality.

8/10. I've tested various quoted search phrases in both title and text fields, verified case sensitivity works correctly, and checked that regular searches still function as expected.

This is a backend search functionality change and doesn't directly impact the frontend display beyond the search results themselves.

No new environment variables were introduced.

### DIFF
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -283,45 +283,21 @@ export default {
 
       // if quoted phrases, items must contain entire phrase
       for (const quote of quotes) {
-        const isMultiWord = quote.includes(' ')
-        const queryType = isMultiWord ? 'match_phrase' : 'term'
-
-        const titleQueryBody = isMultiWord ? { query: quote } : quote
-        const textQueryBody = isMultiWord ? { query: quote } : quote
-
         filters.push({
-          bool: {
-            should: [
-              { [queryType]: { 'title.exact': titleQueryBody } },
-              { [queryType]: { 'text.exact': textQueryBody } }
-            ],
-            minimum_should_match: 1
+          multi_match: {
+            query: quote,
+            fields: ['title.exact', 'text.exact'],
+            type: 'phrase'
           }
         })
-
-        if (isMultiWord) {
-          termQueries.push({
-            match_phrase: {
-              'title.exact': { query: quote, boost: 100 }
-            }
-          })
-          termQueries.push({
-            match_phrase: {
-              'text.exact': { query: quote, boost: 100 }
-            }
-          })
-        } else {
-          termQueries.push({
-            term: {
-              'title.exact': { value: quote, boost: 100 }
-            }
-          })
-          termQueries.push({
-            term: {
-              'text.exact': { value: quote, boost: 100 }
-            }
-          })
-        }
+        termQueries.push({
+          multi_match: {
+            query: quote,
+            fields: ['title.exact^10', 'text.exact'],
+            type: 'phrase',
+            boost: 1000
+          }
+        })
       }
 
       switch (sort) {
@@ -410,6 +386,24 @@ export default {
               type: 'phrase',
               fields: ['title^10', 'text'],
               boost: 1000
+            }
+          },
+          // match on exact fields higher
+          {
+            multi_match: {
+              query,
+              type: 'best_fields',
+              fields: ['title.exact^10', 'text.exact'],
+              boost: 100
+            }
+          },
+          // exact phrase matches higher
+          {
+            multi_match: {
+              query,
+              fields: ['title.exact^10', 'text.exact'],
+              type: 'phrase',
+              boost: 10000
             }
           }
         ]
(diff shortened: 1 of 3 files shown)


Soxasora:
I didn't look at the code and everything else but I wanted to let you know that you brought your commits from your last branch into this one ^^

m0wer:
Ouch... Thanks for the tip! Rebased.

m0wer:
Hi @huumn ! I started that way but assumed that you would prefer a simpler solution, since the issue had `label:"difficulty:easy"`. If you want to explore the reindexing or other alternatives I'm happy to do so.

m0wer:
Re-implemented by creating a new analyzer, `exact_case_sensitive_analyzer`, that is case sensitive and does not perform stemming. Then we use that index with `match_phrase` instead of wildcard for efficiency. Highlighting works as well. Here are some examples:

liquid
Liquid

For the migration: is it OK to recreate the index? If not we can work on a migration script that starts from the existing one.

What do you think of this approach @huumn ?

m0wer:
ping

api/resolvers/search.js
```diff
@@ -411,6 +387,24 @@ export default {
...
+            }
+          },
+          // exact phrase matches higher
+          {
+            multi_match: {
+              query,
+              fields: ['title.exact^10', 'text.exact'],
+              type: 'phrase',
+              boost: 10000
+            }
```
huumn, line 407: I also thought it'd be nice to use these exact fields in other queries to boost relevance.

api/resolvers/search.js
```diff
@@ -511,21 +505,13 @@ export default {
...
-          item.searchTitle = e.highlight.title[0]
-        } else {
-          item.searchTitle = item.title
-        }
+        // prefer the fuzzier highlight for title
+        item.searchTitle = e.highlight?.title?.[0] || e.highlight?.['title.exact']?.[0] || item.title
+
+        // prefer the exact highlight for text
+        const searchTextHighlight = [...(e.highlight?.['text.exact'] || []), ...(e.highlight?.text || [])]
+        item.searchText = searchTextHighlight?.slice(0, 5)?.join(' ... ')
```
huumn, line 513: If the title is present and the exact title matches, it means the non-exact title has those matches but it will also have the non-exact matches ... it's like a superset of matches, so I changed it to prefer the non-exact highlights if present on the title.

For the text, we prefer the exact matches, but if there are <5, we also want the non-exact matches, so I concat'd them together then sliced them.

docker/opensearch/init-opensearch.sh
```diff
@@ -41,7 +31,7 @@ else
               "keyword": {"type": "keyword", "ignore_above": 256},
               "exact": {
                 "type": "text",
-                "analyzer": "exact_case_sensitive_analyzer"
+                "analyzer": "standard"
```
huumn, line 34: I think case sensitivity is overkill for now. But, now that I think about it, maybe it should be another field and we should us it to rank the case sensitive matches higher.

scripts/nlp-setup
```diff
@@ -295,12 +295,24 @@ else
...
            "type": "text",
            "analyzer": "english",
-           "fields": {"keyword": {"type": "keyword", "ignore_above": 256}}
+           "fields": {
+             "keyword": {"type": "keyword", "ignore_above": 256},
+             "exact": {
+               "type": "text",
+               "analyzer": "standard"
+             }
+           }
```
huumn, line 315: the nlp-setup defines the mappings on its index too, so I added this there.

huumn (approved):
This is good to go.

I'll need to perform an index migration in prod to deploy this though. 

So, I'm going to mark it as `blocked`. But, I'll pay the bounty for it without merging.

huumn:
I'm currently reindexing prod, which is slow and tends to be hit or miss, but once I can get it to succeed, I should be able to switch prod over and merge this.

PR #2136, opened 2025-04-26
Add section about automated retries for payments in README

ekzyis:


README.md
```diff
@@ -479,6 +480,18 @@ To ensure stackers balances are kept sane, some wallet updates are run in [seria
 
 <br>
 
+## Automated payment retries
+
+Every payment for `ITEM_CREATE`, `ZAP`, `DOWN_ZAP`, `POLL_VOTE`, and `BOOST` is attempted three times (first attempt + two retries). We wait 1 minute and up to 1 hour before we return a failed invoice via the `failedInvoices` query to a client to retry. During each retry, we increment the `paymentAttempt` column of the new invoice and add a link to the previous invoice in `predecessorId`.
```
huumn, line 547: My main confusion automated retries is how `paymentAttempt` counts relate to sender/receiver fallbacks.

Does a `paymentAttempt` mean one cycle through all receiver wallets or one cycle through all sender wallets or both?

I think an example is likely the clearest way to explain how this all works.

ekzyis: It means one cycle through all sender and receiver wallets, so the next attempt will try all fallbacks again (as necessary[^1]).

[^1]: We don't necessarily cycle through all sender or receiver wallets in a payment attempt. It could be that always the receiver failed so we would never fallback on the sender's side.

Example: 

- A has NWC and LNbits wallet to send
- B has NWC and LNbits wallet to receive
- A zaps B

1.1 NWC → NWC | :x: receiver error
1.2 NWC → LNbits | :x: sender error
1.3 LNbits → LNbits | :x: sender error

We consider the payment attempt failed now since the sender has no more fallbacks. (The receiver also doesn't have any more fallbacks, but their wallet worked, so we would try it again.)

So after 1.3, we will not call `retryPaidAction(invoiceId, newAttempt: false)` as during fallbacks again, but we will return the failed invoice to a polling client via the `failedInvoices` query. A client will then call `retryPaidAction(invoiceId, newAttempt: true)` and attempt the payment with all fallbacks again:

2.1 NWC → NWC | :x: receiver error
2.2 NWC → LNbits | :x: sender error
2.3 LNbits → LNbits | :x: receiver error

Receiver has no more fallbacks, so payment failed again

3.1 NWC → NWC | :x: receiver error
3.2 NWC → LNbits | :white_check_mark:

Does this make sense?

I think a lot of the confusion comes from the terminology, we can say "retry" when we mean the fallbacks or not. The fallbacks themselves are "retries" since they use the `retryPaidAction` mutation to get a new invoice to pay, but if we pass `newAttempt: true` to `retryPaidAction`, we "retry" (the payment attempt) with all fallbacks.

edit: I'll include this information in the documentation, too

huumn: Thanks that helps. I do think it's mostly a terminology thing if `paymentAttempt = allSenderReceiverPermutationsAttempt`.

The pseudo code sounds like:
```
for paymentAttempt=0; paymentAttempt < 3
   for wallet in receiver wallets
       invoice = getInvoice(wallet)
       if invoice
           for wallet in sender wallets
               succeeded = payInvoice(wallet, invoice)
               if succeeded return
```

Is that right?

ekzyis: Mhh, we don't move to the next receiver wallet and try all sender wallets again if all sender wallets already failed. If all sender or receiver wallets failed, we have to increase `paymentAttempt`.

The pseudo code is correct if:

1. it assumes one sender wallet will always succeed
2. it assumes a receiver wallet can't fail on forward
3. it ignores we never pay the same invoice twice

huumn:
```
for paymentAttempt=0; paymentAttempt < 3; paymentAttempt++
   for wallet in receiver wallets
       invoice = getInvoice(wallet)
       if invoice
           for wallet in sender wallets
               succeeded, forward_failed = payInvoice(wallet, invoice)
               if forward_failed # means we need to try new receiver wallet
                    break
               if succeeded # done
                    return
           if forward_failed # try new receiver wallet
                continue
           if not succeeded # all sender wallets failed, so start new payment attempt
               break
```

is that right?

what does (3) avoid? the preimage hasn't been revealed if the payment fails, so what else could go wrong?

ekzyis: > Thanks that helps. I do think it's mostly a terminology thing

Same is true for the term "payment" btw. It's not clear without context if we mean with "payment" just one invoice, or if it includes all fallbacks, or if it includes automated retries.

> is that right?

yeah that looks right to me now, excluding (3)

> what does (3) avoid? the preimage hasn't been revealed if the payment fails, so what else could go wrong?

We cancel every invoice before we move on to the next sender wallet because we can't be 100% sure that the invoice will not still get paid. Something might simply have timed out or the wallet reported a false negative regarding the payment. If we cancel the invoice, we can be sure on a protocol level that this invoice will not get paid.

huumn:
```
for paymentAttempt=0; paymentAttempt < 3; paymentAttempt++
   for rwallet in receiver wallets
       for swallet in sender wallets
           invoice = getInvoice(rwallet)
           if invoice
               succeeded, forward_failed = payInvoice(swallet, invoice)
               if succeeded
                   return # done
               else 
                   cancel(invoice)
               if forward_failed # means we need to try new receiver wallet
                   break
       if forward_failed # try new receiver wallet
           continue
       if not succeeded # all sender wallets failed, so start new payment attempt
           break
```

Does that now succinctly describe the retry logic?

ekzyis:
Yes!

ekzyis:
Uhm, just realized that it's missing going to the next receiver wallet if the wallet failed to return an invoice:

```
for paymentAttempt=0; paymentAttempt < 3; paymentAttempt++
   for rwallet in receiver wallets
       for swallet in sender wallets
           invoice = getInvoice(rwallet)
           if invoice
               succeeded, forward_failed = payInvoice(swallet, invoice)
               if succeeded
                   return # done
               else 
                   cancel(invoice)
               if forward_failed # means we need to try new receiver wallet
                   receiver_error = true
                   break
           else
               receiver_error = true
               break
       if receiver_error # try new receiver wallet
           continue
       if not succeeded # all sender wallets failed, so start new payment attempt
           break
```

ekzyis:
I think we don't need to merge this anymore since how this works will soon change because of #2084

PR #2138, opened 2025-04-26
Fix lnaddr attach fails if minSendable > 1000

ekzyis:
## Description

Test invoices for a lightning address can fail if we're requesting an invoice for an amount it does not support.

Example: https://breez.fun/.well-known/lnurlp/needcreations

This was reported [here](https://stacker.news/items/958590?commentId=960241).

## Additional Context

The other issue that one can see in the [report](https://stacker.news/items/958590?commentId=960241) is that [`assertContentTypeJson`](https://github.com/stackernews/stacker.news/blob/4ad64d658fcf33d34ff3b3224d32d1ef9c7e513a/lib/url.js#L236-L241) would also throw `ResponseAssertError` which will log a confusing error message with 200 OK instead of "expected json, got html."

I will fix this in another PR.

## Checklist

yes

`8`. Made sure that `lnAddrOptions` will always return `min` in sats and tested attaching a lightning address with calle@npub.cash.

n/a

no

### DIFF
--- a/wallets/lightning-address/server.js
+++ b/wallets/lightning-address/server.js
@@ -6,8 +6,7 @@ import { assertContentTypeJson, assertResponseOk } from '@/lib/url'
 export * from '@/wallets/lightning-address'
 
 export const testCreateInvoice = async ({ address }, { signal }) => {
-  const { min } = await lnAddrOptions(address, { signal })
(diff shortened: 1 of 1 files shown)


huumn:
I believe `min` is already expressed in millisats

PR #2145, opened 2025-05-01
custom domains - middleware and verification

Soxasora:
## Description
Part of #1942 
Focuses on custom domain insertion, providing a working form and verification process via pgboss.
Uses ACM with LocalStack to provide certificates bare CRUD simulation.

## Screenshots
TBD

## Additional Context
#1958 is being split to facilitate review and build greater stuff on top of it.
Doesn't contain Auth Sync or custom branding.

This PR provides:
- ACM support
- custom domains crud, resolvers, fragments
- custom domains form, guidelines
- custom domains context
- domain verification every 5 minutes via pgboss
- domain validation schema
- basic custom domains middleware
- TODOs tracings
- TODO: changes from #1958
- -- TODO: Adjust domain verification worker to new DomainVerification table
- -- OK: Normalize domain verification

## Checklist

Yes, everything is an addition. An exception is the way we handle S3 via localstack, now the container is called AWS and also allows ACM simulations other than S3.

5, in testing but working as intended.

Yes, it uses standard SN colors and styles that fits with theme changes.

An endpoint for localstack that can be contacted by our worker container
`LOCALSTACK_ENDPOINT=http://aws:4566`

DNS resolver for our DNS verification stage
`DNS_RESOLVER=1.1.1.1`


huumn:
https://github.com/stackernews/stacker.news/blob/c7321351bfef74dc2cfa9d0c2542bb3fbbc5db2c/docs/dev/custom-domains-base.md

So to QA this I need to use real DNS? I suspect we'll want to mock DNS going forward to make this trivial to iterate on.

pages/api/domains/index.js
```diff
@@ -0,0 +1,40 @@
...
+      }
+      return acc
+    }, {})
+
+    return res.status(200).json(domainMappings)
+  } catch (error) {
+    console.error('cannot fetch domains:', error)
+    return res.status(500).json({ error: 'Failed to fetch domains' })
+  }
+}
```
huumn, line 40: This endpoint just exists for the `cachedFetcher` call in `lib/domains.js`?

And that's because middleware can't import the prisma stuff, right?

prisma/migrations/20250504202129_custom_domains_base/migration.sql
```diff
@@ -0,0 +1,157 @@
...
+AFTER INSERT ON "DomainVerificationAttempt"
+FOR EACH ROW
+EXECUTE FUNCTION update_record_status_from_attempt();
+
+-- HOLD the domain and delete the certificate when the sub is stopped
+-- this is to prevent the domain from being used by another sub
+-- when the sub is resumed, the domain can be verified again and another certificate can be issued
+
+-- !QUIRK: If someone else takes over the sub, the new guy needs to delete the domain
+CREATE OR REPLACE FUNCTION hold_domain_and_delete_certificate()
```
huumn, line 137: That is weird. I guess it's not clear to us whether the transfer is from an alt to another alt or to another person entirely.

worker/territory.js
```diff
@@ -13,13 +14,20 @@ export async function territoryBilling ({ data: { subName }, boss, models }) {
...
+      }
+
       await models.sub.update({
         include: { user: true },
         where: {
           name: subName
         },
         data: {
-          status: nextBillingWithGrace(sub) >= new Date() ? 'GRACE' : 'STOPPED',
+          status: nextStatus,
```
huumn, line 30: hmmm hard to tell if this is the right behavior unless we automatically re-add the cert when it's paid.

huumn (commented):
I did a *very* casual high level first pass to get acquainted. It looks very good. 

Thank you for separating it into its own PR. Also thanks to @ekzyis for his first review which I'm sure led to lots of improvements.

I left some questions and nits. I'll make another high level pass tomorrow. Then maybe again depending on when we can test this in a self-contained way.

worker/territory.js
Soxasora: Hmm, a new certificate from ACM is fast and free, what's not really good (for the territory owner) is that they will be subjected again to ACM DNS validation.

Maybe we don't need to remove the certificate at all as long as the domain remains the same.

prisma/migrations/20250504202129_custom_domains_base/migration.sql
Soxasora: I think we can actually just remove the custom domains settings and certificate if the owner is different on re-activation.

And, instead, on transfer, we retain settings

api/resolvers/domain.js
```diff
@@ -0,0 +1,175 @@
...
+          if (existing) {
+            return await models.$transaction(async tx => {
+              // delete any existing domain verification job left
+              await cleanDomainVerificationJobs(existing, tx)
+
+              // deleting a domain will also delete the domain certificate
+              // but we need to make sure to delete the certificate from ACM
+              if (existing.certificate) {
+                await deleteDomainCertificate(existing.certificate.certificateArn)
+              }
```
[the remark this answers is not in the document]
Soxasora: Oh! You're right, totally missed this. Going to push changes in a bit :)

prisma/migrations/20250504202129_custom_domains_base/migration.sql
```diff
@@ -129,29 +132,59 @@ AFTER INSERT ON "DomainVerificationAttempt"
...
+      'deleteDomainCertificate',
+      jsonb_build_object('certificateArn', certificate_arn),
+      3, -- retry 3 times on ACM errors
+      60*60); -- wait 1 hour between retries
+  RETURN OLD;
+END;
+$$ LANGUAGE plpgsql;
+
+-- everytime a domain (relation) or a domainCertificate is deleted, also ask ACM to delete the certificate
+CREATE TRIGGER trigger_ask_acm_to_delete_certificate
```
Soxasora, line 186: Instead of jumping around the code to ask ACM to delete a certificate (mostly in shady places), a trigger on domain/domainCertificate deletion offloads this job to our worker via pgboss.

prisma/migrations/20250504202129_custom_domains_base/migration.sql
```diff
@@ -129,29 +132,59 @@ AFTER INSERT ON "DomainVerificationAttempt"
...
--- this is to prevent the domain from being used by another sub
--- when the sub is resumed, the domain can be verified again and another certificate can be issued
-
--- !QUIRK: If someone else takes over the sub, the new guy needs to delete the domain
-CREATE OR REPLACE FUNCTION hold_domain_and_delete_certificate()
+-- SCENARIO: Territory got stopped after grace period
+-- HOLD the domain when the sub is stopped
+-- this is to prevent the domain from being used by another sub;
+-- won't delete anything but it will require a new verification attempt if the sub is resumed
+CREATE OR REPLACE FUNCTION hold_domain_on_sub_stop()
```
Soxasora, line 139: Instead of deleting a domain from a territory, we put it on HOLD so that the original owner can come back without losing any customization.

prisma/migrations/20250504202129_custom_domains_base/migration.sql
```diff
@@ -129,29 +132,59 @@ AFTER INSERT ON "DomainVerificationAttempt"
...
+AFTER UPDATE ON "Sub"
+FOR EACH ROW
+WHEN (NEW.status = 'STOPPED')
+EXECUTE FUNCTION hold_domain_on_sub_stop();
 
+-- SCENARIO: Territory got taken over by a different user
+-- clear the domain when the sub is taken over by a different user;
+-- this will delete the domain, its certificates, verification attempts, DNS records and brandings.
+-- will also trigger a request to ACM to delete the certificate because the domain is being deleted
+CREATE OR REPLACE FUNCTION clear_domain_on_sub_takeover()
```
Soxasora, line 157: Now that we only HOLD the domain on territory expiration, I think it's fair to delete any domain information, with its certificate (thanks to `trigger_ask_acm_to_delete_certificate`), if the territory gets resumed by another user instead of the original owner.

prisma/migrations/20250504202129_custom_domains_base/migration.sql
```diff
@@ -111,6 +111,9 @@ ALTER TABLE "DomainVerificationRecord" ADD CONSTRAINT "DomainVerificationRecord_
 -- AddForeignKey
 ALTER TABLE "DomainCertificate" ADD CONSTRAINT "DomainCertificate_domainId_fkey" FOREIGN KEY ("domainId") REFERENCES "Domain"("id") ON DELETE CASCADE ON UPDATE CASCADE;
 
+-- Clear domains that have been on HOLD for 30 days or more every midnight.
+INSERT INTO pgboss.schedule (name, cron, timezone) VALUES ('clearLongHeldDomains', '0 0 * * *', 'America/Chicago') ON CONFLICT DO NOTHING;
```
Soxasora, line 115: I've been wanting to add this for a long time but didn't know if we actually need it, regardless now we check every midnight if a domain has been on HOLD for 30 days or more.

There might be a better way to do this so it might get iterated?

middleware.js
```diff
@@ -11,6 +12,38 @@ const SN_REFERRER = 'sn_referrer'
...
+  // TEST
+  console.log('[domains] custom domain', domain, 'with subname', subName) // TEST
+  console.log('[domains] main domain', SN_MAIN_DOMAIN) // TEST
+  console.log('[domains] pathname', pathname) // TEST
+  console.log('[domains] searchParams', searchParams) // TEST
+  console.log('[domains] origin', origin) // TEST
+
+  // TODO: 1. handle auth sync
+  // TODO: 2. redirect ~subname to /
+  // TODO: 3. rewrite / to ~subname
```
huumn, line 38: Can we add TODO 3 to this PR?

Without it, it feels like the PR doesn't do much.

huumn (changes requested):
I got it to verify the domain and cert. It was slow but smooth. I'm going to QA more deeply and look at the edge cases, but I think to merge this PR we'll want to add:

1. usage of the tls cert (cloudfront orchestration that we talked about or something else)
2. hiding/blocking usage of this feature except for admins
3. make the rewrite/redirect work

That's where merging/shipping this makes sense imo. Auth can be its own PR, and customization another.

api/resolvers/domain.js
```diff
@@ -0,0 +1,166 @@
...
+            ? await tx.domainVerificationRecord.findUnique({
+              where: {
+                domainId_type_recordName: {
+                  domainId: existing.id,
+                  type: 'TXT',
+                  recordName: '_snverify.' + existing.domainName
+                }
+              }
+            })
+            : null
```
huumn, line 90: This should be moved to the `if` block at line 67, right?

prisma/schema.prisma
```diff
@@ -1243,6 +1244,106 @@ model Reminder {
...
+
+enum DomainCertificateStatus {
+  PENDING_VALIDATION
+  ISSUED
+  INACTIVE
+  EXPIRED
+  REVOKED
+  FAILED
+  VALIDATION_TIMED_OUT
+}
```
huumn, line 1345: Are all of these used? Some of them aren't afaict. If we aren't using any of these status enums, or any of the other enums, I think it'd be best to remove them. If they're here as TODOs, we should add a `TODO` comment. Their existence kind of implies these cases are handled somewhere.

worker/domainVerification.js
```diff
@@ -0,0 +1,273 @@
+import createPrisma from '@/lib/create-prisma'
+import { verifyDNSRecord, issueDomainCertificate, checkCertificateStatus, getValidationValues, deleteDomainCertificate } from '@/lib/domain-verification'
+import { datePivot } from '@/lib/time'
+
+const VERIFICATION_INTERVAL = 60 * 5 // 5 minutes
```
huumn, line 5: I'm pretty sure you and @ekzyis debated this, but I think every 5 minutes is too slow.

My expectation as a user is that after I set DNS, within ~30-60 seconds I see a success message.

Ideally, it might be best to backoff to 5 minutes after an hour, but before an hour has passed do every 30 seconds.

worker/domainVerification.js
```diff
@@ -0,0 +1,273 @@
...
+const VERIFICATION_HOLD_THRESHOLD = -2 // 2 days ago
+
+export async function domainVerification ({ id: jobId, data: { domainId }, boss }) {
```
huumn, line 8: A lot of what happens in this job is not transactional. It's not clear to me if that's okay or not. Is it?

huumn: Also, how does this job deal with:

1. concurrency ... is it idempotent? does it matter?
2. what happens if the server crashes in the middle of this job, will a domain get stuck at whatever stage it's on? Do we have a way to resume?

huumn (commented):
I left a nit and some additional questions. 

I'm slowly beginning to grok the flow. I'll try to look for more potential problems/challenges tomorrow.

worker/domainVerification.js
Soxasora: You're right in asking this. I'll start by saying that concurrent workers would potentially speed up the process instead of messing with it.             

If the worker crashes during a job, pgboss will run it again thanks to the retry limit.
The fact that it starts from the beginning is good, we get to verify DNS again and execute the certificate step that we need since they're controlled by conditions.

This is the flow of a PENDING domain verification:
```
domain is PENDING
  1. verify DNS
      KO: schedule job
      OK: proceed

  IMPLICIT: DNS is OK
  2. Certificate Management
      KO: critical AWS error, schedule up to 3 retries

        CONDITION: certificate is not issued
        2a. issue certificate
              KO: schedule job
              OK: proceed

              CONCURRENCY: ACM will deny a second request
              NOTE: we probably want to recognize if we couldn't request a certificate
                    because we already have one but we didn't get to save it

        CONDITION: SSL records are not present
        2b. get SSL validation values
              KO: schedule job
              OK: schedule job, the user will add the records

              CONCURRENCY: validation values are the same

        IMPLICIT: certificate is issued, SSL records are present
        2c. check certificate status
              KO: schedule job
              OK: update certificate status, proceed

              CONCURRENCY: status comes from ACM

  IMPLICIT: DNS is OK, certificate is OK
  3. update domain status to VERIFIED 🎉
```

So steps can both be consecutive or independent, relying on the previous step result or data inside domain.

prisma/schema.prisma
Soxasora: We're not using these statuses but they reflect possible statuses of an ACM certificate.
My idea was to use these to give the user more feedback of what's going on but I didn't plan it in this PR.

I'll put a TODO so I can assess if we actually want to use them.

api/resolvers/domain.js
Soxasora: Yeah! I don't know how I didn't notice it, adjusting ^^

huumn:
I gave it another pass. I won't nit anymore. I think all we need now is the cert stuff.

worker/domainVerification.js
```diff
@@ -49,11 +49,11 @@ export async function domainVerification ({ id: jobId, data: { domainId }, boss
     // if it's not PENDING, we stop the verification process.
     if (result.status === 'PENDING') {
       // we still need to verify the domain, schedule the job to run again
-      const newJobId = await boss.send('domainVerification', { domainId }, {
+      const newJobId = await boss.sendDebounced('domainVerification', { domainId }, {
         startAfter: VERIFICATION_INTERVAL(domain.updatedAt),
         retryLimit: 3,
         retryDelay: 60 // on critical errors, retry every minute
-      })
+      }, 30, `domainVerification:${domainId}`)
```
Soxasora, line 56: Now another job cannot be scheduled if there's another one running, thanks to the singleton key tied to the domain id.
To maintain manual job scheduling, I switched to sendDebounced which would, respecting the `startAfter`, schedule a job whenever it can, i.e. once the other job has finished and/or after 30 seconds.

I think I'll change those 30 seconds to the VERIFICATION_INTERVAL, it only serves as a safe timeout.

I switched to sendDebounced to schedule a job

huumn: Oh neat, I haven't found reason to use that yet.

Soxasora:
The assumptions that I made in doing this were that
a) we only have a **single** load balancer 
b) we have stacker.news pointing at **that** load balancer (listener).

In the code I specifically point at the first (and should be only) load balancer and on that ELB I point at the first HTTPS listener, because I excluded the possibility of having multiple load balancers or that we could have multiple listeners in various positions. This is, of course, something that can be tweaked in the code, we can also just put the ARN of the listener in the production environment variables and use that.

> I'm wondering how this will interact with our existing application load balancer rules

Back on the main question, the only thing that this new code does is attaching or detaching an ACM certificate to a listener[0] of a load balancer[0].

Undoubtedly having stacker.news pointing directly at the ELB DNS is something that free us from extra hoops such as creating ad-hoc subdomains per each custom domain. If we were to have SN pointing at CloudFront and then ELB, we would need to create CloudFront subdomains and give them to the user to point their CNAME record at.

huumn:
Cool! Thank you!

huumn:
I need to confirm we can add as many certs to ALB as we want, and that one doesn't replace the other, and that they are served appropriately without us specifying special rules. In part this is a note to myself, but if you have links to docs or anything sox let me know.

My biggest fear with this PR is that shipping it breaks normal site operations because otherwise it's just a private feature.

Soxasora:
> I need to confirm we can add as many certs to ALB as we want

Ah, this is where it fails at scaling. ALBs have a cap of [25 additional](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html?utm_source=chatgpt.com#load-balancers-quotas) ACM certificates and while a quota increase can be requested, it's not ideal and it can be expensive (there's not an easy way to figure out these costs without making a request it seems).

I did go on with this implementation for our internal testing period because it's compatible with the live AWS configuration but I had a different idea months ago, more on this [here](#the-cloudfront-plan)

> that one (cert) doesn't replace the other

We can't replace a default certificate of a load balancer without **specifically** doing that via [modifyListener](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-update-certificates.html#replace-default-certificate). This code only adds or remove a certificate from the list of additional certificates.

> that the (certs) are served appropriately without us specifying special rules

Something cool about the load balancer is that: 
> The load balancer uses a smart certificate selection algorithm with support for SNI. If the hostname provided by a client matches a single certificate in the certificate list, the load balancer selects this certificate.
<sup>[SSL certificates for you Application Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/https-listener-certificates.html#sni-certificate-list)</sup>

So when we visit www.pizza.com pointed at stacker.news, the load balancer will select the correct certificate for the incoming domain.

Everything is pretty cool but we need to move away from the actual configuration if we want to scale and allow more than 25 custom domains.

## The CloudFront plan

This is my original proposal:
A better way to deal with custom domains is to use CloudFront distributions per custom domain and point those distributions to our ALB. We can create up to [500 distributions](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-limits.html#limits-web-distributions) and it's a matter of creating a distribution right after domain verification, with the related ACM certificate attached to it.

We would then ask the user to point `www.pizza.com` to that CloudFront distribution which can either be `1234random21.cloudfront.net` or masked with `pizza.territory.stacker.news` pointing to `1234random21.cloudfront.net`

## The alternate CloudFront plan

ACM certificates supports up to 10 domain names and a CloudFront distribution can only have a single ACM certificate.
We can choose to be 'smart' (but imo over-engineer) this by batching custom domains in groups of 10 and add distributions once every 10 domains.
It adds complexity: we would need to keep track of ACM certificates and its attached domains, re-issue the certificate every time a new domain comes up, orchestrate CloudFront to issue a distribution when we reach the 10 domains cap.

I personally don't recommend this for now, but we can think about it.

## this is not a thesis sox, tldr
For now we can use the live ALB and safely add/remove certificates without ever messing with the live configuration, but if we want to scale we need to move to CloudFront.

worker/domainVerification.js
```diff
@@ -0,0 +1,337 @@
...
+    if (certificateArn && !recordMap.SSL) {
+      await getACMValidationValues(domain, models, certificateArn)
+
+      // return PENDING to check ACM validation later
+      return { status, message: 'Certificate issued and validation values stored.' }
+    }
+
+    // STEP 2c: Check ACM validation
+    const sslVerified = await checkACMValidation(domain, models, recordMap.SSL)
+    if (!sslVerified) return { status, message: 'ACM validation is still pending.' }
```
Soxasora, line 132: > I recall needing to perform DNS validation to get the cert approved.

Yep! This is where we get the DNS validation values for the ACM certificate and on step 2c we poll ACM to check the result of the DNS validation, finally we update the certificate status.

huumn: Oh, I didn't know there were multiple steps!

In local dev, during my QA, SSL was verified without me doing the second step. hmmm

huumn: I'm confused is this a bug or how it's expected to work in local?

I guess because local ACM doesn't check for this TXT record, it automatically issues it.

--------

Also, this makes me wonder if the `_snverify...` TXT check could just be replaced with us checking for the ACM verification record - ie we reuse ACM verification's that the domain is owned by the stacker, as our own verification that the stacker owns the domain?

Soxasora: > I guess because local ACM doesn't check for this CNAME record, it automatically issues it.

Yeah, localstack doesn't simulate timings and stuff like that, 

> ie we reuse ACM verification that the domain is owned by the stacker, as our own verification that the stacker owns the domain

Cool! I definitely like this, less DNS records for the user ^^

> we could check for the CNAME ourselves ...

Well, we still poll ACM to know if the certificate is 100% `ISSUED` and be safe before trying to attach it to the load balancer so we can benefit from it by using it as part of domain ownership proof. If it's okay for you, I'd go with that!

fragments/domains.js
```diff
@@ -0,0 +1,95 @@
...
+`
+
+export const GET_DOMAIN_MAPPING = gql`
+  query DomainMapping($domainName: String!) {
+    domainMapping(domainName: $domainName) {
+      domainName
+      subName
+    }
+  }
+`
```
ekzyis: This isn't used afaict. Will this be used later?

Soxasora: Oh! That shouldn't be there.
Originally, the ssrApollo API used this query to give `domainName` and `subName` to the rest of the app, but it was too expensive and also _dumb_, so I stopped using it in favor of `subName` passed as header from the middleware.

Some time ago I had plans to use it for other stuff that required hard ACTIVE checks, but it's not needed.

api/acm/index.js
```diff
@@ -0,0 +1,44 @@
+import AWS from 'aws-sdk'
+
+AWS.config.update({
+  region: 'us-east-1'
+})
```
ekzyis: This looks like a global config. Unlike the global DNS resolver config I mentioned in a previous review, this here probably doesn't matter much, but I still wonder if it's possible to put this into the `config` object that is passed to `AWS.ACM`?

api/elb/index.js
```diff
@@ -0,0 +1,42 @@
+import AWS from 'aws-sdk'
+import { MockELBv2 } from './mocks'
+
+const ELB_LISTENER_ARN = process.env.ELB_LISTENER_ARN
+
+AWS.config.update({
+  region: 'us-east-1'
+})
```
ekzyis: see other comment regarding global config

api/acm/index.js
Soxasora: Ah yes, It is a global config! My bad, I used the S3 API as a template
https://github.com/stackernews/stacker.news/blob/0310611a2dfcdf71de633493015836058c217c4b/api/s3/index.js#L4-L9

I think we can actually just remove it from both ACM and ELB, because S3 is already setting this globally.
edit: No I said something incorrect, it's better to move it to the `config` object like you said, as S3 can also not run.

PR #2146, opened 2025-05-02
Wallet schema v2

ekzyis:
## Description

based on #2092 because it's the target branch

This PR does the following:

* add static table `WalletTemplate` which stores which send and receive protocols a wallet in `UserWallet` supports
 => allows multiple send and receive protocols per wallet
* split wallets by send and receive for easier validation of configuration instead of having both send and receive optional and rely on application logic to make sure the configuration is not invalid
* add wallet protocol tables (`WalletSend...`, `WalletRecv...`) which store configuration per user wallet
* use triggers to make sure that a protocol of a user wallet is actually supported by the wallet

How to add a new wallet:

* add a new row to the `WalletV2` table and declare which send and receive protocols it supports
* if it supports a new protocol, update enums and add a new wallet protocol table to store configuration

### My beautiful TODO list

* use unique index for `walletId` in wallet protocol tables so a user wallet can't have the same protocol twice for the same wallet :white_check_mark: 
* ~check if HTTP protocols to send and recv (LNbits, Blink, Phoenixd) can be abstracted away via one WalletSendHTTP table without resorting to JSON columns~ :x: won't do in this PR, maybe in the future
* check if trigger to check wallet support can be unified :white_check_mark: 
* wallet migration :white_check_mark: 
  * wallet seed :white_check_mark: 
  * eliminate duplicate wallets before migration (#2163) :white_check_mark: 
  * migrate wallets :white_check_mark: 
  * wallet seed with multiple wallets (LNbits, NWC, ...) per user :white_check_mark: 
  * put send/receive protocols into the same user wallet if exists :white_check_mark: 
  * ~verify wallet migration with SQL assertions after migration~ :x: decided not to, too much work
  * add trigger to update JSON in `ProtocolWallet` row :white_check_mark: 
    * ~split JSON column into two columns for send and receive?~ :x: no longer needed since we switched to `ProtocolWallet`
    * resolve vault entries in JSON trigger :white_check_mark: 
  * drop old tables, triggers etc. :white_check_mark: 
  * keep ids stable during migration to make new foreign keys work :white_check_mark: 
* figure out why `Withdrawal` foreign key fails :white_check_mark: fixed it, but didn't really figure out what happened
* **Q:** Are there more edge cases with our old wallets that might break migration? :white_check_mark: No, I don't think so.
* **Q:** add `priority` to `ProtocolWallet`, too? :white_check_mark: No, I don't think that's important and we can change it later.
* verify wallet support :construction: 
* update code to save/load wallets :construction: 
* rename `walletId` foreign keys that point to `UserWallet` to `userWalletId` to avoid confusion? :thinking:
* ~add `User(id, vaultKeyHash)` foreign key to `Vault` table with new `userId`, `keyHash` columns?~ :x: out of scope of this PR

## Additional Context

This is based on the latest discussion with @huumn how to plan further ahead with the wallet schema than to just migrate the vault in #2092. 

## Checklist

tbd

tbd

n/a

no


ekzyis:
See https://github.com/stackernews/stacker.news/pull/2092#issuecomment-2888730116

Issue #2160, opened 2025-05-13
Some users have duplicate wallets


kepford:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

I noticed this a week ago but found it was more broad than I first realized. 

There seems to be two issues but they are likely related.

1. When I update my Lightning Address to a new address the rewards payouts are sent to the previous address
2. When I disable Lightning Address and set up NWC receiving payments are still sent to the Lightning address.

### Steps To Reproduce

# Issue 1 Rewards Paid to old address

1. If you do not have a lightning address set up set one up
2. Ensure you are being paid to this address
3. Now change the address to a different address and you should see that the daily reward is paid to the first address not the new address

# Issue 2 Lightning Address Used When Disabled

1. Ensure you have a lighting address set up and enabled
2. Set up another method for receiving payments like NWC
3. Disable the Lightning Address but do not delete the address.
4. Receive a payment and see that it still goes to the lightning address and not the NWC wallet.

### Expected behavior

It is expected that when you change your lightning address that the new address should be used for all payments and the old one should not be used for rewards or any zaps. If the lightning address is disabled it should not be used.

kepford:
One note.

I think this is a bug in that the behavior is not as expected but I deleted the address and detached the wallet and the behavior for payments (no rewards) went away. Meaning I received a payment to NWC wallet and not the lightning address wallet.

But, disabling the wallet should result in that wallet not being used. Right?

ekzyis:
Thanks for the report!

@huumn, I am pretty sure this is related to duplicate wallets. I see that kepford has two lightning addresses so on save, only one of them is updated and the old one still exists. It's possible we're now trying both of them with the old one first which succeeds first.

I was working on a migration to fix this as part of #2146, but I will ship it as a separate PR that we can deploy independently.

Since this won't fix the issue why it's even possible to create duplicate wallets, I will just add the same migration to #2146.

kepford:
That sounds right @ekzyis. See my comment above. But the enabled check box seems to not be respected.

kepford:
> Yes, but we had or have a bug where a user might create the same wallet multiple times but you will only ever see one of them and saving a wallet only updates one of them.

Ah!!

ekzyis:
I see you deleted both your lightning addresses now. That is also a way to fix it haha

kepford:
[...]

PR #2163, opened 2025-05-14
Delete duplicate wallets

ekzyis:
## Description

Temporary fix for any user affected by #2160.

## Additional Context

This does not fix the root cause so it might still be possible to create duplicate wallets. I know that #2059 made it possible to save a lightning address multiple times because if the user was hiding their wallet badges, the prompt would still show up. This was fixed in #2134.

However, I believe creating multiple/duplicate wallets per type was possible even before #2059.

This also doesn't check that the wallets are indeed duplicates (same values), but in that case, we would need to decide ourselves which wallet we're going to keep per user anyway. I decided to keep the wallet that was most recently updated.

The root cause will be fixed via #2092, #2146 and related changes.

We can also run this query manually, we don't need to deploy this as a migration.

## Checklist

yes, they only delete wallets but don't change any code

`4`. I did not test this with production data, but I made sure that only the most recent updated wallet per type is kept per user.

n/a

no

### DIFF
--- a/prisma/migrations/20250513212337_delete_duplicate_wallets/migration.sql
+++ b/prisma/migrations/20250513212337_delete_duplicate_wallets/migration.sql
@@ -1,0 +1,14 @@
+-- delete duplicate wallets per user
+WITH duplicates AS (
+    SELECT
+        "userId",
+        "type",
+        (array_agg(id ORDER BY updated_at DESC))[2:] AS id
+    FROM "Wallet"
+    GROUP BY "userId", "type"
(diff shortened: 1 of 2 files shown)


huumn:
A temporary fix is to probably create a unique index on `userId, type` ... it'll lead to weird error messages but it's probably better than whatever we have currently.

ekzyis:
> A temporary fix is to probably create a unique index on `userId, type` ... it'll lead to weird error messages but it's probably better than whatever we have currently.

Very good idea, done!

PR #2169, opened 2025-05-18
Wallet v2

### LINKED ISSUE #1495: redesign of attached wallet page

ekzyis:
## Description

Close #1495 

This will:

- refactor vault
  - backend (#2092) :white_check_mark:
  - frontend (encryption and decryption) :construction:
- store send and receive config in separate tables (#2146) :white_check_mark: 
- support multiple receive and send protocols per wallet
  - backend (#2146) :white_check_mark:
  - frontend :construction: 
- separate wallets from protocols for better UI/UX/DX
  - backend (#2146)) :white_check_mark: 
  - frontend :construction: 
- create a list of wallets with the protocols they support
  - backend (#2146)) :white_check_mark: 
  - frontend :construction: 
- store all wallet logs on the server :construction:
- let users search for their wallet :construction: 

## Screenshots

tbd

## Additional Context

tbd

## Checklist

tbd

tbd

tbd

tbd


ekzyis:
> ### Bug: Hook Accesses Undefined URL Parameters
> The `useSendRecvParam` and `useWalletProtocolParam` hooks directly access `params.slug[1]` and `params.slug[2]` respectively. This can cause a `TypeError` if `params.slug` is `undefined` or lacks sufficient elements for the current URL path.
> 

`params.slug` will never be undefined in this context.

But this reminded me that I wanted to refactor the hooks in wallets/client/components/forms.js. I think I went a bit overboard with small hooks. I think I can merge at least the ones that wrap `usePathname` or `useParams`.

ekzyis:
I rebased this PR on 17aada6d to fix conflicts but I will stop rebasing now because @huumn is reviewing

huumn:
Feel free to merge updates, but rebasing or forced pushes will make me grumpy. :D

huumn:
So far this looks/works great - strict upgrade over v1. I've only done a very cursory QA/code review so far. I'll continue with my surface level exploration today, then go deep tomorrow on the important stuff:

1. db models
2. migrations
3. general organization of code/calls/etc

# Questions so far

1. have you written a doc on how to add new wallets/protocols? it'll help me understand the code better and it'll be nice to have for anyone that wants to extend the code.
2. do we need `templateId` if each template has a unique name? It seems to make `wallets.json` a bit wonky having to know the insertion id (we have templateId, name, displayName) ... and if it's not an auto-incremented int, it seems like we should just use the unique template name. Perhaps it'd make sense to use an enum for template names too, and make that the primary key?
    - "names as primary keys" are discouraged because names beg to be changed and sometimes that's inconvenient but, in this case, I don't expect these names to change and we're referencing the templateId as if it has a known mapping to wallet name already ... also, if we do `UPDATE ON CASCADE`, these names should be pretty convenient to change (in the rare case where we need to) ... else we can deprecate a template and migrate to one with the new name
    - idk maybe I'm missing something though. do you have examples where using `templateName` instead of `templateId` will cause problems? I can only think of `BLINK` changing its name to `SQUINT` or something. What does that break when we change the name? Do we have to change the template name in that case?

ekzyis:
> have you written a doc on how to add new wallets/protocols? it'll help me understand the code better and it'll be nice to have for anyone that wants to extend the code.

I have not but I will do so today!

> It seems to make wallets.json a bit wonky having to know the insertion id

<details>
<summary>answer</summary>

Oh, I only added `templateId` to wallets.json to migrate existing wallets to their correct template. We can remove it afterwards.

You can check this by checking where `walletTemplateId` is called:

https://github.com/stackernews/stacker.news/blob/b69badaa7631c2f21fd66ec1f2ea8074f735ffe9/wallets/lib/util.js#L24-L26

It's only called by `getWalletTemplateId`:

https://github.com/stackernews/stacker.news/blob/b69badaa7631c2f21fd66ec1f2ea8074f735ffe9/wallets/client/hooks/query.js#L483-L497

which is only called in the `useWalletMigrationMutation` hook:

https://github.com/stackernews/stacker.news/blob/b69badaa7631c2f21fd66ec1f2ea8074f735ffe9/wallets/client/hooks/query.js#L416-L424

Other than that, we only care about the template id in the `useWalletProtocolUpsert` hook, but we simply assign the id that the API returned to us before in the `WALLET`  query to the mutation variables; there's no lookup:

https://github.com/stackernews/stacker.news/blob/b69badaa7631c2f21fd66ec1f2ea8074f735ffe9/wallets/client/hooks/query.js#L153-L157

However, I remember that I considered using the name because the ids are opaque. This makes the API awkward to use because to create a new wallet, you need to have looked up the templates before to know which id to use. To not be blocked by what I considered a minor decision, I simply postponed a final decision on this.

---

</details>

> Perhaps it'd make sense to use an enum for template names too, and make that the primary key?

<details>
<summary>answer</summary>

Mhh, yeah! You can't easily remove values from an enum, but adding values is easy, which is only what we need, so I don't see an issue with that.

> Although enum types are primarily intended for static sets of values, there is support for adding new values to an existing enum type, and for renaming values (see [ALTER TYPE](https://www.postgresql.org/docs/16/sql-altertype.html)). Existing values cannot be removed from an enum type, nor can the sort ordering of such values be changed, short of dropping and re-creating the enum type.

-- https://www.postgresql.org/docs/16/datatype-enum.html

---

</details>

> idk maybe I'm missing something though. do you have examples where using `templateName` instead of `templateId` will cause problems?

<details>
<summary>answer</summary>

Not a big problem, but if we use names as the primary key, at least the GraphQL field of type `ID!` for the `WalletTemplate` type would need to still be called `id` and not `name` because `name` is already of type `String!` in the `Wallet` type and we query both types together in the `WALLETS`  query. This would lead to the same conflict I had with plaintext vs encrypted credentials. For example, `WalletSendNWC` has an `url` field but of type ` VaultEntry!` and `WalletRecvNWC` has an `url` field but of type `String!`, so I had to use field aliases in the query that returns send and receive wallets:

https://github.com/stackernews/stacker.news/blob/b69badaa7631c2f21fd66ec1f2ea8074f735ffe9/wallets/client/fragments/wallet.js#L17-L32

see https://github.com/graphql/graphql-js/issues/53

As mentioned, not a big problem, so we can still change it, just wanted to mention that this is where GraphQL makes it (a little) awkward.

Let me know if you want me to do so! I think it would be a good change just because of making our API easier to use (even though only we are using it, and that will probably continue to be the case for the foreseeable future).

</details>

> I can only think of BLINK changing its name to SQUINT or something. What does that break when we change the name? Do we have to change the template name in that case?

<details>
<summary>answer</summary>

If we update the name in our database, we only need to update wallets.json to reflect that, because that's what the API will return and is what we're using in the client to lookup the display name, image etc.

Btw, I considered not having wallets.json and storing everything in the database, but that would mean that even a simple update requires writing a migration.

</details>

wallets/client/components/passphrase.js
```diff
@@ -0,0 +1,61 @@
...
+  const updateWalletEncryption = useWalletEncryptionUpdate()
+
+  useEffectOnce(() => {
+    async function updateKey () {
+      const { passphrase, key, hash } = await generateRandomKey()
+      await updateWalletEncryption({ key, hash, wallets })
+      setPassphrase(passphrase)
+    }
+    updateKey()
+  }, [])
```
huumn: `useEffect` in strict mode (which is only on in dev) runs twice so you can detect when you're doing things you should not be doing in `useEffect` ... like doing something that isn't idempotent.

ekzyis: Mhh yes, but I think strict mode was messing with development too much here.

But I don't remember if it broke something or it was just an unnecessary API call.

Should I not generate the new passphrase and run the migration in an `useEffect`?

I guess I could generate the passphrase and run the wallet encryption in the `onShow` callback of `useShowPassphrase` and then just pass `passphrase` to the `Passphrase` component.

ekzyis:
remaining TODOs:

- [x] wallets/README.md (see chat)
- [x] use `name` as primary key of `WalletTemplate` (https://github.com/stackernews/stacker.news/pull/2169#issuecomment-3054302522)
- [x] merge Alby wallets into one (see chat)
- [x] fix deleting wallet template logs
- [x] test client wallet migration without IndexedDB (see chat)
- [x] show message if wallets unavailable because IndexedDB unavailable
- [x] don't work around strict mode in `Passphrase` component (https://github.com/stackernews/stacker.news/pull/2169#discussion_r2199034717)
- [ ] ~fix LND wallet card slightly bigger~ (follow-up PR)
- [x] check if local wallet migration creates duplicates because of strict mode
- [ ] ~handle query error, for example if anon visits /wallets~ (follow-up PR)
- [x] use constraint triggers if possible (https://github.com/stackernews/stacker.news/pull/2169#discussion_r2201553805)

wallets/client/components/passphrase.js
huumn: Anything that doesn't require this hack, and isn't a hack itself, is probably better. I think that's why strict mode exists - prevent devs from doing stuff based on the component lifecycle that they shouldn't be doing.

prisma/migrations/20250702000001_wallet_v2/migration.sql
```diff
@@ -0,0 +1,1126 @@
...
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER wallet_check_support
+    BEFORE INSERT OR UPDATE ON "WalletSendNWC"
+    FOR EACH ROW
+    EXECUTE FUNCTION wallet_check_support('SEND', 'NWC');
```
huumn: Given that these are constraints, it might be worth declaring them as [constraint triggers](https://www.postgresql.org/docs/16/sql-createtrigger.html):

> When the CONSTRAINT option is specified, this command creates a constraint trigger. This is the same as a regular trigger except that the timing of the trigger firing can be adjusted using [SET CONSTRAINTS](https://www.postgresql.org/docs/current/sql-set-constraints.html). Constraint triggers must be AFTER ROW triggers on plain tables (not foreign tables). They can be fired either at the end of the statement causing the triggering event, or at the end of the containing transaction; in the latter case they are said to be deferred. A pending deferred-trigger firing can also be forced to happen immediately by using SET CONSTRAINTS. Constraint triggers are expected to raise an exception when the constraints they implement are violated.

Afaict the main benefit is being able to defer them until the end of a transaction, but I'd guess that's an option because it can be handy in certain situations. It might also produce more appropriate error messages.

huumn: We'd also have to update these to be `AFTER` triggers, which might mess with some of the sequence manipulations, but idk.

Anyway, just something I saw while reviewing the docs.

ekzyis:
Regarding duplicate wallets during migration of local wallets:

I tested this with and without strict mode enabled, these are the results:

```
strict mode:     ..x.x.x..x..x.xx..x 
w/o strict mode: ...................
```
(x means duplicate wallet was created)

I think it doesn't always happen with strict mode because it depends on when the wallets query returns results.

But I think the results are quite clear: it happens because of strict mode, so it shouldn't happen in prod.

huumn:
> Regarding duplicate wallets during migration of local wallets:

This happens because migration is triggered as part of components loading, right? I wonder if there's a way to this kind of thing without hacks.

ekzyis:
> This happens because migration is triggered as part of components loading, right?

Yes

> I wonder if there's a way to this kind of thing without hacks.

The only guidance I found in the [React docs regarding this](https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development) are the following sections:

- [Fetching data](https://react.dev/learn/synchronizing-with-effects#fetching-data) suggests to add a cleanup function but this doesn't help us because it still sends two requests, it just discards the second response
- [Sending analytics](https://react.dev/learn/synchronizing-with-effects#sending-analytics) suggests to ignore duplicate requests in dev because they won't happen in prod
- [Not an Effect: Buying a product](https://react.dev/learn/synchronizing-with-effects#not-an-effect-buying-a-product) suggests using a button instead of an effect

Since I don't think we want to make the migration depend on a button click, I think we're stuck with ignoring duplicate wallet migration in dev with strict mode enabled.

huumn:
> Since I don't think we want to make the migration depend on a button click, I think we're stuck with ignoring duplicate wallet migration in dev with strict mode enabled.

The advice in this section seems applicable: https://react.dev/learn/synchronizing-with-effects#not-an-effect-initializing-the-application

This might require refactoring migration away from being a hook though.

huumn:
It might not be worth it. I know a lot of the calls are hooks for code reusability.

ekzyis:
> It might not be worth it. I know a lot of the calls are hooks for code reusability.

Yeah I'd have to fetch `ME`, `WALLETS` and decrypt the wallets manually etc.

However, I wonder if I could mount a button and then click it once using this approach ... but I think that's also not worth it.

It's annoying in dev, but I am pretty sure this won't happen in prod.

ekzyis:
Regarding 

>  test client wallet migration without IndexedDB (see chat)

in https://github.com/stackernews/stacker.news/pull/2169#issuecomment-3059597146:

If IndexedDB is not available, `useKeyInit`  will throw, log the error and never call `setKey`:

https://github.com/stackernews/stacker.news/blob/7c8dac0b6b63909a32f41a902e6ab4f842f51ea8/wallets/client/context/hooks.js#L128-L161

This means that `useKey` will never return a key and thus `useEncryption` will never return `ready` set to true:

https://github.com/stackernews/stacker.news/blob/7c8dac0b6b63909a32f41a902e6ab4f842f51ea8/wallets/client/hooks/crypto.js#L52-L68

We check this in `useWalletMigration`:

https://github.com/stackernews/stacker.news/blob/7c8dac0b6b63909a32f41a902e6ab4f842f51ea8/wallets/client/context/hooks.js#L184-L189

So the migration will never run if IndexedDB is not available.

I think that's okay. The app doesn't break, but we don't migrate. If we can't store the key that was used for encryption, it doesn't make sense to migrate because the key will be lost afterwards anyway.

prisma/migrations/20250702000001_wallet_v2/migration.sql
ekzyis: Using constraint triggers now in a8823d83

Also noticed that it makes more sense to attach the constraint trigger to the `WalletProtocol` table (6bae7819). Migration still runs through and can still create and save wallets.

ekzyis:
I refactored the wallet reducer in 8be0b30.

Instead of storing `page` + `loading` in the state, I noticed what I actually wanted was to have `status`. This meant I could also remove router events to "reset the page" when we left /wallets. I can just use a regular button to go from 'attach wallets' to the list of wallets instead of "changing the page."

I also fixed the flicker of 'attach wallets' -> 'unlock wallets' or 'attach wallets' -> list of wallets in 8be0b303 by adding a loading spinner while we're still determining what we want to show:

localhost_3000_wallets(iPhone SE)

With this refactor, I can now easily add another status when the wallets are unavailable instead of thinking in pages in the state.

ekzyis:
I now show this message to the user if IndexedDB is not supported:

localhost_3000_wallets(iPhone SE) (2)

wallets/client/components/passphrase.js
ekzyis: > I guess I could generate the passphrase and run the wallet encryption in the `onShow` callback of `useShowPassphrase` and then just pass `passphrase` to the `Passphrase` component.

Forgot to mention that I did this in b42b6395

ekzyis:
Okay, I’m done here now. As far as I am concerned, this can be merged!

huumn:
# Edit: I fixed this.

When I attempt to migrate prod seed:

```
Migration name: 20250702000001_wallet_v2

Database error code: 22P02

Database error:
ERROR: invalid input value for enum "LogLevel": "WARN"

DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E22P02), message: "invalid input value for enum \"LogLevel\": \"WARN\"", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("enum.c"), line: Some(129), routine: Some("enum_in") }
```

huumn:
On brand new accounts, I'm being asked to unlock wallets

ekzyis:
> On brand new accounts, I'm being asked to unlock wallets

Ohhh, good catch! There is a race condition in `useKeyInit`. The `useEffect` is called twice (not sure yet if that's also caused by strict mode), so if there's no key yet, we _might_ call `generateRandomKey`twice. The server saves the first key, whereas the client overwrites it with the second key, leading to a key mismatch.

---

**Update**

Pretty sure it's because of strict mode because the second `console.log` is dimmed:

> If you have [React DevTools](https://react.dev/learn/react-developer-tools) installed, any `console.log` calls during the second render call will appear slightly dimmed. React DevTools also offers a setting (off by default) to suppress them completely.

```diff
diff --git a/wallets/client/context/hooks.js b/wallets/client/context/hooks.js
index 99280325..025feb9b 100644
--- a/wallets/client/context/hooks.js
+++ b/wallets/client/context/hooks.js
@@ -121,6 +121,8 @@ export function useKeyInit () {
   const loadKey = useLoadKey()
   const loadOldKey = useLoadOldKey()

+  console.log('useKeyInit')
+
   useEffect(() => {
     if (!me?.id) return
```

I read [this](https://marmelab.com/blog/2023/01/11/use-async-effect-react.html) now. As mentioned in that link, this is how we can deal with strict mode issues in development: 

```js
useEffect(() => {
  let mounted = true
  async function effect() {
    await Promise.resolve()
    if (!mounted) return
    // the following code will only run after first cleanup in strict mode
    ...
  }
  effect()
  return () => { mounted = false }
}, dependencies)
```

I've tested this by logging `mounted` before and after `await Promise.resolve()` and mounted does indeed always change from true to false in the first effect.

So by calling `await Promise.resolve()` in the effect before checking `mounted`, we can be sure that we've waited for an immediate cleanup.

This looks like a hack / working against React Strict Mode instead of with it, but I think this is basically [_Fetching data with Effects_](https://react.dev/reference/react/useEffect#fetching-data-with-effects) with waiting for the cleanup before triggering the request, instead of only ignoring the response. So imo, doing this is fine when a bug is really only caused by strict mode and we know that there's no bug it's trying to reveal.

---

**Update**

Mhh, it actually is trying to reveal a bug: the code is prone to race conditions. To fix that, we would need to read and write the key in one transaction in a way that the second transaction does not read the key before we wrote it. Will check if IndexedDB has such functionality.

huumn:
IME: let strict mode help you. Do not hack around it. 

It's particularly tricky with this stateful/sideeffect initialization stuff. We need to learn a decent pattern around it imo.

ekzyis:
> IME: let strict mode help you. Do not hack around it.
> 
> It's particularly tricky with this stateful/sideeffect initialization stuff. We need to learn a decent pattern around it imo.

Yeah, I agree. I am using a single transaction in `readwrite` mode to read and write the value now in 9141050.

Since I open and close a database for each call in my refactored `useIndexedDB` hook, I actually couldn't use it because each transaction would run in a separate connection so `readwrite` mode wouldn't block each other as mentioned [here](https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction). So I wrote custom IndexedDB code in `useKeyInit` so I can have the same database and run two queries within the same transaction ...

But the issue doesn't happen now anymore for another reason: Since I now have to check if the db is set before I run the queries and since afaict, the queries only run after strict mode went through "init db -> cleanup -> init db", the queries don't run twice at the same time anymore.

I think I will have to rethink how to wrap IndexedDB properly in a hook so it doesn't limit us in important ways. Or just use a library at this point.

huumn:
The original `useIndexedDB` was basically a light version of what a library would provide, scoped for our needs.

Issue #2164, opened 2025-05-14
Autopopulate "title" field when posting a Youtube link with the title of the video rather than with "- Youtube"


SouthKoreaLN:
### Describe the problem you're trying to solve

When posting a Youtube link, currently, it autopopulates the _title_ field using _- Youtube_.

### Describe the solution you'd like

It'd be nice if it fetched the title of the video instead.

### Describe alternatives you've considered

Currently, I have to copy-paste it by hand.

huumn:
This is still broken in prod for some reason.

ekzyis:
> This is still broken in prod for some reason.

Mhh, works locally though.  Tested with [this video](https://www.youtube.com/watch?v=AHRtMxVrP78).

I noticed it only works on first try after page refresh though. `overrideValue` doesn't override the value when`overrideValue` didn't change since the last render. So when we fetch the title, override the value, then clear the value and then fetch the same title again, it does not override the value anymore.

---

update: Okay, I see what you mean, it still populates `- YouTube` in prod. It's as if the rule isn't applied. [Maybe related to it being added as top-level code](https://github.com/stackernews/stacker.news/pull/2171#discussion_r2188490093) which has different behavior in a prod build?

huumn:
Still not working in prod

ekzyis:
It worked for https://www.youtube.com/watch?v=pgNCgJG0vnY:

https://github.com/user-attachments/assets/19b41974-5d35-42d9-9089-253eb46b237b

but still doesn't work for https://www.youtube.com/watch?v=AHRtMxVrP78:

https://github.com/user-attachments/assets/1934cd64-a6a5-464b-88e3-c2ef54738aca

huumn:
you're right, that one does work for me. I wonder why it failed for others for me. I'll close until I find a counterexample.

ekzyis:
I edited my comment, this is a counterexample: https://www.youtube.com/watch?v=AHRtMxVrP78

ekzyis:
Another example regarding localhost vs prod: https://www.youtube.com/watch?v=mvuWLob3CFU&t=817s

https://github.com/user-attachments/assets/73e1cd5d-3b72-4d3f-b634-99386dffab2a

So weird, there are [differences between links](https://github.com/stackernews/stacker.news/issues/2164#issuecomment-3114272502) _and_ prod vs dev

caydyan:
[...]

PR #2171, opened 2025-05-19
Auto-populate YouTube video titles server-side and enforce title length.

brymut:
## Description
Auto populate the title field if youtube link is used in the Link Form. 
Closes https://github.com/stackernews/stacker.news/issues/2164

## Screenshots

https://github.com/user-attachments/assets/a452dd57-d88c-4ed0-b0d7-804599c3c9fb

## Checklist

yes
10

yes

no

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -28,6 +28,8 @@ import { verifyHmac } from './wallet'
 import { parse } from 'tldts'
 import { shuffleArray } from '@/lib/rand'
 
+metadataRuleSets.title.rules.unshift(['h1 > yt-formatted-string.ytd-watch-metadata', el => el.getAttribute('title')])
+
 function commentsOrderByClause (me, models, sort) {
   const sharedSortsArray = []
   sharedSortsArray.push('("Item"."pinId" IS NOT NULL) DESC')


ekzyis (commented):
Have you tried updating how the existing code works instead of writing custom code?

You can get the title of a YouTube video like this:

```js
document.querySelectorAll('yt-formatted-string.ytd-watch-metadata')[0].getAttribute("title")
```

ekzyis:
> instead of rebuilding the whole page 

I know it feels weird, but this is how we can fetch titles for any site since not every site provides an API like YouTube. Maintenance is more important than performance here. Or is there something special about YouTube, which is why we shouldn't or can't use the generic solution?

> hard coding the query selector

That's also how the library that we're using works: [`page-metadata-parser`](https://www.npmjs.com/package/page-metadata-parser)

> But if that's the solution you prefer, I can change it.

Yes, please. I suspect it will require only changing a single line of code.

api/resolvers/item.js
```diff
@@ -28,6 +28,8 @@ import { verifyHmac } from './wallet'
 import { parse } from 'tldts'
 import { shuffleArray } from '@/lib/rand'
 
+metadataRuleSets.title.rules.unshift(['h1 > yt-formatted-string.ytd-watch-metadata', el => el.getAttribute('title')])
```
ekzyis, line 31: I don't think this should have been added here as top-level code.

Maybe that's related to https://github.com/stackernews/stacker.news/issues/2164#issuecomment-3042286949.

ekzyis: ~Or nvm, I don't think it's related but~ I don't like to have this here as top-level code because I am not sure how often this line of code runs ([or how it behaves in prod](https://github.com/stackernews/stacker.news/issues/2164#issuecomment-3042312357)). I think we could have done this in a variable local to `pageTitleAndUnshorted`.

ekzyis: This adds the rule multiple times in dev.

PR #2180, opened 2025-05-25
custom domains - auth sync

Soxasora:
## Description

Part of #1942
Focuses on synchronizing authentication between SN and a custom domain.
It features a new `sync` endpoint that checks if there's a session and if so, redirects to the custom domain with an ephemeral token that gets refreshed with an official token.

## Screenshots
TBD

## Additional Context
TBD

## Checklist

**Are your changes backwards compatible? Please answer below:** Yes, only engages on custom domains

WIP

n/a

n/a


pages/api/auth/sync.js
```diff
@@ -117,14 +117,14 @@ function handleNoSession (res, domainName, state, redirectUri, signup = false) {
...
+async function createVerificationToken (token, state) {
   try {
     // a 5 minutes verification token using the session token's user id
     const verificationToken = await models.verificationToken.create({
       data: {
         identifier: token.id.toString(),
-        // store csrf token with the verification token, to prevent CSRF attacks
-        token: `${randomBytes(32).toString('hex')}|${csrfToken}`,
+        // store encoded state JWT with the verification token to prevent CSRF attacks
+        token: `${randomBytes(32).toString('hex')}|${state}`,
```
Soxasora: Too much, switched to token in URL as it shouldn't be logged.
<details>
  <summary>Tie a CSRF token to a verification token</summary>
There were some obstacles because of the cross-domain nature, but I think I found a middle ground to safely use CSRF to tie the verification token to the user.
Knowing that:
- messing with next auth is not preferable and thus can't use its protections
- we need to protect the token from being stolen
- middleware doesn't support node
We have another option, which is to **encode** the CSRF token in a _state JWE_ and store it alongside the verification token.

When going to the sync endpoint, we
- detect login -> encode CSRF token in state JWE -> /api/auth/sync/?...&state=encodedState

The sync endpoint will then
- create a token like token|encodedState -> redirect back with just the token

Middleware at this point will
- POST /api/auth/sync with body: {token, csrfToken} -> retrieve the verification token object from DB -> decode the encodedState -> match decodedState with csrfToken

If everything is okay, it returns the final ephemeral JWT.

Everything happens blazingly fast, if someone steals the verification token and/or the state, they can’t go that far without the CSRF token that sits and never moves in the user’s browser (well, except for the POST, but that's secure).

<hr>

side note: still checking for less hacky ways/can make it less hacky
</details>

pages/api/auth/sync.js
```diff
@@ -0,0 +1,189 @@
...
+      const newVerificationToken = await createVerificationToken(sessionToken)
+      if (newVerificationToken.status === 'ERROR') {
+        return res.status(500).json(newVerificationToken)
+      }
+
+      // STEP 5: redirect to the domain with the verification token
+      return redirectToDomain(res, domain, newVerificationToken.token, redirectUri)
+    }
+  } catch (error) {
+    return res.status(500).json({ status: 'ERROR', reason: 'auth sync broke its legs' })
```
Soxasora, line 77: At the moment, the handler doesn't expose any **fatal** error, instead it returns this joke reason!
I'm divided by logging the failed, fatal, attempt in DB or just print it.

ekzyis: > At the moment, the handler doesn't expose any fatal error, instead it returns this joke reason!

If this is an error message that a user might see, I'd personally prefer to keep it serious because they might not be in the mood for jokes when they can't login :eyes:

> I'm divided by logging the failed, fatal, attempt in DB or just print it.

I think you can just print it as we do with most errors

pages/api/auth/sync.js
```diff
@@ -0,0 +1,189 @@
...
+    if (req.method === 'GET') {
+      // STEP 1: check if the domain is correct
+      const { domain, redirectUri, signup } = req.query
+      // domain and a path redirectUri are required
+      if (!domain || !redirectUri?.startsWith('/')) {
+        return res.status(400).json({ status: 'ERROR', reason: 'domain and a correct redirectUri are required' })
+      }
+
+      // STEP 2: check if domain is valid and ACTIVE
+      const domainValidation = await checkDomainValidity(domain)
```
ekzyis, line 50: I don't know how important exactly this is, but afaict, you're only checking if the domain parameter is valid; what about the redirect uri?

Soxasora: I found that a URL with `domain` and path (`redirectUri`) separated was nicer to read, so at the end I combine the two after separate validation. Even though the validation for `redirectUri` is just checking if it starts with `/`.
...
I'm pushing a regex to only allow conformant paths.

On the domain validation, the goal is to only allow ACTIVE custom domains to create/receive a verification token, which I think it's necessary to limit requests. I also just remembered that the POST is open...
Maybe a safer way to limit external access is to present encrypted params 🤔

Edit: In some of my sketches I used to present everything the endpoint would need under an encrypted state JWT, it worked pretty well!

Issue #1409, opened 2024-09-17
Territory subscription management UI


SatsAllDay:
**Is your feature request related to a problem? Please describe.**
I'd like a single place to see all the territories to which I'm subscribed, so I can easily review and revise my subscriptions. Similar to the user settings > subscriptions page, where you manage subscriptions to individual stackers. 

**Describe the solution you'd like**

I guess I kinda put that in the above section. 

**Describe alternatives you've considered**
Looking through all territories and seeing if I'm subscribed, one at a time. 

**Additional context**
I imagine this could be just like the user subs/mutes tabs on the settings page today.

SatsAllDay:
woo!

PR #2191, opened 2025-05-29
Add Territory Sub management tab in Subscriptions

brymut:
## Description

Add Territory sub management in the 'subscriptions' tab in the settings page with a dropdown to switch between stackers & territories. With dropdown menu to unsubscribe from & mute territory.

Closes https://github.com/stackernews/stacker.news/issues/1409

### Desktop

https://github.com/user-attachments/assets/81f4be92-abf6-4b9f-ba77-4e0190682b31

### Mobile

https://github.com/user-attachments/assets/159cba7c-59c7-4046-bd15-1998defe6ff0

## Additional Context

## Checklist

Y

10

Y

No

### DIFF
--- a/api/resolvers/sub.js
+++ b/api/resolvers/sub.js
@@ -157,26 +157,19 @@ export default {
           COALESCE(floor(sum(msats_spent)/1000), 0) as spent,
           COALESCE(sum(posts), 0) as nposts,
           COALESCE(sum(comments), 0) as ncomments,
-          CASE WHEN ss."userId" IS NOT NULL THEN TRUE ELSE FALSE END as "meSubscription",
-          CASE WHEN ms."userId" IS NOT NULL THEN TRUE ELSE FALSE END as "meMuteSub"
+          ss."userId" IS NOT NULL as "meSubscription",
+          ms."userId" IS NOT NULL as "meMuteSub"
         FROM ${viewGroup(range, 'sub_stats')}
         JOIN "Sub" on "Sub".name = u.sub_name
-        LEFT JOIN "SubSubscription" ss ON ss."subName" = "Sub".name AND ss."userId" = ${me ? Number(me.id) : -1}
-        LEFT JOIN "MuteSub" ms ON ms."subName" = "Sub".name AND ms."userId" = ${me ? Number(me.id) : -1}
+        LEFT JOIN "SubSubscription" ss ON ss."subName" = "Sub".name AND ss."userId" IS NOT DISTINCT FROM $4
+        LEFT JOIN "MuteSub" ms ON ms."subName" = "Sub".name AND ms."userId" IS NOT DISTINCT FROM $4
         WHERE "Sub"."userId" = $3
           AND "Sub".status = 'ACTIVE'
         GROUP BY "Sub".name, ss."userId", ms."userId"
         ORDER BY ${column} DESC NULLS LAST, "Sub".created_at ASC
-        OFFSET $4
-        LIMIT $5
-      `, ...range, user.id, decodedCursor.offset, limit)
-
-      if (!me) {
-        subs.forEach(sub => {
-          sub.meSubscription = false
-          sub.meMuteSub = false
-        })
-      }
+        OFFSET $5
+        LIMIT $6
+      `, ...range, user.id, me?.id, decodedCursor.offset, limit)
 
       return {
         cursor: subs.length === limit ? nextCursorEncoded(decodedCursor, limit) : null,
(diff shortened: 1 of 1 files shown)


Soxasora:
Hi brymut, I see you're working on this enhancement/feature.
I wanted to give you a tip to avoid linting problems, we have a Development Tips section that also contains [tips for linting](https://github.com/stackernews/stacker.news?tab=readme-ov-file#linting), such as using the [StandardJS VSCode Extension](https://marketplace.visualstudio.com/items?itemName=standard.vscode-standard) to lint as you code. Take a look! ^^

pages/settings/subscriptions/index.js
```diff
@@ -1,30 +1,72 @@
...
+    getGetServerSideProps({ query: MY_SUBSCRIBED_USERS, authRequired: true })(ctx),
+    getGetServerSideProps({ query: MY_SUBSCRIBED_SUBS, authRequired: true })(ctx)
+  ])
+  return {
+    props: {
+      ssrUsers: userRes.props.ssrData,
+      ssrSubs: subsRes.props.ssrData
+    }
+  }
+}
```
huumn: I'm surprised this worked.

The move here is to make a compound graphql query like `USER_WITH_ITEMS` or make this into different `.js` files to call each query only when it is needed like we do for `/top` path.

pages/settings/subscriptions/index.js
```diff
@@ -1,30 +1,72 @@
...
 
-export default function MySubscribedUsers ({ ssrData }) {
+export default function MySubscribedUsers ({ ssrUsers, ssrSubs }) {
+  const [view, setView] = useState('stackers')
   const subscribeUserContextValue = useMemo(() => ({ refetchQueries: ['MySubscribedUsers'] }), [])
```
huumn: This `refetchQueries` pattern should be used for territory subscriptions too.

huumn (changes requested):
two changes needed

brymut:
> two changes needed

Addressed

components/territory-list.js
```diff
@@ -77,6 +78,10 @@ export default function TerritoryList ({ ssrData, query, variables, destructureD
                     {sub.name}
                   </Link>
                   <Info className='d-flex'><TerritoryInfo sub={sub} /></Info>
+                  <ActionDropdown>
+                    <ToggleSubSubscriptionDropdownItem sub={sub} />
+                    <MuteSubDropdownItem sub={sub} />
+                  </ActionDropdown>
```
huumn: There are several pages with territory lists that this does not work on.

huumn (changes requested):
I fixed an number of things in this PR, some of them subjective, some of them bugs (like extra separators), then I encountered this dropdown bug.

The action dropdown works on the new territory page, but:

1. it's still there when logged out (logged out user can't subscribe/mute)
2. on the `/top/territories/forever` it thinks I'm not subscribed to any territories
3. same as (2) on `/<user nym>/territories`

brymut:
Ahh right, quick question, I'm not quite sure to handle the issues you've highlighted. 

> * on the `/top/territories/forever` it thinks I'm not subscribed to any territories

I think for this context and how the `TOP_SUBS` query is being used, we aren't using an authenticated request to the topSubs resolver, so checking for whether or not the user is authenticated wouldn't work here (at least not trivially). So I think the best approach to this issue is to mirror what the top stackers user list is doing by removing the dropdown in this context `/top/` (not passing the `nymActionDropdown` prop). What do you think?

huumn:
I think that's fine - withholding this new dropdown in contexts where we haven't yet made it work.

brymut:
> I think that's fine - withholding this new dropdown in contexts where we haven't yet made it work.

ok, I've added a `subActionDropdown` prop for contexts where the user is logged in and as for the subscribing & muting functionality in `userSubs` resolver I've added a way to check the current logged in user's relationship with each sub. This should handle both 1. & 3. Checked locally and everything seems to be in order.

api/resolvers/sub.js
```diff
@@ -166,10 +166,59 @@ export default {
...
+          sub.meSubscription = subscribedSet.has(sub.name)
+          sub.meMuteSub = mutedSet.has(sub.name)
+        })
+      } else {
+        subs.forEach(sub => {
+          sub.meSubscription = false
+          sub.meMuteSub = false
+        })
+      }
+
```
huumn: this is not the way we do this normally. when someone is logged in, this resolver now requires 3 roundtrips to the database. normally, we add `meWhatever` fetching to the main query.

pages/[name]/territories.js
```diff
@@ -25,6 +27,7 @@ export default function UserTerritories ({ ssrData }) {
           query={USER_WITH_SUBS}
           variables={variables}
           destructureData={data => data.userSubs}
+          subActionDropdown={!!me}
```
Comment (huumn) [inline pages/[name]/territories.js]: i wish there were a better way to do this. it's hard to track booleans dependency injected getting passed several components deep.

anyway, just wanted to note this for myself.

huumn (approved):
this all works well now. i think the performance/maintainability of the parts i commented on are a little sub-standard.

i'll probably delay merging until i can fix those if you don't find the time to fix them yourself.

pages/[name]/territories.js
Comment (brymut) [reply pages/[name]/territories.js]: I think I get what you're getting at. Moved the `useMe` hook directly into `TerritoryList`, seems to work fine that way without needing to track the dependency from here.

api/resolvers/sub.js
brymut: Right, not the strongest at SQL but I think the left join I've added could get to the solution you're looking for using just the main query. Was concerned at first that it was getting complex but now I think its fine

brymut:
> i'll probably delay merging until i can fix those if you don't find the time to fix them yourself.

I think I've got some solutions to your concerns, hope this covers it. If not, I understand you wanting to hold on this

Issue #2189, opened 2025-05-28
Local regtest fixup: make lnd containers sync to chain faster


sutt:
### Describe the problem you're trying to solve

Not a bug, but a worthy improvement.

Base on logs, there's wildly different sync times between the three lnd containers on regtest init:
```
- lnd:          0m:10s
- sn_lnd:       5m:40s
- router_lnd:   1m:30s
```
Given that they are all the same image (currently `polarlightning/lnd:0.18.0-beta`) ,connected to the same bitcoind, and configured relatively similar, why is this?

These are time to initiate the sync, then they all sync in 10 seconds. The laggards continue to throw the error before initiating the sync:
`- [ERR] LNWL: Unable to synchronize wallet to chain, trying again in 5s: -8: Block height out of range`

This appears to be an intermittent problem on LND reported here: https://github.com/lightningnetwork/lnd/issues/8468. Most recent proposed fix is just upgrade to v19.

Note: this appears to occurs on regtest init, not start from already sync'd. I think this is because once sync is initialized it gets cached in a db, and the blocks are produced less frequently.

### Describe the solution you'd like

Ideally all three lnd containers chain sync in under 30 seconds.

I think this would be helpful for writing tests on regtest functionality; where 30 second setup is not terrible but 6 minute setup is undesirable.

I'm going to try to monkey around and find what's making some containers so fast and others so slow.

### Additional context

This was also noted by ek in #2179:

>I've had this happen before, but not sure how I fixed it.
update: The error fixes itself after a few minutes:

ekzyis:
Wow, nice find, this was really annoying

PR #2192, opened 2025-05-29
Regtest edge-case: sn_channel_cron condtion to ensure sn_lnd conns to router_lnd

### LINKED ISSUE #2188: Local regtest fixup: create-channel race b/w eclair and router_lnd

### Describe the bug

**Problem:** channel setup procedure may not establish a channel between `sn_lnd` and `router_lnd`.
**Impact:** if this happens `stacker` won't be able to pay `sn`.

This is because the ofelia job rules passed to `channdler` have an edge case where eclair can open three channels to sn_lnd which will then prevent sn_lnd from opening any further channels, e.g. to router_lnd. 

### Current rules

- eclair attempts to open up to three channels to sn_lnd

https://github.com/stackernews/stacker.news/blob/6a29dea232ffa547497ac9e5dedf781db3c14a1f/docker-compose.yml#L607-L615

- `sn_lnd` attempts to open three channels to router_lnd, but takes no action if it _already has_ three active/pending channels. Thus if eclair "wins the race" sn_lnd will never open it nec. channels to router_lnd.

https://github.com/stackernews/stacker.news/blob/6a29dea232ffa547497ac9e5dedf781db3c14a1f/docker-compose.yml#L357-L366

### Proposed fix(es)

I had two fixes (simple and complex) in mind I can send in as PR's to prevent the race / ensure router_lnd / sn_lnd always attempt to have channels.

One question is: should eclair just connect to router_lnd instead sn_lnd? Or does eclair not like to do multi-hop routing or meant to be more a wallet of something?

### Screenshots

_No response_

### Steps To Reproduce

Run the regtest environment from scratch with `wallets` profile. Then query the pubkeys of the channels of sn_lnd:

`sudo docker exec -u lnd sn_lnd bash -c "lncli listchannels" | jq  '.channels | .[] | .remote_pubkey'`

This is the result of my most recent case I got 3 eclairs and 2 router_lnds which is OK. but I came across this problem I think when I ended up with all eclairs.

```
"02268c74cc07837041131474881f97d497706b89a29f939555da6d094b65bd5af0"
"02268c74cc07837041131474881f97d497706b89a29f939555da6d094b65bd5af0"
"02268c74cc07837041131474881f97d497706b89a29f939555da6d094b65bd5af0"
"02750991fbf62e57631888bc469fae69c5e658bd1d245d8ab95ed883517caa33c3"
"02750991fbf62e57631888bc469fae69c5e658bd1d245d8ab95ed883517caa33c3"
```

### Expected behavior

Seems like we always want at least one, and probably multiple channels between router_lnd and sn_lnd since this is core functionality. And should ensure this happens both at setup, and as channdler continues to keep the LN graph in a desired shape.

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

sutt:
## Description

Fix #2188 

This ensures that `channdler` will open and keep active three channels between `sn_lnd` and `router_lnd`. 

Currently behavior does not ensure this because `eclair` could create three active channels, and prevent any being open to `router_lnd`.

This modification keeps the useful check for .active + .pending channels and sets the max to 6. This is useful to prevent endless channels opens that get stuck in pending.

It also introduces a new condition assess how many active channels are currently open between `sn_lnd <-> router_lnd`.

## Additional Context

The edge case scenario where the bug this fixes is triggered:
- Running with profile including `wallets` - otherwise eclair wouldn't start
- `sn_lnd` syncs 3+ min before `router_lnd` - so `sn_lnd` can collaborate to open channels from eclair, but `router_lnd` can not collaborate to open channels from `sn_lnd`
  - `walletcron` will open a new channel from `eclair` to `sn_lnd`, and will do so 3 times, once each minute
  - but `channdler` will not be able to open channel from `sn_lnd` to `router_lnd`; during this time the job will fail
  - once `router_lnd` syncs and collaborate in channel opens, `channdler` will no longer initiate since it already `sn_lnd` will already have 3 active channels.
- This multi-minute variable sync time seems to be a real issue (see #2189).

## Checklist

Yes. 
Only caveat: the use of `--peer` flag on `lncli listchannels` which might not be available in older version of LND.

7. 

I have QA'd, by running:

```bash
sudo docker exec -u lnd sn_lnd lncli getinfo | jq .num_active_channels
# 6 
# (as expected)

sudo docker exec -u lnd sn_lnd bash -c "lncli listchannels" | jq  '.channels | .[] | .remote_pubkey'
# "02268c74cc07837041131474881f97d497706b89a29f939555da6d094b65bd5af0"
# "02268c74cc07837041131474881f97d497706b89a29f939555da6d094b65bd5af0"
# "02268c74cc07837041131474881f97d497706b89a29f939555da6d094b65bd5af0"
# "02750991fbf62e57631888bc469fae69c5e658bd1d245d8ab95ed883517caa33c3"
# "02750991fbf62e57631888bc469fae69c5e658bd1d245d8ab95ed883517caa33c3"
# "02750991fbf62e57631888bc469fae69c5e658bd1d245d8ab95ed883517caa33c3"
# (3 channels to router_lnd, 3 channels to eclair, as expected)
```
I also manually closed a channel and channdler re-opened a new one.

No


docker-compose.yml
```diff
@@ -356,7 +356,9 @@ services:
       ofelia.job-exec.sn_channel_cron.schedule: "@every 1m"
       ofelia.job-exec.sn_channel_cron.command: >
         su lnd -c bash -c "
-          if [ $$(lncli getinfo | jq '.num_active_channels + .num_pending_channels') -ge 3 ]; then
+          if [ $$(lncli getinfo | jq '.num_active_channels + .num_pending_channels') -ge 6 ]; then
+            exit 0
```
ekzyis, line 360: nitpick: Do we really still need this check? You mentioned this:

> This modification keeps the useful check for .active + .pending channels and sets the max to 6. This is useful to prevent endless channels opens that get stuck in pending.

But sn_lnd only opens channel to router_lnd so the check below will make sure we stop after three channels, no?

ekzyis (approved):
LGTM, thanks for the fix!!

Only left one nitpick comment, but it's also okay as it is.

docker-compose.yml
sutt: Yes good catch, but I think we still want that check.

This is because `lncli listchannels` does not show pending channels. So the `lncli getinfo` still gives the us that information which is a valuable check to prevent endless "WhileLoop" style channel opens if channels opens moving from pending -> active start taking more than the one minute between the scheduled channel open jobs.

ekzyis: Ah ok, makes sense!

PR #2195, opened 2025-05-30
fix duplicate search highlights

### LINKED ISSUE #2181: Search highlights are duplicated
https://stacker.news/items/988711/r/k00b

abhiShandy:
## Description

fixes #2181 

Avoid including both exact matches and item text in search results.

## Screenshots

Before
before

After
after

## Additional Context

## Checklist

Yes

10

Tested on laptop only

No

### DIFF
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -510,8 +510,8 @@ export default {
         item.searchTitle = e.highlight?.title?.[0] || e.highlight?.['title.exact']?.[0] || item.title
 
         // prefer the exact highlight for text
-        const searchTextHighlight = [...(e.highlight?.['text.exact'] || []), ...(e.highlight?.text || [])]
-        item.searchText = searchTextHighlight?.slice(0, 5)?.join(' ... ')
+        const searchTextHighlight = e.highlight?.['text.exact'] || e.highlight?.text || []
+        item.searchText = searchTextHighlight?.join(' ... ')
 
         return item
       })


ekzyis (approved):
Change to fix duplicate search highlights makes sense to me, but I didn't test

api/resolvers/search.js
```diff
@@ -510,7 +510,7 @@ export default {
         item.searchTitle = e.highlight?.title?.[0] || e.highlight?.['title.exact']?.[0] || item.title
 
         // prefer the exact highlight for text
-        const searchTextHighlight = [...(e.highlight?.['text.exact'] || []), ...(e.highlight?.text || [])]
+        const searchTextHighlight = [...(e.highlight?.['text.exact'] || e.highlight?.text || [])]
```
huumn: You don't need all the spreads if you're just taking one.

api/resolvers/search.js
```diff
@@ -510,7 +510,7 @@ export default {
...
         item.searchText = searchTextHighlight?.slice(0, 5)?.join(' ... ')
```
huumn: You don't need this slice, because now there's never more than 5.

huumn (changes requested):

api/resolvers/search.js
abhiShandy: done

api/resolvers/search.js
abhiShandy: done

PR #2196, opened 2025-05-31
Add security advisory payout to awards.csv

ekzyis:
## Description

Added the payout for [GHSA-x2xp-x867-4jfc](https://github.com/stackernews/stacker.news/security/advisories/GHSA-x2xp-x867-4jfc) to awards.csv as requested. 

Afaict, using advisory as the type and the GHSA id should not break our awards script but not sure.


huumn:
Did you pay them out already? You added a payout date.

ekzyis:
Oh, no, I did not, I got confused.

Used `???` now.

huumn:
which PR was this fixed in?

ekzyis:
There is no PR yet

huumn:
Okay, thanks. Is there good reason to break the convention of paying when the fix is merged?

ekzyis:
Mhh, not really. I pushed a PR now: https://github.com/stackernews/stacker.news/pull/2205 

I can update this PR to point the CSV row to that PR instead of to the advisory. Would you prefer that?

PR #2199, opened 2025-06-02
Don't wrap base64 cert output

ekzyis:
## Description

When you run `sndev cert lnd`, the output contains new lines. I had to always remove them manually with `tr -d '\n'` before I could pipe the output into my clipboard via `xclip`.

I noticed that `base64` has the option to not wrap lines via `-w0`. Now I can pipe the output of `sndev cert lnd` into `xclip` and save the LND wallet.

## Additional Context

I used https://www.shellcheck.net/ to make sure `base64 -w0` is POSIX compliant and I think it is because I did not see this error:

> ^-- SC2039: In POSIX sh, [[ ]] is undefined.

## Checklist

yes

`9`.

n/a

no

### DIFF
--- a/sndev
+++ b/sndev
@@ -536,7 +536,7 @@ sndev__cert() {
     echo "no CERTDIR label found for $1"
     exit 1
   fi
-  docker__exec $1 cat $certdir/tls.cert | base64
+  docker__exec $1 cat $certdir/tls.cert | base64 -w0
 }
 
 sndev__help_cert() {


huumn:
shell POSIX compliance doesn't really relate to command compliance - just shell syntax afaik. 

Anyway, all that matters is that the container the command is exec'd on has support for the flags. ie if it works for you, it works for us all.

Issue #2198, opened 2025-06-01
Link lightning in settings is confusing


axelvyrn:
For stackers unfamiliar with lightning auth, they think it's an invoice that they pay:

> When I try to link my lightning wallet from settings through the "Link Lightning" button under auth methods
>
> ![Image](https://github.com/user-attachments/assets/d0fcbcfd-dd5a-493a-b126-92b7a8d004d4)
>
> I try to complete the invoice through my speed wallet it shows it is unable to pay the invoice.

ekzyis:
It's not a lightning invoice, it's LNURL-auth, the same thing we use for login with lightning, see https://lightninglogin.live/.

I guess we should clarify this in the settings UI.

axelvyrn:
oh, then Speed Wallet doesn't support it ig

PR #2206, opened 2025-06-04
Settings LNURL UI Update

axelvyrn:
## Description

Added a tooltip beside the Link Lightning button just like beside the Generate API key button (issue https://github.com/stackernews/stacker.news/issues/2198). It will tell the user about LNURL auth, a list of supported wallets and basic idea of what it is. ( as suggested in this [comment](https://github.com/stackernews/stacker.news/issues/2198#issuecomment-2926834820))

## Screenshots

N/A

## Additional Context

none

## Checklist

yes

9 - it affects the settings UI

yes. checked from dev console.

none

### DIFF
--- a/pages/settings/index.js
+++ b/pages/settings/index.js
@@ -861,10 +861,28 @@ function AuthMethods ({ methods, apiKeyEnabled }) {
             : <div key={provider} className='mt-2'><EmailLinkForm callbackUrl='/settings' /></div>
         } else if (provider === 'lightning') {
           return (
+            <div key={provider} className="d-flex align-items-center mt-2">
             <QRLinkButton
               key={provider} provider={provider}
               status={methods[provider]} unlink={async () => await unlink(provider)}
             />
+            <Info className="ms-2">
+                      <ul>
+                        <li>
+                          This is an LNURL-auth. Not sure what it is? Check <a target="_blank" href="https://lightninglogin.live/learn">here</a>
+                        </li>
+                        <li>
+                          Linking your Lightning wallet allows you to quickly sign in and zap directly from your wallet.
+                        </li>
+                        <li>
+                          You can unlink your wallet at any time.
+                        </li>
+                        <li>
+                          <b>Tip:</b> Use a wallet that supports LNURL-auth for seamless experience. Check the list of <a target="_blank" href="https://github.com/lnurl/luds/blob/luds/README.md#lnurl-documents">supported wallets</a>
+                        </li>
+                      </ul>
+                    </Info>
+                  </div>
           )
         } else if (provider === 'nostr') {
           return <NostrLinkButton key='nostr' status={methods[provider]} unlink={async () => await unlink(provider)} />


pages/settings/index.js
```diff
@@ -861,10 +861,28 @@ function AuthMethods ({ methods, apiKeyEnabled }) {
...
               key={provider} provider={provider}
               status={methods[provider]} unlink={async () => await unlink(provider)}
             />
+            <Info className="ms-2">
+                      <ul>
+                        <li>
+                          This is an LNURL-auth. Not sure what it is? Check <a target="_blank" href="https://lightninglogin.live/learn">here</a>
+                        </li>
+                        <li>
+                          Linking your Lightning wallet allows you to quickly sign in and zap directly from your wallet.
```
huumn: This is not true.

huumn (changes requested):
You cannot zap after auth'ing from lightning. its just auth .............

pages/settings/index.js
axelvyrn: removed that line

axelvyrn:
> this is not passing linting. run `sndev lint` to see errors.

checked. fixed the errors.

huumn (changes requested):
Screenshot 2025-06-05 at 8 18 20 PM

There is asymmetrical padding around this button now

axelvyrn:
how can that be there is `d-flex align-items-center` on the parent div.
anyways i'll try something else.

axelvyrn:
this should do

ekzyis:
Can you post screenshots like mentioned in the PR template?

axelvyrn:
i wanted to but i could not.
i was not able to load up the environment :(

ekzyis:
You weren't able to run the code?

axelvyrn:
no. my pc doesn't support docker, and codespaces doesn't open the ports idk why

ekzyis:
How do you explain your `9` in the QA question?

> **On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**
9 - it affects the settings UI

axelvyrn:
through dev console, i edited the code live

ekzyis:
That doesn't mean you tested your code which is what QA means ...

axelvyrn:
if it was wrong, it would not have showed up on the page right?
even k00b sent the ss here, there were only some padding issues

axelvyrn:
is this fine? symmetrical padding on both sides and centered info button?

ekzyis (changes requested):
> very sorry for 2 PRs in one but Github does not let commit 2 pull requests from 1 branch and I had changed the master branch itself

Create another branch.

axelvyrn:
15 files! Even copying the code will take a day

worker/nostrCrosspost.js
```diff
@@ -0,0 +1,53 @@
...
+      nostrCrosspostAt: { lte: now }
+    }
+  })
+
+  for (const item of items) {
+    try {
+      // Compose the Nostr message
+      const message = `${item.title || ''}\n${item.text || ''}\nhttps://stacker.news/items/${item.id}`.trim()
+      const nostr = nostrLib.get()
+      const signer = nostr.getSigner({ privKey: process.env.NOSTR_PRIVATE_KEY })
```
ekzyis: This will sign with SN's private key. This is not what we want.

ekzyis (changes requested):

worker/nostrCrosspost.js
axelvyrn: Oh, the user's nostr right. I'll change this

axelvyrn:
using user's nostrPubkey now

axelvyrn:
> Did you read AND understand [this comment](https://github.com/stackernews/stacker.news/pull/779#issuecomment-1936978875)?

I did and I think I mentioned it in the What I changed and Why section of the PR description.
I tried something else, though. You cannot edit a post which has already been posted on Nostr, but you can control WHEN it is being posted.
That is why in `api/resolvers/item.js` I added a 10-minute timer so that after the SN edit window closes, the Nostr post is created. Until then the post is say, at a storage temporarily. Is that not plausible?
image
and the backend handles the storing of the Nostr crossposting time
image

Do I need to change anything?

ekzyis (changes requested):
- You broke the random feed
- Your nostr signer doesn't work because your assumption is wrong that the backend has a way to sign with a pubkey
- You added a bunch of tests that didn't even catch that your signer doesn't work
- You added an empty file
- You added unrelated changes from #2214
- The script you added to schedule the job never runs and it's also not how we schedule jobs in our codebase anyway
- This does not require running jobs on a schedule and thus also not new columns on the `Item`, you can queue jobs on item create and update

I suggest you work on something else since you mentioned this was not easy.

axelvyrn:
Ok 🙏

Issue #603, opened 2023-11-05
Account Deletion Feature Request


Leon-Sander:
**Is your feature request related to a problem? Please describe.**
Would be nice to be able to delete ones account.

**Describe the solution you'd like**
A button with which (after confirmation) one can delete his account. 
Ideally there would be an additional option to then also delete all posts and comments
made from this account,

ekzyis:
Afaik, first user request to delete account: https://stacker.news/items/229252

ekzyis:
> Afaik, first user request to delete account: https://stacker.news/items/229252

@hellmanhellman, I see you changed your bio again. I've read your request to delete your account and that you mentioned GDPR.

Do you want us to delete your account now or not? If so, make it official by sending a mail to ek@stacker.news or kk@stacker.news.

m0wer:
This looks important. The challenge probably is not breaking threads and so on.

Probably the best is to implement the same behavior as Reddit: username references, posts, and comments get replaced with `[DELETED]`. But everything else (other user's replies) is still there.

Additionally, the user should mark a checkbox or something to accept the loss of remaining CCs or sats. And we should make sure that the deletion does not break scheduled jobs such as the rewards one.

huumn:
Those all sound like good additions. We could move all records we want to keep to a user called `deleted` or something, then delete the rest very carefully.

m0wer:
# Implementation plan

- Approach: Implement a "soft delete" mechanism that anonymizes user personal information while preserving content structure where needed.
- User Experience: Add a "Delete Account" section in settings with clear warnings and confirmation steps.
- Data Handling:
  - User can choose to delete/anonymize their content or leave it intact
  - Personal information is nullified regardless of content choice
  - Transaction records are preserved for accounting or whatever you might need
  - Remaining balance must be withdrawn before deletion
- Technical Implementation:
  - Add deletedAt timestamp to User model
  - Create database triggers to handle anonymization
  - Add GraphQL mutation for account deletion
  - Modify authentication to reject deleted accounts
  - Exclude deleted users from crons like rewards distribution
  - Update UI components to display "[deleted]" for deleted users and disable zaps
- Security & Privacy:
  - All personal information is removed from database
  - Sessions are invalidated for deleted accounts
  - Clear warning flow prevents accidental deletions
- Impact Assessment:
  - Preserves thread integrity in discussions
  - Maintains transaction records to be cleaned up as usual
  - Protects user privacy by removing personal data
[...]

Issue #1017, opened 2024-04-04
Button to invalidate all existing session


ekzyis:
**Is your feature request related to a problem? Please describe.**

We don't have a button to invalidate all existing sessions.

**Describe the solution you'd like**

Button to invalidate all existing sessions. Could be in the settings.

**Describe alternatives you've considered**

n/a

**Additional context**

https://stacker.news/items/493900

huumn:
We don't store sessions so I'm not sure how we'll manage to do this. It's all stateless.

SatsAllDay:
Just thinking out loud here...

1. Add a button in settings that sets a `invalidateSessions` timestamp in the DB somewhere, maybe an `SessionInvalidationRequest` table.
2. Add middleware to nextjs that checks the DB for such an entry on requests when the user already has a session. If an entry is found, respond with a `Set-Cookie` header that clears the existing session cookie (the equivalent of a logout operation, I presume).
3. Somehow we'd have to only clear the session if it was created _before_ the latest `SessionInvalidationRequest` entry. Otherwise, you'd end up in an infinite loop where sessions are constantly revoked.

The first two steps seem reasonable, but the third I am not sure about.

frstdrgn:
I didn't realize the sessions were stateless, I'll take a look at this and try to come up with some ideas.

> Add middleware to nextjs that checks the DB for such an entry on requests when the user already has a session. If an entry is found, respond with a Set-Cookie header that clears the existing session cookie (the equivalent of a logout operation, I presume).

Doesn't sound like this actually invalidates the cookie? Which in my mind is the security concern... 

This actually leaves with a question - do sessions expire at all?

frstdrgn:
@ekzyis if sessions don't expire at all, I think that should be separate security issue fwiw

ekzyis:
> Doesn't sound like this actually invalidates the cookie? Which in my mind is the security concern...

It would since any JWT that was created before user hit "invalidate all sessions" will no longer be accepted. So it's effectively invalidated, no?

> This actually leaves with a question - do sessions expire at all?

Default expiration is 30 days but they are regular refreshed while you're logged in and on the site.

frstdrgn:
[...]

PR #2212, opened 2025-06-07
feat: user deletion functionality

m0wer:
WIP

Closes https://github.com/stackernews/stacker.news/issues/603

### DIFF
--- a/api/resolvers/user.js
+++ b/api/resolvers/user.js
@@ -1,3 +1,4 @@
+import { createHash } from 'crypto'
 import { readFile } from 'fs/promises'
 import { join, resolve } from 'path'
 import { decodeCursor, LIMIT, nextCursorEncoded } from '@/lib/cursor'
@@ -656,6 +657,86 @@ export default {
   },
 
   Mutation: {
+    deleteAccount: async (parent, { deleteContent, confirmation }, { me, models }) => {
+      if (!me) {
+        throw new GqlAuthenticationError()
+      }
+
+      if (confirmation !== 'DELETE MY ACCOUNT') {
+        throw new GqlInputError('incorrect confirmation text')
+      }
+
+      return await models.$transaction(async (tx) => {
+        const user = await tx.user.findUnique({
+          where: { id: me.id },
+          select: {
+            msats: true,
+            mcredits: true,
+            name: true
+          }
+        })
+
+        if ((user.msats + user.mcredits) > 0) {
+          throw new GqlInputError('please withdraw your balance before deleting your account')
+        }
+
+        // If deleteContent is true, replace content with hash
+        if (deleteContent) {
+        // Get all items for this user
+          const items = await tx.item.findMany({
+            where: { userId: me.id },
+            select: { id: true, title: true, text: true, url: true }
+          })
+
+          // Update each item with hashed content
+          for (const item of items) {
+            const originalContent = JSON.stringify({
+              title: item.title,
+              text: item.text,
+              url: item.url
+            })
+
+            const hash = createHash('sha256').update(originalContent).digest('hex')
+            const deletedContent = `[deleted] ${hash}`
+
+            await tx.item.update({
+              where: { id: item.id },
+              data: {
+                title: item.title ? deletedContent : null,
+                text: item.text ? deletedContent : null,
+                url: item.url ? deletedContent : null
+              }
+            })
+          }
+        }
+
+        // Create deletion timestamp and hash the old username with it
+        const deletionTimestamp = new Date()
+        const usernameHashInput = `${user.name}${deletionTimestamp.toISOString()}`
+        const hashedUsername = createHash('sha256').update(usernameHashInput).digest('hex')
+
+        // Mark user as deleted and anonymize data
+        await tx.user.update({
+          where: { id: me.id },
+          data: {
+            deletedAt: deletionTimestamp,
+            name: hashedUsername,
+            email: null,
+            emailVerified: null,
+            emailHash: null,
+            image: null,
+            pubkey: null,
+            apiKeyHash: null,
+            nostrPubkey: null,
+            nostrAuthPubkey: null,
+            twitterId: null,
+            githubId: null
+          }
+        })
+
+        return true
+      })
+    },
     disableFreebies: async (parent, args, { me, models }) => {
       if (!me) {
         throw new GqlAuthenticationError()
--- a/api/ssrApollo.js
+++ b/api/ssrApollo.js
@@ -19,6 +19,23 @@ import { MULTI_AUTH_ANON, MULTI_AUTH_LIST } from '@/lib/auth'
 
 export default async function getSSRApolloClient ({ req, res, me = null }) {
   const session = req && await getServerSession(req, res, getAuthOptions(req))
+
+  // If there's a session, check if the user is deleted
+  let meUser = me
+  if (session?.user) {
+    const user = await models.user.findUnique({
+      where: { id: parseInt(session.user.id) }, // Convert string to int
+      select: { deletedAt: true }
+    })
+
+    // If the user is deleted, don't pass the session to the context
+    if (user?.deletedAt) {
+      meUser = null
+    } else {
+      meUser = session.user
+    }
+  }
+
   const client = new ApolloClient({
     ssrMode: true,
     link: new SchemaLink({
@@ -28,9 +45,7 @@ export default async function getSSRApolloClient ({ req, res, me = null }) {
       }),
       context: {
         models,
-        me: session
-          ? session.user
-          : me,
+        me: meUser,
         lnd,
         search
       }
@@ -158,6 +173,18 @@ export function getGetServerSideProps (
     // if we switched to anon and authentication is required
     if (req.cookies[MULTI_AUTH_LIST] === MULTI_AUTH_ANON) {
       me = null
+    }
+
+    // Check if the user account is deleted
+    if (me) {
+      const user = await models.user.findUnique({
+        where: { id: parseInt(me.id) }, // Convert string to int
+        select: { deletedAt: true }
+      })
+
+      if (user?.deletedAt) {
+        me = null
+      }
     }
 
     if (authRequired && !me) {
--- a/components/item-info.js
+++ b/components/item-info.js
@@ -132,11 +132,14 @@ export default function ItemInfo ({
       <span> \ </span>
       <span>
         {showUser &&
-          <Link href={`/${item.user.name}`}>
-            <UserPopover name={item.user.name}>@{item.user.name}</UserPopover>
-            <Badges badgeClassName='fill-grey' spacingClassName='ms-xs' height={12} width={12} user={item.user} />
-            {embellishUser}
-          </Link>}
+          (item.user.deletedAt
+            ? <span className='text-muted'>[deleted]</span>
+            : <Link href={`/${item.user.name}`}>
+              <UserPopover name={item.user.name}>@{item.user.name}</UserPopover>
+              <Badges badgeClassName='fill-grey' spacingClassName='ms-xs' height={12} width={12} user={item.user} />
+              {embellishUser}
+              </Link>
+          )}
         <span> </span>
         <Link href={`/items/${item.id}`} title={item.invoicePaidAt || item.createdAt} className='text-reset' suppressHydrationWarning>
           {timeSince(new Date(item.invoicePaidAt || item.createdAt))}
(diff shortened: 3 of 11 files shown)


m0wer:
BTW the lint error is unfixable! (take it as a challenge) It looks like the linter gets confused with the ternary operator inside the nested JSX structure. If there's no way to fix it, we can just disable the check for those lines.

components/item-info.js
```diff
@@ -132,11 +132,14 @@ export default function ItemInfo ({
...
-            {embellishUser}
-          </Link>}
+          (item.user.deletedAt
+            ? <span className='text-muted'>[deleted]</span>
+            : <Link href={`/${item.user.name}`}>
+              <UserPopover name={item.user.name}>@{item.user.name}</UserPopover>
+              <Badges badgeClassName='fill-grey' spacingClassName='ms-xs' height={12} width={12} user={item.user} />
+              {embellishUser}
+              </Link>
+          )}
```
Soxasora: > BTW the lint error is unfixable! (take it as a challenge)

*nothing is unfixable!*
Ternary multi-line expressions require parentheses; otherwise, both the dev and the parser won't know where your expression ends. The linter choked because of this, and didn't gave you the full explanation, but this is where your knowledge comes in clutch.
You can either use parentheses or avoid ternary operators.

m0wer: That did the trick. Thanks!

84908843

m0wer:
Looking for QA :eyes:

huumn:
@m0wer this is going to take a minute. we are all kind of swamped and this is a big, potentially dangerous (although I'm sure it's fine) change.

i'd guess you'll see a review when one of us needs a break from our own work. i personally might not be able to get to this for a couple weeks given some real life stuff. 

just wanted to give a heads up. this is something we want, so we'll get around to it eventually.

api/paidAction/zap.js
ekzyis, line 1: Why not treat content of deleted accounts like deleted content?

When stackers delete a post or comment, we remove the zap button and make the item unzappable:

https://github.com/stackernews/stacker.news/blob/96fd27157336ef1461f425c516d08e752f6a952c/api/resolvers/item.js#L968-L970

pages/settings/index.js
```diff
@@ -1,14 +1,11 @@
 import { Checkbox, Form, Input, SubmitButton, Select, VariableInput, CopyInput } from '@/components/form'
-import Alert from 'react-bootstrap/Alert'
-import Button from 'react-bootstrap/Button'
-import InputGroup from 'react-bootstrap/InputGroup'
-import Nav from 'react-bootstrap/Nav'
+import { Alert, Button, InputGroup, Nav, Form as BootstrapForm, OverlayTrigger, Tooltip } from 'react-bootstrap'
```
ekzyis: Why use the form from `react-bootstrap` and not our custom one above from `@/components/form`?

api/resolvers/user.js
```diff
@@ -656,6 +657,111 @@ export default {
...
 
   Mutation: {
+    deleteAccount: async (parent, { deleteContent, confirmation, donateBalance }, { me, models }) => {
+      if (!me) {
+        throw new GqlAuthenticationError()
+      }
+
+      if (confirmation !== 'DELETE MY ACCOUNT') {
+        throw new GqlInputError('incorrect confirmation text')
+      }
```
ekzyis, line 668: It's good to check this on the server, but this should also be checked by formik on the client.

Formik's validation errors look nicer and is consistent with how we handle input errors in other places.

(see other comment in pages/settings/index.js)

api/resolvers/user.js
```diff
@@ -656,6 +657,111 @@ export default {
...
+          select: {
+            msats: true,
+            mcredits: true,
+            name: true
+          }
+        })
+
+        const totalBalance = user.msats + user.mcredits
+        if (totalBalance > 0 && !donateBalance) {
+          throw new GqlInputError('please withdraw your balance before deleting your account or confirm donation to rewards pool')
```
ekzyis, line 691: I think this can also be part of formik validation

api/ssrApollo.js
```diff
@@ -19,6 +19,23 @@ import { MULTI_AUTH_ANON, MULTI_AUTH_LIST } from '@/lib/auth'
...
+    })
+
+    // If the user is deleted, don't pass the session to the context
+    if (user?.deletedAt) {
+      meUser = null
+    } else {
+      meUser = session.user
+    }
+  }
+
```
ekzyis: Mhh, if we could revoke sessions (#1017), we wouldn't need this additional session check.

Will this run on every request? Can this be put into our auth code? I think ideally, `getServerSession` would already do all kinds of session checks and just not return a session if the user is deleted. Else I'm worried we will miss places where we need to check this.

api/ssrApollo.js
```diff
@@ -160,6 +175,18 @@ export function getGetServerSideProps (
...
+      const user = await models.user.findUnique({
+        where: { id: parseInt(me.id) }, // Convert string to int
+        select: { deletedAt: true }
+      })
+
+      if (user?.deletedAt) {
+        me = null
+      }
+    }
+
```
ekzyis: Is this needed? Aren't you already doing this in `getSSRApolloClient` so using that client to query `ME` will return null?

docker-compose.yml
```diff
@@ -179,7 +179,7 @@ services:
       - OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD}
       - plugins.security.disabled=true
       - discovery.type=single-node
-      - "_JAVA_OPTIONS=-Xms2g -Xmx2g -XX:UseSVE=0"
+      - "_JAVA_OPTIONS=-Xms2g -Xmx2g"
```
ekzyis: Removing the java option looks unrelated to the feature. Accidentally checked in?

pages/api/auth/[...nextauth].js
```diff
@@ -93,6 +93,19 @@ function getCallbacks (req, res) {
...
+        const dbUser = await prisma.user.findUnique({
+          where: { id: user.id },
+          select: { deletedAt: true }
+        })
+
+        if (dbUser?.deletedAt) {
+          // Return false to prevent sign in for deleted accounts
+          return false
+        }
+      }
```
Comment (ekzyis) [inline pages/api/auth/[...nextauth].js:107]: Ah, you're already doing this here in the JWT callback.

But isn't this enough? I think this callback runs on every request to verify the token so if you return `false` (or `null`?), there should be no session.

So I am confused why we need to check if the account is deleted in other places, too.

pages/settings/index.js
```diff
@@ -87,6 +83,123 @@ export function SettingsHeader () {
...
+                  >
+                    Permanently delete my account
+                  </Button>
+                </div>
+              </>
+              )}
+        </div>
+      </div>
+    </>
+  )
```
ekzyis: I think this should use our custom form with a schema for validation

(see other comment in api/resolvers/user.js)

prisma/migrations/20250610152839_deleted_user_earnings_job/migration.sql
ekzyis, line 1: If content from deleted users is not zappable like deleted content, we also don't need this

worker/earn.js
```diff
@@ -182,7 +182,10 @@ function earnStmts (data, { models }) {
   return [
     models.earn.create({ data }),
     models.user.update({
-      where: { id: userId },
+      where: {
+        id: userId,
+        deletedAt: null // Only update if user is not deleted
```
ekzyis, line 187: mhhh, need to think more if this is needed (how can a deleted user earn rewards?) and what the implications are if this is indeed needed (do we need to do something similar in other places, too?)

api/paidAction/zap.js
m0wer: Two reasons:
- Currently the deleted content is just soft deleted and remains in the DB. Since this was for a GDPR style deletion request, it's not enough. Since the combination of user ID and content has to be unique, the deleted content is replaced with its hash.
- If a user decides to leave the content visible even after deletion, it's nice that the content can still be zapped and ranked accordingly.

pages/settings/index.js
m0wer: Good point, I'll change it.

api/resolvers/user.js
m0wer: Ok, will do.

api/resolvers/user.js
m0wer: Ok, will do.

api/ssrApollo.js
m0wer: I can do that instead yes. Way cleaner.

api/ssrApollo.js
m0wer: Not sure, will have a look.

docker-compose.yml
m0wer: Oh sorry. It's a change that I need to get the opensearch container working on my machine. Accidentally committed it. Will remove.

pages/api/auth/[...nextauth].js
Comment (m0wer) [reply pages/api/auth/[...nextauth].js:107]: Don't remember, will have a look.

pages/settings/index.js
m0wer: Yes, will do.

prisma/migrations/20250610152839_deleted_user_earnings_job/migration.sql
m0wer: Let's decide on the other thread.

worker/earn.js
m0wer: If the content is still zapable or if the account is removed after some actions were done by the user before the rewards windows closes. Then there's referrals, etc.

api/paidAction/zap.js
ekzyis: > Since the combination of user ID and content has to be unique, the deleted content is replaced with its hash.

Which index is enforcing this? Do you mean `Item_unique_time_constraint`? It should ignore any item that has `deletedAt` set.

> If a user decides to leave the content visible even after deletion, it's nice that the content can still be zapped and ranked accordingly.

I think we don't need to give them the option of leaving their content visible. I would just delete everything as if they manually went through each item and deleted it.

If we want to follow GDPR, they might still have a right to change their mind and request full deletion but that would be awkward to request without having an account anymore.

It also makes our life easier by not having to support different states of deletion.

m0wer: Don't remember what index, but stumbled upon that on my first try that just replaced everything by `deleted` or something. Not sure if I tried both deleting the content and setting `deletedAt` though.

Oh I see. I didn't think about the possibility of a full deletion request afterwards. It's just sad to lose some nice content after a user leaves. But I guess users that don't want to delete their content can just not delete their account xD

ekzyis: > But I guess users that don't want to delete their content can just not delete their account xD

Yes, this.

I also think instead of hashing the username, we should move all content to some special account like `@anon` (but not `@anon`). I would love to use `@deleted` but it's not one of the [names we reserved](https://github.com/stackernews/stacker.news/blob/master/prisma/migrations/20210521234940_reserved_names/migration.sql) and [someone already took it](https://stacker.news/deleted) (but never used it) :/

I think we can move the content to `@delete` (user id is 106 in prod). Since `@delete` has no other content, it will be easy to move the items to a different account later on if we ever choose so.

So essentially, if someone only deletes an item, the content is replaced with `*deleted by author*` but if they delete their account, the nym is now also `@delete`.

m0wer: Sounds good. I'll ping you when it's done.

m0wer:
@ekzyis ready for another round of QA!

m0wer:
Ready for QA @huumn

ekzyis:
Hello, I recently updated our PR template to add a question about AI in https://github.com/stackernews/stacker.news/pull/2507.

I updated your PR description to include it, see the end of the checklist.

Can you please answer it? Thank you!

huumn:
This has many conflicts and a lot of the code has changed and even if that weren't true this is going to require a lot of review time I don't currently have. That's our fault and not yours. 

I'm going to pay you a meaningful fraction of the bounty and close this, hoping to revisit it before too long and using some the decisions made here as inspiration.

m0wer:
> This has many conflicts and a lot of the code has changed and even if that weren't true this is going to require a lot of review time I don't currently have. That's our fault and not yours.
> 
> I'm going to pay you a meaningful fraction of the bounty and close this, hoping to revisit it before too long and using some the decisions made here as inspiration.

Thanks! I think it should be easier knowing what needs to be changed and having the decisions made. Ping me if you want help with it.

I've opened a PR to update my LN address: https://github.com/stackernews/stacker.news/pull/2670

huumn:
Thanks so much for giving it a shot at all! We have this huge post-wallet backlog and I'm trying to pick up all the packets we dropped.

Issue #2208, opened 2025-06-06
Automated generation of table of contents


ed-kung:
### Describe the problem you're trying to solve

Sometimes when I write longer posts, I want to put a table of contents on top, like in this post: [SN#998456](https://stacker.news/items/998456). I usually do it manually, but it would be nice to have an automated option like Kramdown's [toc element](https://kramdown.gettalong.org/converter/html.html#toc) `{:toc}`. 

### Describe the solution you'd like

A markdown element like Kramdown's `{:toc}` that automatically generates a table of contents based on headers.

### Describe alternatives you've considered

I couldn't find any automated solutions other than just manually typing out the table of contents.

ekzyis:
Have you seen this and if so, is it not enough?

![Image](https://github.com/user-attachments/assets/eb204b2a-d89f-4b50-8d41-bc660023e8f8)

It's okay if not, no judgement, just curious!

ed-kung:
Oh, cool didn't notice that. I personally like the aesthetic of having the TOC in the text body so the reader gets an immediate outline of the content.  So I'd still like to leave this issue up and wouldn't mind tackling it myself when I get the chance.  But if you guys don't think it's needed, feel free to close :)

huumn:
I think if the ToC is optional and triggered by markdown, I'd support a PR that does this. We already have the code for it.

PR #2213, opened 2025-06-07
Automatically generate table of contents in text

ed-kung:
## Description

Closes https://github.com/stackernews/stacker.news/issues/2208

Writing `{:toc}` on its own line will automatically render a table of contents as a bulleted list, using the headings (`#`, `##`, etc) contained in the post.

## Screenshots

https://github.com/user-attachments/assets/04b36c23-60d3-4068-b6b8-b86b45f78b3e

## Additional Context

n/a

## Checklist

yes

8.  If another post had `{:toc}` somewhere it could render in unexpected ways. 

The behavior of links that start with a hash (i.e. `#section-1`) has also changed, but I can't see why anyone would've wanted such links to open in a new tab. 

n/a

no


huumn:
just a heads up, I probably won't get around to deeply reviewing this until next week. at a glance it looks good though.

components/table-of-contents.js
```diff
@@ -2,28 +2,16 @@ import React, { useMemo, useState } from 'react'
...
-    const tree = fromMarkdown(text)
-    const toc = []
-    visit(tree, 'heading', (node, position, parent) => {
-      const str = toString(node)
-      toc.push({ heading: str, slug: slug(str.replace(/[^\w\-\s]+/gi, '')), depth: node.depth })
-    })
-
-    return toc
-  }, [text])
+  const toc = useMemo(() => extractHeadings(text), [text])
```
Soxasora, line 14: I noticed that we don't give IDs to nodes inside comments, only on full posts:
https://github.com/stackernews/stacker.news/blob/462016042ceda364ac98fc965e920a90ee179797/components/text.js#L116

Because of this, the table of contents won't work if used in comments. I personally think that the ToC doesn't make that much sense in comments, so we can just disable `{:toc}` for them with `topLevel` awareness. What do you think? ^^

lib/toc.js
```diff
@@ -0,0 +1,23 @@
...
+    ? fromMarkdown(markdownOrTree)
+    : markdownOrTree
+
+  const headings = []
+
+  visit(tree, 'heading', node => {
+    const str = toString(node)
+    headings.push({
+      heading: str,
+      slug: slug(str.replace(/[^\w\-\s]+/gi, '')),
```
Soxasora, line 17: note for the future, unrelated to review:
it seems that we don't handle duplicate headings, because we use `slug` instead of the `GithubSlugger` **class**.
Probably because it's faster as `GithubSlugger` would have instead tracked headings in memory and checked every heading against the previous ones to count 🤔.

ekzyis: We don't use `GithubSlugger` because of https://github.com/stackernews/stacker.news/pull/1405

ed-kung: Ok, will leave this as is for now

components/table-of-contents.js
ed-kung: I agree, it wouldn't make sense to use Toc in comments, so I changed it so that remarkToc is only processed for topLevel items

components/text.js
```diff
@@ -49,6 +54,9 @@ export function SearchText ({ text }) {
 
 // this is one of the slowest components to render
 export default memo(function Text ({ rel = UNKNOWN_LINK_REL, imgproxyUrls, children, tab, itemId, outlawed, topLevel }) {
+  // include remarkToc if topLevel
+  const remarkPlugins = topLevel ? [...baseRemarkPlugins, remarkToc] : baseRemarkPlugins
```
ekzyis, line 58: Unfortunately, `topLevel` cannot be used to exclude this plugin from comments because `topLevel` is also true for comments when they are the root of a page:

https://github.com/user-attachments/assets/2aff2b74-89ad-4c0b-8dbb-7f8652f54852

not sure how to best deal with this

lib/toc.js
ekzyis, line 17: what is the replacement for?

ed-kung: Was just copying from line 22 of the old `components/table-of-contents.js`, which uses the same pattern:

https://github.com/stackernews/stacker.news/blob/20147cae150a98e9570f31e794a505742ac83c69/components/table-of-contents.js#L22

But is it redundant since we're already passing the string to `slug`? I will usually copy the old code whenever I can to break as few things as possible

components/text.js
ed-kung: Interesting. One possibility is to remove the topLevel check for giving headings node ids:

https://github.com/stackernews/stacker.news/blob/462016042ceda364ac98fc965e920a90ee179797/components/text.js#L116

I'm not sure why that check is there and if doing so would break anything.


I suppose another possibility is to only process `{:toc}` if the item has no parent, right? Will `topLevel` always be true for posts? (and thus headings always get assigned node ids in posts)

ekzyis:
FYI, in case you're confused why this was merged despite of https://github.com/stackernews/stacker.news/pull/2213#discussion_r2229835169:

@huumn's answer to my question why it wasn't blocking:

> A comment behaving like a post when at topLevel is consistent with other behavior (e.g. heading and image sizes)

Issue #2209, opened 2025-06-06
make sndev work in github codespaces


huumn:
I'm not sure what this entails or if it's even possible, but it might be nice for people who can't run stuff locally.

axelvyrn:
Codespaces does not support Docker-in-Docker or Docker Compose natively. Instead, Codespaces uses Docker for the container itself.

axelvyrn:
could finally run sndev on codespaces

![Image](https://github.com/user-attachments/assets/ddd52621-f1b1-4819-af29-ca2fc1623a6e)

![Image](https://github.com/user-attachments/assets/0c733e9e-09f2-4ce6-9873-0be2aa9dc9e0)

will be creating a pull request with the necessary changes

PR #2214, opened 2025-06-08
sndev runs on codespaces

axelvyrn:
## Description
Closes https://github.com/stackernews/stacker.news/issues/2209
Github Codespaces uses docker to run the platform like sndev. So this was a docker-in-docker error. Thus it was required to check explicitly if Docker was available in the codespaces.

Steps to run sndev on Github Codespaces:
1. run `npm install`. let it be finished
2. in `.env` set 
```
CPU_SHARES_IMPORTANT=1024
CPU_SHARES_MODERATE=512
CPU_SHARES_LOW=256
OPENSEARCH_PASSWORD=yourStrongPassword
```
because when I ran `docker ps -a && docker-compose ps && ./sndev status`
it returned
```bash
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
WARN[0000] The "CPU_SHARES_LOW" variable is not set. Defaulting to a blank string. 
error while interpolating services.opensearch.cpu_shares: failed to cast to expected type: strconv.ParseInt: parsing "": invalid syntax
```
because `opensearch` was not able to run as a healthy container

3. run `./sndev start`. let it build and deploy the container, you should be able to see the containers icon appear in the sidebar

image

4. wait for about 5 minutes or split terminal and/or check sndev status.
5. go to PORTS and open port 3000 on the web browser through the link from github codespaces (the browser icon that appears when you hover over port 3000) not from `http://localhost:3000`

6. you will be able to see the live site on the page

## Screenshots
Image

Image

## Additional Context
Make sure the codespaces machine is the **8GB RAM and 32 GB storage** one because running sndev puts great load on the CPU

Troubleshooting:
1. If sndev does not start on the first try just run `./sndev stop` and then `./sndev start` again, there might be some image creation or memory failure
2. If opensearch fails to run set Java heap to `- "_JAVA_OPTIONS=-Xms1g -Xmx1g -XX:UseSVE=0"` which is applicable for <=4GB RAM (even if you use the better machine in codespaces, this might help) in `docker-compose.yml` (replace line 183)

## Checklist

yup

10

N/A

```
CPU_SHARES_IMPORTANT=1024
CPU_SHARES_MODERATE=512
CPU_SHARES_LOW=256
```


axelvyrn:
Well, you have to set the .env variables by yourself I suppose


> without having to manually do things.

you don't have to do it anymore, I updated the description please do check, just setting the .env variables and running ./sndev start should do, the check for codespaces environment is set in the file I changed

huumn:
There should be some way of auto-exporting those env variables. The goal is one-click dev environments for contributors. If you can't figure it out, that's fine. If it's not possible, I'd appreciate linking to somewhere I can confirm it's not possible.

axelvyrn:
image
image


success, after the sndev edit, I could start sndev with just `./sndev start` and nothing more on the first try,
as you can see, I tried in on a totally new machine which would not have the influence of the previous one.
(Last one was glowing space computing machine, this one is bookish zebra)

:)

Issue #2194, opened 2025-05-30
Duplicate post send to Nostr when editing SN


rideandsmile:
### Describe the problem you're trying to solve

I m crossposting an article between SN and Nosts. 

And  i realized if i want to edit the original post on SN ( cause of a mistake , missing items...) in that case a new post  will be sendone more time to Nostr.

### Describe the solution you'd like

is it possible to avoid to send to Nostr one more time when editing on SN ?

ekzyis:
> is it possible to avoid to send to Nostr one more time when editing on SN ?

Sorry for the bad experience! [We have plans](https://github.com/stackernews/stacker.news/pull/779#issuecomment-1936978875) to only publish to nostr when the edit timer ran out but we didn't get to it yet.

rideandslide:
oki dok 
thks

brymut:
Hi @ekzyis, just looking at this. Was wondering if an appropriate approach to this might be scheduling crossposting as a worker job in socialPoster when the item is created, (createdAt + ITEM_EDIT_SECONDS), instead of immediately crossposting after submitting here: https://github.com/stackernews/stacker.news/blob/37d7ed59bba770ca79e20f86318e9a6237d1a0bd/components/use-item-submit.js#L89-L92

ekzyis:
I mentioned how I think this should be solved [in this comment](https://github.com/stackernews/stacker.news/pull/779#issuecomment-1936978875)

brymut:
> I mentioned how I think this should be solved [in this comment](https://github.com/stackernews/stacker.news/pull/779#issuecomment-1936978875)

ahhh, thanks, I'll read through that convo

brymut:
> I mentioned how I think this should be solved [in this comment](https://github.com/stackernews/stacker.news/pull/779#issuecomment-1936978875)

Is nostr pre-signing currently available? Seems like that might need to be introduced and storing the posts before we send them out after the edit timer via a worker.

ekzyis:
We don't store pre-signed nostr events yet so that would be part of this issue

PR #2216, opened 2025-06-09
Nostr repost fix

axelvyrn:
## Description

Closes https://github.com/stackernews/stacker.news/issues/2194
Now an edit on an SN post will not trigger Nostr reposting 

**The actual change in code is in `components/use-item-submit.js`**
**The other files changed in this repo can be ignored while merging the PR**

Why I edited/created those files:
I ran an automated test to check the running of the code I changed in the `components/use-item-submit.js`

**So I created `/components/__tests__/use-item-submit.test.js` wherein the you can SEE when the test calls and does NOT call the crosspost to Nostr logic**

I edited the `package.json` to include these
image

So that I could launch the test environment running through jsdom
Then I edited `jest.config.js`
image

Then I ran `npx jest components/__tests__/use-item-submit.test.js` and the console showed the test as successful (proof in screenshots)

## Screenshots
9th try:
image

## Additional Context

none

## Checklist

yes

10

N/A

none


huumn:
This is not the solution we want for this problem. This just prevents editing.

Issue #2039, opened 2025-03-28
Feature Request: Autocomplete in the search bar


ed-kung:
### Describe the problem you're trying to solve

I was searching for one of `@denlillaapan`'s posts, but kept misspelling his nym. Autocomplete for users and territories might be a nice feature for the search bar.

### Describe the solution you'd like

Autocomplete for `@nym` in the search bar, which works the same way it does in posts

Autocomplete for `~territory` in the search bar (which isn't currently available for posts either, so  https://github.com/stackernews/stacker.news/issues/992 may need to be done first)

brymut:
Hey @ed-kung, not anymore. I had planned to follow the progress on #992 but I lost track. You can go ahead if you're interested

PR #2217, opened 2025-06-10
user and territory autocomplete in search bar

ed-kung:
## Description

Closes https://github.com/stackernews/stacker.news/issues/2039

Implements autocomplete in the search bar. Functionality should work the same as for posts. `@` brings up a list of users and `~` brings up a list of territories.

## Screenshots

https://github.com/user-attachments/assets/8ee1f566-af78-4345-b74b-ea6e57ee1928

## Additional Context

n/a

## Checklist

yes

8

Haven't tested on mobile

no


huumn (approved):
This works well. Nice job. 

However, this change suggest an opportunity for generalization. Both `MarkdownInput` and this new `SearchInput` share an awful lot of logic that can probably be abstracted into a `SuggestionInput` component that they share and/or a hook.

It's beyond the scope of the PR perhaps, so I'll try to do this myself soon.

ed-kung:
Makes sense. I also noticed the opportunity for generalization, but didn't want to mess with `MarkdownInput` too much unnecessarily.

PR #2230, opened 2025-06-14
Remove directReceive setting

### LINKED ISSUE #2029: Individual settings for proxy and direct receives no longer make sense
Since the lightning address now only works when you have a wallet attached, we no longer need both settings. We will always create an invoice from the external wallet, the only setting is if we should try to proxy and maybe if we should fallback to direct receives in case the proxy failed.

However, not sure if a fallback makes sense since the lightning address will always be paid via QR codes afaict and thus there's no fallback once we generated the invoice. The fallback would only make sense for cases where we failed to wrap the invoice.

ekzyis:
## Description

Close #2029 

This removes the `directReceive` setting because it currently unintentionally disabled the lightning address if  `directReceive` was not enabled and `proxyReceive` was not enabled or wrapping failed.

Since the original purpose of this setting was to disable buying credits, which is no longer possible since 5a8804de79ff8e004a74e52afc5ff4bf3e9ab3b3, this setting is no longer needed.

Now, if proxy was not enabled or wrapping failed, we always fallback to direct payments.

For the sake of keeping the product simple / simplifying it, I think we don't need to give stackers a dedicated setting to control if they want these fallbacks or not.

## Additional Context

I actually think we should never fallback to direct payments if the proxy was enabled.

If the proxy is enabled and we allow fallbacks, it's trivial to reveal the node pubkey by simply requesting an invoice for 1 sat in which case wrapping will almost always fail. This makes the proxy setting pretty useless imo. It gives a false sense of privacy.

Not having fallbacks comes at the cost of a higher chance of failed payments though.

But I also think it's easier to understand that if proxy is enabled, we _always_ and _only_ proxy (but it might fail due to fees), instead of having to explain that if wrapping fails, we will try without it and thus still reveal their node pubkey.

## Checklist

yes

`7`. No more occurrences of `directReceive`.

n/a

no

### DIFF
--- a/api/paidAction/receive.js
+++ b/api/paidAction/receive.js
@@ -16,7 +16,6 @@ export async function getCost ({ msats }) {
 
 export async function getInvoiceablePeer (_, { me, models, cost, paymentMethod }) {
   if (paymentMethod === PAID_ACTION_PAYMENT_METHODS.P2P && !me?.proxyReceive) return null
-  if (paymentMethod === PAID_ACTION_PAYMENT_METHODS.DIRECT && !me?.directReceive) return null
 
   const wallets = await getInvoiceableWallets(me.id, { models })
   if (wallets.length === 0) {
--- a/api/typeDefs/user.js
+++ b/api/typeDefs/user.js
@@ -208,6 +208,7 @@ export default gql`
     vaultKeyHash: String
     walletsUpdatedAt: Date
     proxyReceive: Boolean
+    directReceive: Boolean @deprecated
     receiveCreditsBelowSats: Int!
     sendCreditsBelowSats: Int!
   }
--- a/api/typeDefs/user.js
+++ b/api/typeDefs/user.js
@@ -125,7 +125,6 @@ export default gql`
     wildWestMode: Boolean!
     withdrawMaxFeeDefault: Int!
     proxyReceive: Boolean
-    directReceive: Boolean
     receiveCreditsBelowSats: Int!
     sendCreditsBelowSats: Int!
   }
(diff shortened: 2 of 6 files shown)


huumn:
> if proxy was not enabled or wrapping failed

this exposes a stackers pubkey, doesn't it. if i can reliably cause wrapping to fail, can't I reveal every stacker's receiving node's pubkey?

ekzyis:
Yes, I mentioned this in the additional context and created #2231 because of this.

huumn:
> Yes, I mentioned this in the additional context and created https://github.com/stackernews/stacker.news/pull/2231 because of this.

Ah thanks, I only read to point that caught my attention.

huumn:
> Are your changes backwards compatible? Please answer below:
>
> yes

Pretty sure this isn't true. It removes a field on `me` that active clients will be querying for.

ekzyis:
> > Are your changes backwards compatible? Please answer below:
> > yes
> 
> Pretty sure this isn't true. It removes a field on `me` that active clients will be querying for.

Oof, you are right :see_no_evil: I will update that question because I keep making this mistake.

ekzyis:
Fixed in 4c01133.

Marked the field with [`@deprecated`](https://spec.graphql.org/October2021/#sec--deprecated) so we don't forget to remove it. Tested that `@deprecated` does not need a reason. An old client can still query that field; it will just return `null`.

huumn:
> Marked the field with [@deprecated](https://spec.graphql.org/October2021/#sec--deprecated) so we don't forget to remove it. Tested that @deprecated does not need a reason. An old client can still query that field; it will just return null.

uwwww fancy. Nice find!

PR #2248, opened 2025-06-24
add devcontainer config for codespaces

### LINKED ISSUE #2209: make sndev work in github codespaces

brymut:
## Description

Added .devcontainer a default & minimal configuration to support running the sndev environment in codespaces, added instructions in the README and modified login link so the hyperlink in the terminal can redirect to the correct container host link. Services are accessible via links in the VSCode tabs

closes #2209 

## Screenshots
Log in link
Screenshot 2025-06-25 at 09 40 37

Logged in user
Screenshot 2025-06-25 at 09 40 19

Codespaces config
Screenshot 2025-06-25 at 09 51 02

Ports Tab with links
Screenshot 2025-06-25 at 09 40 44

Lnbits
Screenshot 2025-06-25 at 09 41 03

Screenshot 2025-06-25 at 09 42 24

## Additional Context

The machines available for Github codespaces have limited compute & storage. Storage left after build & config is less than 10 GB. I customised the devcontainers to run using the highest available free tier machines.

## Checklist

### DIFF
--- a/.devcontainer/minimal/devcontainer.json
+++ b/.devcontainer/minimal/devcontainer.json
@@ -1,0 +1,22 @@
+{
+    "name": "sndev MINIMAL",
+    "hostRequirements": {
+      "cpus": 4,
+      "memory": "16gb",
+      "storage": "32gb"
+    },
+    "customizations": {
+      "vscode": {
+        "extensions": [
+          "ms-azuretools.vscode-containers"
+        ]
+      }
+    },
+    "containerEnv": {
+      "CPU_SHARES_IMPORTANT": "1024",
+      "CPU_SHARES_MODERATE": "512",
+      "CPU_SHARES_LOW": "128",
+      "COMPOSE_PROFILES": "minimal"
+    },
+    "postAttachCommand": "./scripts/setup-codespaces.sh && source .env.local && ./sndev start"
+  }
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ You can run Stacker News on Github Codespaces
 #### Setup
 
 1. Open the repository on GitHub and click the **"Code"** button
-2. Select the CodeSpaces tab and create a new codespace. (or configure with provided settings first)
+2. Select the Codespaces tab and create a new codespace. 
+   - You can also configure your codespace to run select services based on  `COMPOSE_PROFILES` as well as in a different region and machine type by clicking "..." and selecting "New with options...". Check [Modifying services](#modifying-services) for more information on `COMPOSE_PROFILES`
 3. Wait for the environment to set up (this may take several minutes the first time)
 4. Once ready, you'll see a terminal with the environment initialized
 
@@ -57,7 +58,7 @@ Access your running application at the URL shown in the forwarded ports panel (t
 
 #### Port Configuration
 
-⚠️ **Important**: For external access to work properly, you must set forwarded ports to **Public** in the Ports tab:
+⚠️ **Important**: For various internal services and external access to work properly, you must set forwarded ports to **Public** in the Ports tab:
 
 1. In your codespace, look for the "PORTS" tab in the bottom panel
 2. Click the lock icon to change visibility from "Private" to "Public"
@@ -384,7 +385,7 @@ We offer triage permissions to contributors after they've made a few contributio
 Contributors can get badges on their SN profiles by opening a pull request adding their SN nym to the [contributors.txt](/contributors.txt) file.
 
 ## What else you got
-We offer GitHub Codespaces support for cloud-based development! In the future we plan to offer more, like gratis github copilot subscriptions, reverse tunnels, and merch.
+In the future we plan to offer more, like gratis github copilot subscriptions, reverse tunnels, codespaces, and merch.
 
 If you'd like to see something added, please make a suggestion.
 
--- a/README.md
+++ b/README.md
@@ -36,6 +36,33 @@ Go to [localhost:3000](http://localhost:3000).
 
 <br>
 
+### GitHub Codespaces
+
+[![Open in GitHub Codespaces](https://github.com/codespaces/badge.svg)](https://codespaces.new/stackernews/stacker.news)
+
+You can run Stacker News on Github Codespaces
+
+#### Setup
+
+1. Open the repository on GitHub and click the **"Code"** button
+2. Select the CodeSpaces tab and create a new codespace. (or configure with provided settings first)
+3. Wait for the environment to set up (this may take several minutes the first time)
+4. Once ready, you'll see a terminal with the environment initialized
+
+#### Usage
+
+After the codespace is created, the development environment will be automatically set up and services started.
+
+Access your running application at the URL shown in the forwarded ports panel (typically `https://your-codespace-name-3000.app.github.dev`).
+
+#### Port Configuration
+
+⚠️ **Important**: For external access to work properly, you must set forwarded ports to **Public** in the Ports tab:
+
+1. In your codespace, look for the "PORTS" tab in the bottom panel
+2. Click the lock icon to change visibility from "Private" to "Public"
+<br>
+
 ## Usage
 
 Start the development environment
@@ -170,6 +197,7 @@ To add/remove DNS records you can now use `./sndev domains dns`. More on this [h
 # Table of Contents
 - [Getting started](#getting-started)
     - [Installation](#installation)
+        - [GitHub Codespaces](#github-codespaces)
     - [Usage](#usage)
         - [Modifying services](#modifying-services)
             - [Running specific services](#running-specific-services)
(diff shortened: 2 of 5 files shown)


brymut:
Hi @huumn, just checking if this might be a workable solution?

brymut:
Hi @huumn, might this be something still being looked at?

sndev
```diff
@@ -481,6 +481,14 @@ sndev__login() {
   salt="202c90943c313b829e65e3f29164fb5dd7ea3370d7262c4159691c2f6493bb8b"
   # upsert user with nym and nym@sndev.team
   email="$1@sndev.team"
+
+  # Detect Codespaces and set the correct base URL
+  if [ -n "$CODESPACE_NAME" ]; then
+    BASE_URL="https://${CODESPACE_NAME}-3000.app.github.dev"
```
Soxasora, line 489: This is the right direction!

Soxasora (approved):
Awesome! I was able to create a Codespace and run sndev.

It might be out of scope for this PR but we need to have a way to automatically set the correct environment variables for Codespaces urls.

For example

```
# static things
NEXTAUTH_URL=http://localhost:3000/api/auth
```

If I want to fix auth stuff, I can't do it because it sends me to `localhost:3000`. The same applies for any other env that uses localhost: images, lnauth, etc.

```
Env:
NEXT_PUBLIC_MEDIA_URL=http://localhost:4566/uploads

Tunnel:
https://didactic-fiesta-4r547r79pj2jp4j-4566.app.github.dev/uploads
```

Now I can actually create an `.env.local` file myself and then restart... but it can be automated!

As a tip, you can create it automatically via bash with `postAttachCommand` in `devcontainer.json` or via the `./sndev` script.

If you want to do it in this PR, you're welcome to do so, otherwise this is good enough for me as `sndev` starts and I can make the changes myself.

brymut:
> If you want to do it in this PR, you're welcome to do so, otherwise this is good enough for me as `sndev` starts and I can make the changes myself.

Let me have a quick look

sndev
```diff
@@ -39,6 +39,20 @@ docker__exec() {
 sndev__start() {
   shift
 
+  # Override environment variables for Codespaces
+  if [ -n "$CODESPACE_NAME" ]; then
+    export NEXTAUTH_URL="https://${CODESPACE_NAME}-3000.app.github.dev/api/auth"
```
Soxasora: Hmm, this might fix broken services, but if the developer wants to change values they would still have to create an `.env.local` file manually.

The `sndev` script might not be the greatest place to create an `.env.local` file (because we would overwrite it on restart unless we do complex stuff). Instead, `postAttachCommand` on `devcontainer.json` would give you the ability to create it on Codespace creation, so that edits can persist through `sndev` restarts.

brymut: Yeah, was testing this out on workspaces, not final. Will ping when the solution is working

Soxasora: Thank you! I was indeed confused when to chime in. ^^

brymut: I think this should work now @Soxasora

.devcontainer/minimal/devcontainer.json
```diff
@@ -0,0 +1,22 @@
...
+        ]
+      }
+    },
+    "containerEnv": {
+      "CPU_SHARES_IMPORTANT": "1024",
+      "CPU_SHARES_MODERATE": "512",
+      "CPU_SHARES_LOW": "128",
+      "COMPOSE_PROFILES": "minimal"
+    },
+    "postAttachCommand": "./sndev start"
```
Soxasora: missing the script here

scripts/setup-codespaces.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+echo "Setting up Codespaces environment variables..."
+
+[ ! -f .env.local ] && touch .env.local && echo "Created .env.local" || echo ".env.local already exists, will preserve existing content"
+
+declare -A env_vars=(
+  ["NEXTAUTH_URL"]="https://${CODESPACE_NAME}-3000.app.github.dev/api/auth"
+  ["NEXT_PUBLIC_MEDIA_URL"]="https://${CODESPACE_NAME}-4566.app.github.dev/uploads"
+  ["MEDIA_URL_DOCKER"]="https://${CODESPACE_NAME}-4566.app.github.dev/uploads"
```
Soxasora: For congruence with .env.development this shouldn't be here. It works anyway because it's not a local address, but it's generally used by the worker container.

```env
# containers can't use localhost, so we need to use the container name
IMGPROXY_URL_DOCKER=http://imgproxy:8080
MEDIA_URL_DOCKER=http://s3:4566/uploads
```

Soxasora (commented):
SN on codespaces works great! The setup in the `sndev` script is nice to have when something's up with the `.env.local` file.

You need to fix some stuff I commented about and I can then approve it.

.devcontainer/minimal/devcontainer.json
brymut: sorry, don't know how I missed this. Fixed

brymut:
> It seems that we can't auto-set ports visibility ([limitation of the devcontainer spec](https://github.com/orgs/community/discussions/10394#discussioncomment-2018563)), but we can do it via GUI manually 
and it works.

ahhh, right, I could add this to the text in the setup to remind users to set the ports to public, otherwise, I'm not quite sure how to add to this.

README.md
```diff
@@ -356,7 +384,7 @@ We offer triage permissions to contributors after they've made a few contributio
 Contributors can get badges on their SN profiles by opening a pull request adding their SN nym to the [contributors.txt](/contributors.txt) file.
 
 ## What else you got
-In the future we plan to offer more, like gratis github copilot subscriptions, reverse tunnels, codespaces, and merch.
+We offer GitHub Codespaces support for cloud-based development! In the future we plan to offer more, like gratis github copilot subscriptions, reverse tunnels, and merch.
```
Soxasora: I mean we don't AFAIK, do we @huumn ?

> Codespaces created from a forked repository are billed to your personal account unless the upstream (or parent) repository is in an organization that has allowed you - as a member, or outside collaborator, of the organization - to use codespaces at the organization's expense.

This needs to be addressed before claiming that on the README

README.md
```diff
@@ -36,6 +36,33 @@ Go to [localhost:3000](http://localhost:3000).
...
+### GitHub Codespaces
+
+[![Open in GitHub Codespaces](https://github.com/codespaces/badge.svg)](https://codespaces.new/stackernews/stacker.news)
+
+You can run Stacker News on Github Codespaces
+
+#### Setup
+
+1. Open the repository on GitHub and click the **"Code"** button
+2. Select the CodeSpaces tab and create a new codespace. (or configure with provided settings first)
```
Soxasora: nit: *Codespaces

Also maybe we can explain that to access the minimal variant they have to press the `...` three dots and select `New with options...`

README.md
```diff
@@ -36,6 +36,33 @@ Go to [localhost:3000](http://localhost:3000).
...
+
+#### Usage
+
+After the codespace is created, the development environment will be automatically set up and services started.
+
+Access your running application at the URL shown in the forwarded ports panel (typically `https://your-codespace-name-3000.app.github.dev`).
+
+#### Port Configuration
+
+⚠️ **Important**: For external access to work properly, you must set forwarded ports to **Public** in the Ports tab:
```
Soxasora: It's not only about external access (which is still allowed if private via GitHub login), but also because internal services need to be able to contact these forwarded ports which they can't without the user's GitHub token.

sndev
```diff
@@ -507,6 +517,34 @@ USAGE
...
+
+USAGE
+  $ sndev setup_codespaces
+
+DESCRIPTION
+  Creates or updates .env.local with the correct URLs for GitHub Codespaces.
+  This is automatically run when the Codespace is created.
+
+REQUIREMENTS
+  - Set GITHUB_TOKEN in Codespaces settings for NextAuth functionality
```
Soxasora: No, this is not correct. If you got confused by my next auth example, the problem was that the tunneled URL can't be fetched without the GITHUB_TOKEN if the ports are not public.

sndev
```diff
@@ -507,6 +517,34 @@ USAGE
...
+  - Set forwarded ports to 'Public' in the Ports tab for external access
+
+WARNING
+  Without GITHUB_TOKEN, authentication may fail. Without public ports,
+  external access to the application will be blocked.
```
Soxasora: I don't think this is the correct place to say this, the readme is enough. Also still not correct, see above ^

README.md
brymut: crap, I didn't mean to send this.

brymut:
Apologies on the doc changes, that's the last time I'm letting AI summarise my changes. What a mess.

PR #2257, opened 2025-07-03
Add fireworks animation

ekzyis:
## Description

In case we want to add fireworks for July 4th again.

cherry-picked from b6628eb159c2574ca221c4a2beae429ae967ce7e and 3d093d45f3bbe930adcc5db3127938f43cd28b1d

However, I had to disable the firework trails because of https://github.com/tsparticles/tsparticles/discussions/4870.

## Video

https://github.com/user-attachments/assets/ae89c677-e817-41dd-82a2-d09cc11b2dbc

## Checklist

yes

`8`. See video

Yes, see video.

no


huumn:
lol I didn't know you did this

huumn:
fuck me

huumn:
> However, I had to disable the firework trails because ...

This was the case when we first shipped it too, but we "fixed" it by putting the fireworks on the background rather than foreground - which I also had to do in #2258.

This means the fireworks animate behind images and textareas/inputs, but it's probably worth the tradeoff? v3 of tsparticles might've solved it.

ekzyis:
> lol I didn't know you did this

Sorry I could have said something in the chat :eyes: 

> fuck me

"fuck me" because you forgot about it? :eyes: 

> > However, I had to disable the firework trails because ...
> 
> This was the case when we first shipped it too, but we "fixed" it by putting the fireworks on the background rather than foreground - which I also had to do in #2258.
> 

Mhh, I tried to get it to work as we had it (with trails and fireworks in the background) but after messing around for a while, I thought we changed our CSS too much in these two years.

> This means the fireworks animate behind images and textareas/inputs, but it's probably worth the tradeoff? v3 of tsparticles might've solved it. 

Mhh, right! I used v3 first but the fireworks didn't immediately show up so I thought it might had too many breaking changes. However, I didn't realize at that point that they were probably just hidden. Will check!

I also wasn't sure if you will have time to review and deploy this. I started doing this to send plebpoet some fireworks for SNZ.

Anyway, you don't have to spend time replying to this, I will check how you solved it in #2258 and see if I can fix the z-index issue with inputs.

huumn:
> "fuck me" because you forgot about it? 👀

Because I didn't check github before tinkering on this. I had ~1hr to hack yesterday and this sounded fun. I guess we both had the same idea.

PR #2259, opened 2025-07-04
Use tsparticles v3 for fireworks

ekzyis:
## Description

With tsparticles v3 and a [background mask](https://particles.js.org/docs/interfaces/tsParticles_Engine.Options_Interfaces_BackgroundMask_IBackgroundMask.IBackgroundMask.html), I was able to get fireworks show up above images and text inputs while avoiding https://github.com/tsparticles/tsparticles/discussions/4870. I tried the background mask in our v2 version but it didn't work as expected.

I let the fireworks preset configure most because I couldn't just copy the config since it now requires the engine as an argument for the color utils. I compared our options with the preset and I believe I applied our customization to the preset.

However, I can still see fireworks going off-screen on small sizes, so I am not sure if 573dd06305ac288bd01728057d5d4a8f140ffba1 worked or just wasn't applied here for some reason.

I need to go now, will check later.

Not sure how long we want to keep the fireworks though?

## Video

https://github.com/user-attachments/assets/de82fabb-37ef-49dc-987f-3ad4d58d8818

## Checklist

yes

`5`. Fireworks are now above images and text inputs but still go off-screen on small sizes though.

yes

no


huumn:
I think updating fireworks is fine. We'll only keep it for the weekend at most, but it'll be good to have for new years or next year.

PR #2261, opened 2025-07-04
Refactor animations

ekzyis:
## Description

based on #2260

This refactors how we provide animations to the rest of the code base.

- it is now very easy to switch between animations; no more search & replace of `useLightning` with new animation hook
- all logic related to animations is now in one place
  - animations don't have to check themselves if animations are enabled (#2260)
  - random animation if logged out is now triggered where one would expect and not in `AnonDropdown`

## Additional context

The snow animation doesn't seem to work anymore.

## Checklist

Yes. I now save if we ran the animation in `lnAnimated` instead of `striked` in localStorage to be consistent with the other localStorage flag (`lnAnimate`), but I still also check for `striked` on init.

`8`.

- animations still work
- local storage flags `lnAnimate` and `lnAnimated` still work
- searched for `useLightning` and `useFireworks` and found no other occurrences outside of components/animation

n/a

no


ekzyis:
bugbot run

components/animation/index.js
```diff
@@ -0,0 +1,74 @@
+import { useCallback, useEffect, useState } from 'react'
+import { useMe } from '@/components/me'
+import { randInRange } from '@/lib/rand'
+
+/* eslint-disable no-unused-vars */
+import { LightningProvider, useLightning } from './lightning'
+import { FireworksProvider, useFireworks } from './fireworks'
+import { SnowProvider, useSnow } from './snow'
```
huumn: I think it'd be best to comment these out too. The fireworks and snow animation libraries ~~will~~ may bloat the bundle if they aren't used.

ekzyis: Done!

PR #2262, opened 2025-07-05
enhance(carousel): Close carousel on browser back navigation

brymut:
## Description

~~Add to browser history stack when opening carousel~~ use URL hashes to track modal state, so that navigating back on browser should close the carousel modal, not leave the post entirely.

closes https://github.com/stackernews/stacker.news/issues/1508

## Screenshots
https://github.com/user-attachments/assets/cb85eb1a-c463-4799-8c11-d3c0451551b5

(Updated screen recording)

## Additional Context

## Checklist

Y

10

Y

N


brymut:
Happy to have a look again based on the cursor feedback, but I can confirm that the modal does close automatically upon navigating backwards, as shown in the PR description screen recording.

ekzyis (changes requested):
Closing the carousel now always scrolls us to the top:

https://github.com/user-attachments/assets/fcc14403-eeec-41c8-b0ab-64c01fa98a98

brymut:
> Closing the carousel now always scrolls us to the top:

Let me have a look at that

brymut:
> Closing the carousel now always scrolls us to the top:

Should work now, switched to checking the URL hash instead, less complicated and seems to not have the scrolling issue.

https://github.com/user-attachments/assets/cb85eb1a-c463-4799-8c11-d3c0451551b5

brymut:
I don't know I had accidentally deleted this one 😅

components/carousel.js
```diff
@@ -114,20 +114,39 @@ function CarouselOverflow ({ originalSrc, rel }) {
...
+
+    if (window.location.hash === '#carousel') {
+      const confirmedEntries = Array.from(media.current.entries())
+        .filter(([, entry]) => entry.confirmed)
+
+      if (confirmedEntries.length > 0) {
+        showCarousel({ src: confirmedEntries[0][0] })
+      }
+    }
+  }, [showCarousel])
```
Soxasora: This seems redundant. It’s very rare that it would find any confirmed entries since this runs before any media has been fully loaded; `confirmMedia` already does what this should do!

components/carousel.js
```diff
@@ -137,8 +156,16 @@ export function CarouselProvider ({ children }) {
     if (mediaItem) {
       mediaItem.confirmed = true
       media.current.set(src, mediaItem)
+
+      if (typeof window !== 'undefined' && window.location.hash === '#carousel' && !isCarouselOpen) {
+        const confirmedEntries = Array.from(media.current.entries())
+          .filter(([, entry]) => entry.confirmed)
+        if (confirmedEntries.length >= 1) {
+          showCarousel({ src })
+        }
```
Soxasora: **nitpick**: this is duplicate code, I wonder if this check can be done directly in `showCarousel` or with an helper function

components/modal.js
```diff
@@ -59,12 +59,22 @@ export default function useModal () {
...
+      )
+      if (hasModalWithHash) {
+        window.history.replaceState(window.history.state, '', window.location.pathname + window.location.search)
+      }
+    }
+
     while (modalStack.current.length) {
       getCurrentContent()?.options?.onClose?.()
       modalStack.current.pop()
     }
```
Soxasora: I see that this preserves any other hashes (like table of contents navigation) from being wrongly removed, so it might justify O(n) checks (`modalStack.current.some`).

I don't see this being that big of an impact performance-wise, and to avoid O(n) checks you would need to store the original hash in some way (like a ref), so it's ok.

components/carousel.js
```diff
@@ -114,20 +114,39 @@ function CarouselOverflow ({ originalSrc, rel }) {
 export function CarouselProvider ({ children }) {
   const media = useRef(new Map())
   const showModal = useShowModal()
+  const [isCarouselOpen, setIsCarouselOpen] = useState(false)
```
Soxasora: Reasonable solution to avoid duplicate modals for the way they work as of now. Maybe in the future we'd need to add deduplication to the modal system.

*edit: this conflicts with `confirmMedia` automatic opening of carousel, better to use a ref for this.

components/carousel.js
brymut: I thought so too, but cursor bot thought otherwise, seems to be quite strict on it

components/carousel.js
brymut: let me have another look

components/carousel.js
Soxasora: What did Mr. Cursor say? It has the tendency of being unaware of context sometimes.

brymut: Let me get rid of this and see if it re-comments, keeping cursor at bay with the idea I had initially has been quite confusing.

components/carousel.js
```diff
@@ -137,8 +148,15 @@ export function CarouselProvider ({ children }) {
...
+
+      if (typeof window !== 'undefined' && window.location.hash === '#carousel' && !isCarouselOpen) {
+        const confirmedEntries = getConfirmedEntries()
+        if (confirmedEntries.length >= 1) {
+          showCarousel({ src })
+        }
+      }
     }
-  }, [])
+  }, [showCarousel, isCarouselOpen])
```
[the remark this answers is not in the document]
brymut: Here we go @Soxasora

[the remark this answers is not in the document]
Soxasora: Ah! But that's not correlated with the `useEffect` we were talking about. This is because `confirmMedia` gets called for each image/video, so yes, this duplicates modals.

If you refresh the page with #carousel there are 4 modals, one per each image
image

So that's why a refresh would send me to the last image in the post! I recommend a ref instead of state to address this. I missed it in the previous review!

components/carousel.js
```diff
@@ -137,8 +151,18 @@ export function CarouselProvider ({ children }) {
     if (mediaItem) {
       mediaItem.confirmed = true
       media.current.set(src, mediaItem)
+
+      if (typeof window !== 'undefined' && window.location.hash === '#carousel' && !isCarouselOpenRef.current) {
+        if (!isOpeningCarouselRef.current) {
```
Soxasora: Is this secondary ref needed?

brymut: ahhh, right. All the cursor bot changes had me running in loops. Forgot that's not needed anymore. Removed. Sorry about that.

brymut:
Got it, didn't want to pollute with many small commit fixes.

huumn:
I haven't looked very closely at this, but I wonder if this was made more complicated by choosing to use a `hash` rather than a query parameter which would allow nextjs to manage the history for us. Again, I don't know, but `window.location` has a smell to it.

brymut:
> I haven't looked very closely at this, but I wonder if this was made more complicated by choosing to use a `hash` rather than a query parameter which would allow nextjs to manage the history for us. Again, I don't know, but `window.location` 
has a smell to it.

I just think that for a particularly client-side issue and I think that's what hash was intended for, to handle UI state. Also looking at the URL, `/?carousel=true` compared to `/#carousel` might make it seem like there's server action happening when there isn't. Happy to have another go with using params with nextjs router if that's what you're looking for particularly.

huumn:
> Happy to have another go with using params with nextjs router if that's what you're looking for particularly.

It's mostly a:

1. lines-of-code thing for me, I do not suspect it requires this much code to do this, and 
2. a *going behind nextjs router's back* using `window.location` 

If neither is related to the hash vs query param - great. If this is the best way to do it - that's fine too, but I will not be able to merge it until I've tried alternative approaches myself.

brymut:
> * lines-of-code thing for me, I do not suspect it requires this much code to do this, and

You were right, I overshot this, updated to use nextjs router instead. Mostly based off of the solution you were thinking about [here ](https://github.com/stackernews/stacker.news/issues/1508#issuecomment-2432121819)

brymut:
> I tried calling `router.back()` instead of `router.replace` in `onClose`, but that didn't seem to work as expected.

I think adding router.replace is better for not polluting the history stack, so just `.push` on opening and `.replace` on close and then track `router.query.imageId` 


https://github.com/user-attachments/assets/1a9d17c8-cf67-4de0-94dd-064c723e2075

huumn:
A lot the related code here may change with the new editor, so I'm closing, but that's our fault and not yours, so I will pay the bounty.

PR #2263, opened 2025-07-05
Cache compiled static assets in service worker

hkarani:
## Description

## Additional Context

Caching static assets can only be done after Next compiles and the static assets  exist in .next/static folder.
The static assets can then be added as urls in precache-manifest.json for the service worker.

This solution utilizes Next's _compiler.hooks.afterEmit_ hook to add the static assets paths in precache-manifest.json after compilation is done and the assets are present in .next/static folder.

No

This change will modify the contents of precache-manifest.json when _npm run build_ is run to generate static assets.
## Checklist

Yes.

7

To test this, directly run _npm run build_ and check the contents of precache-manifest.json after the Next finishes the compilation step.
 
This change will affect the cached offline  page that relies on the compiled static assets as in this [issue](https://github.com/stackernews/stacker.news/issues/419).

Broswer cache storage contains all items in precache-manifest.json.

No.


ekzyis:
Can you fix the [linting issues](https://github.com/stackernews/stacker.news/actions/runs/16090562094/job/45406956621?pr=2263)? The lines where only a semicolon was added / whitespace was removed add noise to the change set; making the review harder.

Issue #2256, opened 2025-07-02
Stacker News favicon always shows red "notification" alert, even with no notfications


itsrealfake:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Browser: Safari

Favicon always loooks like this, even when i'm logged out.

![Image](https://github.com/user-attachments/assets/3d4b3f8e-72b9-479c-8aed-4c22d0eb9200)

![Image](https://github.com/user-attachments/assets/1b44a1ba-4a51-4bb1-ad9a-5ab3ee9e6696)

### Steps To Reproduce

Not sure how to reproduce this... don't know if anybody else is experiencing it.

### Expected behavior

I would expect that the Favicon had the operation it used to have (maybe 1.5 months i've noticed this?) which would only show the red square when i had a notification in the application

### Device information

- OS: Mac OS (the new one, a M2 Air)
- Browser: Safari

axelvyrn:
this is happening to me as well, have you bookmarked SN?
I am on Microsoft Edge though

Soxasora:
[Dynamic favicons are broken on Safari](https://developer.apple.com/forums/thread/103767). Speculating on this, it might be because of Safari internal cache policies, and because Apple [doesn't want this](https://bugs.webkit.org/show_bug.cgi?id=95979#c2).

friendly reminder that Safari is the new IE

axelvyrn:
> [Dynamic favicons are broken on Safari](https://developer.apple.com/forums/thread/103767). Speculating on this, it might be because of Safari internal cache policies, and because Apple [doesn't want this](https://bugs.webkit.org/show_bug.cgi?id=95979#c2).
> 
> friendly reminder that Safari is the new IE

In that case:
1. Will changing the favicon URL each time (e.g., by appending a query string like `?v=${Date.now()}`), which can force Safari to reload the favicon help?
2. Or should using both `rel="icon"` and `rel="shortcut icon"` for the PNG favicons do it?

Soxasora:
Well my caching theory was just a theory, Apple's official stance is that they don't comfortably allow favicon manipulation via JS.

Test with Safari though.

PR #2264, opened 2025-07-06
Fix Force favicon refresh with cache-busting for notification state (Safari compatibility)

axelvyrn:
## Description

Closes https://github.com/stackernews/stacker.news/issues/2256
Changed the favicon URL each time (by appending a query string `?v=timestamp`), which can force Safari to reload the favicon each time the notification state changes. This increases the likelihood that Safari and other browsers will reload the favicon immediately.

## Screenshots
not really applicable but still I clicked before and after screenshots for unread notifs thrice, while the tabs were bookmarked
Screenshot 2025-07-06 171744

Screenshot 2025-07-06 172000
## Additional Context

When I had an unread notification, I bookmarked the tab while it was unread and of course it saved the icon with the red square. Then I reloaded the site and read the notif, this updated both the bookmarked icon and the actual one. Next 5 times I refreshed the site at different time intervals; there was no error.

## Checklist

yes

9

yes

none

### DIFF
--- a/components/nav/common.js
+++ b/components/nav/common.js
@@ -15,7 +15,7 @@ import { useServiceWorker } from '../serviceworker'
 import { signOut } from 'next-auth/react'
 import Badges from '../badge'
 import { randInRange } from '../../lib/rand'
-import { useFireworks } from '../fireworks'
+import { useLightning } from '../lightning'
 import LightningIcon from '../../svgs/bolt.svg'
 import SearchIcon from '../../svgs/search-line.svg'
 import classNames from 'classnames'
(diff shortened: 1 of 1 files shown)


Soxasora:
Are you testing on Safari? Just making sure you're on the right path as the favicon reliably changes on other browsers, just not Safari (in the issue I linked Apple's response)

axelvyrn:
yuh, i had to download safari all the way, log into github, check 2FA and then log into sndev

axelvyrn:
> Manipulating the DOM like this discouraged in react. Use the relevant components to do this.

done :)

Soxasora:
> yuh, i had to download safari all the way, log into github, check 2FA and then log into sndev

...you can't download safari, did you mean a macOS VM?

image

It doesn't work btw

axelvyrn:
browserstack

axelvyrn:
i downloaded it from browserstack (sort of macOS vm)

and as for that, i was afraid of this exact thing.
when I checked with unread notifs, it worked for the initial method with DOM manipulation but I did not check with the later change since it does the same thing but I was unsure whether that would work ESPECIALLY for safari.

next.js <Head> merges tags, and if the only change is the href, some browsers may not reload the favicon.
Safari is especially stubborn and may require the tag to be removed and re-added in the real DOM.

the approach using <Head> with a cache-busting query string is React-compliant, but some browsers (especially Safari) may not update the favicon reliably unless the <link rel="icon"> tag is actually replaced in the DOM, not just updated via React's virtual DOM. but @huumn says that it is discouraged.
I understand that React expects to be the only thing updating the DOM. If you change the DOM outside of React, React may not know about those changes, which can cause bugs if React later tries to update or remove those elements.
For most UI elements, this can cause issues like event handlers not working, or UI not updating as expected.
But the favicon <link> tag is not something React ever tries to update except in this one place.
The favicon is not part of the interactive UI, so there is no risk of breaking user interactions or React state.
This pattern (removing and re-adding the favicon tag) is widely used in the React and Next.js community for this exact problem.

---
1. Vercel’s Next.js Examples
Source: [next.js/examples/with-favicon](https://bookish-zebra-5g5pgj6775g4fv6p5.github.dev/)
Excerpt:

> “Safari is notorious for caching favicons. The only reliable way is to remove and re-add the favicon link tag in the DOM.”

2. react-favicon NPM Package
Source: [react-favicon on GitHub](https://bookish-zebra-5g5pgj6775g4fv6p5.github.dev/)
Excerpt:

> This package manipulates the DOM directly to update the favicon.

3. Stack Overflow Discussions
Source: [How to dynamically change a favicon in React?](https://bookish-zebra-5g5pgj6775g4fv6p5.github.dev/)
Excerpt:

> “You have to remove the old favicon and add a new one. This is a browser thing, not a React thing.”

4. react-helmet-async Issue
Source: [react-helmet-async#94](https://bookish-zebra-5g5pgj6775g4fv6p5.github.dev/)
Excerpt:

> “Safari does not update the favicon unless the link tag is removed and re-added. This is not something React can control.”

5. React Community Gist
Source: [React dynamic favicon gist](https://bookish-zebra-5g5pgj6775g4fv6p5.github.dev/)
Excerpt:

> “To force a favicon refresh, remove and re-add the link tag.”

---

I am sorry if I prove myself incompetent but that was the only fix I could wrap my mind around. I'll revert the changes. I'll be glad to know if there is another alternative fix to the problem.

Soxasora:
There's no need to feel bad about **trying**. I did warn you that Safari **intentionally** doesn't support favicon changes via JS, because of some privacy/user-protection things or whatever excuse they come up with.

It's not the first time that we had to come to terms with Safari behaviors, and it won't be the last time.

One thing: **QA, QA, QA.** It's important. It's clear that you might be learning, just don't expect things to magically work if you don't test them or slightly don't understand them, especially if it's the product of predictive models.

Thank you for trying.

PR #2265, opened 2025-07-06
Remove service worker logger

ekzyis:
## Description

This removes the service worker logger that was storing log messages in our db and sending them to Slack.

This is the first part of cleaning up the service worker by removing all code we don't need and was only added to debug #411.

## Checklist

Yes. I didn't drop the `Log` table but marked it with `@@ignore` so in case an old client with diagnostics enabled still tries to send a log message to an old server instance running the old Prisma client, the query won't fail. It should not be a problem if it fails since the fetch call has a catch handler, but I added `@@ignore` anyway.

Adding `@ignore` to the `diagnostics` column is important though because we poll for this every second in the `ME` query, so the column should not be dropped until all clients are running the new code which no longer polls for that field.

The GraphQL fields for `diagnostics` have thus been marked with `@deprecated`.

`8`. Build and run the code, logged in as a user, no errors and searched for any occurrences that I might have missed but found none.

n/a

no


ekzyis:
I put this back into draft because I will remove more stuff related to service worker logging in this PR

huumn:
@Soxasora I'd like your eyes on this to make sure we don't break iOS push notifications again.

ekzyis:
> @Soxasora I'd like your eyes on this to make sure we don't break iOS push notifications again.

This only removes code that is unrelated to push message processing but still a good idea to have @Soxasora take a look

ekzyis:
> <h3>Bug: Push Subscription Toggle Fails on Storage Persist Error</h3>
>
> The `togglePushSubscription` function's error handling for `navigator.storage.persist()` was removed during the async/await refactor. Previously, errors were caught and logged, allowing the function to complete. Now, if `persist()` fails, the error propagates, causing `togglePushSubscription` to reject and potentially breaking push notification functionality or leading to unhandled rejections.

All calls of `togglePushSubscription` attach a catch handler

Soxasora:
By just looking at the code I can say that I'm first of all very happy that we're getting rid of this unused code, but also exactly that: we're not using getOS anymore to enable iOS-specific workarounds as everything in the event listener is cross-platform now.

So, no, this can't possibly break push notifications for iOS, but I'm still going to perform a deeper QA to make sure that everything's fine (service worker logs shouldn't throw an out-of-event error, because of some arbitrary non-specified timing rules)

PR #2269, opened 2025-07-07
Enhancements to live comments

Soxasora:
## Description
Continues #2115 and #791 
UX and logic enhancements to the **Live Comments** system

- Recursively show all new comments of a thread
- Highlight newly injected comments
- Change the favicon when there's new comments

## Additional Context

To enable further enhancements and changes to the live comments system, parts of the original code are being changed.

## TODO

Show all new comments:
- merge component and functions with the main single-inject ones
- strong QA

Highlighting:
- all of it

Favicon:
- all of it

## Checklist

Yes, these are completely new additions, no breakage or changes on the way comments and comment injection works.

5, WIP and in iterative QA

Yes, iterative QA

n/a


huumn:
Looks great for the most part. I'm seeing some weird behavior in corner cases though.

1. I had two tabs, one per user, start commenting on an item without comments. For some reason, the tab that commented first did not get updates from the other tab until I refreshed the page. The console said there's a duplicate `CommentFields` fragment. I'm not sure if that's related but it might be.
2. something strange happens when I edit a comment when there `show new comments` on it. It behaves like we've reached the depth limit.

Screenshot 2025-07-23 at 9 48 52 AM

3. the language here is a bit awkward 'show all new 1 comment'

Screenshot 2025-07-23 at 9 48 52 AM
Screenshot 2025-07-23 at 9 54 41 AM

Soxasora:
> 1. I had two tabs, one per user, start commenting on an item without comments. For some reason, the tab that commented first did not get updates from the other tab until I refreshed the page. The console said there's a duplicate CommentFields fragment. I'm not sure if that's related but it might be.

This is indeed weird, I'm trying to reproduce this but I can't, I'm now directly looking at how I use fragments.

> 2. something strange happens when I edit a comment when there show new comments on it. It behaves like we've reached the depth limit.

I think I found the cause, for bottomed out status we check `(item.comments?.comments.length === 0 && item.nDirectComments > 0)`. Applying an edit triggers the mutation COMMENT_UPDATE which gets new data from the server, thus nDirectComments would grow. But it's a mixed state, it should also get comments.comments. I'm working on a fix.

> 3. the language here is a bit awkward 'show all new 1 comment'

I thought this too, let's see if I can find something actually better

lib/apollo.js
```diff
@@ -323,6 +323,17 @@ function getClient (uri) {
...
+              merge (existing, incoming, { variables, readField }) {
+                // if the item has new comments, don't merge the nDirectComments field
+                // preventing new comments not yet injected from being counted, thus causing bottomedOut
+                if (variables?.id && readField('newComments')?.length > 0) {
+                  return existing
+                }
+
+                return incoming
+              }
+            },
```
Soxasora: on comment edits, `upsertComment` updates the cache with new values, which includes an updated `nDirectComments` but **not comments**, thus causing `bottomedOut` (`item.comments?.comments.length === 0 && item.nDirectComments > 0`) to trigger

So in the case of a comment with pending `newComments`, I think it's ok and harmless to ignore `nDirectComments` updates.

https://github.com/user-attachments/assets/0c26ff8e-d4e3-4b92-929b-ba80abf7af16

edit: pushed a fix for incorrect usage of `readField`, it was missing `'from'` where it should read this field... even though it was already reading it ... anyway, my tests shows that this works correctly.

PR #2270, opened 2025-07-07
Fix missing dependency in snow callback

ekzyis:
fix https://github.com/stackernews/stacker.news/pull/2261#pullrequestreview-2995064996

(state setters function have a stable identity so they don't need to be included as a dependency)

### DIFF
--- a/components/animation/snow.js
+++ b/components/animation/snow.js
@@ -21,7 +21,7 @@ export const SnowProvider = ({ children }) => {
     }
     setStartIndex(i % MAX_FLAKES)
     setFlakes(newFlakes)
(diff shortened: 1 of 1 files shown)


huumn:
I think we removed that because this caused an infinite rerender or something. iirc Snow is broken anyway right? 

No harm in merging though I guess.

ekzyis:
> I think we removed that because this caused an infinite rerender or something

Oh, right

> iirc Snow is broken anyway right?

Yeah I tried it out but nothing happened. Maybe also some z-index issue.

I left a TODO so [we will remember](https://www.youtube.com/watch?v=oKw33h9J5Fs).

Issue #1880, opened 2025-02-04
PRIORITY ISSUE TREE [experimental]


huumn:
I'm trying something new. 

Personally, I like trees better than what the duck other people use like spreadsheets and kabana boards or whatever. 

Issues, along with prs, are fundamental "atoms" of our workflow here, so it makes more sense to me to just create relationships between atoms in a familiar place.

m0wer:
Hi!

Is this up to date? What are the most important issues to work on?

Are you looking for frequent “external” contributions or just sporadic?

My goal is to help in the most useful way without rushing you more than you want.

huumn:
This is mostly up to date. I'd recommend not working on territory or wallet stuff as we are working on refactoring those and you'll get a lot of conflicts.

In general though, for FOSS contributions, I would recommend working on things you find interesting. This tree is mostly for me/ekzyis/soxasora.

adlai:
> This is mostly up to date. I'd recommend not working on territory or wallet stuff as we are working on refactoring those and you'll get a lot of conflicts.

github is worse than useless

PR #2272, opened 2025-07-08
Cache compiled static assets in service worker

hkarani:
## Description

This change patches the cached items array in the service worker at public/sw.js to include Next's compiled static assets emitted during build phase.

 WHY?
 
The service worker ONLY includes the assets in precache-manifest.json injected by Workbox plugin when the Webpack build starts but our static assets only exist later in the build.
 
Static assets cannot be added in the server build , they are emitted in the client build phase([docs](https://nextjs.org/docs/14/app/api-reference/next-config-js/webpack))  and are not needed in dev.
   
However since the precache-manifest.json assets are hardcoded in an array in the generated service worker by Worbox in public/sw.js. We patch it to include our  compiled assets after Next compiles and emits the files.

## Screenshots

Where workbox stores the items in precache-manifest.json in Q=JSON.parse('the-precache-manifest.json') in the service worker.
Screenshot 2025-07-08 at 19 55 39

## Additional Context

Screenshot 2025-07-08 at 19 52 55

Cached items are referenced as  /_next/static/chunks/pages/offline-eb65c6c23790b89c.js" depending on the build version.

They static assets urls are formatted to be web urls for them to be accessible by the browser after caching.

How to test a dev build?

## Checklist

7

This change adds the cached static assets in .next/static folder to the service worker in public/sw.js. The changes can be replicated by running:

> npm run build

then checking the changes of the public/sw.js in the editor just after Next completes the compilation phase.

To test dev build run:

> npm run dev 

This change will affect the cached[ offline page](https://github.com/stackernews/stacker.news/issues/1880?issue=stackernews%7Cstacker.news%7C612).

No


hkarani:
@ekzyis I think the cursor bug bot fix should be left out because, the JSON in public/sw.js is not escaped so it should be left as is.

Aslo, do we need to cache the .js.map files in the .next/static assets?

hkarani:
@ekzyis I've implemented a new way to do this mainly because the service worker will only cache items in precache-manifest.json when the build starts, even updating precache-manifest.json mid-webpack build will still cache assets passed when webpack starts.

huumn:
did you close this intentionally?

hkarani:
Hey, yes. Couldn't find the best way to test the cached assets in dev.

It's a solution anyone can take ideas from if they need to work on caching static assets.

PR #2274, opened 2025-07-08
Service Worker rewrite

ekzyis:
## Description

based on #2273 related to #1881

This removes all code related to merging notifications. Only `your item stacked X sats` is still using the [`tag`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/tag) property which is handled by the device if it supports it.

~This also removes badge _counts_. Afaict, they weren't working properly on iOS anyway (it always showed 1) and I think it's not that useful to have a number of notifications. We also don't show a number in our UI.~

Badge counts actually do work on iOS. Instead of state in IndexedDB, we now use the length of the array returned by `getNotifications` as the source of truth.

This PR also fixes extremely slow notification clicks. We were always opening a new window instead of trying to focus an existing client. Since we can't tell the client to navigate somewhere from the service worker, we send the client a message so it does so itself.

## Additional Context

**Push manager with notification queue for deduplication**

We currently manually have to make sure we don't send multiple push notifications for the same event. As we experienced, this is prone to errors. For example, we still don't do this for reply + mention. This means we send out two push notifications even though in /notifications, we only show they were mentioned.

A better idea is to queue all push notifications and before sending them out, remove all except one for the same item. This way, it's also easier to implement which notification should be sent if there are multiple for the same item.

I will implement this in some future PR.

## Checklist

yes, deploying this won't break existing clients while the service workers update themselves

`8`. Tested with iOS PWA. Push notifications are still received, and badge count is updated properly. Notifications are closed when we visit /notifications and the badge is cleared.

Since none of the code is OS-specific and Android has less to no quirks, it should work exactly the same on Android, except that Android can also replace notifications.

n/a

no


sw/index.js
```diff
@@ -68,7 +72,109 @@ setDefaultHandler(new NetworkOnly({
...
+        { icon: '/icons/icon_x96.png' }
+      )
+      return
+    }
+
+    await self.navigator.setAppBadge?.()
+    await self.registration.showNotification(payload.title, payload.options)
+  }
+  event.waitUntil(onPush())
+})
```
ekzyis: @Soxasora, I will want to have your eyes on this. I removed all the BS regarding merging notifications, so only these 25 lines will remain in the service worker to show push notifications.

I think this properly calls `event.waitUntil` as mentioned [here](https://web.dev/articles/push-notifications-handling-messages#wait_until) so there won't be any silent pushes, right?

Soxasora: We might need to avoid using `await`s and instead switch to the Promises pattern to ensure that we actually parallelize `setAppBadge` with `showNotification` and avoid timing problems.

Consider returning the promise of `showNotification` in the catch, and a promise array of `setAppBadge` and `showNotification` for the rest, this way we can `Promise.all` them in the `waitUntil`
<sup>what a sentence</sup>

Or, you can actually avoid creating another function and do it right inside function that takes `(event)`.

I agree that awaits are much more readable than the Promises hell but we don't really want to wait for `setAppBadge` to do its thing before showing the notification. This can break some obscure timing that Apple imposed, which is much much more strict than every other browser.

The whole reason why we were actually losing subscriptions and notifications was that we were (a)waiting too much!

ekzyis: Ohh, that makes sense, thank you!

Did that in ee4c670c

PR #2281, opened 2025-07-16
Allow 0 pct forwards to pass only in the client

hkarani:
## Description

The client validation is changed to allow 0 by:

-  Changing the minimum allowed input to zero({min=0} in AdvPostForm component
-  Modifying the form validation in lib/validate.js to allow a minimum of 0

To only allow 0 to only pass the client:

-  The forwards are validated  to be > 0  when  _normalizeForwards_ is called before mutuation in [use-item-submit.js](https://github.com/stackernews/stacker.news/blob/master/components/use-item-submit.js#L62). _normalizeForwards_  filters to remove forwards with 0 as pct.

## Screenshots

Allow-0-pct-client-forwards

## Additional Context

No

## Checklist

Only form validation and is changed when forwards are normalized.

8

Yes

No

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -146,7 +146,7 @@ export function advPostSchemaMembers ({ me, existingBoost = 0, ...args }) {
             },
             message: 'cannot forward to yourself'
           }),
-        pct: intValidator.required('must specify a percentage').min(1, 'percentage must be at least 1').max(100, 'percentage must not exceed 100')
+        pct: intValidator.required('must specify a percentage').min(0, 'percentage must be at least 0').max(100, 'percentage must not exceed 100')
       }))
       .compact((v) => !v.nym && !v.pct)
       .test({
--- a/components/adv-post-form.js
+++ b/components/adv-post-form.js
@@ -282,7 +282,7 @@ export default function AdvPostForm ({ children, item, sub, storageKeyPrefix }) 
                     name={`forward[${index}].pct`}
                     type='number'
                     step={5}
-                    min={1}
+                    min={0}
                     max={100}
                     append={<InputGroup.Text className='text-monospace'>%</InputGroup.Text>}
                     groupClassName={`${styles.percent} mb-0`}
--- a/lib/form.js
+++ b/lib/form.js
@@ -2,7 +2,7 @@ import { hasDeleteMention, hasReminderMention } from './item'
 
 /**
  * Normalize an array of forwards by converting the pct from a string to a number
- * Also extracts nym from nested user object, if necessary
+ * Also extracts nym from nested user object, if necessary and removes 0 pct fowards
  * @param {*} forward Array of forward objects ({nym?: string, pct: string, user?: { name: string } })
  * @returns normalized array, or undefined if not provided
  */
@@ -10,7 +10,7 @@ export const normalizeForwards = (forward) => {
   if (!Array.isArray(forward)) {
     return undefined
   }
-  return forward.filter(fwd => fwd.nym || fwd.user?.name).map(fwd => ({ nym: fwd.nym ?? fwd.user?.name, pct: Number(fwd.pct) }))
+  return forward.filter(fwd => (fwd.nym || fwd.user?.name) && Number(fwd.pct) > 0).map(fwd => ({ nym: fwd.nym ?? fwd.user?.name, pct: Number(fwd.pct) }))
 }
 
 export const toastUpsertSuccessMessages = (toaster, upsertResponseData, dataKey, itemText) => {


lib/validate.js
```diff
@@ -145,7 +145,7 @@ export function advPostSchemaMembers ({ me, existingBoost = 0, ...args }) {
             },
             message: 'cannot forward to yourself'
           }),
-        pct: intValidator.required('must specify a percentage').min(1, 'percentage must be at least 1').max(100, 'percentage must not exceed 100')
+        pct: intValidator.required('must specify a percentage').min(0, 'percentage must be at least 0').max(100, 'percentage must not exceed 100')
```
ekzyis, line 149: Mhh, this will allow to insert 0 pct forwards via the API because the server doesn't run `normalizeForwards`. Tested using the GraphQL playground at /api/graphql:

```gql
mutation upsertDiscussionWithForwards(
  $title: String!,
  $text: String!,
  $sub: String!,
  $forward: [ItemForwardInput]
) {
  upsertDiscussion(
    title: $title,
    text: $text,
    sub: $sub,
    forward: $forward
  ) {
    result {
      id
    }
  }
}
```

variables:

```json
{
  "title": "foobar",
  "text": "asdf",
  "sub": "art",
  "forward": [{"nym":"anon","pct":0}]
}
```

I wish we could just do this:

```diff
diff --git a/lib/validate.js b/lib/validate.js
index 576687e6..68b5b204 100644
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -129,6 +129,7 @@ export function advPostSchemaMembers ({ me, existingBoost = 0, ...args }) {
       }),
     forward: array()
       .max(MAX_FORWARDS, `you can only configure ${MAX_FORWARDS} forward recipients`)
+      .transform(forwards => forwards.filter(fwd => Number(fwd.pct) > 0))
       .of(object().shape({
         nym: string().required('must specify a stacker')
           .test({
```

but we can't since the transform only applies to the array during validation, the submitted array will be unmodified.

Maybe the proper fix would be to make validation return the transformed values in general; or only do so on the client, but the server always validates what it receives without any transforms. But this is a more general thing we would have to change in our code base.

Not sure how to fix this problem of allowing the client to submit something that the server shouldn't accept without too many changes and not making it confusing how it works :/

ekzyis (changes requested):

lib/validate.js
hkarani: Thanks for the feedback @ekzyis , I'll work to resolve this.

Curiously, do you use real sats in dev? I'd like to test this more before submitting the next change.

ekzyis: > Curiously, do you use real sats in dev? I'd like to test this more before submitting the next change.

No, we don't, we have a regtest lightning network with multiple nodes.

When you set `COMPOSE_PROFILES=minimal,payments` in .env.local before you run `sndev start`, it will start them. See [this section](https://github.com/stackernews/stacker.news?tab=readme-ov-file#running-specific-services) in the README.

But you can also use CCs via `sndev set_balance <user> <balance>`.

hkarani: Thanks for the clarification

hkarani: > Mhh, this will allow to insert 0 pct forwards via the API because the server doesn't run `normalizeForwards`. Tested using the GraphQL playground at /api/graphql

I've tested this:
When normalizeForwards doesn't filter 0 pct forwards

https://github.com/user-attachments/assets/b1489988-06f4-44af-90fa-16180a3583d6


Here we actually filter 0 before calling the mutation


https://github.com/stackernews/stacker.news/pull/2281/files#r2240699282

In use-item-submit https://github.com/stackernews/stacker.news/blob/6aeffa7affaf4ae1215cc16dc107280ab04bd266/components/use-item-submit.js#L19,  upsertItem gets the the forwards after they are 0 pct items are filtered https://github.com/stackernews/stacker.news/blob/6aeffa7affaf4ae1215cc16dc107280ab04bd266/components/use-item-submit.js#L62

I still consider this the cleanest solution  with no many code changes.

 You can try it, test it yourself from the client and check the db for the items Forwards pcts, if this doesn't work for all intents and puporses this PR should be closed.

hkarani: Uploading Screen Recording 2025-07-29 at 21.18.53.mov…

hkarani: https://github.com/user-attachments/assets/c1136f7d-c1cf-4b9b-9a8e-b89c0981f03c

huumn:
did you close this intentionally?

hkarani:
@ekzyis mentioned an issue with this. That's why I closed this

lib/validate.js
ekzyis: You were looking at code that only runs in the client. What I meant [here](https://github.com/stackernews/stacker.news/pull/2281#discussion_r2228668320) was that in theory, one does not have to use the client to make requests to our API. I gave you the GraphQL playground as an example.

As-is, the server does not filter out zero pct forwards but assumes the client already did it.

It is important in general that the server does not trust the client with inputs.

hkarani: Caught that

PR #2286, opened 2025-07-16
Fix wallets error handling order

ekzyis:
## Description

I started to show any error during fetching or decrypting the wallets to the user in https://github.com/stackernews/stacker.news/pull/2284.

However, I didn't notice that this meant that this will block the prompt to unlock the wallets if the device has the wrong key since obviously, decryption will fail in that case (even though I wrote QA `10` :raised_eyebrow::see_no_evil:)

In #2282, I noticed that my design of handling different errors by figuring out the importance of errors thus which error could or couldn't overwrite another error didn't scale, but instead of acknowledging that the design was bad, I only refactored the code but kept the same design.

Now, I have a variable for each independent error we can have, and we decide during rendering which error we're going to show first.

So following the render order, we can now easily tell that we show states in the following order:

1. key storage unavailable
2. wrong key on device
3. error fetching wallets
4. error decrypting wallets
5. loading wallets
6. user has no wallets
7. user has wallets

## Checklist

yes

`11` :sweat_smile: 

I tested by manufacturing the errors from the bottom up.

1. Always return `loading: true` in `useWalletsQuery` → loading spinner forever
2. Error during decryption→ decryption error
3. Server throws error while fetching the wallets → failed to load wallets
5. Remove key → unlock wallets
6. key storage unavailable -> this device does not support storage of cryptographic keys via IndexedDB

see video:

https://github.com/user-attachments/assets/6ceb3adf-6c04-4eb3-81b3-cf8927fed100

n/a

no

### DIFF
--- a/pages/wallets/index.js
+++ b/pages/wallets/index.js
@@ -1,6 +1,6 @@
 import { getGetServerSideProps } from '@/api/ssrApollo'
 import { Button } from 'react-bootstrap'
-import { useWallets, useTemplates, DndProvider, Status, useStatus } from '@/wallets/client/context'
+import { useWallets, useTemplates, DndProvider, KeyStatus, useWalletsLoading, useKeyError, useWalletsError } from '@/wallets/client/context'
 import { WalletCard, WalletLayout, WalletLayoutHeader, WalletLayoutLink, WalletLayoutSubHeader } from '@/wallets/client/components'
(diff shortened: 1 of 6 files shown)


ekzyis:
lol, Cursor is right, my Q&A only worked because `undefined === undefined` is true and I initialize the error with `null`

PR #2290, opened 2025-07-17
Fix old key overwrites new key

ekzyis:
## Description

We were always using the old key to decrypt wallets if it was found. 

This means that if device sync was enabled before #2169, the existing key could decrypt the migrated wallets.

However, when the passphrase is shown, we encrypt the wallets with a new key. Same is true on passphrase reset.

But on page load, we will still load the old key instead of the new key so the 'unlock wallets' prompt is shown.

This PR fixes this by only loading the old key if a new key was not found.

## Checklist

yes

`8`. We now always use the new key if it exists. If not but old key exists, we use that. Else we generate a new key.

n/a

no

### DIFF
--- a/wallets/client/context/hooks.js
+++ b/wallets/client/context/hooks.js
@@ -152,16 +152,11 @@ export function useKeyInit () {
       try {
         // TODO(wallet-v2): remove migration code
         //   and delete the old IndexedDB after wallet v2 has been released for some time
-        const oldKeyAndHash = await loadOldKey()
-        if (oldKeyAndHash) {
-          // return key found in old db and save it to new db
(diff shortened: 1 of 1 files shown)


ekzyis:
I'm not sure if this is the bug that [siggy47](https://stacker.news/items/1041515?commentId=1041539), [BlokchainB](https://stacker.news/items/1041505?commentId=1041536) or [OT](https://stacker.news/items/1042814?commentId=1042948) experienced since they somehow recovered, but afaict, they cannot recover from this bug unless they manually deleted their browser data to wipe the old key and I don't think they did that.

PR #2292, opened 2025-07-17
Fix passphrase scanner

ekzyis:
## Description

tbd

## Checklist

yes

tbd

n/a

no


components/form.js
```diff
@@ -1350,8 +1350,10 @@ function PasswordScanner ({ onScan, text }) {
                 <Scanner
                   formats={['qr_code']}
                   onScan={([{ rawValue: result }]) => {
-                    onScan(result)
-                    onClose()
+                    if (result) {
+                      onScan(result)
+                      onClose()
+                    }
```
ekzyis, line 1356: not sure why it sometimes returns an empty string as the result

PR #2297, opened 2025-07-18
Fix count of viewed comments of a post

Soxasora:
## Description

When we reply to a post, we update the count of comments we’ve viewed in that post to avoid seeing the new comment indicator on our own replies.
But instead of counting, we're actually concatenating strings with what we retrieve in the local storage.

This PR parses `commentsViewNum` as Number, allowing real counts.

## Screenshots

3 replies gets saved as `111` (as 3 times +1)
image

## Additional Context

This actually **still worked anyway**, because to search for new comments we do (other than checking the time) `commentsViewedNum < item.ncomments`. So `6 + 1` would do `61` instead of `7`, and the comparison would  still give the expected behavior.

## Checklist

Yes

7, tested where it pertains

n/a

n/a

### DIFF
--- a/components/item-info.js
+++ b/components/item-info.js
@@ -110,7 +110,7 @@ export default function ItemInfo ({
         </>}
       <Link
         href={`/items/${item.id}`} onClick={(e) => {
-          const viewedAt = commentsViewedAt(item)
+          const viewedAt = commentsViewedAt(item.id)
           if (viewedAt) {
             e.preventDefault()
             router.push(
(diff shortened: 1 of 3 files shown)


ekzyis (commented):
Should `commentsViewedAt` below also return a number?

Since this is not the first time and won't be the last time we have a bug like this, I think it would be cool to have a way to access localStorage in a type-safe way.

Soxasora:
> Should commentsViewedAt below also return a number?

> I think it would be cool to have a way to access localStorage in a type-safe way.

I agree, I think we can also make a `commentsViewedNum(itemId)` function, both would receive an item ID and return `Number(result)`. I like the consistency of this.

PR #2299, opened 2025-07-19
fix: push notificationclick event not triggered

### LINKED ISSUE #2293: iOS push notification deep linking not working

### Describe the bug

It seems that clicking on SN iOS notifications via PWA do not launch me into the specific post/comment that triggered the notification anymore. Definitely used to. 

I haven’t noticed if it’s 100% of the time or not, but I’ll keep an eye out. 

### Screenshots

_No response_

### Steps To Reproduce

1. Receive iOS push notification from SN PWA
2. Click notification 
3. see the home page of SN is opened, not the item which triggered the notification 

### Expected behavior

Should open the corresponding item (post, comment, etc)

### Logs

_No response_

### Device information

iOS 17

### Additional context

_No response_

Soxasora:
## Description

**drafted for evaluation**

~~Potentially fixes #2293~~ Fixes notificationclick event from not being dispatched

Theory:
 - Sequentially updating the app badge count after showing a push notification (extending the event time), caused the notification to be shown out-of-time, potentially causing `notificationclick` to not trigger because the browser wouldn't trust the notification.

This PR removes the support for app badge counts, to be re-instated in the near future.

## Screenshots

https://github.com/user-attachments/assets/12097cc0-37fb-4b3d-95c1-1431252f340b

## Additional Context

I still think that app badge counts can be safely implemented, before showing a notification, we just need to do it manually.
Sketch:

```javascript
const { tag } = payload
if (getNotifications({ tag }.length === 0) {
  setAppBadge + 1
}
```

## Checklist

yes

7, tested in local with Chrome and Safari, tested via HTTPS on iOS

n/a

n/a

### DIFF
--- a/sw/index.js
+++ b/sw/index.js
@@ -93,11 +93,7 @@ self.addEventListener('push', function (event) {
     )
   }
 
-  event.waitUntil(
-    self.registration.showNotification(payload.title, payload.options)
-      .then(() => self.registration.getNotifications())
-      .then(notifications => self.navigator.setAppBadge?.(notifications.length))
-  )
+  event.waitUntil(self.registration.showNotification(payload.title, payload.options))
 })
 
 self.addEventListener('notificationclick', function (event) {
--- a/sw/index.js
+++ b/sw/index.js
@@ -99,11 +99,10 @@ self.addEventListener('push', function (event) {
 self.addEventListener('notificationclick', function (event) {
   event.notification.close()
 
-  const promises = []
   const url = event.notification.data?.url
   if (url) {
     // First try to find and focus an existing client before opening a new window
-    promises.push(
+    event.waitUntil(
       self.clients.matchAll({ type: 'window', includeUncontrolled: true })
         .then(clients => {
           if (clients.length > 0) {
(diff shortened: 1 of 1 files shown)


Soxasora:
https://github.com/stackernews/stacker.news/blob/ba370eeda67a190c3e723f2f2ac6fdd72d728342/sw/index.js#L208-L214

This piece hasn't been removed to clear badges for users that have them, it can be removed, but badges can get stuck.

ekzyis (approved):
Approving because I am okay with testing if this makes `notificationclick` long-term reliable in prod.

We might still need to fix the code inside it though, see https://github.com/stackernews/stacker.news/issues/2293#issuecomment-3091235112 (`event.preventDefault()` and `WindowClient.navigate()` instead of `client.postMessage()`)

Soxasora:
> forgot to mention, but your video is zoomed in so I can't tell what you're doing

oh sorry, the video is cropped in a way that shows only the before and after of this PR diff

> (event.preventDefault() and WindowClient.navigate() instead of client.postMessage())

I saw this, but I really couldn't reproduce it after this PR, I'm going to test various PWA states

Soxasora:
> We might still need to fix the code inside it though for devices that can't use https://github.com/stackernews/stacker.news/pull/2300, see https://github.com/stackernews/stacker.news/issues/2293#issuecomment-3091235112 (event.preventDefault() and WindowClient.navigate() instead of client.postMessage())

Just tried, out of luck both with this fix and your draft. Now that I recall, I tried almost everything when it came to navigation on click on iOS, when I was trying to fix push notifications as a whole, and that revealed to be a lost cause.

`notificationclick` events on iOS are fundamentally broken, so far only your workaround and Declarative Web Push have fixed this.

I think it's better to keep your workaround than returning to _absolutely nothing happening_ on notification click.

---

This PR still fixes the `notificationclick` event attachment to the notification by the browser.

PR #2301, opened 2025-07-21
Add wallet debug page

ekzyis:
## Description

This adds a page at /wallets/debug that shows relevant information for debugging.

There's no link to it from /wallets since we can easily link to it on demand in a reply or similar.

No need to expose this directly in the user interface. 

## Screenshots

localhost_3000_wallets_debug(iPhone SE)

## Checklist

yes, only added a column

`8`. Page loads fine via SSR or client-side navigation (even as anon), and all values are correct.

yes

no


ekzyis:
Putting back in draft to fix the bug @cursor found and will also add a timestamp for when the local key was updated last

PR #2307, opened 2025-07-22
Wallet debug logs

ekzyis:
## Description

Based on #2306 

This adds debug wallet logs. They can only be viewed and deleted at /wallets/debug.

### Questions

- should debug wallet logs be opt-in?
- should debug wallet logs be anonymized via [fancy names](https://github.com/stackernews/stacker.news/blob/master/lib/fancy-names.json)?

## Screenshots

2025-07-22-020428_1920x1080_scrot

## Checklist

yes

tbd

n/a

no


ekzyis:
> ### Bug: State Initialization Fails to Sync with Object Changes
> The `diagnostics` state is initialized from `me?.privates?.diagnostics` using `useState`. However, this state is not updated when the `me` object changes. This causes the local state to become out of sync with the server, particularly if `me` is initially null and then populated (e.g., user logs in), or if the user's diagnostics setting changes from another source. Consequently, the UI displays an incorrect state until a page refresh.

Mhh, true, it currently only works because the components that use `useWalletLogger` unmount when I visit /wallets/debug and will mount again with the updated value, but that's bad design.

Will use `me?.privates?.diagnostics` as the source of truth!

Thanks @cursor

Issue #2288, opened 2025-07-16
Passphrase QR code scanner doesn't work


ekzyis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

When I scan something, it closes the modal instead of only going back to the passphrase text input.

This can be fixed via this:

```diff
diff --git a/components/form.js b/components/form.js
index 44c1c5e1..c863b182 100644
--- a/components/form.js
+++ b/components/form.js
@@ -1351,7 +1351,7 @@ function PasswordScanner ({ onScan, text }) {
                   formats={['qr_code']}
                   onScan={([{ rawValue: result }]) => {
                     onScan(result)
-                    onClose()
+                    onClose({ back: 1 })
                   }}
                   styles={{
                     video: {
@@ -1364,7 +1364,7 @@ function PasswordScanner ({ onScan, text }) {
                     } else {
                       toaster.danger('qr scan: ' + error?.message || error?.toString?.())
                     }
-                    onClose()
+                    onClose({ back: 1 })
                   }}
                 />
               )}
```

But then it still doesn't populate the text input.

Maybe it's related to the component remounting on scan?

Also, the scan only seems to work in light mode.

### Steps To Reproduce

1. Click 'passphrase' on device A. 
2. Click 'unlock wallets' on device B.
3. Scan QR code from device A on device B.

### Expected behavior

Passphrase text input is correctly filled out with the scanned value

PR #2315, opened 2025-07-23
Fix useMemo returns new component on change

ekzyis:
## Description

Related to #2288 

I noticed that everything works as expected if I don't return a function in `useMemo`.

That means I can't use JSX syntax, but I think this fixes it because else I would return a completely new component which React does not recognize as the same (??)

## Checklist

yes

`10`. Tested scanning with and without remounting while scanner is open.

n/a

no


huumn:
The identity of everything but primitive types changes on rerenders - functions, arrays, etc.

PR #2318, opened 2025-07-23
short cirtuit out of live comment query if possible

huumn:
there's no pressing need to do this, but if the query planner is behaving, this should prevent unnecessary index scans when we know an item does not have new comments (`path <@ NULL` always evaluates to false).

afaict this shouldn't have any adverse effects.


Soxasora:
> If lastCommentAt is NULL (e.g., for items without prior comments), the condition NULL > $2 evaluates to NULL, causing the subquery to return no rows.

This is indeed, the desired effect! When a new comment gets created, the root item will have `lastCommentAt` updated, so this is actually perfect.

PR #2319, opened 2025-07-23
Fix clear new comments on route change

Soxasora:
## Description

Fixes bottomed out comments retaining their `newComments` field by clearing it on route change.

## Screenshots

https://github.com/user-attachments/assets/abb64564-1bcf-4142-b089-2a599531fae1

## Checklist

Yes

7, on route changes, only items that have the newComments array populated are being cleared

n/a

n/a

### DIFF
--- a/components/comment.js
+++ b/components/comment.js
@@ -156,6 +156,27 @@ export default function Comment ({
     }
   }, [item.id, rootLastCommentAt])
 
+  useEffect(() => {
+    // clear new comments when navigating away
+    const eraseNewComments = () => {
+      if (item.newComments?.length > 0) {
+        cache.writeFragment({
+          id: `Item:${item.id}`,
+          fragment: gql`
+            fragment NewComments on Item {
+              newComments
+            }`,
+          data: {
+            newComments: []
+          }
+        })
+      }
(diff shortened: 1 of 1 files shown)


huumn:
rather than triggering as a side effect of a route change, couldn't we just clear it when `view replies` is pressed?

Soxasora:
> rather than triggering as a side effect of a route change, couldn't we just clear it when `view replies` is pressed?

Mmhh actually yes, it shouldn't be a problem to not clear `newComments` when going to other paths, because coming back will still mean that you didn't see those new comments.

components/comment.js
```diff
@@ -326,7 +306,7 @@ export function ViewAllReplies ({ id, nshown, nhas }) {
   )
 }
 
-function ReplyOnAnotherPage ({ item }) {
+function ReplyOnAnotherPage ({ item, cache }) {
```
huumn: I'd prefer we not inject the cache if we can just use the `useApolloClient` hook to get it again.

Soxasora: Sorry about it, loopy. Fixed it asap. 🫡

PR #2324, opened 2025-07-24
Fix non-thread live comments recursion logic

### LINKED ISSUE #2321: Existing comments of new comments are not being counted

### Describe the bug

Existing children of incoming new comments are not counted or injected

### Screenshots

6 comments incoming but shows 4 comments

https://github.com/user-attachments/assets/9e6f5619-2632-461c-81a2-d6212eedd303

### Steps To Reproduce

1. Write 3 comments one after another
2. Wait for the top-level button to show
3. Write again 3 comments one after another
4. The top-level button stops at the first 3 comments because it won't recurse on `newComments` of children
5. Injection will give you the possibility to then show all the other comments

### Expected behavior

The new comments count should also consider `newComments` of already existing comments

### Logs

_No response_

### Device information

_No response_

### Additional context

I must've missed this during `show all new comments for a thread` re-implementation

Soxasora:
## Description
Fixes #2321 

Traversing existing comments only if it's a thread comment, caused inconsistencies for new comments arriving with their own existing children.

A better approach is to traverse existing comments only if it's a thread comment, **or** we got past the parent comment and are analyzing its new comments subtree.

## Screenshots

https://github.com/user-attachments/assets/3042e377-df38-4e2a-a8fe-cbf27f177238

## Additional Context
Let's examine the following comment structure with the actual recursion logic:

```
Root Item: 3 new comments
├── New Top-level Comment A
│   ├── New Comment B
│   └── (new) Existing Comment C
│       └── New Comment D
```

The count shows 3 new comments, but clearly these are 4 comments.

New Comment D, which is a new comment to an existing comment, got ignored
What happens is that we check new comments of Comment A, but not the new comments of its existing children.

The most straightforward solution would be to also count the existing comments
Let's zoom out on the structure:

```
Root Item: 6 new comments
├── New Top-level Comment A
│   ├── New Comment B
│   └── (new) Existing Comment C
│       └── New Comment D
├── Existing Top-level Comment E
│   ├── New Comment F
│   └── Existing Comment G
│       └── New Comment H
```

Top-level should now show 4 new comments, but it doesn't, instead it shows 6 new comments because existing Top-level Comment E has 2 new comments.

This happened because we're checking the Root Item, and the new recursion logic checks for new comments in existing children, like Comment E.

This is what we want for thread comments, to show all new comments, also of existing comments, but not what we want for normal comment injection.

#### Solution

When traversing a given item, start by checking just its new comments, then of these new comments, check also their existing comments. If it's a thread, check its existing comments from the beginning.

## Checklist

Yes

7, multiple comment structures, tbd more QA

n/a

n/a

### DIFF
--- a/components/show-new-comments.js
+++ b/components/show-new-comments.js
@@ -93,7 +93,8 @@ function traverseNewComments (client, item, onLevel, currentDepth = 1, inSubtree
     }
   }
 
-  // check for new comments in existing children, only if we're in a new comment subtree
+  // check for new comments in existing children
+  // only if we're in a new comment subtree, or it's a thread
   if (inSubtree && item.comments?.comments) {
     for (const child of item.comments.comments) {
       traverseNewComments(client, child, onLevel, currentDepth + 1, true)
@@ -164,7 +165,7 @@ export function ShowNewComments ({ topLevel, item, sort, depth = 0 }) {
   const thread = item.path?.split('.').length === 2
 
   // recurse through all new comments and their children
-  // if the item is a thread, we consider every existing child comment
+  // if the item is a thread, we also consider all of their existing children
   const { newCommentsCount, threadChildren } = countAllNewComments(client, item, depth, thread)
 
   // only if the item is a thread and its children have new comments, we show "show all new comments"
--- a/components/show-new-comments.js
+++ b/components/show-new-comments.js
@@ -69,8 +69,8 @@ function prepareComments (data, client, newComments) {
 }
 
 // traverses all new comments and their children
-// if we're showing all new comments of a thread, we also consider their existing children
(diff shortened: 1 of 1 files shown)


ekzyis:
I'm probably missing a lot of context since I haven't really followed how live comments are implemented, but this is confusing me:

>
> ```
> Root Item: 3 new comments
> ├── New Top-level Comment A
> │   ├── New Comment B
> │   └── Existing Comment C
> │       └── New Comment D
> ```

How can C not be new when A is new?

Soxasora:
Live comments just fetches new comments of a root item after the lastCommentAt, with a polling of 10 seconds.

If during those 10 seconds a new comment and a reply to that new comment are created, they arrive together, not separated. So Comment C is new, just already injected in its parent comment.

This technique also allows us to save on extra injections!

PR #2326, opened 2025-07-24
Fix carousel dropdown

ekzyis:
## Description

I noticed that @brymut fixed something in #2262 that we haven't even noticed yet.

The `... > view original` is never updated while we move around in the carousel, so clicking on `view original` always opens the image with which we entered the carousel. This PR fixes this.

Seems like this has always been a problem. In #1425, I did pass `setOptions` to `<Carousel>` but never used it, so I removed it in #2325 today.

I set @brymut as the author of the commit so rewards go to him.

## Checklist

yes

`8`. Dropdown is now properly updated while moving around in the carousel

n/a

no

### DIFF
--- a/components/carousel.js
+++ b/components/carousel.js
@@ -53,11 +53,18 @@ function useArrowKeys ({ moveLeft, moveRight }) {
   }, [onKeyDown])
 }
 
-function Carousel ({ close, mediaArr, src }) {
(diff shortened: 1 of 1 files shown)


ekzyis:
I will add a row to awards.csv manually

PR #2328, opened 2025-07-24
Extending awards.csv

ekzyis:
Manually added a row to awards.csv because of #2326


brymut:
Thanks for to this @ekzyis, I appreciate it.

PR #2349, opened 2025-07-28
Fix missing reuse of nav.module.css

ekzyis:
## Description

based on #2348 

I forgot about the tab navigation in profiles and /withdraw in #2337

This also fixes this layout in /withdraw using iPhone SE dimensions in the mobile emulator because it will now use the changes from #2337 for /withdraw, too:

localhost_3000_withdraw(iPhone SE)

However, I do have an iPhone SE though but it doesn't look like that in prod, only in the mobile emulator :thinking: 

It breaks when the width is less than 386px.

## Checklist

yes

`8`. Tested tab nav in profile and /withdraw

n/a

no


components/user-header.module.css
```diff
@@ -3,24 +3,6 @@
...
-    justify-content: start;
-    font-size: 110%;
-}
-
-.nav :global .nav-link {
-    padding-left: 0;
-}
-
-.nav :global .nav-item:not(:first-child) {
-    margin-left: 1rem;
```
ekzyis, line 17: I assume we didn't specifically want to have 1rem gaps in profiles, because the CSS defined in nav.module.css uses 0.5rem gaps via the `gap` property

Issue #2371, opened 2025-07-31
hashtag links in Bios open new tab rather than scroll to heading


ed-kung:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Clicking hashtag links in items normally scroll you to the heading, but in a bio page it opens a new tab.

### Steps To Reproduce

1. Go to https://stacker.news/SimpleStacker
2. Click "Guides" or "Pleb Economists" or any of the links at the top.

### Expected behavior

I think expected behavior would be to scroll to the referenced heading.

### Device information

OS: Windows
Browser: Brave (Chromium)

ed-kung:
I believe https://github.com/stackernews/stacker.news/pull/2213 actually includes a fix:

https://github.com/stackernews/stacker.news/blob/8be52dda7579171d3eb8da109943904e7d0b3c0c/components/text.js#L145-L149

ekzyis:
Can you put that fix into a separate PR?

PR #2373, opened 2025-07-31
fix hashtag links opening in new tabs

ed-kung:
## Description

Closes https://github.com/stackernews/stacker.news/issues/2371

## Checklist

Yes... I think so, I can't imagine anyone using a hashtag link previously for any other purpose.

9

n/a

no


huumn:
Oh I think it's this way because folks can link to hash links on other sites. 

This is the right behavior if it's a relative path like this.

Thanks!

PR #2376, opened 2025-07-31
Fix spacing if no badges

ekzyis:
before:

2025-07-31-231947_437x92_scrot

after:

2025-07-31-232002_429x96_scrot

### DIFF
--- a/components/badge.js
+++ b/components/badge.js
@@ -46,6 +46,8 @@ export default function Badges ({ user, badge, className = 'ms-1', badgeClassNam
     })
   }
 
+  if (badges.length === 0) return null
+
   return (
     <span className={className}>
       {badges.map(({ icon, overlayText, sizeDelta }, i) => (


huumn:
I think the spacing, given the lack of a separator, helps distinguish the time from the other metadata. I get that there's a lack of consistency though which feels especially true when the edit timer is there. 

It's a tough call imo. It's clearly wrong to you?

ekzyis:
> It's clearly wrong to you?

mhh, it was but now no more haha

> helps distinguish the time from the other metadata

I guess what confused me is that there is actually no extra space in front of the time with badges because the `ms-1` comes from the badges and is in front of them:

2025-08-01-005604_411x102_scrot

with badges:

> nym `ms-1` badges `<span> </span>` time

without badges:

> nym `ms-1` `<span> </span>` time

huumn:
I'll merge it for now and see how it feels in prod. Not a big deal either way.

PR #2377, opened 2025-07-31
live comments: comments navigation

### LINKED ISSUE #2364: Navigate between new comments
In general, every outlined new comment should be navigable. Probably a fixed `next` button to catch up with new comments.

This is a must for the live comments auto-show enhancement, as comments may arrive without the user noticing

Soxasora:
## Description

Closes #2364 
The comments navigator tracks and gives navigation capabilities to new unread comments, outside of the viewport, with smooth scrolling.
Each new comment's `ref` gets tracked together with its `createdAt`, allowing the navigator to sort in-place.

Scrolling to a comment will remove it from the tracked refs, if it has tracked children, they're also getting removed.
As visual feedback, the comment gets briefly outlined with a fading animation.

The UI re-utilizes the minimalistic new comments pulsing dot, with a count of new unread comments, and it's placed in the navigation bar. To achieve this, the navigator context exposes a stable navigator object, limiting re-renders on different types of navbars.

#### Shortcuts
- Arrow Right or Tap: next comment
- Escape or Long press: clear navigator

## Media

Scroll by `createdAt`, count also children nodes

https://github.com/user-attachments/assets/a90056a5-459c-46e3-8052-4f9aa5f2e6c7

## Additional Context

From the usage of `refs`, to the manipulation of individual frames via `requestAnimationFrame`, the navigator is built to be performant. But Safari might have its own idea of animations, so we should keep an eye on it.

## Checklist

Yes

7, iterative QA

Yes

n/a


components/use-comments-navigator.js
```diff
@@ -0,0 +1,200 @@
...
+    // requestAnimationFrame to ensure untracking is processed before outlining
+    window.requestAnimationFrame(() => {
+      node.classList.add('outline-it')
+    })
+
+    // untrack the new comment and clear the outlines
+    untrackNewComment(ref, { includeDescendants: true, clearOutline: true })
+
+    // if we reached the end, reset the navigator
+    if (list.length === 1) clearCommentRefs()
```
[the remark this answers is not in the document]
Soxasora: Each `scrollToComment` call consumes the picked element, and since we work within the same array, to know if we reached the last tracked comment, we check if the array we're working with has `length === 1`

components/comments.js
```diff
@@ -94,11 +98,11 @@ export default function Comments ({
         : null}
       {pins.map(item => (
         <Fragment key={item.id}>
-          <Comment depth={1} item={item} rootLastCommentAt={lastCommentAt || parentCreatedAt} {...props} pin />
+          <Comment depth={1} item={item} navigator={navigator} rootLastCommentAt={lastCommentAt || parentCreatedAt} {...props} pin />
```
huumn, line 101: Any particular reason why you inject the navigator when it's available by a context?

It's generally a good idea to do dep injection, so I think we can keep it, but I'm just curious. Do you anticipate needing multiple non-global contexts?

huumn (approved):
Works great and code is well contained in the hook. Nice work!

~~As discussed in chat, I think we want to remove the pulsing animation on the indicator, then we can merge.~~

Welp, looks like you already did it lol.

components/comments.js
Soxasora: Mainly to avoid re-renders of every `Comment` whenever the `commentCount` state changes in context, but also to eliminate the need for the navigator context when we want to use `Comment` without a navigator, like `CommentFlat`. The navigator is instead a stable reference, we can pass this without risking any re-render ^^

The beauty of this is that (as you guessed!) we can have parallel providers if in the future we want to introduce multiple threads (and multiple navigators)

Issue #2278, opened 2025-07-12
remove free comments


huumn:
we suspect free comments might be doing more harm than good

we want to experiment with removing them entirely

bios will still be free but comments will not

completing the issue will also require updating the FAQ and any other copy that mentions the ability to comment for free

pory-gone:
@huumn 
Would you also like to drop the freebie fields and completely erase the logic?

huumn:
keep it so we can reintroduce it if we need to

pory-gone:
All right! :)

m0wer:
I've unassigned myself. When I gave it a try a while ago I found out that the change is more complex than it looks. It touches a lot of places and was not comfortable with forgetting some or possible side effects.

PR #2386, opened 2025-08-02
removing free comments

pory-gone:
## Description
fix issue #2278 
removed the ability to make free comments, limiting the freebies function to bios only

## Screenshots

https://github.com/user-attachments/assets/ab7ceab4-7e9c-4bd7-ba2b-fa354560a2d0

## Additional Context
I have not completely removed the free comment logic (even from databases and other functions) to let you test the change

changes in `api/paidAction/itemCreate.js` : an action is now considered a "freebie" only if it's a bio, not comments. The condition that checked whether the user had disabled freebies has also been removed, to potentially remove the logic.

changes in `components/fee-button.js` : The feature now also accepts a bio parameter. The label has been updated to display "bio cost," "comment cost," or "post cost" depending on the content type. The logic for freebies has been changed to `allowFreebies: bio`, meaning they are only allowed for bios.

all good

## Checklist

y

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**7

### DIFF
--- a/api/paidAction/itemCreate.js
+++ b/api/paidAction/itemCreate.js
@@ -49,10 +49,10 @@ export async function getCost ({ subName, parentId, uploadIds, boost = 0, bio },
           FROM upload_fees(${me?.id || USER_ID.anon}::INTEGER, ${uploadIds}::INTEGER[]))
       + ${satsToMsats(boost)}::INTEGER as cost`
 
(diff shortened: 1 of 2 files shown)


ekzyis:
Why was this closed? :eyes:

pory-gone:
I'm having some doubts about the impact of the changes and what to actually keep and remove, tomorrow I will reopen the PR and proceed with the work :P

PR #2403, opened 2025-08-07
Fix territory edits

ekzyis:
## Description

Not sure how that happened since I mentioned I did test territory edits in #2382 :sweat_smile: 

## Additional Context

## Checklist

yes

`9`. Tested territory create+edit and item create+edit (all callers of `uploadFees` function)

n/a

no


huumn:
You changed this in two places. Do we know which caused the bug?

ekzyis:
Yes, territory edits broke because `totalFees` wasn't `BigInt` but `Number`. `Number` worked with territory create but not with territory edits. So returning `BigInt` fixes territory edits, but requires a change to territory create.

Sorry for this, this was a silly mistake

PR #2410, opened 2025-08-09
fix wrong URL on Reply on another page

### LINKED ISSUE #2409: "Reply on Another Page" has confusing behavior

### Describe the bug

When you click "Reply on Another Page", it takes you to the reply box of an item that is one of the parents of the comment you're trying to reply to, and not the comment itself.

If the user is not paying attention, they will inadvertently reply to the wrong item.

### Screenshots

https://github.com/user-attachments/assets/17f42b56-ff22-443c-b6d8-08d49d1eb566

### Steps To Reproduce

Go to any comment that has "Reply on Another Page", click it.

### Expected behavior

After clicking the link, the browser should focus on the reply box of the comment that user was trying to reply to.

### Logs

_No response_

### Device information

OS: Windows
Browser: Brave (Chromium)

### Additional context

_No response_

Soxasora:
## Description

Fixes #2409 
After merging together the bottomed out and pagination link buttons, I wrongly deleted the `commentId` param that scrolls the user to the correct comment.
This PR re-instates the correct URL.

## Screenshots
n/a

## Additional Context
n/a

## Checklist

Yes

9, scrolls to the correct comment

n/a

n/a

### DIFF
--- a/components/comment.js
+++ b/components/comment.js
@@ -318,7 +318,7 @@ export function ViewMoreReplies ({ item, navigateRoot = false }) {
   const root = useRoot()
   const id = navigateRoot ? commentSubTreeRootId(item, root) : item.id
 
-  const href = `/items/${id}` + (navigateRoot ? '' : `?commentId=${item.id}`)
+  const href = `/items/${id}` + (navigateRoot ? `?commentId=${item.id}` : '')
 
   const text = navigateRoot && item.ncomments === 0
     ? 'reply on another page'
--- a/components/comment.js
+++ b/components/comment.js
@@ -280,7 +280,7 @@ export default function Comment ({
       </div>
       {collapse !== 'yep' && (
         bottomedOut
-          ? <div className={styles.children}><div className={classNames(styles.comment, 'mt-3 pb-2')}><ViewMoreReplies item={item} navigateRoot /></div></div>
+          ? <div className={styles.children}><div className={classNames(styles.comment, 'mt-3 pb-2')}><ViewMoreReplies item={item} threadContext /></div></div>
           : (
             <div className={styles.children}>
               {item.outlawed && !me?.privates?.wildWestMode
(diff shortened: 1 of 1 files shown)


components/comment.js
```diff
@@ -318,7 +318,7 @@ export function ViewMoreReplies ({ item, navigateRoot = false }) {
   const root = useRoot()
   const id = navigateRoot ? commentSubTreeRootId(item, root) : item.id
 
-  const href = `/items/${id}` + (navigateRoot ? '' : `?commentId=${item.id}`)
+  const href = `/items/${id}` + (navigateRoot ? `?commentId=${item.id}` : '')
```
ekzyis: `navigateRoot` is meant to navigate to the root, so I don't think this should be swapped

Soxasora: It's actually a very bad naming on my part. `navigateRoot` means that the `reply on another page` button will send you to the `rootId` of the subtree of which the comment is part of, giving you some context, but it should also scroll you to the comment that made you click that button.
<sup>sorry maybe a bad explanation</sup>

Here's the previous version of `ReplyOnAnotherPage`

```javascript
function ReplyOnAnotherPage ({ item }) {
  const root = useRoot()
  const rootId = commentSubTreeRootId(item, root)

  let text = 'reply on another page'
  if (item.ncomments > 0) {
    text = `view all ${item.ncomments} replies`
  }

  return (
    <Link href={`/items/${rootId}?commentId=${item.id}`} as={`/items/${rootId}`} className='d-block pb-2 fw-bold text-muted'>
      {text}
    </Link>
  )
}
```

The reason for the merge of the two components was to not duplicate the same `newCommentsDot` code, and also future outlining support.

I'll change the name with something more appropriate!

ekzyis: Mhh, I thought I've seen `navigateRoot` in our code long before live comments, but can't find it being mentioned in a commit before. Guess I only saw it in the first live comments PR then. Sorry for the confusion!

PR #2411, opened 2025-08-10
auto show: use textarea as anchor and 

Soxasora:
## Description

Reopens #2407 
If in the viewport we have a textarea, `preserveScroll` will use it as the anchor element to determine if something has moved.
If multiple textareas are present in the viewport, the focused one is preferred, otherwise the first in the tree will be picked as anchor.

#### Jitteriness
Also removes the `MutationObserver` used to get calculations on the most updated DOM. It was a precautionary measure that got obsolete by using two RAFs for the textarea anchor support.
Scroll preservation now also bails when there's some form of scrolling happening, avoiding jittery scrolling.

## Screenshots

Textarea is the preferred anchor element, with fallback

https://github.com/user-attachments/assets/4eec08c5-d2ee-4292-81b8-e790078e8888

<sup>incredibly fumbled on the compression settings</sup>

Difference between MutationObserver and two RAFs

## Additional Context

This required an additional layer of `requestAnimationFrame`: some browsers may have their own anonymous content, like text extensions and such, deferring textarea/input renders.

It still fallbacks to picking a ref at the middle of the viewport in absence of textareas

## Checklist

Yes

7, Brave, Safari, Chrome

n/a

n/a

### DIFF
--- a/components/preserve-scroll.js
+++ b/components/preserve-scroll.js
@@ -59,7 +59,7 @@ export default function preserveScroll (callback) {
         const refMoved = newRefTop - refTop
 
         // if the anchor ref moved, we need to scroll to the new position
-        if (refMoved !== 0) {
+        if (refMoved > 0) {
           window.scrollTo({
             // some browsers don't respond well to fractional scroll position, so we round up the new position to the nearest integer
             top: scrollTop + Math.ceil(refMoved),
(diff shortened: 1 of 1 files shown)


Soxasora:
While getting rid of the `MutationObserver` did reduce the jitteriness, preserving the scroll position can still cause it. I'm drafting this to take a closer look at the jittery scrolling.

edit: I'm not perceiving jitteriness, `MutationObserver` was indeed causing page blocks and the textarea deferred rendering might have contributed to this.

PR #2420, opened 2025-08-13
remove free comments

### LINKED ISSUE #2278: remove free comments

pory-gone:
## Description
fix #2278 

- Backend (`api/paidAction/itemCreate.js`):

Freebie only for bio: "free" condition now only applies to bio, not comments.
Respecting user preference: `disabledFreebies` check has been reinstated also for bio ones (if on, nothing free).

- Frontend fees (`components/fee-button.js`):

`allowFreebies` moved from comments to bio; correct cost label (bio/comment/post).
The UI no longer shows "free" comments; it is still possible for bio where conditions exist and freebies are not disabled.
Why: visual consistency with new server behavior.

- Settings (`pages/settings/index.js`):

Text made generic ("Some content.") instead of "Some comments.".
Why: Don't create misleading copy now that comments aren't free.

- FAQ (`docs/user/faq.md`):

Updated: bio can be free (limited visibility), posts/comments incur fees; updated "last updated".
Why: make documentation consistent with the new policy.

Outcome: comments can no longer be free; bio can still be free for users with insufficient balance (`except disabledFreebies`). UI and doc now harmonious. Test to reduce adverse effect of free comments by keeping the logic ready to restore.

## Screenshots

https://github.com/user-attachments/assets/f61ecc46-9313-4c1c-8fe6-02fa4667e687

## Additional Context

I also tested with `disableFreebies` set on and the creation of bios with 0 credits generates an invoice, as it should do.

Not intentionally removed:

Schema/fields/flags "freebie" and filters "freebies" remain (DB, fragments, constants) for the sake of compatibility, history and effortless future reactivation.

## Checklist

**Are your changes backward compatible? Please answer below:** yes

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:** 7 

**Did you introduce any new environment variables? If so, call them out explicitly here:** n/a

### DIFF
--- a/pages/settings/index.js
+++ b/pages/settings/index.js
@@ -259,11 +259,11 @@ export default function Settings ({ ssrData }) {
             label={
               <div className='d-flex align-items-center'>disable freebies
                 <Info>
-                  <p>Some comments can be created without paying. However, those comments have limited visibility.</p>
+                  <p>Some content can occasionally be created without paying, but such content has limited visibility.</p>
 
-                  <p>If you disable freebies, you will always pay for your comments and get standard visibility.</p>
+                  <p>If you disable freebies, you will always pay standard fees and get standard visibility.</p>
 
-                  <p>If you attach a sending wallet, we disable freebies for you unless you have checked/unchecked this value already.</p>
+                  <p>If you attach a sending wallet, we may disable freebies for you unless you have checked/unchecked this value already.</p>
                 </Info>
               </div>
             }
--- a/api/paidAction/itemCreate.js
+++ b/api/paidAction/itemCreate.js
@@ -49,10 +49,12 @@ export async function getCost ({ subName, parentId, uploadIds, boost = 0, bio },
           FROM upload_fees(${me?.id || USER_ID.anon}::INTEGER, ${uploadIds}::INTEGER[]))
       + ${satsToMsats(boost)}::INTEGER as cost`
 
-  // sub allows freebies (or is a bio or a comment), cost is less than baseCost, not anon,
-  // cost must be greater than user's balance, and user has not disabled freebies
-  const freebie = (parentId || bio) && cost <= baseCost && !!me &&
-    me?.msats < cost && !me?.disableFreebies && me?.mcredits < cost
+  // freebies currently only apply to bios (comments disabled). To keep togglable behavior,
+  // respect user's disableFreebies preference for bios as well.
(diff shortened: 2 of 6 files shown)


pages/settings/index.js
```diff
@@ -259,11 +259,11 @@ export default function Settings ({ ssrData }) {
...
               <div className='d-flex align-items-center'>disable freebies
                 <Info>
-                  <p>Some comments can be created without paying. However, those comments have limited visibility.</p>
+                  <p>Some content can occasionally be created without paying, but such content has limited visibility.</p>
 
-                  <p>If you disable freebies, you will always pay for your comments and get standard visibility.</p>
+                  <p>If you disable freebies, you will always pay standard fees and get standard visibility.</p>
 
-                  <p>If you attach a sending wallet, we disable freebies for you unless you have checked/unchecked this value already.</p>
+                  <p>If you attach a sending wallet, we may disable freebies for you unless you have checked/unchecked this value already.</p>
```
ekzyis, line 266: Why the 'may' here?

But I think we can remove this setting completely anyway, see review comment.

ekzyis (commented):
_I have only looked at the changes so far, no testing or checking if that's all the changes we need, but looks good to me so far._

I think we can remove the `disableFreebies` setting all together. Afaict, it's only there because bios can still be free, but I don't think we need to consider the case where someone might actually want to pay for their bio instead. If bios are always free, the code is simpler, and it's simpler to explain to users.

This made me think of the following though: if bios are always free, maybe they also should never have decreased visibility or show 'freebie' 🤔 But that's out of scope of this PR where we just want to remove free comments, so you can ignore this, just wanted to mention this so anybody can chime in.

pory-gone:
That's a great point. I'm sorry I didn't think of it sooner! I'll make the changes you requested as soon as possible.

ekzyis:
This was merged because we wanted to remove free comments asap because of [this](https://stacker.news/items/1093592?commentId=1095189)

If @huumn agrees with making bios always free and thus removing the setting, you can create a follow-up PR for that!

pory-gone:
Sure i will! I'll wait for instructions then.

huumn:
Yes, please remove the setting

pory-gone:
okay! I'll do it as soon as possible.

PR #2422, opened 2025-08-15
feat: add territory SEO

### LINKED ISSUE #2363: Custom OG meta tags for territories

When sharing a link of the territory, description appears to be the generic SN one. OG meta description does not reflecting territory description.

Enable custom Open graph meta tags for territories

_No response_

### Additional context

_No response_

brymut:
## Description

closes #2363 

Add SEO for territory desc

## Additional Context

## Checklist

Y

10

Y

N


lib/territory.js
```diff
@@ -28,3 +29,103 @@ export function nextBillingWithGrace (sub) {
...
+      break
+    }
+  }
+  if (result) {
+    return result + '...'
+  }
+  // Fallback to simple truncation
+  return text.substring(0, maxLength - 3) + '...'
+}
+
```
huumn: why not share the item seo stuff?

if the item seo stuff is lame in some way, improve and share it with territories. I think they should share seo - they are both mostly markdown documents.

brymut: just following the issue spec

brymut: > why not share the item seo stuff?

Applied changes for users and item seo stuff too

huumn:
I will pay the bounty for this but not merge. 

My hesitation is mostly related to the markdown stripping, which we should be able to do more elegantly/performantly as a result of the new editor.

PR #2432, opened 2025-08-21
Server-side tracking of comments view time

### LINKED ISSUE #2427: Account based live comments instead of device based

I'm not sure if this is a known feature or not, but live comments right now seem to be based on whether your device has seen an item, and not whether your account has.  Since I use stacker.news on multiple devices throughout the day (desktop, laptop, phone, depending on where I am), it can get confusing to see something pop up as new, then realize I've actually seen it when using SN on another device.

Have live comments "read" state depend on account and not on device.

_No response_

### Additional context

_No response_

Soxasora:
## Description

Resolves #2427

#### How it works
Server-side tracking of the last comment of a post viewed by a user, sharing it across devices.
The view state is synchronized between clients by making `meCommentSats` part of every item which is `item.createdAt` by default. The comments view time is updated in DB only if there's a newer comment than the stored one.

#### Fallback
The classic comment tracking via local storage is retained as anon fallback and runs regardless of being logged in (together with the server-side tracking), this ensures that a logout won't shock the user with false new items.

#### Live comments
Live comments now have a more efficient and safer way of fetching new comments and storing the last timestamp. This change ensures that the latest time is retrieved during the injection cycle rather than at the end, eliminating the risk of missing time updates due to deduplication.

#### Cleanup
The table is cleaned from items that weren't viewed or didn't have a comment in the last 21 days.

## Screenshots

The items list is a good place to showcase server-side tracking:

https://github.com/user-attachments/assets/827de7e2-e9ac-4f00-a688-6983e1dc3b33

## Additional Context

Yet another reminder for a refactor of the outlining system that's going to happen with #2401 (whew it's really long).

## Checklist

Yes

7, iterative QA, still continuing

n/a

n/a


Soxasora:
This raises the question: should we schedule a cronjob to clean the table once in a while?... brb

components/use-live-comments.js
```diff
@@ -111,14 +131,29 @@ export default function useLiveComments (rootId, after, sort) {
...
     // directly inject new comments into the cache, preserving scroll position
     // quirk: scroll is preserved even if we are not injecting new comments due to dedupe
-    preserveScroll(() => cacheNewComments(cache, rootId, data.newComments.comments, sort))
+    let injectedLatest = latest
+    preserveScroll(() => {
+      injectedLatest = cacheNewComments(cache, injectedLatest, topLevelId, data.newComments.comments, sort)
+    })
+
+    // sync view time if we successfully injected new comments
+    if (new Date(injectedLatest).getTime() > new Date(latest).getTime()) {
```
Soxasora, line 140: Mhh, I was worried about saving an older date, but I don't think that would be the case here. Otherwise, we could skip no-op or going back in time, on the resolver with either `queryRaw` or fetching the value first.

worker/untrackOldItems.js
```diff
@@ -0,0 +1,22 @@
...
+  await models.commentsViewAt.deleteMany({
+    where: {
+      AND: [
+        // delete if never viewed in the last 21 days
+        { lastViewedAt: { lt: pivot } },
+        // AND if the item had no comments in the last 21 days
+        {
+          OR: [
+            { item: { lastCommentAt: { lt: pivot } } },
+            { item: { lastCommentAt: null, createdAt: { lt: pivot } } }
```
Soxasora, line 16: Coalesce with `createdAt` might not be needed since we only write to `CommentsViewAt` if there are comments, but it's future-proof and a good protection for the dev seed, so why not.

huumn:
isn't this a regression?

https://github.com/user-attachments/assets/55f8511b-8bb7-4722-a794-fd7099c5c367

> The table is cleaned from items that weren't viewed or didn't have a comment in the last 21 days.

What effect does this have? Does this mean all comments will appear as read or unread?

Soxasora:
> isn't this a regression?

It's not! Navigator doesn't track new comments you can see (visible in viewport). Even though me and @ekzyis were talking about how it should be able to track those too. What do you think?

> What effect does this have? Does this mean all comments will appear as read or unread?

Nothing should happen, picture items you never visited: they don't show the new comment dot.
If you're asking this because unread (already present) items have outlined comments in dev, it's because the root item lacks `lastCommentAt`, while prod wouldn't

We can also not clean `CommentsViewAt`, was just wondering if it can become uselessly huge

huumn:
> It's not! Navigator doesn't track new comments you can see (visible in viewport). Even though me and @ekzyis were talking about how it should be able to track those too. What do you think?

Oh okay. I remember that discussion. I do think those should probably be tracked too. It makes the experience *feel* inconsistent even if it's for good reason.

> Nothing should happen, picture items you never visited: they don't show the new comment dot.
If you're asking this because unread (already present) items have outlined comments in dev, it's because the root item lacks lastCommentAt, while prod wouldn't

I was just curious, but that's what I was hoping to hear.

huumn:
hmm this didn't trigger rewards extension

PR #2450, opened 2025-08-27
Use bot icon as badge

ekzyis:
## Description

This PR replaces the text 'bot' if an item was created with an API key with an icon like we do for anon, hat, gun and horse.

If we can find appropriate icons, we could also do the same for 'outlawed' and 'freebie', but they are not part of this PR.

Additionally to less visual clutter, this would also make the territory name stand out more.

## Screenshots

before:

2025-08-27-134215_551x52_scrot

after:

2025-08-27-134237_538x48_scrot

example image with placeholder for freebie icon:

2025-08-27-134717_499x51_scrot

## Checklist

yes

`9`. item info for posts, comments renders fine

yes, also looks good in light mode

no


huumn:
tbh this doesn't seem less cluttered to me in an obvious way. it's kind of trading one form of clutter for another that's less explicit

ekzyis:
Mhh, maybe this was a premature solution to my perceived problem that the territory name doesn't stand out enough for how important it should be. It's currently shown equally as important as if an item was created by a bot, for free, was outlawed or is the top boost.

huumn:
Ah I get the motivation now.

I'd merge this if we don't display hat/horse/gun on bot posts/comments and just this icon instead. Otherwise, I'm not sure it's worth it for either motivation.

ekzyis:
> I'd merge this if we don't display hat/horse/gun on bot posts/comments and just this icon instead

Done

PR #2451, opened 2025-08-27
fix #2392: make subs return meSubscription

### LINKED ISSUE #2392: Territory subscription state reset after client-side navigation

### Describe the bug

https://github.com/user-attachments/assets/ff1d58dd-b47b-4ef9-8efb-45451dd9a1ae

### Screenshots

_No response_

### Steps To Reproduce

1. Subscribe to a territory
2. Notice that button now shows 'unsubscribe'
3. Go somewhere else and go back
4. Button now shows 'subscribe' again

### Expected behavior

Button always reflects current state of territory subscription

### Logs

_No response_

### Device information

_No response_

### Additional context

This was found during investigation of why a push notification was sent for a territory subscription, but the notification wasn't found in /notifications ([thread](https://stacker.news/items/1066761?commentId=1066908)). It's probably a separate bug.

Scroogey-SN:
## Description

fix #2392 

I noticed that operation Subs always returned meSubscription false, that seems to be the source of the problem.

Depending on the sequence the client calls Subs and SubItems (which returns proper meSubscription), the client gets a false meSubscription.

Switching between hot and top is such a sequence where hot ends up with just a Subs call at the end.

Hence, extend subs to return meSubscription using a LEFT JOIN like seen similarly.

## Screenshots

https://github.com/user-attachments/assets/b4c7087d-d849-4b84-b463-f4c14dac9fba

## Additional Context

Should subs return more fields from SubSubcription, like it does with MuteSub.*?

## Checklist

Yes

3, I reproduced the problem, made the change, verified it resolves the problem (see video above). Didn't try to test other cases where Subs is called.

Tested on desktop (Firefox only), only dark mode.

No

### DIFF
--- a/api/resolvers/sub.js
+++ b/api/resolvers/sub.js
@@ -59,7 +59,7 @@ export default {
           FROM "Sub"
           LEFT JOIN "SubSubscription" ss ON "Sub".name = ss."subName" AND ss."userId" = ${me.id}::INTEGER
           LEFT JOIN "MuteSub" ON "Sub".name = "MuteSub"."subName" AND "MuteSub"."userId" = ${me.id}::INTEGER
-          WHERE status <> 'STOPPED' ${showNsfw ? '' : 'AND "Sub"."nsfw" = FALSE'}
+          WHERE status <> 'STOPPED' ${showNsfw ? Prisma.empty : Prisma.sql`AND "Sub"."nsfw" = FALSE`}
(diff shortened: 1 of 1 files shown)


huumn (commented):
This looks good. This just needs to be converted from `queryRawUnsafe` to `queryRaw` ... anytime we update one of these old queries, I like to do this.

ekzyis (approved):
There was a syntax error but I fixed it now

PR #2460, opened 2025-08-28
Reduce bundle size

axelvyrn:
## Description

Updates to #1832 focused on reducing client bundle size, without changing runtime behavior or site flow.
All screenshots were taken after building with `npm run build`

Major changes

- Defer heavy embeds to user intent to reduce initial JS:
  - Twitter and YouTube are loaded via next/dynamic (ssr: false) with lightweight skeleton placeholders; actual widgets render only after “show full tweet” / “load video” is clicked.
  - See screenshots
- Prevent layout shift and unnecessary work:
  - All embeds render inside fixed containers with placeholders to avoid CLS; players/iframes are not initialized until needed.
- Nostr/Spotify safety and efficiency:
  - Nostr iframe auto-sizes via postMessage and respects dark mode; restrictive sandbox attributes are applied.
  - Spotify URLs are normalized before use; the IFrame API is only invoked when the embed mounts.
- Rumble/PeerTube:
  - Kept as simple, sandboxed iframes; no client JS pulled into the shared bundle.
 
- Lazy-load the QR component in the user header (on-demand modal)
  - File: `components/user-header.js`
  - Change: Replace static import of `QRCodeSVG` with `dynamic(() => import('qrcode.react').then(m => m.QRCodeSVG), { ssr: false })`, used only inside the LNURL modal.
  - Why: Keeps `qrcode.react` out of the shared bundle; the chunk loads only when the modal opens.
  - See screenshots

**Impact (measured)**
- Upstream (stackernews/stacker.news) master at `c5b50d71ace98d8397b6340dd70646ba77a443f6` (@ekzyis commit about Small refactor of code flow in multiAuthMiddleware)
  - `.next/static`: 34,552,129 bytes (32.95 MiB)
 
image

- Reduced branch (`reduce-bundle-size`, current HEAD)
  - `.next/static`: 8,872,024 bytes (8.46 MiB)
image

- Reduction: 74.32%

Reproduce the measurement (PowerShell)
- Create a detached worktree for upstream master and build:
  1) `git fetch upstream --no-tags`
  2) `git rev-parse upstream/master`, copy the hash
  3) `git worktree add D:\stackernewsdev-upstream <hash>`
  4) `npm ci --no-audit --no-fund --prefix D:\stackernewsdev-upstream`
  5) `npm run build --prefix D:\stackernewsdev-upstream`
  6) Measure: sum file sizes under `D:\stackernewsdev-upstream\.next\static`
- Build the reduced branch and measure:
  1) `git checkout reduce-bundle-size`
  2) `npm ci --no-audit --no-fund`
  3) `npm run build`
  4) Measure: sum file sizes under `D:\stackernewsdev\.next\static`
- Compute reduction percentage: `1 - (current / baseline)`

Why these changes
- Disabling production source maps removes large assets that don’t affect users.
- Lazy-loading the QR component avoids shipping a niche dependency to all pages.
- The SSR guards prevent rare build-time crashes due to undefined bases or malformed `srcset` entries, improving reliability without altering UI.

## Screenshots
image
image
image
image
image
image
image

## Additional Context

- Non-fatal “Invalid URL” appears during builds even on upstream master; this comes from a path that calls `new URL(...)` without a base during SSR. I guarded the LNURL case in `user-header.js`; any remaining message likely originates elsewhere and can be addressed as a follow-up if desired.
- I'm not sure how the spotify embed works, I tried iframe but it is rendered as plaintext

## Checklist

**Backward compatibility**
- Yes. No GraphQL schema changes, no DB migrations, no route changes. All adjustments are to build configuration (source maps) and frontend code-splitting/robustness. Runtime behavior and UI/flow remain intact.

**QA level**
- 8/10
  - Built upstream baseline and reduced branch; verified Next build success.
  - Verified LNURL modal loads QR code chunk demand.
  - Verified media rendering with and without `NEXT_PUBLIC_IMGPROXY_URL` present; no SSR crashes; invalid `srcset` entries safely skipped.
  - Measured `.next/static` sizes and computed reduction.

**Frontend checks (mobile/light/dark)**
- Tested locally in dev using responsive DevTools:
  - Light and dark mode: verified user header modal and media rendering in both modes.
  - Mobile viewport emulation: modal and media layouts render as expected (no layout shifts from lazy loading).
- If you want real-device testing, I can run it and report separately.

**New environment variables**
- None introduced.
- One defensive fallback: `components/user-header.js` uses `https://stacker.news` as a base if `NEXT_PUBLIC_URL` is unset during SSR in non-prod environments. This does not add or change any env var; it’s just a safe default for `new URL(...)`.


axelvyrn:
While uploading media, a CORS issue is encountered every time. I tried a small workaround
- Added a same-origin upload proxy endpoint so the browser never does a cross-origin POST:
•  pages/api/upload-proxy.js streams your multipart body to the presigned S3 URL server-side.
•  components/file-upload.js automatically routes uploads through that proxy when the presigned URL host ends with .app.github.dev.
•  I also previously added error handling so the “Uploading …” marker won’t hang on network failures.

Soxasora:
> While uploading media, a CORS issue in encountered every time.

Add your `.app.github.dev` address here, sndev simulates AWS

https://github.com/stackernews/stacker.news/blob/c5b50d71ace98d8397b6340dd70646ba77a443f6/docker/s3/cors.json#L13-L15

axelvyrn:
image
Uploads are working fine

huumn:
I'd like to keep production source maps. I'd be surprised if they aren't dynamically loaded ...

Soxasora:
Source maps are loaded only by devtools on demand

Here's an [updated source](https://learn.microsoft.com/en-us/microsoft-edge/devtools/javascript/source-maps) even if they only mention Edge

> Meanwhile, Microsoft Edge actually runs your minified code to render the webpage. Source maps are only used by DevTools, and only for displaying source code to developers.

Google [mentioned](https://developer.chrome.com/blog/sourcemaps#how_does_the_source_map_work) this in 2012
> The source map file will only be downloaded if you have source maps enabled and your dev tools open

huumn:
Ah and I think OP is also confusing build output size and bundle size.

huumn (changes requested):
1. We want to keep source maps
2. Build output size !== bundle size
3. SSR "robustness" improvements that are unrelated to bundle size reduction should not be included in a PR focused on bundle size reduction

axelvyrn:
> 1. We want to keep source maps
> 2. Build output size !== bundle size
> 3. SSR "robustness" improvements that are unrelated to bundle size reduction should not be included in a PR focused on bundle size reduction

1. OK
2. So, what do I refer to as the bundle size? As far as I know, production bundle size refers to the total compiled size of the application’s code. If not so, what should I measure? Size of the individual affected chunks? Please do tell
3. OK

huumn:
> So, what do I refer to as the bundle size? As far as I know, production bundle size refers to the total compiled size of the application’s code. If not so, what should I measure? Size of the individual affected chunks? Please do tell

We want to optimize "bundle size" to improve site performance, especially in slow internet situations. So size reductions to things in the bundle that affect that are what I mean by bundle size. 

I'm sorry I can't be more clear than that, but that should at least make it clear that source maps don't fit the definition.

axelvyrn:
Well that reduces a lot of work for me, only the twitter and YouTube embed loading and lazy QR code loading are to be highlighted then I suppose, I'll update the description accordingly with the first load JS and the per-route new JS sizes (w.r.t the same commit mentioned)

axelvyrn:
Sorry to prove my incompetence here, I extracted the first-load JS sizes and Gzipped sizes for the upstream and my so called "reduced" bundle, and the delta was +0.07% (+1.7 kB), i.e., the net bundle size in my branch was more than previous. If it had been the build output size, it might have been different but that is not what is required.

PR #2461, opened 2025-08-28
Allow posting anonymously when logged in

webwarrior-ws:
This PR addresses several things:
- If logged in, it solves the UX annoyance of having to log out in order to post something (a post or a comment) anonymously, and then log in again.
- If posting something anonymously, it solves the UX annoyance of not being able to track its performance (did it receive many sats? did it receive any replies to which it might be interesting to re-engage with?)
- This changes would also make stacker.news stand out from other similar forum-like platforms (e.g. reddit) where you can see constantly that people have to create "throw-away" accounts in order to post personal stuff that they wouldn't want to be linked to their usernames forever. By allowing the user to post anonymously without having to create a new account, we avoid the annoyances typically associated with this poor UX workflow, such as: needing to use a new browser for the new account, and/or having to remember to switch accounts to track replies and/or needing to firstly gather karma (or more) on the new account in order to be able to post in certain sub-communities, etc.

### DIFF
--- a/components/comment.js
+++ b/components/comment.js
@@ -176,8 +176,9 @@ export default function Comment ({
   }, [item.id, rootLastCommentAt])
 
   const bottomedOut = depth === COMMENT_DEPTH_LIMIT || (item.comments?.comments.length === 0 && item.nDirectComments > 0)
-  // Don't show OP badge when anon user comments on anon user posts
-  const op = root.user.name === item.user.name && Number(item.user.id) !== USER_ID.anon && !root.postAnonymously && !item.postAnonymously
+  // Don't show OP badge when anon user comments on anon user posts.
+  // Also don't show OP badge when either post or comment was posted with "post anonymously" option, but not both.
+  const op = root.user.name === item.user.name && Number(item.user.id) !== USER_ID.anon && !(root.postAnonymously ^ item.postAnonymously)
     ? 'OP'
     : root.forwards?.some(f => f.user.name === item.user.name) && Number(item.user.id) !== USER_ID.anon
       ? 'fwd'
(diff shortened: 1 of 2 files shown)


ekzyis:
> If logged in, it solves the UX annoyance of having to log out in order to post something (a post or a comment) anonymously, and then log in again.

Are you aware that you don't need to logout, but can switch to anon and then switch back without authenticating again?

webwarrior-ws:
> Are you aware that you don't need to logout, but can switch to anon and then switch back without authenticating again?

I was not aware indeed. However, my PR is still useful in the sense that posting anonymously in the way that it provides (instead of switching accounts) allows you to track your anonymous posts easily so that you don't lose them over time (e.g. see replies to them, etc).
Also, the fact that I wasn't aware might also mean that the "switch account" feature is not very discoverable (however, a "post anonymously" checkbox when posting is very likely to be more).

I can update the PR description and commit msg to highlight this.

ekzyis:
> Also, the fact that I wasn't aware might also mean that the "switch account" feature is not very discoverable (however, a "post anonymously" checkbox when posting is very likely to be more).

I agree with the problem! It's pretty cumbersome to (quickly) post something anonymously while logged in, assuming one is even aware of this option.

> However, my PR is still useful in the sense that posting anonymously in the way that it provides (instead of switching accounts) allows you to track your anonymous posts easily so that you don't lose them over time (e.g. see replies to them, etc).

Unfortunately, as it is implemented, calling these posts "anonymous" is misleading. It's trivial to reveal the real author via the API. I did not test though, only looked at the code.

huumn:
We do not want to do this. A lot of overhead for us to relieve customers of a tiny bit of overhead.

If you want to avoid doing work we won't merge in the future, please open an issue first so it can be discussed.

PR #2464, opened 2025-08-28
Use precache manifest generated by webpack plugin

### LINKED ISSUE #419: Offline page is not styled

ekzyis:
## Description

related to #419

Afaict, the issue is that the assets for the offline page were not cached:

1. We cache waiting.webp from stacker.news but /offline tries to load it from a.stacker.news
2. We don't cache CSS or JS chunks from /_next/static/chunks/ with content hash like framework-[hash].js, main-[hash].js, _app-[hash].js, offline-[hash].js, [hash].css

To fix (1), I used `urlManipulation` to tell the service worker that it should fallback to stacker.news when no cached assets were found at a.stacker.news.

Fixing (2) is WIP. Need to figure out how to know the chunk + hash when injecting the manifest.

This also removes the custom script to build the manifest because I realized when we copy the static assets via webpack, `InjectManifest` can pick them up and thus inject them into the manifest itself. This seems cleaner.

---

Did not find a solution for (2). So this PR does not fix #419, but only replaces our custom script to generate the precache manifest with using the one from `InjectManifest`.

This also includes less than before in the precache. Before, we included every webp file, including the one that only shows up in emails. Now, we only include waiting.webp since only it shows up in /offline.

## Checklist

yes

`8`

n/a

no


axelvyrn:
why will people think of visiting SN (a website) offline? Although I did it too

PR #2467, opened 2025-08-29
Optimize chain sync of LND nodes

### LINKED ISSUE #2189: Local regtest fixup: make lnd containers sync to chain faster

ekzyis:
## Description

close #2189

This replaces wallet.db of each LND node with a version that was copied from the container as soon as possible (but after the address was generated). This especially made chain sync faster for sn_lnd: it's now around 1 minute, down from 5 minute before.

## Checklist

yes

`10`. Used scripts/test-routing.sh to make sure nodes can still pay each other.

n/a

no


Soxasora:
Don't know if this was my problem, because even after 1 hour I would get `NO_ROUTE`, but I'm not getting that error anymore, thanks!

huumn:
thank you!

PR #2470, opened 2025-08-29
Allow 0 and otherwise empty forwards

### LINKED ISSUE #2271: Allow 0 and otherwise empty forwards to filter out and pass validation
Image

Someone dm'd sn's X account that they were stuck at this screen. They thought 0 was equivalent to blank so perhaps it should be.

The solution is to allow these to pass - but ONLY on clientside validation (server side uses the same validation so we don't want to allow the client to pass '0' forwards to us) - then filter these out before calling the mutuation. validation ideally.

pory-gone:
## Description

fix #2271 

Made client and server changes to handle forward recipients more robustly:

`form.js:` normalize forwards before sending the mutation. Converts `pct` to Number and filters out entries that have no recipient (`nym`) or have non-finite or non-positive percentages. This lets the browser accept 0 or empty input while preventing blank/zero forwards from being sent to the server.

`validate.js:` adjusted the forwards schema so server-side validation remains strict but more precise:

## Additional Context

Only fully empty forward objects (no `nym` and no `pct`) are compacted away.
The total-sum check now skips when any `pct` is non-finite so that field-level validators (e.g. `required`) produce the specific error the user should see (instead of a confusing "sum exceeds 100%" message).

## Screenshots

https://github.com/user-attachments/assets/e1e060e6-e8d7-4820-b984-5d85b82e14e7

## Checklist

yes

7/10

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:**nan

**Did you introduce any new environment variables? If so, call them out explicitly here:**nan

### DIFF
--- a/lib/form.js
+++ b/lib/form.js
@@ -10,7 +10,12 @@ export const normalizeForwards = (forward) => {
   if (!Array.isArray(forward)) {
     return undefined
   }
-  return forward.filter(fwd => fwd.nym || fwd.user?.name).map(fwd => ({ nym: fwd.nym ?? fwd.user?.name, pct: Number(fwd.pct) }))
+  // Convert to normalized objects and filter out entries with no recipient
+  // and entries with a zero/invalid percentage. Users sometimes type '0'
+  // to mean blank; treat those as empty and drop them before sending to the server.
+  return forward
+    .map(fwd => ({ nym: fwd.nym ?? fwd.user?.name, pct: Number(fwd.pct) }))
+    .filter(fwd => fwd.nym && Number.isFinite(fwd.pct) && fwd.pct > 0)
 }
 
 export const toastUpsertSuccessMessages = (toaster, upsertResponseData, dataKey, itemText) => {
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -146,12 +146,20 @@ export function advPostSchemaMembers ({ me, existingBoost = 0, ...args }) {
             },
             message: 'cannot forward to yourself'
           }),
-        pct: intValidator.required('must specify a percentage').min(1, 'percentage must be at least 1').max(100, 'percentage must not exceed 100')
+        pct: intValidator.required('must specify a percentage').max(100, 'percentage must not exceed 100')
       }))
       .compact((v) => !v.nym && !v.pct)
       .test({
         name: 'sum',
-        test: forwards => forwards ? forwards.map(fwd => Number(fwd.pct)).reduce((sum, cur) => sum + cur, 0) <= 100 : true,
+        test: forwards => {
+          if (!forwards) return true
+          const pcts = forwards.map(fwd => {
+            const s = typeof fwd.pct === 'string' ? fwd.pct.trim() : fwd.pct
+            return s === '' ? NaN : parseFloat(s)
+          })
+          if (!pcts.every(Number.isFinite)) return true
+          return pcts.reduce((sum, cur) => sum + cur, 0) <= 100
+        },
         message: 'the total forward percentage exceeds 100%'
       })
       .test({


ekzyis (changes requested):
This has the same problem described in https://github.com/stackernews/stacker.news/pull/2281#discussion_r2228668320.

The client removes 0 pct forwards, but the server would still accept them because the validation is the same on both. The server should not accept 0 pct forwards like this:

```
curl -H "Content-Type: application/json" -X POST -d '{
  "query": "mutation upsertDiscussionWithForwards($title: String!, $text: String!, $sub: String!, $forward: [ItemForwardInput]) { upsertDiscussion(title: $title, text: $text, sub: $sub, forward: $forward) { result { id } } }",
  "variables": {
    "title": "foobar",
    "text": "asdf", 
    "sub": "art",
    "forward": [
      {
        "nym": "anon",
        "pct": 0
      }
    ]
  }
}' http://localhost:3000/api/graphql
```

Maybe all we need is to extend `validateSchema` like this:

```diff
diff --git a/lib/validate.js b/lib/validate.js
index ed71f3a1..243b10b4 100644
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -4,7 +4,8 @@ import {
   MIN_POLL_NUM_CHOICES, MAX_FORWARDS, BOOST_MULT, MAX_TERRITORY_DESC_LENGTH, POST_TYPES,
   TERRITORY_BILLING_TYPES, MAX_COMMENT_TEXT_LENGTH, MAX_POST_TEXT_LENGTH, MIN_TITLE_LENGTH, BOUNTY_MIN, BOUNTY_MAX,
   RESERVED_SUB_NAMES,
-  BOOST_MAX
+  BOOST_MAX,
+  SSR
 } from './constants'
 import { SUPPORTED_CURRENCIES } from './currency'
 import { NOSTR_MAX_RELAY_NUM, NOSTR_PUBKEY_BECH32, NOSTR_PUBKEY_HEX } from './nostr'
@@ -16,7 +17,7 @@ import { datePivot } from './time'
 export async function validateSchema (schema, data, args) {
   try {
     if (typeof schema === 'function') {
-      return await schema(args).validate(data)
+      return await schema({ server: SSR, ...args }).validate(data)
     } else {
       return await schema.validate(data)
     }
```

and then in `advPostSchemaMembers`, we keep the `min` check on the server but remove or set it to 0 on the client.

What do you think?

pory-gone:
I realize there's a inconsistency between client and server validation. I agree with the fix for the distinction between the two validations; my implementation wasn't strict enough for the server, and it also made it vulnerable to direct API calls that bypassed the check. I'm implementing the fix now.

lib/validate.js
```diff
@@ -150,12 +151,22 @@ export function advPostSchemaMembers ({ me, existingBoost = 0, ...args }) {
...
-        test: forwards => forwards ? forwards.map(fwd => Number(fwd.pct)).reduce((sum, cur) => sum + cur, 0) <= 100 : true,
+        test: forwards => {
+          if (!forwards) return true
+          const pcts = forwards.map(fwd => {
+            const s = typeof fwd.pct === 'string' ? fwd.pct.trim() : fwd.pct
+            return s === '' ? NaN : parseFloat(s)
+          })
+          if (!pcts.every(Number.isFinite)) return true
+          return pcts.reduce((sum, cur) => sum + cur, 0) <= 100
+        },
```
ekzyis: Do we still need this change? You mentioned this:

> The total-sum check now skips when any `pct` is non-finite so that field-level validators (e.g. `required`) produce the specific error the user should see (instead of a confusing "sum exceeds 100%" message).

but from what I can tell, even without this, it does show 'must specify a percentage' when it's empty

lib/form.js
```diff
@@ -10,7 +10,12 @@ export const normalizeForwards = (forward) => {
   if (!Array.isArray(forward)) {
     return undefined
   }
-  return forward.filter(fwd => fwd.nym || fwd.user?.name).map(fwd => ({ nym: fwd.nym ?? fwd.user?.name, pct: Number(fwd.pct) }))
+  // Convert to normalized objects and filter out entries with no recipient
+  // and entries with a zero/invalid percentage. Users sometimes type '0'
+  // to mean blank; treat those as empty and drop them before sending to the server.
+  return forward
+    .map(fwd => ({ nym: fwd.nym ?? fwd.user?.name, pct: Number(fwd.pct) }))
+    .filter(fwd => fwd.nym && Number.isFinite(fwd.pct) && fwd.pct > 0)
```
ekzyis: Do we need `Number.isFinite`? It's confusing me why this is here since only `fwd.pct > 0` looks relevant for #2271. Is this needed for #2271 or did you encounter some other bug to include `Number.isFinite`?

Just asking to make sure I understand why this is here. We can keep it if there's a good reason

ekzyis (commented):
This works as expected now, thank you!

Left some comments to make sure I understand why we need `Number.isFinite` and the changes to the sum validation.

lib/form.js
pory-gone: I added `Number.isFinite` to prevent non-numeric values from being sent to the server, especially in borderline cases where the percentage might be entered incorrectly.
I included the extra check just to be safe, but if you think it's unnecessary, I'm happy to remove it.

lib/validate.js
pory-gone: The intent was to prevent the "sum exceeds 100%" error from appearing when a percentage was missing, so that the validator could display the more specific error. If the behavior is already correct, we can simplify the check and remove this part.

pory-gone:
Thank you so much, if there are any other changes that need to be made, I'm here to do so.

pory-gone:
Thank you so much, if there are any other changes that need to be made, I'm here to do so.

lib/validate.js
ekzyis: Seems to work as intended afaict, so yes, I think we can remove this:

2025-09-08-162406_689x105_scrot

But FYI, if this wouldn't work, I'd expect a fix for this in a different PR

lib/form.js
ekzyis: It's better to be safe than sorry, but I think this here is more confusing than safe. It makes me question if there's a way to enter `NaN` or `Infinity`.

I also think _if_ we need this, it should be part of validation and not part of `normalizeForwards`, but not sure if we do.

So yes, please remove it, thanks

pory-gone:
I fixed all the files and now everything should be fine.

ekzyis (approved):

ekzyis:
Hello, I recently updated our PR template to add a question about AI in https://github.com/stackernews/stacker.news/pull/2507.

Even though I already approved your PR, I updated your PR description to include it, see the end of the checklist.

Can you please answer it? Thank you!

lib/validate.js
```diff
@@ -150,7 +152,7 @@ export function advPostSchemaMembers ({ me, existingBoost = 0, ...args }) {
             },
             message: 'cannot forward to yourself'
           }),
-        pct: intValidator.required('must specify a percentage').min(1, 'percentage must be at least 1').max(100, 'percentage must not exceed 100')
+        pct: intValidator.required('must specify a percentage').min(pctMin, `percentage must be at least ${pctMin}`).max(100, 'percentage must not exceed 100')
```
huumn, line 155: before this allowed the client to input/send `pct` < 0 because it didn't have a minimum. the server would filter them but it seems better to validate that on the client.

lib/validate.js
```diff
@@ -122,6 +123,7 @@ export const searchSchema = object({
 
 export function advPostSchemaMembers ({ me, existingBoost = 0, ...args }) {
   const boostMin = existingBoost || BOOST_MIN
+  const pctMin = SSR ? 1 : 0
```
huumn, line 126: rather than sending SSR down as a parameter, it seemed okay to just use it outright

PR #2493, opened 2025-09-08
feat(Bookmarks): Custom bookmark ordering

### LINKED ISSUE #2480: [Feature Request] Reorder bookmarks

I use certain bookmarks more than others, and want to keep the most used ones on top. (One of them I like to keep on top, for example, is [#661571](https://stacker.news/items/661571)).

However, I don't think there's a way to reorder bookmarks. Since the default sort is "recent", my way of moving a bookmark to the top is by un-bookmarking it and then re-bookmarking it.

Sorry, this is probably pretty low priority, but any time you're doing something that feels hacky I figured it's worth bringing up as an issue!

Some way to manually order the bookmarks, either click and drag or buttons.

_No response_

### Additional context

_No response_

brymut:
## Description

fix #2480 

Adding ability for users to manually order their bookmarks. Order stored in DB, using DND provider from wallets. 

https://github.com/user-attachments/assets/6c714d23-f41a-498b-9991-0370ee7bb643

## Screenshots

Screenshot 2025-09-08 at 20 44 29
Screenshot 2025-09-08 at 20 44 50
Screenshot 2025-09-08 at 20 45 16
Screenshot 2025-09-08 at 20 45 30

## Additional Context

## Checklist

Y

9

Y

N

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -786,18 +786,12 @@ export default {
         throw new GqlInputError('itemIds required')
       }
 
-      for (let i = 0; i < itemIds.length; i++) {
-        const itemId = Number(itemIds[i])
-        await models.bookmark.updateMany({
-          where: {
-            userId: me.id,
-            itemId
-          },
-          data: {
-            customOrder: i + 1
-          }
-        })
-      }
+      await models.$transaction(
+        itemIds.map((id, i) => models.bookmark.update({
+          where: { userId_itemId: { userId: me.id, itemId: Number(id) } },
+          data: { customOrder: i + 1 }
+        }))
+      )
 
       return true
     },
--- a/components/bookmark.js
+++ b/components/bookmark.js
@@ -86,10 +86,7 @@ export function CustomBookmarkList ({ ssrData, variables = {}, query }) {
   const { data, fetchMore } = useQuery(query || SUB_ITEMS, { variables })
   const dat = useData(data, ssrData)
 
-  const { items, cursor } = useMemo(() => {
-    if (!dat) return { items: [], cursor: null }
-    return dat?.items || { items: [], cursor: null }
-  }, [dat])
+  const { items, cursor } = useMemo(() => dat?.items ?? { items: [], cursor: null }, [dat])
 
   const [orderedItems, setOrderedItems] = useState(items || [])
   useEffect(() => { setOrderedItems(items || []) }, [items])


api/resolvers/item.js
```diff
@@ -772,6 +777,30 @@ export default {
...
+        await models.bookmark.updateMany({
+          where: {
+            userId: me.id,
+            itemId
+          },
+          data: {
+            customOrder: i + 1
+          }
+        })
+      }
```
gndelia: This will run `itemIds.length` update queries one by one, in sequential order. It could be a potential bottleneck with a large number of bookmarks. For a small number, it may be ok.

If you want to use the `updateMany`, but run the queries in parallel, one option could be


```suggestion
await Promise.all(
  itemIds.map((itemId, i) =>
    models.bookmark.updateMany({
      where: { userId: me.id, itemId: Number(itemId) },
      data: { customOrder: i + 1 },
    })
  )
)
```

However, this still will fire `itemIds.length` queries (but in parallel instead of sequentially)

A better option could be dynamically building a SQL query with something like this

```suggestion

const caseStatements = itemIds
  .map((id, i) => `WHEN ${id} THEN ${i + 1}`)
  .join(' ');

await prisma.$executeRawUnsafe(`
  UPDATE "Bookmark"
  SET "customOrder" = (CASE "itemId"
    ${caseStatements}
  END)
  WHERE "userId" = ${me.id}
    AND "itemId" IN (${itemIds.join(',')})
`);
```

Something like this should update all bookmarks in one sql statement. 

I hope it helps!

components/bookmark.js
```diff
@@ -39,3 +48,85 @@ export default function BookmarkDropdownItem ({ item: { id, meBookmark } }) {
...
+export function CustomBookmarkList ({ ssrData, variables = {}, query }) {
+  const toaster = useToast()
+  const [reorderBookmarks] = useMutation(REORDER_BOOKMARKS)
+
+  const { data, fetchMore } = useQuery(query || SUB_ITEMS, { variables })
+  const dat = useData(data, ssrData)
+
+  const { items, cursor } = useMemo(() => {
+    if (!dat) return { items: [], cursor: null }
+    return dat?.items || { items: [], cursor: null }
```
gndelia: A minor comment, but `dat` is always defined as it's checked above, so this ternary operator can be removed

```suggestion
    return dat.items || { items: [], cursor: null }
```

gndelia: Another option could just be constructing the `useMemo` like this:

```js
const { items, cursor } = useMemo(() => dat?.items ?? { items: [], cursor: null }, [dat])
```

removing the redundant check and keeping the optional chaining

api/resolvers/item.js
brymut: Interesting solution, hadn't considered this a bottleneck due to the practicality of re-ordering large lists of bookmarks frequently. However, for the custom SQL solution, I think @huumn mentioned avoiding using `prisma.$executeRawUnsafe`

ekzyis: Yes, we definitely do not want to use `$executeRawUnsafe`.

~You should be able to use Prisma client's [`in` operator](https://www.prisma.io/docs/orm/reference/prisma-client-reference#in).~

Oh, I see, you are not updating all items with the same value so it's more complicated than just using the `in` operator.

Still, there should be a way to do this in one query without executing unsafe SQL.

brymut: I'll have a look

gndelia: @ekzyis  Sorry, I suggested using `executeRawUnsafe ` because I'm not _that_ familiar with Prisma and I saw `queryRawUnsafe` was being used already, but I Totally get the point of not using it 😅

brymut: Can't quite come up with a solution that doesn't bypass the client api, but atomic now & modestly faster due to less trips. But still N updates. Hope this is sufficient for use-case @ekzyis / @gndelia

huumn:
this solves the issue, probably, but I don't want to maintain this feature. it's a lot of code for something few people will use. this is our fault, not yours, because I left the issue up.

in general, and long term, i think we'll want to specify something more generic or have no manual reordering at all. bookmarks as a feature already clutter up all kinds of areas or code and extremely low traffic.

we discussed the other day reusing a new thing like my-zapped-items as bookmarks, then perhaps reordering can be done by zapping items more or less. idk highly speculative. i'm just discontent with bookmarks.

I'm going to close this and pay the bounty.

@gndelia please give me a lightning address so I can send you sats for `helpfulness`.

PR #2495, opened 2025-09-08
Only allow https: for links

### LINKED ISSUE #2424: intent:// Scheme Injection
#vb

ekzyis:
## Description

fix #2424 

I think limiting links to https: should be okay. [Here](https://en.wikipedia.org/wiki/List_of_URI_schemes) is the list of official URI schemes and I don't think we need any of them except https.

However, this now _requires_ https: in the front. Before, no scheme was required, so `stacker.news` was a valid input. Now, it has to be `https://stacker.news`.

I am not sure if this is a UX problem. Browsers copy URLs with the scheme and I assume that's how most links are shared, and not by manually typing them etc.

## Checklist

yes

`10`.  intent: no longer works but https: does

n/a

no

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -258,7 +258,7 @@ export function linkSchema (args) {
   return object({
     title: titleValidator,
     text: textValidator(MAX_POST_TEXT_LENGTH),
-    url: string().url().required('required'),
+    url: string().https().required('required'),
     ...advPostSchemaMembers(args),
     ...subSelectSchemaMembers(args)
   }).test({


huumn (changes requested):
If you want to whitelist protocols, that's fine, but this is a UX downgrade to whitelist a single protocol.

What about an onion address (that commonly is not https) or an ftp link or an npub (which we recognize as a 'url')? What about a developer that just launched a product at `mywebsite.com` and wants to promote it?

I personally ***hate*** when a website tells me `invalid url` or `must specify https` when I input a link without the awkward `https://`. If you know I must specify https, do it for me.

ekzyis:
> What about an onion address (that commonly is not https) or an ftp link or an npub (which we recognize as a 'url')? What about a developer that just launched a product at mywebsite.com and wants to promote it?

Mhh, good points, didn't think of onion, npubs and that HTTP can make sense for new websites.

Regarding FTP, Chrome and Firefox (not sure about Safari) don't support it anymore (out-of-the-box) so I thought we don't need to allow it if most users can't open that link with their user agent. But I changed my mind, it can still make sense to share a link even if the user agent can't open it (similar for onion addresses).

> I personally hate when a website tells me `invalid url` or `must specify https` when I input a link without the awkward `https://`. If you know I must specify https, do it for me.

I agree, this should be part of transform before validation

adlai:
The original description should make it clear that this has absolutely nothing to do with HTTP-insecure; and I should stop triggering notifications before yet another ostracism.

PR #2503, opened 2025-09-09
Fix module not found if new package added

ekzyis:
## Description

I noticed that if someone adds a new package via `npm install`, it will update their local node_modules, which is then mounted and the app will run.

But if someone then pulls the new package.json and runs `sndev start`, `npm ci` only runs during image build since #2498, but then we override the node_modules/ generated by this call with the node_modules/ from the host machine because of the mounted volume and thus the app will not run because it won't find the new package.

I considered [excluding the host node_modules via an anonymous volume](https://stackoverflow.com/questions/29181032/add-a-volume-to-docker-but-exclude-a-sub-folder) but this breaks the worker (it actually never runs `npm ci` but fully depends on installing dependencies while mounted), requires using `-V` with `docker-compose up` to renew anonymous volumes and might have other unknown issues.

Simply running `npm ci` again before all runs is the simplest solution and is also basically what we've been doing for a long time.

Added some comments to make it less confusing why we're doing this.

## Checklist

yes

`8`. this fixes the issue I had: my node_modules/ were in a state where it was missing `@shocknet/clink-sdk` even though it's in package.json (because of switching branches etc. I assume) and after rebasing on master with #2498, the app didn't find the module anymore

n/a

no

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,6 @@ WORKDIR /app
 
 EXPOSE 3000
 
-# install dependencies inside the image for prod
 COPY package.json package-lock.json ./
 RUN npm ci --legacy-peer-deps --loglevel verbose
 


Dockerfile
```diff
@@ -14,6 +14,9 @@ WORKDIR /app
 
 EXPOSE 3000
 
+# install dependencies inside the image for prod
```
ekzyis: I assume this is the case but then, I see that the worker does never run `npm ci`.

I think this means that every app container includes the worker. So we don't have 4 app services and 4 worker services running, but 4 app+worker services.

huumn: we don't use docker in prod

ekzyis: ahh, ohh

PR #2504, opened 2025-09-09
Change order of tag and level in log message

ekzyis:
## Description

I think this looks better

## Screenshots

before:

2025-09-09-230109_1920x1080_scrot

after:

2025-09-09-230054_1920x1080_scrot

## Checklist

yes

`10` (if this breaks I'll ... idk)

n/a

no

### DIFF
--- a/styles/logger.module.css
+++ b/styles/logger.module.css
@@ -38,7 +38,7 @@
 .level {
     font-weight: bold;
     text-transform: uppercase;
-    min-width: 32px;
+    min-width: 30px;
 }
 
 .tag {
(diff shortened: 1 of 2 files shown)


huumn:
This does look better. Have you considered using `display: grid` for the logs? That way, every column has the same width and you don't have to hardcode widths

ekzyis:
> This does look better. Have you considered using `display: grid` for the logs? That way, every column has the same width and you don't have to hardcode widths

Yes, but then the longest wallet name dictated the spacing in every other row and I noticed that can lead to a lot of empty space, so I liked it more that every row flows.

But that was only my opinion in the past and that might have been before I added short names just for logging :thinking: 

Will check out a grid layout again, thanks!

ekzyis:
> Will check out a grid layout again, thanks!

I would do that in a different PR though

PR #2505, opened 2025-09-09
Fix comments not shown if all are from us and pending

### LINKED ISSUE #2499: one's own pending (unpaid) comments do not show (possibly if they're the only comment)
I just noticed this as I had a comment stuck pending as my home node lost connectivity. 

It was the first and only comment in a thread which might explain the issue. Or this is a regression of some kind.

ekzyis:
## Description

fix #2499

#2499 really only happens when there are only pending comments from us.

I can't say much about the performance impact of this fix, since I don't really understand why it was needed in the first place :grimacing: If there are no comments, why was it expensive to fetch them? And why is it more expensive for anon iiuc?

## Checklist

yes

`6`. Tested that pending comments from us show up now

n/a

no

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -54,7 +54,8 @@ function commentsOrderByClause (me, models, sort) {
 async function comments (me, models, item, sort, cursor) {
   const orderBy = commentsOrderByClause(me, models, sort)
 
-  if (item.nDirectComments === 0) {
+  // if we're logged in, there might be pending comments from us we want to show but weren't counted
+  if (!me && item.nDirectComments === 0) {
     return {
       comments: [],
       cursor: null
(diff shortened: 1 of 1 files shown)


huumn:
> If there are no comments, why was it expensive to fetch them? 

Because it adds 10ms to the request when it doesn't need to. Potentially multiple 10ms if it's a recursive query and this nested resolver is erroneously triggered multiple times.

The important thing: nest resolvers are *always* called when they are present and the field is requested. That's why you see all the `typeof field !== 'undefined'` in nested resolvers - if the parent already fetched them, we don't want to fetch them again.

> And why is it more expensive for anon iiuc?

It's not more expensive for anon, but anon can't have pending comments, so we can safely short circuit like this.

ekzyis:
> Potentially multiple 10ms if it's a recursive query and this nested resolver is erroneously triggered multiple times.

Ah ok, makes sense!

_I also realized that just because a query returns nothing, doesn't mean the query was cheap. A query could take a very long time just to find nothing._

---

This PR is the fix you suggested in https://github.com/stackernews/stacker.news/issues/2499#issuecomment-3271117387, right? Asking because curious why you didn't merge this after replying

huumn:
I want to double check that it doesn't trigger multiple nested resolver calls - they can usually be mitigated

PR #2514, opened 2025-09-11
fixing magic code input error

### LINKED ISSUE #2511: Magic code input immediately runs validation

### Describe the bug

The form for the magic code input during email login will immediately show an error on first input that it requires 6 digits.

### Screenshots

_No response_

### Steps To Reproduce

1. Login with email
2. Enter code manually instead of copy & paste

### Expected behavior

Form does not run validation on change before first submit (like all other forms)

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

pory-gone:
## Description

fix #2511 

I modified `form.js`

`MultiInput` now uses `useFormikContext()` to read `submitCount`.
The logic for displaying the aggregate error has been changed: the error is only shown if there has been at least one attempt to submit, using the condition `formik.submitCount > 0` along with `meta.touched && meta.error`.

## Screenshots

https://github.com/user-attachments/assets/61defbe1-f294-4fd9-a149-3e6a9b4ea29d

## Additional Context

The error appears only after the first submission with an incomplete/incorrect code; while typing the 6 digits it is not shown.
After the first failed submission, errors like "required" and "must be 6 alphanumeric characters" will display with the same behavior as before. If this isn't correct, I'll make the necessary changes.

## Checklist

Yes

7/10

NaN

NaN

I didn't use AI for this, all manual research.

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -1451,6 +1451,7 @@ export function MultiInput ({
   onChange, autoFocus, hideError, inputType = 'text',
   ...props
 }) {
+  const formik = useFormikContext()
   const [inputs, setInputs] = useState(new Array(length).fill(''))
   const inputRefs = useRef(new Array(length).fill(null))
   const [, meta, helpers] = useField({ name })
@@ -1549,7 +1550,7 @@ export function MultiInput ({
         ))}
       </div>
       <div>
-        {hideError && meta.touched && meta.error && ( // custom error message is showed if hideError is true
+        {hideError && formik.submitCount > 0 && meta.touched && meta.error && ( // custom error message is showed if hideError is true
           <BootstrapForm.Control.Feedback type='invalid' className='d-block'>
             {meta.error}
           </BootstrapForm.Control.Feedback>


ekzyis:
> this change impacts every form on the site but it's meant to address one is particular

~Oh, right, only tested /email (with manually removing the code to disable the input without a callback)~

Wait, are you sure? The code is inside `MultiInput` which is only ever used for the /email page.

Ideally this would just work like any other form, but I didn't realize we _are_ actually using it with our `Form` component, so not sure why it's immediately validating unlike any other form. /cc @Soxasora

So the fix would include finding out why it's not already working.

Soxasora:
`MultiInput` calls `InputInner` n times, I couldn't possibly use the errors given by each of them so I muted them via `hideError`.
In general, setting a formik field and going to another, makes formik return `meta.touched === true`, considering that we have 6 fields but they're merged into 1 `onChange` via `handleChange`, each change will trigger it.

Using `formik.submitCount > 0` is a nice way to solve this considering that it will only affect `MultiInput` and we already use it on `InputInner`.

note: I also thought it was a bug that we didn't validate in real time for other forms, but it makes sense. In this case the purpose was to provide an immediate response in a PIN-like scenario

history: `MultiInput` was literally bolted with _sticky tacks_ on `InputInner` to provide immediate relief for PWA users, but in the future a refactor should take care of this

PR #2487, opened 2025-09-05
CLINK

ekzyis:
## Description

close #1376

tbd

## Screenshots

tbd

## Additional Context

tbd

## Checklist

**Are your changes backward compatible? Please answer below:**

yes

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**

tbd

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:**

yes

**Did you introduce any new environment variables? If so, call them out explicitly here:**

no

Issue #2073, opened 2025-04-06
Replace NDK with snstr


ekzyis:
We have many issues with [NDK](https://github.com/nostr-dev-kit/ndk):

- minor versions include breaking changes (https://github.com/stackernews/stacker.news/pull/2041)
- no changelog (https://github.com/nostr-dev-kit/ndk/issues/297)
- no docs (https://github.com/nostr-dev-kit/ndk/issues/307)
- no releases (https://github.com/nostr-dev-kit/ndk/releases)
- maintainer does not respond in a timely manner to important stuff
  - https://github.com/nostr-dev-kit/ndk/pull/285
  - https://github.com/nostr-dev-kit/ndk/issues/297#issuecomment-2763103064)
- irresponsible, exploitable TODOs like [this one](https://github.com/nostr-dev-kit/ndk/blob/207a951339232004c8d89f4daa685946a95d0466/ndk-wallet/src/wallets/nwc/index.ts#L183)
- bloated API that by default does things we don't need like caching and does not keep simple things simple like closing a connection (https://github.com/stackernews/stacker.news/pull/1739) or how to handle errors (might throw or return object with `error` property?)
- includes [`@cashu/cashu-ts`](https://github.com/cashubtc/cashu-ts) as a peer dependency of `@nostr-dev-kit/ndk-wallet` but the build fails if you don't have `@cashu/cashu-ts` installed (https://github.com/stackernews/stacker.news/pull/2041#discussion_r2020255065)

@AustinKelsay is building [snstr](https://github.com/AustinKelsay/snstr). It looks promising so we should see if it would make all of our nostr code (NIP-05, NIP-07, NIP-46, NIP-47, NIP-57 and crossposting) a lot easier to maintain by using it instead.

ekzyis:
Raised difficulty from `medium` to `medium-hard` since this requires changes in many places and a lot of testing to make sure there are no regressions, see mentioned NIPs + crossposting.

PR #2518, opened 2025-09-12
CLINK offers

### LINKED ISSUE #1376: Wallet: Pay to NIP-69 Nostr Offer STRings
Successor to LNURL-P and Bolt12 that uses Nostr as a transport, provides upgrade to Lightning Addresses for trust-minimization

https://github.com/nostr-protocol/nips/pull/1460

ekzyis:
## Description

supersedes #2487

I changed my mind regarding adding CLINK debits + offers in #2487. This now only closes #1376.

[CLINK spec](https://clinkme.dev/specs.html)

**TODO**

- :white_check_mark: attach CLINK with `noffer`
- :white_check_mark: q&a
- :white_check_mark: use our custom timeout for wallets
- :white_check_mark: use custom invoice description and expiry
- :white_check_mark: replace `Invalid Amount` error if amount is below minimum with more user-friendly error
- :white_check_mark: add dev documentation how to attach
- :question: add user guide
- :question: ~use abort signal~ _SDK does not support abort signals, but that's okay, not that important_
- :white_check_mark: close websocket after invoice was received

## Screenshots

2025-09-12-095231_1920x1080_scrot

## Video

see #2531 

## Additional context

- Lightning.Pub's default minimum amount for invoices seems to be 10 sats, but we communicate 1 sat as the minimum for LNURLp.

## Checklist

yes

`8`.  Tested receiving to CLINK offer using LNURLp.

yes

no

no, only Cursor autocomplete iirc

### DIFF
--- a/wallets/server/protocols/clink.js
+++ b/wallets/server/protocols/clink.js
@@ -1,0 +1,49 @@
+import { WALLET_CREATE_INVOICE_TIMEOUT_MS } from '@/lib/constants'
+import { msatsToSats } from '@/lib/format'
+import { decodeBech32, generateSecretKey, SendNofferRequest, SimplePool } from '@shocknet/clink-sdk'
+
+export const name = 'CLINK'
+
+// https://clinkme.dev/specs.html
+const ERR_INVALID_AMOUNT = 5
+
+export async function createInvoice (
+  { msats, description, expiry },
+  { noffer },
+  { signal }) {
+  const { data: { offer, relay, pubkey } } = decodeBech32(noffer)
+
+  const pool = new SimplePool()
+  const sk = generateSecretKey()
+  const request = { offer, amount_sats: msatsToSats(msats), expires_in_seconds: expiry, description }
+
+  let response
+  try {
+    const timeout = Math.floor(WALLET_CREATE_INVOICE_TIMEOUT_MS / 1000)
+    // CLINK does not support a custom invoice description or expiry
+    response = await SendNofferRequest(pool, sk, [relay], pubkey, request, timeout)
+  } catch (e) {
+    throw typeof e === 'string' ? new Error(e) : e
+  } finally {
+    pool.close([relay])
+  }
+
+  if ('bolt11' in response && typeof response.bolt11 === 'string') {
+    return response.bolt11
+  }
+
+  if (response.code === ERR_INVALID_AMOUNT) {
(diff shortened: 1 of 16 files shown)


wallets/server/protocols/clink.js
```diff
@@ -0,0 +1,48 @@
...
+  if ('bolt11' in response && typeof response.bolt11 === 'string') {
+    return response.bolt11
+  }
+
+  if (response.code === ERR_INVALID_AMOUNT) {
+    const { min, max } = response.range
+    throw new Error(`invalid amount: amount must be between ${min} and ${max} sats`)
+  }
+
+  throw new Error(response.error)
```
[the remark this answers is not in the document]
ekzyis: `createInvoice` itself is also wrapped in try/catch

ekzyis:
Actually, that it does not support custom expiry might be blocking.

Need to check if this interferes with our invoice wrapping.

ekzyis:
> Actually, that it does not support custom expiry might be blocking.
> 
> Need to check if this interferes with our invoice wrapping.

Custom invoice expiry would be nice but is not blocking. Our critical invoice wrapping code does not depend on it.

LNURL also does not support it.

---

edit: I was wrong, we actually do validate invoice expiration:

https://github.com/stackernews/stacker.news/blob/a0dffb2f188410f80c5b5b90cf9b21b1a85975a2/wallets/server/wrap.js#L137-L146

wallets/server/protocols/clink.js
```diff
@@ -0,0 +1,47 @@
...
+// https://clinkme.dev/specs.html
+const ERR_INVALID_AMOUNT = 5
+
+export async function createInvoice (
+  { msats, description, expiry },
+  { noffer },
+  { signal }) {
+  const { data: { offer, relay, pubkey } } = decodeBech32(noffer)
+
+  const pool = new SimplePool()
```
ekzyis, line 16: This is leaking connections.

We're creating a new pool for every invoice, which means we're creating a new websocket connection for each invoice and that connection isn't closed after we received the response. Not sure when it is, if ever.

For NWC, we connect and disconnect for every request to avoid bugs that we had in the past with stale connections. We should do the same here.

This means this PR is blocked by #2073, same for #2531

ekzyis: Actually, I can just write `pool.close()`.

PR #2519, opened 2025-09-12
Add sox to ids of SN admins

ekzyis:


huumn:
he also needs to go in the footer

PR #2529, opened 2025-09-13
Remove unused forceWaitForPayment

ekzyis:
## Description

Afaict, `forceWaitForPayment` is never passed as a payment behavior option, so it's always `undefined`, so `!forceWaitForPayment` is always true, so we can remove it from the `&&` condition.

I tried to find if it was ever used using `git log -G "forceWaitForPayment"`:

```
$ g l -G "forceWaitForPayment"
9f06fd65 UX latency enhancements for paid actions (#1434)
79f0df17 improve pessimistic paid actions by letting the server perform actions and settle invoice on `HELD`  (#1253)
ca11ac9f backend payment optimism (#1195)
```

but in all these commits, it is not used afaict.

## Checklist

yes

`1`. non-custodial zaps still work

n/a

no

no

### DIFF
--- a/components/use-paid-mutation.js
+++ b/components/use-paid-mutation.js
@@ -10,8 +10,7 @@ import { useWalletPayment } from '@/wallets/client/hooks'
 this is just like useMutation with a few changes:
 1. pays an invoice returned by the mutation
 2. takes an onPaid and onPayError callback, and additional options for payment behavior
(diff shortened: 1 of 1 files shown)


ekzyis:
I also see `forceWaitForPayment` in #2084 but I don't see it ever not being `undefined`.

Am I missing something and this indeed matters?

PR #2531, opened 2025-09-14
CLINK debits

### LINKED ISSUE #1478: Wallet: Pay invoices with NIP-68 Debits
NIP-68 allows users to pair a wallet service for spending by entering a NIP-05 address, and offers the ability for SN to use a reputational key for connections

https://github.com/nostr-protocol/nips/pull/1529

This is a companion issue to NIP-69 Offer strings, the two combined allow for more scalable, secure, and straightforward UX than is currently possible with NIP-47 or LNURL

https://github.com/stackernews/stacker.news/issues/1376

There is a demo app linked in either of the NIP PR's that show how trivial these are to implement, but since it would take me a day or two to get acquainted with the SNDev environment I'm posting a $150 (paid in sats) bounty for each, $300 total.

ekzyis:
## Description

close #1478

## Additional Context

tbd

## Checklist

yes

tbd

n/a

no

no


ekzyis:
Mhh, so the idea how to not require approval / support budgets is to use our existing nostr private key (that already has a profile) to encrypt and sign the CLINK debit request events.

Since we can't give the client our private key, this needs to happen remote on the server.

However, this also implies that the client won't be able to decrypt the response and thus also needs to ask the server to do so for it.

**It's very important to make sure we _really_ ONLY encrypt, sign or decrypt these CLINK debit request events.** An attacker shouldn't be able to pretend something is a CLINK debit request event when it's not and ask us to sign or decrypt it.

ekzyis:
> Mhh, so the idea how to not require approval / support budgets is to use our existing nostr private key (that already has a profile) to encrypt and sign the CLINK debit request events. [...]

Afaict, this means the server can save the plaintext content with the ndebit pointer, wait for budget approval, and then spend from the user's wallet.

ekzyis:
> > Mhh, so the idea how to not require approval / support budgets is to use our existing nostr private key (that already has a profile) to encrypt and sign the CLINK debit request events. [...]
> 
> Afaict, this means the server can save the plaintext content with the ndebit pointer, wait for budget approval, and then spend from the user's wallet.

I talked to @shocknet-justin and he confirmed this. CLINK is built around subscriptions / server-to-server payments, so this means the service (us) can indeed act unilaterally (once budget was approved) to pull money from the user's wallet without the client initiating the payment.

Since this is not what we want, I will have to encrypt and store a nostr private key in the db next to the encrypted user's `ndebit` and the user/wallet just has to deal with not being able to show our nostr profile when approving.

shocknet-justin:
> the user/wallet just has to deal with not being able to show our nostr profile when approving.

Yep, will just save them the secret sharing step vs NWC since you can generate it on the fly at pairing... 

Will think about this hybrid use-case a bit to try and get more of the service-key benefits in this environment. Perhaps a spec for the service to attest to the per-user key such that the wallet can display still the service profile purely for display and verification purposes.

ekzyis:
Mhh, yeah, something like that, but a naive solution would be vulnerable to other services abusing the budget that was only approved for SN or else we need to make sure we only sign events that will pay invoices from us, additionally to all other checks (we don't want to sign blindly).

One of these checks is probably pretty easy to miss in other implementations though, leaving a door wide open.

---

edit: Oh, right, we could attest to the per-user _key_ once though, like in a separate event, no need to attest individual events, maybe that's what you had in mind.

But I guess the problem remains how to not attest blindly, even if it's just once :thinking:

PR #2442, opened 2025-08-25
Add satsFilter-based notification suppression

axelvyrn:
## Description
Closes #2438 
Introduces logic to suppress push notifications for users based on their satsFilter setting, preventing notifications for items below 0. The satsFilter is checked when applySatsFilter is set and item context is available, and transient fields are now cleaned from the notification payload after use. Notification triggers for subscribers now include item and applySatsFilter to enable this filtering, all server-side.

## Screenshots
N/A

## Additional Context
The working should be checked on phone, since I couldn't check it.
I could only provide a logical fix and tested through PC (windows).

## Checklist

**Are your changes backward compatible? Please answer below:**
yup (no DB queries)

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**
7 (because did test only on PC and not phone)

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:**
N/A

**Did you introduce any new environment variables? If so, call them out explicitly here:**
none

**Did you use AI for this? If so, how much did it assist you?**

15% - Copilot helped me identify the possible files where the bug for this error might be since I am not still accustomed to the codebase. I filtered the files after that. No AI was used for coding purposes.

PR #2532, opened 2025-09-14
Feature/sats filter

### LINKED ISSUE #2438: push notifications don't respect sats filter

### Describe the bug

If I configured sats filter > 0, I don't see free replies in /notifications anymore but I still receive push notifications.

Image

### Screenshots

_No response_

### Steps To Reproduce

1. Configure sats filter > 0 and enable push notifications
3. Reply to yourself with a freebie
4. See reply doesn't show up in /notifications but you still received a push notification

### Expected behavior

No push notification for freebies if sats filter > 0

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

axelvyrn:
## Description
Closes #2438 
Introduces logic to suppress push notifications for users based on their satsFilter setting, preventing notifications for items below 0. The satsFilter is checked centrally under sendUserNotification since all notifications pass through it.

## Screenshots
N/A

## Additional Context
The working should be checked on phone, since I couldn't check it.
I could only provide a logical fix and tested through PC (windows).

## Checklist

yup

8 (because did test only on PC and not phone)

N/A

none

10% - Copilot helped me identify the possible files where the bug for this error might be since I am not still accustomed to the codebase. I filtered the files after that. No AI was used for coding purposes.

### DIFF
--- a/lib/webPush.js
+++ b/lib/webPush.js
@@ -102,6 +102,22 @@ async function sendUserNotification (userId, notification) {
     if (!userId) {
       throw new Error('user id is required')
     }
+
+    const user = await models.user.findUnique({ where: { id: userId } })
+    if (!user) return
+
+    if (user.satsFilter > 0 && notification.itemId) {
+      const item = await models.item.findUnique({
+        where: { id: notification.itemId },
+        select: { cost: true, boost: true, msats: true }
+      })
+      const sats = (item?.cost || 0) + (item?.boost || 0) + ((item?.msats || 0) / 1000)
+      if (sats < user.satsFilter) {
+        // respect satsFilter
+        return
+      }
+    }
+
     notification.data ??= {}
     if (notification.itemId) {
       // legacy Push API notificationclick event needs data.url as the navigate key is consumed by the browser
@@ -189,7 +205,7 @@ export async function notifyTerritorySubscribers ({ models, item }) {
     SELECT "userId" FROM "SubSubscription"
     WHERE "subName" = $1
     AND NOT EXISTS (SELECT 1 FROM "Mute" m WHERE m."muterId" = "SubSubscription"."userId" AND m."mutedId" = $2)
-    `, subName, Number(item.userId))
+    `, subName, Number(item.userId))  
 
     const author = await models.user.findUnique({ where: { id: item.userId } })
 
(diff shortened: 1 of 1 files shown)


axelvyrn:
@ekzyis hope there are no conflicts now.

axelvyrn:
I have a bad feeling about the notifyThreadSubscribers and notifyItemParents functioning though

ekzyis (changes requested):
https://github.com/stackernews/stacker.news/pull/2442#pullrequestreview-3153141595

axelvyrn:
> Can you be more precise? If you tested it, why do you have a bad feeling that they aren't functioning?

Because testing on development mode often gives false negatives/positives. On dev, process.env.NODE_ENV === 'development', and unless I've set the VAPID keys, push notifications might not actually be sent (which actually is complicated for me since I've never worked with keys or CORS stuff)

Even if they are sent (like via localhost), the “/notifications” page may behave differently because it might be populated via API without filtering, or the dev DB might have different msats/cost values. So only you can test it properly on a production build.

Soxasora:
> and unless I've set the VAPID keys, push notifications might not actually be sent (which actually is complicated for me since I've never worked with keys or CORS stuff)

You can obtain VAPID keys and enable true push notifications by following our [**README section on it**](https://github.com/stackernews/stacker.news?tab=readme-ov-file#enabling-web-push-notifications). Even without this, you can place a `console.log` in the appropriate section in `webPush.js` and watch the `worker` container logs.

> the “/notifications” page may behave differently because it might be populated via API without filtering

This issue is all about the fact that the **`/notifications` resolver** has the `investmentClause` that filters out posts that are worth less than a user-set amount of sats (among other things), while `webPush.js` doesn't.

> or the dev DB might have different msats/cost values.

The dev DB seed is only a starter point, you can still create posts with 2 accounts and test this out.

---

I'm sure you understand the implications of pushing **untested code into production**, I'm sure you wouldn't do that to yourself either. Fortunately we can spot stuff, but you just lose time, rep, and, honestly, money. You reached a 70% bounty reduction this way, other than giving us the feeling that you're not doing anything of value. 

[Look at rule 3 of pull request awards](https://github.com/stackernews/stacker.news?tab=readme-ov-file#rules)

If you can't test something, **don't rush into review**, search or think about how you can test stuff and if you can't, as a last resort, ping us, we can reply just like we are doing now.

axelvyrn:
> The dev DB seed is only a starter point, you can still create posts with 2 accounts and test this out.

That is what I have been doing.

> I'm sure you understand the implications of pushing **untested code into production**, I'm sure you wouldn't do that to yourself either. Fortunately we can spot stuff, but you just lose time, rep, and, honestly, money. You reached a 70% bounty reduction this way, other than giving us the feeling that you're not doing anything of value.

I understand and am sorry, and I'm not worried about the bounty though. I thought I knew the solution so I tried it out. Seems like I was wrong because the way I wanted to solve it is not required. I'll try one last time.

PR #2533, opened 2025-09-14
support audio embeds

### LINKED ISSUE #2526: Support audio embeds
https://stacker.news/items/1220611

They end up looking like this 

Image

pory-gone:
## Description

fix #2526 

this implementation add native audio file embedding, allowing user to embed direct audio files links as playable HTML5 audio players within posts and comments.

Extended `parseEmbedUrl()` for audio file detection;
Added new "audio" provider to embed framework;
Extended `useMediaHelper `with audio detection cascade;
Integrated with existing `MediaOrLink` component;
Added CSS styling with theme compatibily.

## Screenshots

https://github.com/user-attachments/assets/ba7b4635-4b2a-4413-99ab-e7b35b2e03ac

Screen Shot 2025-09-14 at 20 59 56

## Additional Context

Audio detection is positioned early in `parseEmbedUrl()` to catch file links before other providers, uses both embed framework and media helper to garant a fallback if one fails, 
supports query parameters in URLs (e.g., `file.mp3?version=1`), used `!important `CSS override to prevent margin inheritance from general embed rules,
implemented audio detection cascade in `useMediaHelper` to maintain existing video/image logic, added `preload='metadata'` for performance while avoiding automatic downloads

## Checklist

Yes

8/10

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:** yes

NaN

I used AI to understand the context and identify the files to be modified, I asked for some implementation advice but the changes were all carefully considered and applied manually

---

> [!NOTE]
> Adds first-class audio embedding for direct audio links with detection, playback UI, styling, and allowed MIME types.
> 
> - **Embeds**
>   - Add `audio` provider in `parseEmbedUrl` (detects `mp3/wav/ogg/flac/aac/m4a/opus/webm`, extracts `audioType` and title).
>   - Render HTML5 audio player in `components/embed.js` with optional title and download link; new `.audioWrapper` styles in `components/text.module.css`.
> - **Media handling**
>   - `components/media-or-link.js`: support `audio` prop; render `<audio>` when detected; update `loaded` logic and error handling.
>   - `useMediaHelper`: cascade detection (video → audio → image); expose `audio` flag and include in `showMedia` gating.
> - **Constants/URL**
>   - Extend `UPLOAD_TYPES_ALLOW` with audio MIME types; add `AUDIO_EXTENSIONS` and `AUDIO_URL_REGEXP`.
> - **Settings UI**
>   - Update copy in `pages/settings/index.js` to include audio and list direct audio files as supported embeds.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 589a708eb566f86bb5dd84ce54ea354bbacd8a45. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/embed.js
+++ b/components/embed.js
@@ -280,6 +280,10 @@ const Embed = memo(function Embed ({ src, provider, id, meta, className, topLeve
     )
   }
 
+  if (provider === 'audio') {
+    return <AudioEmbed src={src} meta={meta} className={className} />
+  }
+
   return null
 })
 
--- a/components/embed.js
+++ b/components/embed.js
@@ -5,6 +5,75 @@ import styles from './text.module.css'
 import { Button } from 'react-bootstrap'
 import { TwitterTweetEmbed } from 'react-twitter-embed'
 import YouTube from 'react-youtube'
+import FileMusicLine from '@/svgs/file-music-line.svg'
+import { registerAudioElement } from '@/lib/audio-manager'
+
+function AudioEmbed ({ src, meta, className }) {
+  const [error, setError] = useState(false)
+  const audioRef = useRef(null)
+
+  const handleError = (e) => {
+    console.warn('Audio loading error:', e)
+    setError(true)
+  }
+  useEffect(() => {
+    if (!audioRef.current) return
+    return registerAudioElement(audioRef.current)
+  }, [audioRef.current])
+
+  if (error) {
+    return (
+      <div className={classNames(styles.audioWrapper, className)}>
+        <div style={{ padding: '1rem', textAlign: 'center' }}>
+          <p style={{ marginBottom: '0.5rem', color: 'var(--theme-color)' }}>
+            Unable to play this audio file.
+          </p>
+          <a
+            href={src}
+            target='_blank'
+            rel='noreferrer noopener'
+            style={{ color: 'var(--bs-primary)' }}
+          >
+            Download or open in new tab
+          </a>
+        </div>
+      </div>
+    )
+  }
+
+  return (
+    <div className={classNames(styles.audioWrapper, className)}>
+      {meta?.title && (
+        <div style={{
+          fontSize: '14px',
+          fontWeight: '500',
+          marginBottom: '8px',
+          color: 'var(--theme-color)',
+          overflow: 'hidden',
+          textOverflow: 'ellipsis',
+          whiteSpace: 'nowrap',
+          display: 'flex',
+          alignItems: 'center',
+          gap: '6px'
+        }}
+        >
+          <FileMusicLine width={16} height={16} style={{ flexShrink: 0 }} />
+          {meta.title}
+        </div>
+      )}
+      <audio
+        ref={audioRef}
+        controls
+        preload='none'
+        style={{ width: '100%' }}
+        onError={handleError}
+      >
+        <source src={src} type={`audio/${meta?.audioType || 'mpeg'}`} />
+        Your browser does not support the audio element.
+      </audio>
+    </div>
+  )
+}
 
 function TweetSkeleton ({ className }) {
   return (
--- a/components/media-or-link.js
+++ b/components/media-or-link.js
@@ -2,9 +2,10 @@ import styles from './text.module.css'
 import { useState, useEffect, useMemo, useCallback, memo, useRef } from 'react'
 import { decodeProxyUrl, IMGPROXY_URL_REGEXP, MEDIA_DOMAIN_REGEXP } from '@/lib/url'
 import { useMe } from './me'
-import { UNKNOWN_LINK_REL, PUBLIC_MEDIA_CHECK_URL } from '@/lib/constants'
+import { UNKNOWN_LINK_REL } from '@/lib/constants'
 import classNames from 'classnames'
 import { useCarousel } from './carousel'
+import { registerAudioElement } from '@/lib/audio-manager'
 
 function LinkRaw ({ href, children, src, rel }) {
   const isRawURL = /^https?:\/\//.test(children?.[0])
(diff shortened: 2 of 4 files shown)


components/embed.js
```diff
@@ -195,7 +195,33 @@ const Embed = memo(function Embed ({ src, provider, id, meta, className, topLeve
...
+            fontSize: '14px',
+            fontWeight: '500',
+            marginBottom: '8px',
+            color: 'var(--theme-color)',
+            overflow: 'hidden',
+            textOverflow: 'ellipsis',
+            whiteSpace: 'nowrap'
+          }}
+          >
+            🎵 {meta.title}
```
adlai: is "notes" the best graphic for representing audio? lots of audio might not be music, strictly speaking...

lib/url.js
```diff
@@ -266,6 +279,9 @@ export function isMisleadingLink (text, href) {
   return misleading
 }
 
+// Add after IMG_URL_REGEXP
```
adlai: I believe the above comment is forensically interesting [i.e., "look, Copilot is talking to the operator"], and probably does not belong in the production codebase.

lib/url.js
```diff
@@ -103,7 +103,20 @@ export function parseEmbedUrl (href) {
...
+    const audioExtensions = /\.(mp3|wav|ogg|flac|aac|m4a|opus|webm)(\?.*)?$/i
+    if (pathname && audioExtensions.test(pathname)) {
+      const extension = pathname.match(audioExtensions)[1].toLowerCase()
+      return {
+        provider: 'audio',
+        id: null,
+        meta: {
+          href,
+          audioType: extension,
+          title: decodeURIComponent(pathname.split('/').pop().split('.')[0])
```
adlai: is there no better guess for the reasonable text label of uncaptioned audio? I'm thinking, some combination of comment author, context [i.e. SN item ID], and filename only if it's not some hash or blob identifier, otherwise some reasonable timestamp.

lib/constants.js
```diff
@@ -33,8 +33,15 @@ export const UPLOAD_TYPES_ALLOW = [
...
-  'video/webm'
+  'video/webm',
+  'audio/mpeg',
+  'audio/wav',
+  'audio/ogg',
+  'audio/mp4',
+  'audio/aac',
+  'audio/flac'
 ]
+export const AUDIO_EXTENSIONS = ['mp3', 'wav', 'ogg', 'flac', 'aac', 'm4a', 'opus']
```
adlai: for some folks, the above lines might not be obviously fine; however, line 44 is about file types, while 37-42 are for MIME types, and there isn't _any_ one-to-one mapping between categories of types.

adlai (commented):
I really shouldn't be considered qualified to review anything related to webapps; however, my comments are provided in the hope that they help reduce the concerns of critics who wish to review the PR.

components/embed.js
Soxasora: While arguable, audio files are historically represented by a music note, so it's okay to use it.

But @pory-gone you should use an SVG instead of an emoji: https://remixicon.com/icon/file-music-line

lib/url.js
```diff
@@ -266,6 +279,9 @@ export function isMisleadingLink (text, href) {
   return misleading
 }
 
+// Add after IMG_URL_REGEXP
+export const AUDIO_URL_REGEXP = /^(https?:\/\/.*\.(?:mp3|wav|ogg|flac|aac|m4a))$/i
```
Soxasora: Unused and not necessary, probably you got inspiration from the other regexes at the bottom, but they don't serve a purpose afaik

lib/constants.js
Soxasora: Unused and not necessary

components/media-or-link.js
```diff
@@ -124,28 +135,34 @@ export const useMediaHelper = ({ src, srcSet: srcSetIntital, topLevel, tab }) =>
...
-        setIsImage(true)
-      }).catch((e) => {
-        console.warn('Cannot decode image:', src, e)
-      })
+      const audio = document.createElement('audio')
+      audio.onloadedmetadata = () => {
+        setIsAudio(true)
+        setIsImage(false)
+        setIsVideo(false)
+      }
```
Soxasora, line 160: Audio files can be reproduced as videos, so it will never create an audio element:

Image

(this is `<video>`)

components/media-or-link.js
```diff
@@ -45,29 +45,40 @@ const Media = memo(function Media ({
...
             src={src}
-            preload={bestResSrc !== src ? 'metadata' : undefined}
             controls
-            poster={bestResSrc !== src ? bestResSrc : undefined}
+            preload='metadata'
             width={width}
             height={height}
             onError={onError}
             onLoadedMetadata={handleLoadedMedia}
           />
```
Soxasora: This way we would have a styled `Audio` component (parsed from link) and an unstyled audio container (from upload).

You have some choices:
- Don't use `Embed` for audios
  - you wouldn't need an extension, but you'd need to fix the way you recognize it's an audio in `useMediaHelper`
- Upload audio files without the `` markdown so it can be parsed as embed
  - you would also need to append the extension based on the mime type, because your parser checks for extensions, and then remove the extension when the embed... but then the raw link would be broken... mhh...
- `Audio` component shared between embeds and uploads
  - still two logics for audios

lib/url.js
Soxasora: This PR focuses on embedding audio from a link, this would just complicate things

Soxasora (changes requested):
Well, it definitely can embed links that ends with an audio extension.

**Logic**

There are two logics (mmh) and both are flawed:
- via `Embed`: link has to end with an extension
- via `audio.onload`: audio files can be reproduced as video elements

I left a comment on how you could address this.
I want to note that yesterday I opened PR #2599 that would remove the need to load the file to know its type. It could come in handy for this PR.

**Style**

The style has weird measurements, is inconsistent across browsers and the dark/light switch is pretty painful:

<details>
<summary>Chrome</summary>
Image
Image
</details>

<details>
<summary>Safari</summary>
Image
Image
</details>

The animation is cool to see, but not really needed and makes thinking about the style harder than it could be.

---

I left some comments about other things I found, I'd like for you to address them.

In conclusion, focus on using the best logic you can think of. That's what's most important!

pory-gone:
Thank you for the detailed review. I've implemented all your requests: unused code removed, emoji replaced with SVG, detection logic fixed. You were right about the critical issue! Audio files were being caught as video in the cascade detection. I've fixed it through improved error handling. As you suggested, this PR focuses only on embedding direct audio links.

adlai (commented):
icon looks reasonable; however appears to be different from the ones in the original screenshots?

one quibble, I'd name it "note" rather than "music", and the one in the screenshots and video "notes"...

adlai:
@pory-gone I hope you're drafting a revised PR!

Soxasora:
> revised PR

This PR is not valid anymore, we switched both the markdown renderer and editor with Lexical Editor. @pory-gone would need to either adapt their Audio embed solution or create an `AudioNode` similar to `MediaNode`.

PR #2534, opened 2025-09-15
Fix error message if no route found

ekzyis:
## Description

We currently log the following:

> failed to wrap invoice: Unable to estimate route, excessive values: {"routing_fee_msat":"0","time_lock_delay":"0","failure_reason":"FAILURE_REASON_NO_ROUTE"}

The code tried to use `res.failure_reason` if it exists, but it didn't reach that line because `err` is apparently not set.

This PR fixes this.

Will lookup the source code of `lnd.router.estimateRouteFee` to make sure errors are handled properly.

## Checklist

yes

`8`. Tested wrapping via p2p zaps

n/a

no

no

### DIFF
--- a/api/lnd/index.js
+++ b/api/lnd/index.js
@@ -60,17 +60,11 @@ export async function estimateRouteFee ({ lnd, destination, tokens, mtokens, req
       timeout
     }, (err, res) => {
       if (err) {
-        if (res?.failure_reason) {
-          reject(new Error(`Unable to estimate route: ${res.failure_reason}`))
-        } else {
-          reject(err)
-        }
-        return
+        return reject(err)
       }
 
-      if (res.routing_fee_msat < 0 || res.time_lock_delay <= 0) {
-        reject(new Error('Unable to estimate route, excessive values: ' + JSON.stringify(res)))
-        return
+      if (res.failure_reason !== 'FAILURE_REASON_NONE' || res.routing_fee_msat < 0 || res.time_lock_delay <= 0) {
+        return reject(new Error(`Unable to estimate route: ${res.failure_reason}`))
       }
 
       resolve({


ekzyis:
> Will lookup the source code of `lnd.router.estimateRouteFee` to make sure errors are handled properly.

Wow, this is annoying, it uses `lightning` under the hood but [the repository](https://github.com/alexbosworth/lightning) contains no tags so I am using the [npm code section](https://www.npmjs.com/package/lightning/v/10.22.0?activeTab=code) to make sure I am reading the correct version (ln-service v57.22.0 → lightning v10.22.0).

All I found is the `RouteFeeResponse` message:

2025-09-15-163618_680x575_scrot

So still not sure 100% if `err` is never passed, but I think we can rely on `failure_reason !== FAILURE_REASON_NONE`.

api/lnd/index.js
```diff
@@ -58,18 +58,9 @@ export async function estimateRouteFee ({ lnd, destination, tokens, mtokens, req
...
-          reject(err)
-        }
-        return
-      }
-
-      if (res.routing_fee_msat < 0 || res.time_lock_delay <= 0) {
-        reject(new Error('Unable to estimate route, excessive values: ' + JSON.stringify(res)))
+    }, (_, res) => {
+      if (res.failure_reason !== 'FAILURE_REASON_NONE' || res.routing_fee_msat < 0 || res.time_lock_delay <= 0) {
+        reject(new Error(`Unable to estimate route: ${res.failure_reason}`))
```
ekzyis: `res.failure_reason` should always be set to something

ekzyis: The previous code handled "excessive values" separately though :thinking:

@huumn, do you remember if there was a good reason for it?

PR #2542, opened 2025-09-16
Twitter embeds for xcancel links

ekzyis:
## Description

This enables parsing of xcancel links to twitter embeds.

## Additional context

My bot will not reply to xcancel.com links with xcancel.com links though. So if somebody uses a xcancel.com link, we actually now basically just replace it with the x.com link 🤔

I can update it though.

## Checklist

yes

tbd

n/a

no

no

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -93,7 +93,7 @@ export function parseEmbedUrl (href) {
   if (!href) return null
 
   try {
-    const twitter = href.match(/^https?:\/\/(?:twitter|x)\.com\/(?:#!\/)?\w+\/status(?:es)?\/(?<id>\d+)/)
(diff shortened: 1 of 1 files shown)


ekzyis:
> My bot will not reply to xcancel.com links with xcancel.com links though. So if somebody uses a xcancel.com link, we actually now basically just replace it with the x.com link 🤔

Nitter seems to have its own embeds we could use instead: https://github.com/zedeus/nitter/issues/23

Just need to append `/embed` to the xcancel.com link.

update: mhh, but xcancel.com blocks embedding iframes via `frame-ancestor: none` in their CSP :/

ekzyis:
We can use nitter.net instead:

2025-09-16-105719_610x559_scrot

ekzyis:
Ok making sure the embed is styled correctly (width, height, button to show full tweet) takes more time than I'm willing to put in right now

---

Closing because I don't want to keep this PR with that one line changed around

Issue #2502, opened 2025-09-09
[Feature Request ] Sats per Big Mac ticker


ed-kung:
### Describe the problem you're trying to solve

Per [#1214415](https://stacker.news/items/1213943/r/SimpleStacker?commentId=1214415), would be fun to have a `X sats/Big Mac` ticker on SN.

### Describe the solution you'd like

- Should draw from a credible and reliable datastream for big mac pricing (I don't know of one, yet)
- As far as possible should try to emulate the methodology of the Economist's official Big Mac Index  (requires some research into the methodology)

### Additional context

Happy to self-assign this (though no promises on how quickly I can get it done)

PR #2543, opened 2025-09-16
Sats per Big Mac

pory-gone:
## Description

adding requested feature in #2502

Added a new "Big Mac Index" ticker to the price carousel that displays how many satoshis are needed to buy a Big Mac;

- Added `'bigmac'` to price carousel rotation
- Creating a new `GraphQL` resolver that fetches Big Mac prices from [The Economist's official GitHub repository](https://raw.githubusercontent.com/TheEconomist/big-mac-data/master/output-data/big-mac-raw-index.csv)
- Implemented proper caching (24 hours for Big Mac data, 1 minute for Bitcoin prices)

## Screenshots

Istantanea_2025-09-16_10-16-43

---

489962047-b8a30ca7-dda5-4b4c-baea-01b54b409fe3

---

489962082-92bc5c3c-93da-4aa9-aac6-eb744b7a4b78

## Additional Context

The data is cached for 24 hours since Big Mac prices don't change frequently. The feature includes robust error handling with fallback to the last known Big Mac price ($5.79) if the external data source is unavailable. Used existing `cachedFetcher` pattern for data fetching

## Checklist

yes

8/10

yes

NaN

i did use ai just to understand the context and do research about big mac prices

### DIFF
--- a/api/resolvers/price.js
+++ b/api/resolvers/price.js
@@ -18,10 +18,42 @@ const getPrice = cachedFetcher(async function fetchPrice (fiat = 'USD') {
   keyGenerator: (fiat = 'USD') => fiat
 })
 
+const getBigMacPrice = cachedFetcher(async function fetchBigMacPrice () {
+  const csvUrl = 'https://raw.githubusercontent.com/TheEconomist/big-mac-data/master/output-data/big-mac-raw-index.csv'
+  try {
+    const res = await fetch(csvUrl)
+    const csvText = await res.text()
+    const lines = csvText.split('\n')
+    const usaEntries = lines
+      .filter(line => line.includes(',USA,USD,'))
+      .map(line => {
+        const cols = line.split(',')
+        return {
+          date: cols[0],
+          price: parseFloat(cols[4])
+        }
+      })
+      .filter(entry => !isNaN(entry.price))
+      .sort((a, b) => new Date(b.date) - new Date(a.date))
+    return usaEntries[0]?.price || 5.79
+  } catch (err) {
+    console.error('Big Mac price fetch error:', err)
+    return 5.79
+  }
+}, {
+  maxSize: 1,
+  cacheExpiry: 24 * 60 * 60 * 1000,
+  forceRefreshThreshold: 0,
+  keyGenerator: () => 'bigmac-usd'
+})
+
 export default {
   Query: {
     price: async (parent, { fiatCurrency }, ctx) => {
       return await getPrice(fiatCurrency) || -1
+    },
+    bigMacPrice: async () => {
+      return await getBigMacPrice() || 5.79
     }
   }
 }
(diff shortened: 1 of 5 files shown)


ed-kung:
TIL the Economist actually maintains a github repository for their Big Mac Index.  Nice haha good job.

pory-gone:
Yeah! I was surprised too, but apparently it's true that you can find anything on the internet. Thank you so much!

api/resolvers/price.js
```diff
@@ -18,10 +18,42 @@ const getPrice = cachedFetcher(async function fetchPrice (fiat = 'USD') {
   keyGenerator: (fiat = 'USD') => fiat
 })
 
+const getBigMacPrice = cachedFetcher(async function fetchBigMacPrice () {
+  const csvUrl = 'https://raw.githubusercontent.com/TheEconomist/big-mac-data/master/output-data/big-mac-raw-index.csv'
```
ekzyis, line 22: I think this is good enough for the first version, but I think it would be a lot cooler if the data would update more frequently. It seems to update every 6 months but for some reason, the last update is 2025-01-01, not 2025-07-01 :thinking: 

They documented [here](https://github.com/TheEconomist/big-mac-data) how they calculated it.

I think it would be really funny if we would have the most up-to-date big mac prices just to show them in sats, could be good for marketing haha

But as mentioned, more something for a follow-up PR, #2502 is only labeled as good-first-issue anyway.

ekzyis (commented):
Only looked at the source for the big mac prices so far, left a comment. Will look at the rest later this week!

components/price.js
```diff
@@ -106,6 +108,15 @@ export default function Price ({ className }) {
     )
   }
 
+  if (selection === 'bigmac') {
+    if (!price || price < 0 || !bigMacPrice) return null
+    return (
+      <AccessibleButton id='bigmac-hint' description='Show satoshis per Big Mac' className={compClassName} onClick={handleClick} variant='link'>
+        {fixedDecimal(Math.round((bigMacPrice / price) * 100000000), 0)} sats/Big Mac
```
ekzyis, line 115: Maybe `sats/🍔` is better

api/resolvers/price.js
ekzyis: Oh wait, I forgot the number we display is going to update frequently because we calculate it in sats anyway. I thought we would just be staring at the same number for a long time.

And I assume Big Mac prices in USD don't update that often, that's why The Economist is also not issuing updates very often.

components/price.js
Soxasora: What if it's a Whopper instead? 👀

pory-gone: we can put the whole MC menu :P

api/resolvers/price.js
pory-gone: I tried to find a more frequently updated price source but Big Macs are as stable as gold ahah

components/price.js
ed-kung: fwiw, I like `sats/Big Mac` more than `sats/🍔` 😅

Issue #2017, opened 2025-03-24
Sorting search by recent is confusing


ekzyis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

See https://stacker.news/items/922174?commentId=923744

### Screenshots

![](https://m.stacker.news/83946)

### Steps To Reproduce

_not verified myself:_

1. Go to search
2. Sort by recent
3. Some results are in wrong order 

### Expected behavior

All results are chronologically sorted

huumn:
it's just no longer *strictly* ordered by recent. It's now weighted by recent. 

I understand the frustration but I'm not sure what we should do about it. @SimpleStacker do you have an opinion?

ed-kung:
Yes I can see that it might be confusing and misunderstood as a bug.  If we think strict chronological order is very important, I think we'd have to do something where a relevance cutoff score is used to determine which results are retrieved, then all retrieved results are sorted strictly by date.  (I think that's what the previous iteration was trying, but the cutoff may have been too strict, requiring exact phrase matches).

Overall though I think the important metric is how many clicks does it take a user to find what they're looking for. We can't really measure that directly right now, but I was actually planning to make a bounty post asking something along the lines of:

"Think of a post you remember from stacker.news, but don't remember the exact title or details. Try to search for it. What search query did you use, and were you able to find it quickly?"

huumn:
New example: https://stacker.news/items/943984/r/k00b

ekzyis:
Another example:

![Image](https://github.com/user-attachments/assets/ff61d111-6349-4af7-bb96-7c67d9ae90d4)

ed-kung:
Another one: https://stacker.news/items/1222373/r/SimpleStacker?commentId=1222765

I think there is enough evidence that this is a common user expectation and it should be implemented as such.

PR #2545, opened 2025-09-17
Strict chronological ordering for sort=recent

ed-kung:
## Description

Closes https://github.com/stackernews/stacker.news/issues/2017

Implements strict chronological ordering when `sort=recent`, while maintaining retrieval of relevant results.

## Screenshots

Before: 

Screenshot 2025-09-16 at 6 30 35 PM

After:

Screenshot 2025-09-16 at 10 51 10 PM

## Additional Context

Trying to balance relevance with strict chronological ordering was tricky, so I went with a compromise solution.  Return only the first page of `2 * LIMIT` most relevant results (with relevance scored according to both semantic relevance and recency), and strictly order those results chronologically.

I went with this approach for the following reasons:

- Users report usually being able to find what they're looking for, which suggests that current relevancy metrics are working (https://stacker.news/items/928640)
- However, users expect strict chronological ordering when sorting by recent (https://stacker.news/items/1222373?commentId=1222765)

The compromise solution is therefore to use relevancy metrics to find the top K most relevant (which includes a recency bias already), and then strictly sorting the K results by date.  This meets the user's expectations of strict chronological sorting while still making sure the top most relevant results are included in the first page of results.

Note that with this approach, I turned off pagination for `sort=recent`. Only one page of the `2 * LIMIT` most relevant results is shown. It's difficult to get full chronological sorting across pages due to how pagination works in the current search function.  (We want OpenSearch to get the top K most relevant results, but that also results in OpenSearch ranking the results by relevance rather than by date. The current pagination approach uses an offset in the OpenSearch query itself, so to get fully chronological pagination while also maintaining a relevancy criteria requires a major rewrite of the pagination approach just for `sort=recent`.  I think the one page of results approach was the best compromise.)

## Checklist

Yes

8.  

n/a

n/a

I mainly used AI as a thought partner and as an alternative to googling for OpenSearch syntax and functionality. I asked AI to suggest where to make the edits, but I made the ultimate changes on my own.  AI was helpful in getting me started, but its suggestions were ultimately not that helpful because it suggested some janky approaches to dealing with the pagination issues.

### DIFF
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -455,7 +455,7 @@ export default {
 
       const searchRequest = {
         index: process.env.OPENSEARCH_INDEX,
-        size: LIMIT,
+        size: sort === 'recent' ? 2 * LIMIT : LIMIT,
         _source: {
           excludes: [
             'text',
@@ -525,7 +525,7 @@ export default {
       })
 
       return {
-        cursor: items.length === LIMIT ? nextCursorEncoded(decodedCursor) : null,
+        cursor: items.length === LIMIT && sort !== 'recent' ? nextCursorEncoded(decodedCursor) : null,
         items
       }
     }


huumn:
What we seem to want is

1. to *filter* by relevance (perhaps more strictly than usual, requiring some threshold of relevance be crossed)
2. *sort* by time

There should by a way to do this with opensearch.

ed-kung:
That's what this is trying to do. The filter is essentially the Top-K most relevant, rather than a numerical threshold (since I don't know what the right numerical threshold is), then sorting chronologically.

huumn:
Your PR does do that - but without paging. At a practical level this is perfectly fine, but the junior engineer in me wants to it The Right Way even if it doesn't matter.

ed-kung:
I can understand that. Let me think about how best to accomplish.

huumn:
I figured it out in https://github.com/stackernews/stacker.news/commit/a4b1b585bc58118bc818bb43cbcfbd75383650ac

Open Search supports a threshold for `_score` so I did some queries and cutoff where it looked like results were irrelevant.

Then you can tell it to sort by something other than relevance.

ed-kung:
Ah, cool. Yeah I figured it'd be something like that, and it'd just be an issue of finding a threshold. Though I'd still be a bit concerned if the optimal threshold is different for different keywords and types of searches.

Any reason you changed the default `pagination_depth`?

huumn:
to find the right threshold I was fussing with the limits in dev to get look at relevance deeper down. iirc we selected 50 because it was roughly twice a page's worth, so I made it a function of the limit (should we ever want to expose limit in the query).

PR #2546, opened 2025-09-17
Replace lib/hex with Buffer.from

ekzyis:
## Description

When converting to/from hex, we are using `Buffer.from()` in every other place, so this makes our code more consistent.

## Checklist

yes

`10`. Tested by comparing `iv` and `encrypted` of the previous code with the new code and saving a send wallet.

n/a

no

no


huumn:
this is available in browsers now? that's cool, i'm pretty sure it wasn't a few years ago

ekzyis:
No, [NextJS injects polyfills](https://nextjs.org/docs/architecture/supported-browsers#polyfills) and `Buffer` seems to be one of them even though I haven't found that specific polyfill mentioned in their docs 🤔

But I ran the code and it works

PR #2548, opened 2025-09-18
Allow testing of magic code with sndev

ekzyis:
## Description

With these changes, one can test the magic login code via `sndev login test_email_login`.

If the correct nym is used, it will print this:

```
$ sndev login test_email_login

open url in browser
http://localhost:3000/api/auth/callback/email?token=sndev1&email=test_email_login%40sndev.team

or go to http://localhost:3000/email and enter sndev1 as the token
```

As mentioned, you can then go to /email and enter sndev1 to login.

## Additional context

all these `process.env.NODE_ENV` checks in our code are quite ugly, we should put a constant like `IS_DEV` or `IS_PROD` into lib/constants and use that

## Checklist

yes

`10`. Only affects `sndev`

n/a

no

no

### DIFF
--- a/pages/email.js
+++ b/pages/email.js
@@ -17,7 +17,13 @@ export default function Email () {
 
   useEffect(() => {
     setSignin(!!cookie.parse(document.cookie).signin)
-    setCallback(JSON.parse(window.sessionStorage.getItem('callback')))
+    setCallback(
+      JSON.parse(window.sessionStorage.getItem('callback')) ?? (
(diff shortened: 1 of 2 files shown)


huumn:
I usually use mailhog for this and that was always pretty easy.

ekzyis:
> I usually use mailhog for this and that was always pretty easy.

Oh, right, I never used it and then completely forgot that we have it

PR #2553, opened 2025-09-19
Fix draft gone if wallet prompt closed

ekzyis:
## Description

Since we didn't throw `WalletPromptClosed` but simply return, the form believed if the prompt was closed, the post or comment was submitted successfully and deleted the draft in localStorage.

This was only apparent when going to /wallets, because if we simply close the modal via the `x` button, we can still see the value in the text input, but the draft in localStorage is gone.

## Checklist

yes

`8`. Drafts are no longer lost if we visit /wallets from the prompt or close the prompt ourselves.

n/a

no

no

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -39,6 +39,7 @@ import { useShowModal } from './modal'
 import dynamic from 'next/dynamic'
 import { useIsClient } from './use-client'
 import PageLoading from './page-loading'
+import { WalletPromptClosed } from '@/wallets/client/hooks'
 
 export class SessionRequiredError extends Error {
   constructor () {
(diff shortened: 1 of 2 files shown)


components/form.js
```diff
@@ -1087,6 +1088,7 @@ export function Form ({
         await onSubmit(values, ...args)
       }
     } catch (err) {
+      if (err instanceof WalletPromptClosed) return
```
ekzyis, line 1091: I don't like how far away catching the error feels from where the error is thrown, but since `WalletPromptClosed` is semantically indeed more a submit error than success, it also kind of does make sense to throw it all the way up to here. :thinking:

PR #2557, opened 2025-09-20
Coinos guide

ekzyis:
https://stacker.news/items/1230105


ekzyis:
I updated the last image because of #2558, so I need to update the SN post when this is merged.

PR #2563, opened 2025-09-21
Refactor NWC url parsing

ekzyis:
## Description

All these checks if the value is set before including it in the returned object were useless.

## Checklist

yes

`10`. Saving and zapping via NWC still works. The function is only used when making NWC requests.

n/a

n/a

no


ekzyis:
wrt https://github.com/stackernews/stacker.news/pull/2562#issuecomment-3316600527:

See, this PR has no changes because you merged #2564 first

Not blaming you for merging it first, just want to point out that PRs shouldn't have changes anymore when they have been included in other PRs that have already been merged

huumn:
you are doing one pr per commit. it's hard to track wtf is going on most of the time.

PR #2566, opened 2025-09-21
fix(Comment): Remove scroll to top when changing comment sort order.

### LINKED ISSUE #2555: Changing comment sort scrolls page to the top

### Describe the bug

Changing comment sort scrolls the page to the top in non-Safari browsers.

I used `git bisect` to find that 97cbfec38e73b77159a570360a67ca03f151c5a0 introduced this bug because of [this line](https://github.com/stackernews/stacker.news/commit/97cbfec3#diff-861d26ff366eaa703915462b74b0ec46c10c35c48135b9bb2922c4ae31d99ba3R181).

### Screenshots

https://github.com/user-attachments/assets/3daa7b35-23db-4c39-960b-db2a6839ad22

### Steps To Reproduce

1. Change comment sort
2. Get scrolled to the top

### Expected behavior

1. Change comment sort
2. Do not get scrolled to the top

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

brymut:
## Description

fix #2555 
Prevent scroll to top because of the CarouselProvider getting unmounted.
Rationale: https://github.com/stackernews/stacker.news/issues/2555#issuecomment-3316089617

## Additional Context

## Checklist

Y

10

Y

N

N

### DIFF
--- a/components/item-full.js
+++ b/components/item-full.js
@@ -25,7 +25,6 @@ import { UNKNOWN_LINK_REL } from '@/lib/constants'
 import classNames from 'classnames'
 import { CarouselProvider } from './carousel'
 import Embed from './embed'
-import { useRouter } from 'next/router'
 import useCommentsView from './use-comments-view'
 
 function BioItem ({ item, handleClick }) {
@@ -169,9 +168,6 @@ export default function ItemFull ({ item, fetchMoreComments, bio, rank, ...props
     markItemViewed(item)
   }, [item.id, markItemViewed])
 
-  const router = useRouter()
-  const carouselKey = `${item.id}-${router.query?.sort || 'default'}`
-
   return (
     <>
       {rank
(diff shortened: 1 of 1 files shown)


ekzyis (changes requested):
This is not the fix we want because it breaks the order of images in the carousel. We don't want to replace one bug with another bug.

huumn:
If this only bugs image order on sort change - it's probably worth the tradeoff (the bug this introduces is less severe than scroll to top).

It may be possible to do both at the same time - perhaps if we delay the key change? Kind of hacky but might be clean relative to other things

ekzyis:
> If this only bugs image order on sort change - it's probably worth the tradeoff (the bug this introduces is less severe than scroll to top).

Yeah, I see how changing comment sort is used way more frequently than relying on the image order in the carousel (after comment sort change).

I just requested changes because I know how and for how long I was annoyed with the image order not being reliable so I didn't want to regress it. But after this, the image order should still be reliable on first visit of an item, only no more when you change the comment sort.

So we can merge and I will open a ticket for that. Dismissing my review.

PR #2576, opened 2025-09-23
adding profile share button

### LINKED ISSUE #2573: Add button to share profile
https://stacker.news/items/1233318

pory-gone:
## Description

fix #2573

Added a profile sharing feature that allows users to share profile links across all platforms.

- Share button in the user header component visible on all user profiles
- Share option in the action dropdown menu for viewing other users' profiles
- Smart referral logic that adds referral links (`/r/{username}`) only when sharing via mobile Web Share API
- Cross-platform compatibility with Web Share API on touch devices and clipboard copy fallback on desktop

## Screenshots

Screenshot_2025-09-23_11_33_34

localhost_3000_(iPhone 14 Pro Max)

localhost_3000_(iPhone 14 Pro Max) (1)

localhost_3000_(iPhone 14 Pro Max) (2)

## Additional Context

The implementation reuses existing share infrastructure from `share.js` for consistency

## Checklist

Yes

7/10

Yes

NaN

i used AI to check if a share-logic was already implemented

### DIFF
--- a/components/user-header.js
+++ b/components/user-header.js
@@ -21,6 +21,7 @@ import { numWithUnits } from '@/lib/format'
 import Badges from './badge'
 import SubscribeUserDropdownItem from './subscribeUser'
 import ActionDropdown from './action-dropdown'
+import Dropdown from 'react-bootstrap/Dropdown'
 import CodeIcon from '@/svgs/terminal-box-fill.svg'
 import MuteDropdownItem from './mute'
 import copy from 'clipboard-copy'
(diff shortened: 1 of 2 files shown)


Soxasora:
I personally don't like having two share buttons, have you considered getting inspiration from posts?
So something like this, but maybe you can find a better placement:
image

You can just re-use the `Share` component instead of repeating the logic this way.

components/user-header.js
```diff
@@ -245,6 +264,36 @@ function SocialLink ({ name, id }) {
   }
 }
 
+const referrurl = (ipath, me, isMobile = false) => {
```
Soxasora: This repeats what is already done in the `Share` component and `share` function, kewl name though

Soxasora (changes requested):
Meant to request changes, you can boil this down to a bunch of lines

pory-gone:
I understand! With this new implementation I reuse the share component and I eliminated the share profile button stand-alone, adding the button for the function in the drop-down menu. If you prefer something more similar to what you recommended, I'll get to work.
screenshot

Soxasora:
It's pretty good, I don't know how to feel about the icon in the dropdown but it's okay.
What about your own profile? This way the user can't share their profile (unless they grab the URL)

pory-gone:
okok i get the point, i'll try a different idea

pory-gone:
okok i get the point, i'll try a different idea

pory-gone:
I've updated the implementation. I was thinking of a mix of what you showed me and the one similar to the one on posts, I'm not sure about the icon yet.

- Removed the duplicated logic in `ShareProfileDropdownItem`
- Reused the existing `Share` component instead of repeating code
- Added a share button next to "edit nym" so users can share their own profile
- Kept the dropdown option for sharing other users' profiles

The share button now works consistently with how it's implemented for posts. Let me know if you want any other changes.

Istantanea_2025-10-23_19-27-17

Istantanea_2025-10-23_19-27-43

PR #2592, opened 2025-10-01
fix cancel on edit

### LINKED ISSUE #2568: Cancel on edit does not prompt for confirmation

pory-gone:
## Description

fix #2568 

Implemented confirmation prompt when canceling comment edit.

- Created reusable `ConfirmModal` component with presets for different confirmation types.
- Added confirmation modal logic to `comment.js eoggleEdit `function that checks for unsaved changes in `localStorage`.
- Updated both` comment.js` and `reply.js` to use the new reusable component.
- Added `storageKeyPrefix` to `comment-edit.js` to properly track form changes

## Screenshots

https://github.com/user-attachments/assets/6c2ceda3-10e7-4dc1-9608-bc8029538f51

## Additional Context

The implementation follows the exact same pattern used in `reply.js` to maintain consistency.
the confirmation logic is handled at the parent component level `comment.js` 
This implementation *DOES NOT* fix the multi-tab problem highlighted by @adlai, a broader refactor of the editing system would be needed, which goes beyond the scope of this issue.

## Checklist

Yes

7/10

NaN

No

I used AI to analyze the existing codebase and find interested files

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit c4b2bd637618ddf40491e985d6f156c69a958d68. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/comment-edit.js
+++ b/components/comment-edit.js
@@ -37,6 +37,7 @@ export default function CommentEdit ({ comment, editThreshold, onSuccess, onCanc
           }}
           schema={commentSchema}
           onSubmit={onSubmit}
+          storageKeyPrefix={`comment-edit-${comment.id}`}
         >
           <MarkdownInput
             name='text'
--- a/components/comment.js
+++ b/components/comment.js
@@ -28,6 +28,8 @@ import LinkToContext from './link-to-context'
 import Boost from './boost-button'
 import { gql, useApolloClient } from '@apollo/client'
 import classNames from 'classnames'
+import { useShowModal } from './modal'
+import { CancelWorkConfirm } from './confirm-modal'
 
 function Parent ({ item, rootText }) {
   const root = useRoot()
@@ -100,6 +102,7 @@ export default function Comment ({
   rootText, noComments, noReply, truncate, depth, pin, setDisableRetry, disableRetry,
   navigator
 }) {
+  const showModal = useShowModal()
   const [edit, setEdit] = useState()
   const { me } = useMe()
   const isHiddenFreebie = me?.privates?.satsFilter !== 0 && !item.mine && item.freebie && !item.freedFreebie
@@ -246,7 +249,26 @@ export default function Comment ({
                     </>
                   }
                   edit={edit}
-                  toggleEdit={e => { setEdit(!edit) }}
+                  toggleEdit={e => {
+                    if (edit) {
+                      const editText = window.localStorage.getItem(`comment-edit-${item.id}-text`)
+                      if (editText?.trim() && editText.trim() !== item.text.trim()) {
+                        showModal(onClose => (
+                          <CancelWorkConfirm
+                            onConfirm={() => {
+                              window.localStorage.removeItem(`comment-edit-${item.id}-text`)
+                              setEdit(false)
+                            }}
+                            onClose={onClose}
+                          />
+                        ))
+                      } else {
+                        setEdit(false)
+                      }
+                    } else {
+                      setEdit(true)
+                    }
+                  }}
                   editText={edit ? 'cancel' : 'edit'}
                 />}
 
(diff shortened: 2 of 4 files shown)


adlai:
Thank you for the shout-out; and was my own [limited human bafflement by the nightmare of modern client-side HTML5 codebases](../issues/2568#issuecomment-3341576356) helpful in any way, given that you also had an AI tool review the entire codebase?

components/confirm-modal.js
```diff
@@ -0,0 +1,50 @@
+import Button from 'react-bootstrap/Button'
+
+export default function ConfirmModal ({
```
Soxasora, line 3: I suppose this can be used directly instead of also creating `CancelWorkConfirm` and `DeleteConfirm` (which you didn't use but I get why you created it)

Maybe this can be called generically `ObstacleModal` but I'm just thinking out loud, there's no need to do this.

components/comment.js
```diff
@@ -246,7 +249,26 @@ export default function Comment ({
                     </>
                   }
                   edit={edit}
-                  toggleEdit={e => { setEdit(!edit) }}
+                  toggleEdit={e => {
+                    if (edit) {
+                      const editText = window.localStorage.getItem(`comment-edit-${item.id}-text`)
+                      if (editText?.trim() && editText.trim() !== item.text.trim()) {
```
Soxasora: The first trim (editText?.trim()) can be avoided, so that any change can count in the check.

Soxasora (approved):
Pretty cool! Left some nitpicks, but other than that it works correctly.

Also thank you for providing a centralized modal for confirmations

components/comment-edit.js
```diff
@@ -37,6 +37,7 @@ export default function CommentEdit ({ comment, editThreshold, onSuccess, onCanc
           }}
           schema={commentSchema}
           onSubmit={onSubmit}
+          storageKeyPrefix={`comment-edit-${comment.id}`}
```
ekzyis, line 40: is this somehow related to #2568?

pory-gone: This use local storage to save a draft of the user's edit, the draft is used to verify that there is actually text to be modified as `reply.js`. If you prefer not to use local storage we could open a modal when you want to close the edit.

Soxasora: Not really needed and kind of out of the scope of the PR but I appreciate the fact that it makes the draft experience consistent in all typing modes, and it's also the way textboxes are handled in other places.

So that's why I didn't comment on it. If useless, it can just be removed and the obstacle modal would be there only to guard the state setting

components/confirm-modal.js
ekzyis: > which you didn't use

@pory-gone Please don't add code that is dead on arrival

You're more than welcome to refactor something, but please not while trying to fix something and not in a way that it's half-done.

components/comment.js
ekzyis: Mhh, I agree with @Soxasora 

I don't think any user will care about losing an edit full of whitespace, but I think the inconsistency when the prompt shows up might confuse users. It might not be clear that it additionally checks if there's only whitespace.

components/comment-edit.js
ekzyis: > the draft is used to verify that there is actually text to be modified as `reply.js`

Ahh ok, that makes sense!

Sorry, I could have figured this out if I just looked at the other code. I tend to now just immediately ask if something is _really_ related if it's not immediately clear since I wasted too much time in my life trying to figure out _how_ something in someone else's code is related to a ticket or bug, just for it to end up not being related at all (and maybe even have a bug)

ekzyis (changes requested):
Relevant changes look good, but I did not test them yet.

Please remove the dead code, then I will take another look, thanks

pory-gone:
Thanks for your feedback! I removed the `DeleteConfirm` component from `confirm-modal.js` as it was unused and the first `.trim()` check in `comment.js` as suggested by @Soxasora - now any change (including whitespace) will trigger the confirmation

Issue #850, opened 2024-02-18
Prevent layout shift when images load


huumn:
When we run the `imgproxy` job we should also be [getting the image dimensions](https://docs.imgproxy.net/usage/getting_info) and storing them so we can set the image width/height on `img` tags and prevent layout shift.

ekzyis:
I have seen this still happen.

This also causes "scroll to comment" to have bad UX since a layout shift causes the comment to get scrolled out of view again. 

Reopening and will try to record a video when I see this again. Might only happen on Safari iOS.

PR #2599, opened 2025-10-11
enhance: improve media type recognition with HEAD or magic bytes

### LINKED ISSUE #1433: AVIF images are rendered as video

### LINKED ISSUE #2341: Attempt to render links as media triggers HTTP Basic Auth

### Describe the bug

When someone replies with a link that requires HTTP Basic Auth, our code to render it as media will trigger a login prompt, see screenshot.

### Screenshots

Image

### Steps To Reproduce

1. Paste link that requires HTTP Basic Auth into a comment
2. Get login prompt when we try to render link as media

### Expected behavior

- no login prompt for external sites ever

### Logs

_No response_

### Device information

_No response_

### Additional context

https://stacker.news/items/1056229?commentId=1056335

Soxasora:
## Description
fixes #2341
fixes #1433
can address #850 for images

We check if a link is a media file by downloading it fully, and then we download it again when we want to render it.
This PR improves media type recognition on links by fetching `HEAD` or, as a fallback, the first [magic bytes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types#other_methods_of_conveying_document_type) via [magic-bytes.js](https://github.com/LarsKoelpin/magic-bytes).

Client and imgproxy can use an endpoint placed in the **capture** micro service (avoids CORS) to know if they're dealing with an image or a video.

Also checks if a link has HTTP Basic Auth.

## Screenshots
Loading an `image/avif` file from the browser to check and render media vs checking the magic bytes
Proof of concept

https://github.com/user-attachments/assets/beb2de5b-6721-4e30-a38d-ce0ef6a1ede9

Loading a URL with CORS

https://github.com/user-attachments/assets/e63a86a7-81c6-46ba-a914-303e7bb3128d

## Additional Context
The endpoint lives in the capture micro service, because of this, the compose profile must have "capture".
Maybe we can add an extra fallback for when the capture instances go offline, if ever

---

We can get rid of the `HEAD` fetch if there's the possibility of false informations, it's just a cheap way to get `Content-Type`

---

> Can address #850 for images

The first magic bytes of an image can also contain informations about dimensions, and we could use it to avoid render jumps before imgproxy takes over. It's not implemented in this PR

---

This job could also be done in-house but we would deal with lots of magic numbers this way, so a popular and well-maintained library seemed a better idea.

---

This doesn't get rid of the heuristics involved in the imgproxy worker, it's still something that we know for sure that it works. But it's definitely redundant now.

## Checklist

Yes

7, pretty good actually

n/a

The following env vars have been introduced

```ini
MEDIA_CHECK_URL_DOCKER=http://capture:5678/media
-- url for imgproxy, communication between containers
NEXT_PUBLIC_MEDIA_CHECK_URL=http://localhost:5678/media
-- url for client-side fetches, e.g. media-or-link.js
```

The last one has been introduced in `.env.production` too but I don't think that file is even used

The readFirstBytes function is partially vibed, there were some things not really clear to me in that moment about the part of reading the small chunk with Reader, so it came in help.

---

> [!NOTE]
> Adds a /media endpoint in the capture service for HEAD/magic-bytes media detection and integrates it in the client and imgproxy worker, with new env/config wiring.
> 
> - **Capture service**:
>   - Add `GET /media/:url` route with CORS, using `HEAD` then magic bytes (`magic-bytes.js`) to detect `image/*` or `video/*` (`capture/index.js`, `capture/media-check.js`).
>   - New deps: `cors`, `magic-bytes.js` (`capture/package.json`).
> - **Worker (imgproxy)**:
>   - Use `MEDIA_CHECK_URL_*` to query capture media check before falling back to direct `HEAD`/`GET` (`worker/imgproxy.js`).
> - **Frontend**:
>   - Replace DOM-based media probing with fetch to `PUBLIC_MEDIA_CHECK_URL` in `components/media-or-link.js`.
> - **Config/Env**:
>   - Add `MEDIA_CHECK_URL_DOCKER` and `NEXT_PUBLIC_MEDIA_CHECK_URL` (.envs).
>   - Expose `process.env.NEXT_PUBLIC_MEDIA_CHECK_URL` via webpack DefinePlugin (`next.config.js`).
>   - Update docker compose: add `images` profile for `capture` service (`docker-compose.yml`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e4e2e6fd3cecea5eae4235717057448d31acef50. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/worker/imgproxy.js
+++ b/worker/imgproxy.js
@@ -32,7 +32,7 @@ const imageUrlMatchers = [
   u => u.host === 'i.imgur.com'
 ]
 const exclude = [
-  u => u.protocol !== 'https:',
+  u => process.env.NODE_ENV === 'production' && u.protocol !== 'https:',
   u => u.host.endsWith('.onion') || u.host.endsWith('.b32.ip') || u.host.endsWith('.loki'),
   u => ['twitter.com', 'x.com', 'nitter.it', 'nitter.at', 'xcancel.com'].some(h => h === u.host),
   u => u.host === 'stacker.news',
@@ -149,11 +149,13 @@ const isMediaURL = async (url, { forceFetch }) => {
 
   // primary: media check service
   try {
+    console.log('[imgproxy] media check service url:', `${MEDIA_CHECK_URL}/${encodeURIComponent(url)}`)
     const res = await fetch(`${MEDIA_CHECK_URL}/${encodeURIComponent(url)}`)
     if (res.ok) {
       const data = await res.json()
       isMedia = data.isImage || data.isVideo
       cache.set(url, isMedia)
+      console.log('[imgproxy] media check service response:', data)
       return isMedia
     }
   } catch (err) {
--- a/copilot/capture/manifest.yml
+++ b/copilot/capture/manifest.yml
@@ -45,6 +45,7 @@ variables:                    # Pass environment variables as key value pairs.
   CAPTURE_URL: https://stacker.news/
   MAX_PAGES: 10
   TIMEOUT: 3000
+  NEXT_PUBLIC_URL: https://stacker.news
 
 #secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
 #  GITHUB_TOKEN: GITHUB_TOKEN  # The key is the name of the environment variable, the value is the name of the SSM parameter.


components/media-or-link.js
```diff
@@ -130,29 +130,35 @@ export const useMediaHelper = ({ src, srcSet: srcSetIntital, topLevel, tab }) =>
...
-    video.src = src
+    checkMedia()
 
     return () => {
-      video.onloadedmetadata = null
-      video.onerror = null
-      video.src = ''
+      // abort the fetch
+      aborted = true
+      try { controller.abort() } catch {}
```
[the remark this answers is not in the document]
Soxasora: My mistake, we don't need to abort on imgproxy, and the way I implemented it was actually useless.
Client-side code still needs it to abort when the component unmounts.

worker/imgproxy.js
```diff
@@ -15,6 +14,7 @@ if (!imgProxyEnabled) {
 const IMGPROXY_URL = process.env.IMGPROXY_URL_DOCKER || process.env.NEXT_PUBLIC_IMGPROXY_URL
 const IMGPROXY_SALT = process.env.IMGPROXY_SALT
 const IMGPROXY_KEY = process.env.IMGPROXY_KEY
+const MEDIA_CHECK_URL = process.env.MEDIA_CHECK_URL_DOCKER || process.env.NEXT_PUBLIC_MEDIA_CHECK_URL
```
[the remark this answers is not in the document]
Soxasora: Well... yeah, everything would show as links.
I was wondering if a fallback to the traditional system might be acceptable considering the dislocation of the endpoint to another service.

capture/media-check.js
```diff
@@ -0,0 +1,101 @@
...
+      controller.abort(new Error(`timeout after ${timeout / 1000}s`))
+    }, timeout)
+  }
+
+  return controller.signal
+}
+
+function hasBasicAuth (res) {
+  const wwwAuth = res.headers.get('www-authenticate') || ''
+  if (res.status === 401 && /basic/i.test(wwwAuth)) return true
```
ekzyis: Shouldn't it bail on any auth error?

capture/media-check.js
```diff
@@ -0,0 +1,101 @@
...
+
+export default async function mediaCheck (req, res) {
+  const url = req.params.url
+  if (typeof url !== 'string' || !/^(https?:\/\/)/.test(url)) {
+    return res.status(400).json({ error: 'Invalid URL' })
+  }
+
+  try {
+    // trying with HEAD first, as it's the cheapest option
+    // TODO: should we trust it?
```
ekzyis: Outdated TODO?

If not: We can trust that false positives from `HEAD` are negligible, but false negatives are not.

So if HEAD says yes, it's media, we can trust it but not if it says it's not media. That might just be a default response and not according to the spec.

components/media-or-link.js
```diff
@@ -130,29 +130,35 @@ export const useMediaHelper = ({ src, srcSet: srcSetIntital, topLevel, tab }) =>
...
+    const controller = new AbortController()
+
+    const checkMedia = async () => {
+      try {
+        const res = await fetch(`${PUBLIC_MEDIA_CHECK_URL}/${encodeURIComponent(src)}`, { signal: controller.signal })
+        if (!res.ok) return
+
+        const data = await res.json()
+        // bail if the fetch was aborted
+        if (aborted) return
```
ekzyis: if the fetch was aborted, didn't `await fetch` throw?

Soxasora: Yeah we can specifically use `AbortError`. I placed the aborted flag out of precaution, there was this anxiety about state setters going off on unmount.

Pushing in a bit!

capture/media-check.js
Soxasora: We could probably just check 401/403 yes!

Soxasora:
Thanks ^^
Like suggested, I think we can also give it its own micro-service, which could also house many more tools in the future

components/media-or-link.js
```diff
@@ -130,29 +130,31 @@ export const useMediaHelper = ({ src, srcSet: srcSetIntital, topLevel, tab }) =>
...
     }
-    video.src = src
+    checkMedia()
 
     return () => {
-      video.onloadedmetadata = null
-      video.onerror = null
-      video.src = ''
+      // abort the fetch
+      try { controller.abort() } catch {}
```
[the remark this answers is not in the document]
Soxasora: I was indeed wondering if we should also retain the old system as fallback, but then it would also mean triple checks (HEAD, GET, client)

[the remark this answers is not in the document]
Soxasora: Well, since it's actually only a problem of service availability, then I suppose we could use the fallback only in such cases

huumn (approved):
This looks good. I need to redeploy `capture` before merging this.

.env.development
```diff
@@ -116,6 +116,11 @@ NEXT_PUBLIC_EXTRA_LONG_POLL_INTERVAL_MS=300000
 IMGPROXY_URL_DOCKER=http://imgproxy:8080
 MEDIA_URL_DOCKER=http://s3:4566/uploads
 
+# media check with capture container
+MEDIA_CHECK_ROUTE=media
```
ekzyis: Do we need this env var for something?

capture/media-check.js
```diff
@@ -0,0 +1,104 @@
...
+    return { bytes: buf, headers: res.headers }
+  } else {
+    const ab = await res.arrayBuffer()
+    const buf = new Uint8Array(ab.slice(0, byteLimit))
+    return { bytes: buf, headers: res.headers }
+  }
+}
+
+export default async function mediaCheck (req, res) {
+  let url = req.params.url
```
ekzyis, line 76: Not sure what the implications are of not doing it, but if you use `encodeURIComponent` on the client, shouldn't you use `decodeURIComponent` here?

components/media-or-link.js
```diff
@@ -130,29 +130,31 @@ export const useMediaHelper = ({ src, srcSet: srcSetIntital, topLevel, tab }) =>
...
-      img.decode().then(() => { // decoding beforehand to prevent wrong image cropping
-        setIsImage(true)
-      }).catch((e) => {
-        console.warn('Cannot decode image:', src, e)
-      })
+    const controller = new AbortController()
+
+    const checkMedia = async () => {
+      try {
+        const res = await fetch(`${PUBLIC_MEDIA_CHECK_URL}/${encodeURIComponent(src)}`, { signal: controller.signal })
```
ekzyis, line 137: ~Since `src` is a full URI, and not just a URI component, shouldn't this use `encodeURI`, not `encodeURIComponent`?~

Actually, I think `encodeURIComponent` is correct

capture/media-check.js
Soxasora: Ah yes much probably, without decoding, fetch would try to use special characters and fail. Thanks!

.env.development
Soxasora: Eh... not really. It's to have the possibility to change the route on the capture container, which needs to be redeployed anyway on changes.
I'll just hardcode it

capture/media-check.js
Soxasora: ohhh wait I forgot that Express [already decodes the parameters](https://expressjs.com/en/4x/api.html#req.params)! That's why I didn't decode there.

> Express automatically decodes the values in req.params (using decodeURIComponent).

README.md
```diff
@@ -144,7 +144,7 @@ COMPOSE_PROFILES=minimal
 To run with images and payments services:
 
 ```.env
-COMPOSE_PROFILES=images,payments
+COMPOSE_PROFILES=images,capture,payments
```
ekzyis: I think instead of saying that you need to also pass `capture` to test images (not obvious), I think the `capture` container should also start if `images` is passed.

This is what I meant in the meeting.

Soxasora: I agree!

capture/media-check.js
ekzyis: Can you add this as a comment here? This isn't obvious

ekzyis (changes requested):
This needs https://capture.stacker.news/ to return CORS headers to work, see discussion in chat

capture/media-check.js
```diff
@@ -0,0 +1,104 @@
...
+  if (typeof url !== 'string' || !/^(https?:\/\/)/.test(url)) {
+    return res.status(400).json({ error: 'Invalid URL' })
+  }
+
+  try {
+    // in development, the capture container can't reach the public media url,
+    // so we need to replace it with its docker equivalent, e.g. http://s3:4566/uploads
+    if (url.includes(process.env.NEXT_PUBLIC_MEDIA_URL) && process.env.NODE_ENV === 'development') {
+      url = url.replace(process.env.NEXT_PUBLIC_MEDIA_URL, process.env.MEDIA_URL_DOCKER)
+    }
```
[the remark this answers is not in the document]
Soxasora: I think everything would fail if NEXT_PUBLIC_MEDIA_URL is undefined.

capture/media-check.js
```diff
@@ -0,0 +1,105 @@
...
```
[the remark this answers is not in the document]
Soxasora: oh wait that's right lol
but then again... would someone p0wn itself in dev?

capture/index.js
```diff
@@ -51,10 +53,18 @@ const args = [
 let browser
 const app = express()
 
+app.use(cors({
+  origin: process.env.NEXT_PUBLIC_URL,
+  methods: ['GET', 'OPTIONS'],
+  credentials: false
+}))
+
```
huumn: Hmmm I think this applies to capture too?

Soxasora: Yeah, I don't think that it's gonna be a problem, for `NextSeo` in particular, but if everything worked before then maybe it's better to restrict CORS to the `media` endpoint

docker-compose.yml
```diff
@@ -87,7 +87,7 @@ services:
     restart: unless-stopped
     depends_on: *depends_on_app
     profiles:
-      - capture
+      - images
```
ekzyis, line 90: I didn't mean to remove the `capture` profile because it can still be useful to only test the `capture` container without any additional dependencies that are only needed for images.

> [...] I think the capture container should also start if images is passed.

-- https://github.com/stackernews/stacker.news/pull/2599#discussion_r2466889938

Soxasora: My mistake, sorry, didn't read correctly. It makes much more sense this way

ekzyis:
Okay, looks good to me now.

Please upload a video where you test this, so I can see that you tested it and how you tested it.

Then I will test this myself again.

Soxasora:
I uploaded a video where I test the endpoint by loading a post with a number of URLs (that are images), and a new post with just a single URL.
I also took a video of the capture container being loaded by the images profile, but I realized that I *captured* (tss) nothing worth being shown (docker loading)

PR #2622, opened 2025-10-30
remove lifpay wallet

### LINKED ISSUE #2613: Remove Lifpay wallet
https://stacker.news/items/1262093?commentId=1262391

https://www.nobsbitcoin.com/lifpay-temp-suspends-operations/ 

Stackers who already have a lightning address from Lifpay:

a) get their lightning address migrated to the wallet for custom lightning addresses

_or_

b) get their lightning address deleted

pory-gone:
## Description

fix #2613 

Lifpay Wallet Completely Removed
- Data Migration: Users with a Lifpay Lightning address are migrated to the "Custom Lightning Address" wallet (LN_ADDR).
- Cleanup: Lifpay wallets without a Lightning address are deleted.
- Removal: Lifpay is removed from the WalletTemplate and WalletName enums.

This ensures there are no new Lifpay attachments, no remaining Lifpay templates, and that existing users continue to receive via their Lightning address in LN_ADDR, if any.

## Additional Context

- The enum change required deleting/recreating the foregin key constraint between Wallet.templateName and WalletTemplate.name to avoid P3006 errors.
- The migration performs data updates before the enum change to avoid "invalid input value for enum" errors.

## Checklist

No, the WalletName enum loses a value (LIFPAY), which is not backwards compatible.

7/10

NaN

No

Yes, to draft the migration order.


ekzyis (commented):
> No, the WalletName enum loses a value (LIFPAY), which is not backwards compatible.

I agree that removing the enum value from `WalletName` will cause downtime.

Therefore, I think we can keep this simple and only migrate the lightning addresses and remove the wallet from the UI.

Then the migration should also be a lot simpler. I haven't tried it myself yet, but I expected something simpler.

ekzyis:
> The enum change required deleting/recreating the encryption key between Wallet.templateName and WalletTemplate.name to avoid P3006 errors.

encryption key?

pory-gone:
I agree with just doing the migration by removing lifpay only from the UI, it's definitely easier that way.

pory-gone:
> > The enum change required deleting/recreating the encryption key between Wallet.templateName and WalletTemplate.name to avoid P3006 errors.
> 
> encryption key?

Sorry, I was referring to the foreign key constraint

pory-gone:
In this commit I simplified the migration, kept lifpay in the database and otherwise followed the instructions given

ekzyis (changes requested):
do not create two migrations to do one thing

prisma/migrations/20251031121927_migrate_lifpay_to_lnaddr/migration.sql
```diff
@@ -0,0 +1,21 @@
...
+  FROM "Wallet" w
+  JOIN "WalletProtocol" p
+    ON p."walletId" = w.id
+   AND p."send" = false
+   AND p."name" = 'LN_ADDR'
+  JOIN "WalletRecvLightningAddress" la
+    ON la."protocolId" = p.id
+  WHERE w."templateName" = 'LIFPAY'
+    AND la."address" IS NOT NULL
+    AND btrim(la."address") <> ''
```
ekzyis, line 13: Is this needed for something?

ekzyis (commented):
Is this migration not enough?

```sql
UPDATE "Wallet"
SET "templateName" = 'LN_ADDR'
WHERE "templateName" = 'LIFPAY'
```

ekzyis:
> **Did you use AI for this? If so, how much did it assist you?**
>
> Yes, to draft the migration order.

What did you mean with "migration order"? Do you mean the migration itself?

What about the PR description? Did you write it yourself?

PR #2638, opened 2025-11-10
Hybrid query for related posts

### LINKED ISSUE #2513: use hybrid query for related posts
currently related posts only return semantic matches and could be improved by increasing relevance of semantic matches that also contain term matches (as we already do in normal search)

pory-gone:
## Description

fix #2513

Added term matching queries to the related posts feature to improve relevance scoring, now it combines semantic matching with keyword-based term matching, giving higher scores to posts with both

## Screenshots

https://github.com/user-attachments/assets/20aaa1e3-830a-4977-a8d8-2ea49ac26af2

## Additional Context

Everything works with or without the ML model. When no ML model is configured, it uses a boolean query combining `more_like_this` and term matching. When an ML model is available, all three methods are combined in a hybrid query.

## Checklist

Yes

8

NaN

NaN

Used AI for research on hybrid query

### DIFF
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -107,38 +107,151 @@ export default {
 
       let osQuery = moreLikeThisQuery
 
-      // Use hybrid query combining neural and more_like_this if model is available
-      if (process.env.OPENSEARCH_MODEL_ID) {
-        let qtitle = title
-        let qtext = title
-        if (id) {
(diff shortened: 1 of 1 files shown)


huumn:
This was already fixed here: 18638d886006e1cc6d1b617a9afc2c913210ed59

Looks like I forgot to close the issue.

pory-gone:
Oh that's great! Thanks anyway

PR #2652, opened 2025-11-21
fix_timer_bug

### LINKED ISSUE #2646: QR codes no longer show expiry
https://github.com/user-attachments/assets/a8fa88b6-b789-4150-aaf1-0133edbca2c4

Seems to be a general problem. Also happens for zap invoices.

pory-gone:
## Description

Fixed #2646 

QR code payment status timer was stuck at 00:00, the cause was a timezone interpretation bug. GraphQL Date scalar returns ISO 8601 strings without the 'Z' UTC suffix (e.g., "2025-11-21T12:31:08"), causing browsers to interpret them as local time instead of UTC. This made invoices appear expired immediately in non-UTC timezones. This fix adds 'Z' to `expiresAt` timestamps when missing. 

## Screenshots

https://github.com/user-attachments/assets/1e835967-5e3e-492e-a466-94976138ba6b

## Additional Context

Also fixed a shadow variable bug in `cancelAndReject` that prevented proper state updates. added polling with `FAST_POLL_INTERVAL_MS` to keep payment status updated.

## Checklist

Yes

8/10

nan

nan

I used AI to create a strategy to catch and understand the bug to apply the correct changes

### DIFF
--- a/components/payIn/hooks/use-qr-pay-in.js
+++ b/components/payIn/hooks/use-qr-pay-in.js
@@ -29,7 +29,7 @@ export default function useQrPayIn () {
       let updatedPayIn
       const cancelAndReject = async (onClose) => {
         if (!updatedPayIn && cancelOnClose) {
-          const updatedPayIn = await payInHelper.cancel(payIn, { userCancel: true })
+          updatedPayIn = await payInHelper.cancel(payIn, { userCancel: true })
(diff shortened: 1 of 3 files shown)


huumn:
Thank you for identifying the issue! For raw queries, we return timestamp fields as strings which are often stored without time zones, and the serialization of `Date` scalar would simply pass along the string. The solution, I think, is to always add a time zone on serialization.

I'll create an award for your work.

PR #2667, opened 2025-12-04
enhance: add dynamic import for stack trace decoding in error boundary.

brymut:
## Description

Addresses  #1832 

Reducing bundle size by importing `decodeMinifiedStackTrace` when copy error button is actually clicked. Rare user path so should have minimal impact. Noticed in a different project so figured I could bring it here and looks like it has significant bundle savings.

## Screenshots

**Before fix:**
Screenshot 2025-12-04 at 16 49 01
**After fix:**
Screenshot 2025-12-04 at 16 55 55

From 1.01MB -> 917kB, which is about 10%

For testing the functionality I added a window function to force an error in `ErrorBoundary` constructor method:
```
constructor (props) {
    super(props)

    // Define a state variable to track whether is an error or not
    this.state = {
      hasError: false,
      error: undefined,
      errorInfo: undefined
    }

    // DEV error trigger
    if (typeof window !== 'undefined' && process.env.NODE_ENV === 'development') {
      window.__triggerErrorBoundary = () => {
        this.setState({
          hasError: true,
          error: new Error('Test error triggered via window.__triggerErrorBoundary()')
        })
      }
    }
  }
 ```
 Vid:

https://github.com/user-attachments/assets/6dedfcf6-e8ba-499e-96c1-4f89008372b9

## Checklist

9

N

Y, coming up with a way to test that wasn't too hacky.

### DIFF
--- a/components/error-boundary.js
+++ b/components/error-boundary.js
@@ -1,4 +1,4 @@
-import { Component, useState } from 'react'
+import { Component } from 'react'
 import { StaticLayout } from './layout'
 import styles from '@/styles/error.module.css'
 import Image from 'react-bootstrap/Image'
@@ -61,10 +61,8 @@ export default ErrorBoundary
 // This button is a functional component so we can use `useToast` hook, which
 // can't be easily done in a class component that already consumes a context
 const CopyErrorButton = ({ errorDetails }) => {
-  const [copying, setCopying] = useState(false)
   const toaster = useToast()
   const onClick = async () => {
-    setCopying(true)
     try {
       const { decodeMinifiedStackTrace } = await import('@/lib/stacktrace')
       const decodedDetails = await decodeMinifiedStackTrace(errorDetails)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Soxasora** [comment]
This is an amazing find! It actually works great.

I’ll test the difference in bundle size in a fake-production environment and get back to you.

File: components/error-boundary.js [hunk 1]
```diff
@@ -61,16 +61,25 @@ export default ErrorBoundary
...
       console.error(err)
       toaster?.danger?.('failed to copy')
+    } finally {
+      setCopying(false)
     }
   }
-  return <Button className='mt-3' onClick={onClick}>copy error information</Button>
+  return (
+    <Button className='mt-3' onClick={onClick} disabled={copying}>
+      {copying ? 'copying...' : 'copy error information'}
```
Soxasora: while this is nice to look at, the toaster already does the job of giving visual feedback, so this is unnecessary.

brymut: Fair, removed.

brymut:
> I’ll test the difference in bundle size in a fake-production environment and get back to you.

Actually, would be cool to know how you do your fake prod environment if you don't mind with a quick overview of what you skip/include.

Soxasora:
In this particular case there was no need to run a full production environment. So I just used `@next/bundle-analyzer`, wrapped our `next.config.js` using `withBundleAnalyzer` and ran `npm run build`. You’d be automatically directed to the generated bundle overview.

brymut:
Ahh fair, considered using bundle analyzer too but wasn't too sure on what metric to use to measure the savings between the parsed/gzipped for the client. That's why I used the first load reporting next has for its builds.

PR #2674, opened 2025-12-13
adds AI summary tab

zanynik:
## Description

Adds AI summary tab. (LLMs are best known for good summarization & linking sources - I think it can be used to better guide discussion which sometimes can get lost in a huge nest of comments.)

## Screenshots
image

## Additional Context

The API Key is in .env.local , which is obviously not uploaded. So in production, a decision has to be made on which AI and API KEY will be used. Current version uses GEMINI API KEY.

## Checklist

YES

YES

YES

GEMINI_API_KEY

YES


huumn:
welcome to the repo. if you want a feature, the best thing to do is propose it and let us decide if we want it or not, then we'll welcome prs for it.

i don't want this feature nor do I want a dep on some cloud model.

PR #2682, opened 2025-12-19
Switch to Njump.to

4G0R4:
## Description

Njump to is a proxy that redirects to a random njump instance. More info at https://github.com/dergigi/njump-to

## Screenshots
image

## Additional Context
Sometimes njum.me is overflooded and does not load

## Checklist

Yes

10

N/A

No

No

### DIFF
--- a/components/embed.js
+++ b/components/embed.js
@@ -29,7 +29,7 @@ export const NostrEmbed = memo(function NostrEmbed ({ src, className, topLevel, 
     if (!iframeRef.current) return
 
     const setHeightFromIframe = (e) => {
-      if (e.origin !== 'https://njump.to' || !e?.data?.height || e.source !== iframeRef.current.contentWindow) return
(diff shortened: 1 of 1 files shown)


huumn:
I'm not sure embeds (iframes) will work with redirects. If it does work, I'll need to review njump.to to confirm I can trust the njump sources if we're giving them a CSP hallpass.

huumn:
... the embeds are janky. we shouldn't need them really and should be able to fetch nostr stuff ourselves ... it's just more code to maintain ... so if we shouldn't/can't do these embeds, it might be time to remove them

4G0R4:
True I did not evaluate the embeddings visualization; I simply replaced anything njump. Removing embeds (for nostr notes) could be a good move, as it forces stackers to fill the description input with some content to provide more value to the post.

PR #2683, opened 2025-12-19
Allow parsing item mentions from markdown links

Soxasora:
## Description

If a `LinkNode` has an item URL, it will get converted to item mention retaining the link text.
`ItemMentionNode` will have the default link theme in this case

## Screenshots

https://github.com/user-attachments/assets/95e4b2da-7e5b-4abf-bd3e-f093e681126b

## Additional Context

I wonder if we should stop using a different theme for mentions (bold) and instead use the default link theme.

## Checklist

8

---

> [!NOTE]
> Converts internal item/comment markdown links into `ItemMentionNode`s (preserving text) and exports them as links when custom text is used; adjusts priorities and theming.
> 
> - **Markdown/MDAST visitors**:
>   - Add `MdastItemMentionLinkVisitor` (priority `20`) to detect internal item/comment `link` nodes via `parseInternalLinks` and create `ItemMentionNode`s using extracted child text.
>   - Inline `MdastAutolinkVisitor` priority to `25` and remove PRIORITIES constant.
>   - Register `MdastItemMentionLinkVisitor` in `importVisitors` before regular `MdastLinkVisitor`.
> - **Lexical `ItemMentionNode`**:
>   - Default text to `#<id>` when none provided; export `isCustomText` and use it to:
>     - Render with link theme when custom text, else mention theme.
>     - Export to MDAST as a `link` with the custom text; otherwise export plain text URL.
>   - Support DOM import/export/clone unchanged aside from theming/text default.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6a61acd98b895c18b8c7ac51e6a2431fe0f2eafd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/lexical/mdast/visitors/mentions.js
+++ b/lib/lexical/mdast/visitors/mentions.js
@@ -66,6 +66,22 @@ export const LexicalTerritoryMentionVisitor = {
   }
 }
 
+// extract text by traversing the mdast node children
+function extractText (children) {
+  if (!children || children.length === 0) return ''
+
+  return children.map(child => {
+    if (child.type === 'text') {
+      return child.value
+    }
+
+    if (child.children) {
+      return extractText(child.children)
+    }
+    return ''
+  }).join('')
+}
+
 export const MdastItemMentionLinkVisitor = {
   testNode: 'link',
   priority: 20,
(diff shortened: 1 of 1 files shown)


huumn:
> I wonder if we should stop using a different theme for mentions (bold) and instead use the default link theme.

that's what we were doing before the release. i think removing the bold is good

Soxasora:
> that's what we were doing before the release. i think removing the bold is good

Yeah, I think I made them bold during early development to distinguish mentions from item/territory/user mentions (was lacking the prefix at the time)

Soxasora:
Mhh... do we want formatted item mentions? I can do them, but item mentions have to be restructured a bit, because `DecoratorNode`s won't allow children.

huumn:
We don't style item mentions usually

PR #2687, opened 2025-12-19
Update editor on quote reply

Soxasora:
## Description

The editor initializes itself with `appendValue` (quote reply) if available, but doesn't keep track of `appendValue` changes, breaking quote reply on full items that already have the reply box open.

This PR fixes this via `QuoteReplyPlugin` that simply reacts to `appendValue` changes and replaces text content.

## Additional Context

I didn't create an extension because `appendValue`, not being a stable prop, would have caused re-renders of the editor

## Checklist

8, actually really good, got to use `quote reply` more

---

> [!NOTE]
> Add QuoteReplyPlugin and plumb appendValue so the editor updates when the quote-reply content changes.
> 
> - **Editor**:
>   - Pass `appendValue` to `EditorContent` and initialize state with `appendValue || text.value`.
> - **Plugins**:
>   - Add `QuoteReplyPlugin` that reacts to `appendValue` changes and reinitializes the editor state.
>   - Wire `QuoteReplyPlugin` into the editor plugin stack.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5130c564a590d9e6431802c5704e5adfd65e9338. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/editor/plugins/core/formik.js
+++ b/components/editor/plugins/core/formik.js
@@ -1,8 +1,8 @@
 import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext'
 import { useEffect } from 'react'
 import { useField, useFormikContext } from 'formik'
-import { $isMarkdownEmpty } from '@/lib/lexical/utils'
-import { $getRoot, COMMAND_PRIORITY_HIGH, createCommand } from 'lexical'
+import { $isMarkdownEmpty, $getMarkdown } from '@/lib/lexical/utils'
+import { COMMAND_PRIORITY_HIGH, createCommand } from 'lexical'
 
 export const SUBMIT_FORMIK_COMMAND = createCommand('SUBMIT_FORMIK_COMMAND')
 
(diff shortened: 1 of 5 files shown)


Soxasora:
> Editor initial state now uses only existing Formik text.value (no longer initializes with appendValue).

I mean we probably *should* initialize with `appendValue` anyway if it's present, to avoid flashes from `text` (if present)

huumn:
First off, get some rest! You deserve it and this isn't super high priority.

I don't think this behavior is right. Every time I call `quote reply` it deletes anything I've already put in the text field. I think we want it to `append` the `Value`.

Soxasora:
I thought about `append`ing the `Value` too lol. But for some reason I also thought that it was meant to replace the whole editor.

I'll take some rest now, will switch to the native `insertText` tomorrow.

Soxasora:
I was thinking of implementing quote replies exactly like that. In absence of selection it will just append the content.

PR #2707, opened 2025-12-24
Prop key to update value every time button is clicked

### LINKED ISSUE #2703: Cannot click the same zap suggestion more than once

### Describe the bug

Zap suggestions cannot be clicked more than once, even if the zap amount changes and is not the same

### Screenshots

_No response_

### Steps To Reproduce

1. long press zap
2. select 100
3. change to something else
4. select 100
5. doesn't work

### Expected behavior

Selecting a zap suggestion should insert it regardless of it being previously selected

### Logs

_No response_

### Device information

_No response_

### Additional context

Reported by [0xbitcoiner on Stacker News](https://stacker.news/items/1370225)

pory-gone:
## Description
closes #2703 

Fixed forcing input remounting via a prop key that is updated every time the suggestion is clicked, so the value is always updated even when selecting the same option.

## Screenshots

https://github.com/user-attachments/assets/53e0b3db-be56-4ce0-8045-ec2e9298c6dc

## Checklist

Yes

8/10

Yes

NaN

AI was not used for fixing this bug

### DIFF
--- a/components/item-act.js
+++ b/components/item-act.js
@@ -27,7 +27,7 @@ const Tips = ({ setOValue }) => {
     <Button
       size='sm'
       key={num}
-      onClick={() => { setOValue({ value: num, ts: Date.now() }) }}
+      onClick={() => { setOValue(num) }}
     >
       <UpBolt
         className='me-1'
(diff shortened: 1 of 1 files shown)


huumn (changes requested):
This works!

It's a bit of a hack though, which is cool when we can't find an alternative, but there is an alternative here.

<details><summary>Spoiler Alert: the alternative</summary>

Define the `onChange` handler of `Input` and `setOValue(undefined)` when called.

This bug exists because `overrideValue` works as a dependency in a `useEffect` that overrides the form's input value. So when the same `overrideValue` is passed, the `useEffect` in `Input` does not run.

By unsetting `oValue` when the input value changes in response to user input, the `useEffect` will run when the `oValue` is set again.

</details>

pory-gone:
hello! First of all, Happy Christmas Eve to the SN team. I apologize for taking the "easy way out" with this little trick. I'll correct it as you explained as soon as possible.

huumn:
Merry Christmas Pory!

PR #2708, opened 2025-12-24
Fix inconsistent editor behavior with some Android IME stacks

### LINKED ISSUE #2691: Odd backspace behavior on Android 16 PWA since MDAST update

### Describe the bug

When removing characters on Android 16 using the built-in keyboard, the backspace button seems to get "stuck" at times, not removing characters repeatedly, both when just pressing it or when holding. I've also seen text re-appearing after removal. Attached screen capture shows all 3 symptoms.

### Screenshots

https://github.com/user-attachments/assets/146adf81-2c06-4e23-9158-051d283bec25

### Steps To Reproduce

Type something in the editor, press backspace to remove.

### Expected behavior

Backspace removes a single character at a time, no stalling

### Logs

_No response_

### Device information

- OS: Android 16 (GrapheneOS build 2025121701)
- Browser: Vanadium / Trichrome 143.0.7499.146.0
- Keyboard: built-in default keyboard, default settings

### Additional context

Also see: https://stacker.news/items/1357458

### LINKED ISSUE #2704: Keyboard Event Normalization System
Mobile keyboards, particularly those on Android, emit events very inconsistently with one another and make `contenteditable=true` inputs a total snake pit of bugs. Most `contenteditable=true` editors, including Lexical, dgaf or have crazy workarounds. We need a workaround of our own.

This is an excerpt from ChatGPT and I taking a walk around the issue:

> On Android, many soft keyboards effectively run everything through IME/composition-style editing (even for plain Latin typing), and the DOM event sequences you get can vary wildly across keyboards (Gboard vs Samsung Keyboard vs SwiftKey, voice input, swipe typing, “smart” correction layers, etc.). A lot of “expected” beforeinput → DOM mutation → input flows either change shape or don’t happen at all.

I'd guess The Mother Solution is to proxy all keyboard input through some kind of normalization layer where every input event is mapped to some kind of standard (perhaps emulating something like an iOS keyboard).

### LINKED ISSUE #2710: Selection is unreliable on alternative Android keyboards

### Describe the bug

When the caret is placed on a word that is targeted by autocorrect, selection gets glitchy.
This does not happen on Google Keyboard, and it's possible that this bug exists only in the same IMEs that #2708 targeted

### Screenshots

https://github.com/user-attachments/assets/fb8ce0da-af18-4dc6-83db-c59be51287d3

### Steps To Reproduce

1. write something
2. place the caret on a word
3. try to select another word with long press

### Expected behavior

That I can select another word

### Logs

_No response_

### Device information

- OS: Android 16
- Browser: Chrome PWA
- Keyboard: Yandex Keyboard

### Additional context

It seems like the selection is being restored to the word that the IME was targeting.
#2708 doesn't fix this

riccardobl:
Fixes #2691 
Fixes #2704
Fixes #2710

This PR targets what appears to be a single underlying cause behind editor inconsistencies with some Android IMEs. 

While the symptoms look different, my findings suggest that they likely come from race conditions between certain Android IME implementations and browser engines competing over the same contenteditable fields. 

In specific cases (bugged IME stacks + low latency/high performance environments), their events can overlap and their internal states interleave, causing a lot of breakage. 

Rather than working around the fallout, this PR  prevents the bad interaction by temporarily suppressing one of the two parties in the few key cases where the conflict occurs.

The patches are implemented as lexical plugins.

**Below are the details of each bug and its fix:**

# Letter ressurection when the text node is empty

Browser engines may inject placeholder content during editing to keep an element open and editable ([eg. chromium in delete_selection_command.cc around line 1281](https://chromium.googlesource.com/chromium/src/%2B/refs/heads/main/third_party/blink/renderer/core/editing/commands/delete_selection_command.cc#1281)). If this happens while an IME composition is still in-flight, the composition flow can get disrupted and the last character might resurrect.

To avoid this, the editor now injects its own placeholder instead of relying on the browser. This ensures the field never becomes truly empty (even though it appears empty to the user).

# Letters resurrection when deleting

This happens when the user presses and holds backspace. On some Android IME stacks, the words being deleted might become candidates for auto completion and trigger composition events that contain text that was just deleted.  This is fixed by suppressing all composition events for a short time window after certain events that should not trigger them.

# linebreak not deletable

It seems that for some IME stacks line breaks can become non-editable while typing. For Android WebView using the CSS property `user-select: all` seems to fix the problem, but this didn’t work with the most recent Chrome version, so instead I fixed this by switching the editor to use RichTextExtension instead of PlainTextExtension. 

This causes Lexical to represent lines as paragraphs `<p></p>` rather than relying on `<br>`

### DIFF
--- a/components/editor/plugins/patch/softkey-unborker.js
+++ b/components/editor/plugins/patch/softkey-unborker.js
@@ -32,14 +32,21 @@ function applySoftkeyWorkaround (el) {
   let compositionTimeout = null
   let selectionTimeout = null
 
-  const beginSelectionSuppression = (e) => {
+  const endSelectionSuppression = (e) => {
     if (!e) return
-    isSelectionSuppressed = true
+    isSelectionSuppressed = false
 
     if (selectionTimeout != null) {
       clearTimeout(selectionTimeout)
       selectionTimeout = null
     }
+  }
+
+  const beginSelectionSuppression = (e) => {
+    if (!e) return
+    endSelectionSuppression()
+
+    isSelectionSuppressed = true
 
     selectionTimeout = setTimeout(() => {
       isSelectionSuppressed = false
@@ -92,6 +99,7 @@ function applySoftkeyWorkaround (el) {
 
   document.addEventListener('selectionchange', beginSelectionSuppression, true)
   document.addEventListener('selectionchange', filterSelection, true)
+  el.addEventListener('cut', endSelectionSuppression, true)
 
   return () => { // cleanup
     el.removeEventListener('beforeinput', beginCompositionSuppression, true)
@@ -102,6 +110,7 @@ function applySoftkeyWorkaround (el) {
 
     document.removeEventListener('selectionchange', beginSelectionSuppression, true)
     document.removeEventListener('selectionchange', filterSelection, true)
+    el.removeEventListener('cut', endSelectionSuppression, true)
 
     if (compositionTimeout != null) {
       clearTimeout(compositionTimeout)


riccardobl:
# Notes for testing

Since this is (allegedly) a race condition, reproducing it reliably requires a fast, low-latency environment. That’s likely why a Google Pixel 9/10  with GrapheneOS and a keyboard that has this issue (like the AOSP keyboard) is a good testing ground.

But it is even easier to reproduce on a powerful workstation using the Android Studio emulator. The issue becomes non-reproducible vs reproducible just by changing performance settings of the virtual device:

## Non-reproducible (low-powered device)

* Device: `pixel_9a`
* API: 36
* Services: Android Open Source
* System Image: Intel x86_64 Atom System Image
* CPU Cores: 1
* Graphics acceleration: Hardware
* RAM: 2 GB
* VM heap size: 1 GB

## Reproducible (high-powered device)

* Device: `pixel_9a`
* API: 36
* Services: Android Open Source
* System Image: Intel x86_64 Atom System Image
* CPU Cores: **4** <<<
* Graphics acceleration: Hardware
* RAM: **8 GB** <<<
* VM heap size: 1 GB

Basically when emulating a low-powered device the issue often won’t show up, likely because the competing events won't be able to fire at the same time, while with higher CPU/RAM it reproduces consistently.

In both cases this was tested with:

* Browser: Chrome or the preinstalled Android WebView
* Keyboard: Android AOSP

(if the software keyboard doesn't appear in the emulator, just run `adb shell settings put secure show_ime_with_hard_keyboard 1` )

components/editor/plugins/patch/softkey-emptyguard.js
```diff
@@ -0,0 +1,104 @@
...
+ * by inserting placeholders. This might hijack the IME composition and cause unexpected behavior.
+ */
+
+// The character used as a guard. Ideally this would be non-printable, but I’ve found
+// that some Android keyboards disable auto-correct and other features when a
+// non-printable character is present in the text. A space seems to work best.
+// TODO: experiment with other characters if issues arise.
+const GUARD_CHARACTER = ' '
+
+const GUARD_REGEX = new RegExp(GUARD_CHARACTER, 'g')
```
riccardobl, line 19: this and related logic is used only when the GUARD_CHARACTER is not a space, left this in place so you can test out non printable characters if you want, but it seems some keyboards don't like to deal with non printable characters and they automatically turn off autocompletion

components/editor/plugins/patch/softkey-emptyguard.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  } else {
+    const children = el?.getChildren?.() || []
+    for (const child of children) {
+      ensureGuard(child)
+    }
+  }
+}
+
+function clearEditorIfEmpty (root) {
```
riccardobl: This is to make sure that if you empty the field, it is not just visually empty (ie. has only guard characters) but empty for real.
This keeps the placeholder behavior and also preserves any logic that depends on detecting an empty field.

huumn:
This required big insight and I'm thankful for your effort.

Was this left in draft intentionally or is it ready for review?

### edit 

I'd also be curious if you were able to recreate selection issues and this fixes those too. If not, we may need to extend composition suppression after other events too. ie We had reports of selecting some text and text elsewhere in the editor being selected.

riccardobl:
Thanks, yes it should be ready for review :+1: .


> I'd also be curious if you were able to recreate selection issues and this fixes those too. If not, we may need to extend composition suppression after other events too. ie We had reports of selecting some text and text elsewhere in the editor being selected.

I wasn't able to reproduce this, but it is possible that it was a side effect of accumulated divergences between what was seen by the IME, what was in the DOM and the internal representation in lexical, as the user was editing the post.

Soxasora:
> ([eg. chromium in delete_selection_command.cc around line 1281](https://chromium.googlesource.com/chromium/src/%2B/refs/heads/main/third_party/blink/renderer/core/editing/commands/delete_selection_command.cc#1281))

These are amazing discoveries that I personally wouldn't have found as fast as you did.

In our meetings we talked about switching to `RichTextExtension` as a measure to counteract line breaks, and I'm glad that you found it to be exactly what we need.

And, most importantly, everything is working great so far![^1]

I'll do some more QA tomorrow and leave a final review, thanks!

[^1]: but, due to the switch to `RichTextExtension`, plugins such as `formik` and `local-draft` need to be adapted to allow for a different kind of `\n` newlines management. rich text creates paragraphs instead of line breaks, which Lexical exports with `\n\n` if the paragraph is empty.

components/editor/plugins/patch/softkey-emptyguard.js
```diff
@@ -0,0 +1,114 @@
...
+    })
+
+    const disposeRootListener = editor.registerRootListener((root, prevRoot) => {
+      editor.update(() => {
+        if (root) {
+          ensureGuard(root)
+          clearEditorIfEmpty(root)
+        }
+      }, { tag: GUARD_TAG })
+    })
```
Soxasora: This doesn't actually do anything, because `registerRootListener` [returns the root DOM element](https://lexical.dev/docs/concepts/listeners#registerrootlistener) and `ensureGuard` and `clearEditorIfEmpty` actually expect a `LexicalNode` object. Nothing breaks because your code just ignores this and goes on.

components/editor/plugins/patch/softkey-emptyguard.js
```diff
@@ -0,0 +1,114 @@
...
+    const disposeListener = editor.registerUpdateListener(({ editorState, tags }) => {
+      if (tags.has(GUARD_TAG)) return
+      editor.update(() => {
+        const root = $getRoot()
+        if (root) {
+          ensureGuard(root)
+          clearEditorIfEmpty(root)
+        }
+      }, { tag: GUARD_TAG })
+    })
```
Soxasora: What is happening here is that you're creating an *update waterfall*:

- update
  - listener picks it up
    - update: `ensureGuard` + `clearEditorIfEmpty`
      - listener picks it up
        - `GUARD_TAG` prevents the loop that would instead happen

I think that what you actually needed here was a [`registerNodeTransform`](https://lexical.dev/docs/concepts/transforms) in order to react to `TextNode`s transformations (keystroke, delete, etc).

Now of course `ensureGuard` and `clearEditorIfEmpty` would still cause a loop, because you're transforming text nodes to have the `GUARD_CHARACTER`, but! you're checking if the `TextNode` ends with a ` ` whitespace, so the loop doesn't actually happen.

Another cool thing is that by directly targeting `TextNode`s (`registerNodeTransform(TextNode, () => {})`) you don't even need to use recursion to find one. You're not receiving ***every update happening inside Lexical*** anymore.

components/editor/plugins/patch/softkey-emptyguard.js
Soxasora: From QA, whitespace at the end and the way you reposition selection is actually non-intrusive (at all!).

components/editor/plugins/patch/softkey-unborker.js
```diff
@@ -0,0 +1,104 @@
...
+ * Workaround for what appears to be an unresolved race condition in some Android IME
+ * stacks: a composition event containing the word is emitted immediately after
+ * the deletion event that should have removed it. As a result, the word resurrects.
+ *
+ * This workaround suppresses all composition events for a short time window after we
+ * detect events that should not trigger them.
+ */
+
+// how long the composition events will be suppressed
+const suppressionWindow = 80 // ms (tweak as necessary)
```
Soxasora, line 14: I'm okay with what seems to be a magic number, this does the job.

I'm pretty sure that a `correct` timing for suppression is not even a thing, performance conditions can vary.

Soxasora (commented):
QA is great and I cannot fault this even though we're in `deep workaround` territory.
I'm sure Android users will rejoice!

I don't know if you ever worked with Lexical, but if you didn't this is actually really impressive.
The whitespace to avoid placeholders is honestly ingenious and fruit of your research within what-can-be obscure Chrome code.

I left some comments on how you could do things in a much more efficient way. Everything works correctly, but we can choose the right tools.

Great job! 🫡

---

Before merging we need to migrate to `RichTextExtension` ourselves, and prepare our core plugins to expect some differences in newlines being returned by getting the root text content.

riccardobl:
thanks for the review @Soxasora . I've applied the suggestions :+1:

Soxasora:
That was fast ⚡️
Let me QA.

Soxasora:
> I'd also be curious if you were able to recreate selection issues and this fixes those too. If not, we may need to extend composition suppression after other events too. ie We had reports of selecting some text and text elsewhere in the editor being selected.

I couldn't reproduce that bug in this PR (I'll try harder), but selection keeps being overall glitchy (#2710) and needs some more research 🤔

Soxasora (commented):
QAs the same, the writing experience finally feels normal. Let’s hope the whitespace won’t be annoying for users.
Chances are that no one will ever interact with it unwillingly because the selection manipulation feels really smooth!

I have a nit: I think we should limit these plugins to Android. We shouldn't hinder iOS experience because of Android (and vice versa).

riccardobl:
> I couldn't reproduce that bug in this PR (I'll try harder), but selection keeps being overall glitchy (#2710) and needs some more research 🤔

This seems like another instance of the same issue: when the selection changes a `selectionchange` event is fired, the IME sees this before its internal state is updated and so it tries to bring it back causing another  `selectionchange` to fire , but as it does that, its state update and so it breaks.

The PR should fix this too now, by just suppressing unnatural `selectionchange` that happens in short succession.

> I have a nit: I think we should limit these plugins to Android. We shouldn't hinder iOS experience because of Android (and vice versa).

Done :+1:  . There might be some privacy setups that fake the user agent though, but i am not sure how common they are  (I suppose that browsing the web pretending to use a different os causes issues elsewhere too...) but if there are users for which the patch doesn't activate, it might be worth reverting this change and leave the patch active for everyone or add a toggle in the settings.

components/editor/plugins/patch/softkey-unborker.js
```diff
@@ -63,17 +63,21 @@ function applySoftkeyWorkaround (el) {
   }
 
   el.addEventListener('beforeinput', beginSuppression, true)
+  document.addEventListener('selectionchange', beginSuppression, true)
```
Soxasora: Did you add the event listener to the `document` because the Lexical root DOM element doesn't get the `selectionchange` events?

Soxasora:
> if there are users for which the patch doesn't activate, it might be worth reverting this change and leave the patch active for everyone or add a toggle in the settings.

Or do checks based on things users can't spoof, this behavior is unique to some Android IMEs so we can target it for this purpose. But I don't think it's needed for now, let's wait for users' feedback.

components/editor/plugins/patch/softkey-unborker.js
riccardobl: yes, `selectionchange` normally fires only on document

riccardobl:
While doing some additional testing, I noticed some interference with the clipboard cut action, so I relaxed the filters, added a dedicated suppression flag for selection and disabled it on `clipboard cut`.

I also updated the empty-guard logic so it only triggers when the text node contains 1 or less characters.

These changes make the patch less aggressive and should reduce the risk of unintended side effects.

Issue #2589, opened 2025-09-29
Users are notified of mentions in code blocks


ekzyis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

[This comment](https://stacker.news/items/1240888) mentions a few users.

They are not highlighted as a mention because it's in a code block (which is good) but notifications were still sent out. This is the case because detecting if somebody was mentioned on the server is not context aware:

https://github.com/stackernews/stacker.news/blob/288892e28b4bc8f2c179d04ef63516e64489d60b/api/paidAction/lib/item.js#L5-L22

### Steps To Reproduce

1. Mention someone in a code block
2. Users are notified

### Expected behavior

Users are not notified

ekzyis:
This is probably easier to fix after #2501

adlai:
why is it a _problem_, again?

I wasn't using code blocks to avoid notifications, in case anyone thought that was my intent. Monospace looks better in some cases.

Soxasora:
> in case anyone thought that was my intent

We didn't think that, but mentions should not be parsed from code blocks

```java
@Entity
public class Zap
```

because code might contain @ signs and it wouldn't be great if something like this ^ would trigger a mention to user Entity

adlai:
ok, definitely good to avoid common pitfalls

could it be conditionally enabled with a suffix to the opening delimiter?

some markdown dialects allow suffixing the syntax to the opening triple backtick, and if there isn't any appropriate syntax already, then SN could designate some magic "quote" suffix that _would_ trigger notifications. obviously it should only happen if the person creating the comment has deliberately added that suffix, so maybe it should trigger some confirmation or extra cost on the item.

PR #2726, opened 2025-12-28
Don't detect mentions in code blocks

judemont:
## Description
fix: #2589
Don't detect mentions in code blocks and don't notify users.
## Checklist

Yes, they are.
8/10, I have manually tested 

N/A

No

I used AI for the regex string.

### DIFF
--- a/api/payIn/lib/item.js
+++ b/api/payIn/lib/item.js
@@ -30,8 +30,9 @@ export async function getItemResult (tx, { id }) {
 }
 
 export async function getMentions (tx, { text, userId }) {
+  const sanitizedText = text.replace(/```[\s\S]*?```/g, ' ')
   const mentionPattern = /\B@[\w_]+/gi
-  const names = text.match(mentionPattern)?.map(m => m.slice(1))
+  const names = sanitizedText.match(mentionPattern)?.map(m => m.slice(1))
   if (names?.length > 0) {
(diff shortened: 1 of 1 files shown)


huumn:
@Soxasora would it be possible to use some of the editor tooling to pull out mentions/images etc? This regex stuff is brittle and we already do this robustly in the editor.

Soxasora:
> would it be possible to use some of the editor tooling to pull out mentions/images etc?

Yes, we can [create a `lexicalState`](https://github.com/stackernews/stacker.news/blob/f4b4333977bdedf466793b85555a590338ae62de/lib/lexical/server/interpolator.js#L14) on the spot (like with item creation, we can bundle `lexicalState` with `item`) or pick the one we use for items via [DataLoader](https://github.com/stackernews/stacker.news/blob/f4b4333977bdedf466793b85555a590338ae62de/api/resolvers/item.js#L1440).
This opens a huge amount of possibilities, in this case we can either search nodes of type `userMentionNode` inside the `lexicalState` object or load this state on [an headless editor](https://github.com/stackernews/stacker.news/blob/f4b4333977bdedf466793b85555a590338ae62de/lib/lexical/server/headless.js#L47) and run lexical dollar functions such as `$getNodesByType()`. 

I need to continue the docs that I sadly de-prioritized for bugfixing, sorry about that!

PR #2766, opened 2026-01-24
Editor Stage 3: WYSIWYG

Soxasora:
## Description

Introduces `RichTextExtension` to our already established Lexical Markdown Editor.

## Additional Context

The implementation is relatively crude, switching from markdown to rich and viceversa re-renders the editor with the correct set of extensions, plugins and styling.

The cleanest move would be to have another instance of the editor (like we already doing with Preview) in order to avoid many conditionals and re-renders.

## Checklist

### DIFF
--- a/components/editor/contexts/toolbar.js
+++ b/components/editor/contexts/toolbar.js
@@ -1,7 +1,8 @@
 import { createContext, useContext, useMemo, useState, useCallback } from 'react'
 
 const INITIAL_STATE = {
-  previewMode: false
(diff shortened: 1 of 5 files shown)


huumn:
Woot! Out with the old

Soxasora:
Basically this prototype was based on swapping `MarkdownTextExtension` with `RichTextExtension`, re-rendering the editor when switching back and forth. I didn't really like this, so I figured might as well restart with a clean slate!

PR #2783, opened 2026-02-07
fix: compare wallet network with SN node on attach

### LINKED ISSUE #1028: Network not checked for LND autowithdrawals

While working on #994, I noticed that I could connect a mainnet node even though the SN node on my local machine was running on regtest. This would obviously fail on payments.

I tried to fix it in #994 by comparing the networks of both nodes but did not find a solution that I was satisfied with. I also didn't want to hardcode networks for dev and prod using `NODE_ENV` since I like to also test production mode locally sometimes.

Make sure that connected node is on same network as SN node. Ideally, don't depend on environment variables but actually compare the networks of both nodes.

A simple check that checks that the BOLT11 invoice prefix is `lnbc` for production and `lnbcrt` for development (note the overlapping prefixes during implementation).

**Additional context**

n/a

joelklabo:
## Summary

- When a user attaches a receive wallet (LND, CLN, LNbits, NWC, etc.), the test invoice's BOLT11 network prefix is now compared against the SN node's actual network
- Prevents silently connecting a mainnet wallet to a regtest/testnet SN instance (or vice versa), which would cause all payments to fail
- The SN node's network is auto-detected from a single test invoice and cached for the process lifetime — no hardcoded `NODE_ENV` checks

### How it works

1. Extract the network prefix from the BOLT11 invoice returned by the user's wallet (`lnbc` = mainnet, `lntb` = testnet, `lnbcrt` = regtest)
2. Create a small test invoice on the SN node's LND to determine its network (cached after first call)
3. If the networks differ, return a clear error: *"wallet is on mainnet but SN node is on regtest"*
4. If the SN node check fails for any reason, the wallet attach proceeds (graceful degradation)

Closes #1028

## Test plan

- [x] All CI checks pass (lint, shellcheck, unit tests, security)
- [x] BOLT11 prefix regex handles all network variants: `lnbcrt` (regtest), `lntbs`/`lntb` (testnet), `lnbc` (mainnet)
- [x] Network mismatch returns descriptive error message with both network names
- [x] SN node detection failure falls through gracefully (doesn't block wallet attach)
- [x] Network result is cached after first call — no repeated LND queries

### DIFF
--- a/wallets/server/resolvers/protocol.js
+++ b/wallets/server/resolvers/protocol.js
@@ -10,6 +10,34 @@ import { notifyNewStreak, notifyStreakLost } from '@/lib/webPush'
 import { decodeCursor, LIMIT, nextCursorEncoded } from '@/lib/cursor'
 import { walletLogger } from '@/wallets/server/logger'
 import { WalletValidationError } from '@/wallets/client/errors'
+import { createInvoice as lndCreateInvoice } from 'ln-service'
+
+/**
+ * Extract the network prefix from a BOLT11 invoice string.
+ * BOLT11 format: "ln" + network + amount + "1" + data
+ * Network prefixes: "bcrt" (regtest), "tb" (testnet), "bc" (mainnet)
+ * We match longest first to avoid "bc" matching "bcrt".
+ */
+function bolt11Network (bolt11) {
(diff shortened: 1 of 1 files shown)


joelklabo:

PR #2785, opened 2026-02-07
fix: don't convert Fediverse handles to mailto links

### LINKED ISSUE #93: Smarter URL detection, not every @ in the middle makes it an e-mail address

joelklabo:
## Summary

- Fixes #93 — Fediverse-style handles like `@Cointastical@BitcoinHackers.org` and `!group@server.org` are no longer incorrectly converted to `mailto:` links
- Adds a new `emailAutolinkTransform` to the markdown processing pipeline that detects and unwraps these false-positive email autolinks
- Regular email addresses (e.g. `user@example.com`) are unaffected

## How it works

GFM autolink detection aggressively matches `text@domain.tld` as an email address. When the original text is `@user@domain.tld`, GFM splits it into:
1. Text node: `"...@"` (the leading `@` gets consumed into preceding text)
2. Link node: `mailto:user@domain.tld`

The new transform checks if a `mailto:` link is immediately preceded by a text node ending with `@` or `!` (Fediverse mention/group prefixes). If so, it unwraps the link back to plain text and merges adjacent text nodes.

## Changes

- `lib/lexical/mdast/transforms/links.js` — new `emailAutolinkTransform` function
- `lib/lexical/mdast/transforms/index.js` — export the new transform
- `lib/lexical/utils/mdast.js` — add transform to the pipeline (runs before other link transforms)

## Test plan

- [x] All CI checks pass (lint, shellcheck, unit tests, security)
- [x] Transform only triggers when preceding text ends with `@` or `!` — no false positives on regular emails
- [x] Adjacent text nodes are properly merged after unwrapping
- [x] Transform runs before mention/nostr transforms to avoid interference
- [x] Existing email autolink behavior unchanged for standard addresses

### DIFF
--- a/lib/lexical/mdast/transforms/index.js
+++ b/lib/lexical/mdast/transforms/index.js
@@ -1,5 +1,5 @@
 export { mentionTransform } from './mentions.js'
 export { nostrTransform } from './nostr.js'
-export { misleadingLinkTransform, malformedLinkEncodingTransform } from './links.js'
+export { misleadingLinkTransform, malformedLinkEncodingTransform, emailAutolinkTransform } from './links.js'
 export { footnoteTransform } from './footnotes.js'
 export { tocTransform } from './toc.js'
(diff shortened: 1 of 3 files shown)


joelklabo:

PR #2788, opened 2026-02-07
Fix duplicate detection for redirected URLs

### LINKED ISSUE #254: Handle redirects for dup checks
Some people post Subtask newslettter articles using the open.substack.com URL format, while others post using the author subdomain URL scheme.  Either way, it's the same article.  The open.substack.com methods gets an HTTP 302 redirect to the author subdomain URL scheme.

Open scheme: https://open.substack.com/pub/[author]/p/[article]
Canonical (author subdomain scheme): https://[author].substack.com/p/[article]

Example:

https://open.substack.com/pub/inbitcoinwetrust/p/bitcoin-price-is-not-the-most-exciting
https://inbitcoinwetrust.substack.com/p/bitcoin-price-is-not-the-most-exciting

If the solution chosen is to simply rewrite the link (i.e., change the value that goes in the input field from the open.substack.com to the [author].substack.com scheme, then a dup check needs to be performed on that author subdomain scheme, so as to not introduce a new problem like the one reported for issue #244.

joelklabo:
## Summary

Closes #254

When users post URLs that redirect to a canonical form (e.g. `open.substack.com/pub/author/p/article` → `author.substack.com/p/article`, or `medium.com/@user/article` → `user.medium.com/article`), the duplicate checker now catches these as duplicates of the canonical URL.

## How it works

The `pageTitleAndUnshorted` resolver already follows redirects via `snFetch(..., { redirect: 'follow' })`. This PR captures `response.url` (the final URL after redirects) and returns it to the client as a new `redirected` field. The client then runs a duplicate check against this redirected URL, just like it already does for unshorted URLs.

## Changes

- `api/typeDefs/item.js`: Add `redirected` field to `TitleUnshorted` GraphQL type
- `api/resolvers/item.js`: Capture `response.url` after redirect, return when it differs from input
- `components/link-form.js`: Query `redirected` field and trigger dup check with it

## Test plan

- [x] All CI checks pass (lint, unit tests, shellcheck, security)
- [x] Only sets `redirected` when `response.url` differs from input URL — no redundant dup checks
- [x] Follows same pattern as existing `unshorted` dup check
- [x] No changes to the `dupes` query itself — same regex matching logic applies

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -649,6 +649,10 @@ export default {
       const res = {}
       try {
         const response = await snFetch(url, { protocol: 'http', redirect: 'follow' })
+        // capture the final URL after following redirects
+        if (response.url && response.url !== url) {
+          res.redirected = response.url
+        }
         const html = await response.text()
         const doc = domino.createWindow(html).document
         const titleRuleSet = {
--- a/api/typeDefs/item.js
+++ b/api/typeDefs/item.js
@@ -16,6 +16,7 @@ export default gql`
   type TitleUnshorted {
     title: String
     unshorted: String
+    redirected: String
   }
 
   extend type Mutation {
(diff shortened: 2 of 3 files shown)


joelklabo:

PR #2790, opened 2026-02-07
fix: pre-delete wallet logs to prevent transaction timeout on delete

### LINKED ISSUE #2629: Wallet delete can fail with 'Transaction already closed' error
When I try to delete a wallet, I get 'Transaction already closed' error:

Image

I think this happens more often if the wallet has been created a long time ago which makes me think it might be related to wallet logs.

If not clear, this is NOT a transaction timeout. This means we are for some reason trying to run `prisma.wallet.findMany()` on a transaction that has already been closed.

Missing `await` somewhere?

The delete does usually go through on second try.

joelklabo:
## Summary

Fixes #2629 — Wallet delete can fail with 'Transaction already closed' error.

When a wallet with many log entries is deleted, the CASCADE delete through `Wallet` → `WalletProtocol` → `WalletLog` can exceed Prisma's interactive transaction timeout (default 5s). This causes the transaction to close before `updateWalletBadges` can run `tx.wallet.findMany()`, producing the "Transaction already closed" error.

### Root cause

The `deleteWallet` function runs in a single Prisma interactive transaction:
1. `tx.wallet.delete()` — triggers CASCADE through protocols and logs
2. `updateWalletBadges()` — queries `tx.wallet.findMany()` on the same transaction

For wallets created long ago with extensive log history, step 1 can exceed the transaction timeout, causing step 2 to fail.

### Fix

Pre-delete wallet logs with `tx.walletLog.deleteMany()` before the wallet delete. This is a single efficient `DELETE ... WHERE` statement that clears the bulk of cascadeable rows, keeping the actual `wallet.delete()` cascade fast and the overall transaction within timeout bounds.

### Test plan

- [x] `standard` lint passes
- [x] `npm test` — all 18 tests pass
- [x] The `deleteMany` uses Prisma's relation filter (`protocol.walletId`) to target only logs belonging to the wallet being deleted
- [x] Wallet ownership is still verified by the `wallet.delete` call (`userId: me.id`)

### DIFF
--- a/wallets/server/resolvers/wallet.js
+++ b/wallets/server/resolvers/wallet.js
@@ -173,6 +173,15 @@ async function deleteWallet (parent, { id }, { me, models }) {
   if (!me) throw new GqlAuthenticationError()
 
   await models.$transaction(async tx => {
+    // delete wallet logs first to avoid cascade timeout on old wallets
+    // with many log entries (see #2629)
+    await tx.walletLog.deleteMany({
+      where: {
+        protocol: {
+          walletId: Number(id)
+        }
+      }
+    })
     await tx.wallet.delete({ where: { id: Number(id), userId: me.id } })
     await updateWalletBadges({ userId: me.id, tx })
   })


joelklabo:

PR #2795, opened 2026-02-07
fix: reserve placeholder space for unsized media to prevent layout shift

### LINKED ISSUE #2581: Layout shift while scrolling to comments?

### Describe the bug

Sometimes, when I click on a comment in my notifications, I am not scrolled to the comment, but somewhere else.

I think there's a layout shift while scrolling or after we scrolled.

I only noticed this in the iOS PWA, but it might just be more noticeable there because of the smaller width so there's more chance for vertical layout shift.

### Screenshots

_No response_

### Steps To Reproduce

1. ???
2. Scrolling works as expected reliably (afaict) when I go back and click on my notification again

### Expected behavior

Scrolling to comment always works

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

joelklabo:
## Summary

Fixes #2581

When unproxied images without known dimensions (e.g., autolinked images not yet processed by imgproxy) load inside comments, the `.sn-media` container collapses to zero height. When the image finally loads, the layout shifts and breaks scroll-to-comment navigation — the user lands at the wrong position.

### Root Cause

In `createDOM()` and `exportDOM()`, when `width`/`height` are not available, the CSS custom properties `--width` and `--height` are never set. This makes the container's `aspect-ratio: var(--width) / var(--height)` and `width: min(...)` declarations invalid, causing the container to collapse until the lazy-loaded image renders.

### Fix

- Added `sn-media--unsized` CSS class to media containers in `createDOM()` and `exportDOM()` when dimensions are unknown
- Added CSS rules for `.sn-media--unsized` that reserve a 200×150px (4:3 aspect ratio) placeholder — matching the existing `MediaLoading` component's dimensions
- This ensures the container occupies space before the image loads, significantly reducing layout shift magnitude
- The existing `preserveScroll` mechanism handles any remaining shift from the placeholder→actual size transition

### Changes

| File | Change |
|------|--------|
| `lib/lexical/nodes/content/media.jsx` | Add `sn-media--unsized` class when dimensions are unknown in `createDOM()` and `exportDOM()` |
| `styles/text.scss` | Add `.sn-media--unsized` CSS rule with placeholder dimensions |

## Test Plan

- [x] `npm run lint` passes (standard)
- [x] `npm test` passes (18/18 tests)
- [x] Verified that sized media containers (with `--width`/`--height`) are unaffected — they don't get the `--unsized` class
- [x] Verified that unsized media containers get `sn-media--unsized` class and reserve 200×150px placeholder space
- [x] The fix is backwards-compatible: existing content with saved Lexical state gets the class via `createDOM()`, and new SSR HTML gets it via `exportDOM()`

### DIFF
--- a/lib/lexical/nodes/content/media.jsx
+++ b/lib/lexical/nodes/content/media.jsx
@@ -169,8 +169,12 @@ export class MediaNode extends DecoratorNode {
     }
 
     const { width, height } = this.getWidthAndHeight() || {}
-    width && span.style.setProperty('--width', width)
-    height && span.style.setProperty('--height', height)
+    if (width && height) {
+      span.style.setProperty('--width', width)
+      span.style.setProperty('--height', height)
+    } else {
+      span.classList.add('sn-media--unsized')
+    }
 
     const loading = document.createElement('span')
     loading.className = 'sn-media__loading'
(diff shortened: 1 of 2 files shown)


joelklabo:

PR #2797, opened 2026-02-07
fix: restore trailing underscores in autolinked URLs

### LINKED ISSUE #180: Bug report: Underscores at the end of URLs
Underscores at the end of URLs are not rendering as part of Stacker News links.

This is one example, note the last underscore in the Twitter link in white text rather than blue text: https://stacker.news/items/48458

joelklabo:
## Summary

Fixes #180 — GFM autolink treats underscores as trailing punctuation and strips them from bare URLs, breaking links like `https://twitter.com/some_user_`.

### Root Cause

The `micromark-extension-gfm-autolink-literal` tokenizer explicitly lists `_` (code 95) in its trailing punctuation set. When it encounters a bare URL ending with `_`, it strips the trailing underscore(s) and they appear as plain text after the link node in the MDAST tree.

### Fix

Added a new mdast transform (`trailingUnderscoreAutolinkTransform`) that runs before the misleading-link check. It:

1. Walks the MDAST tree looking for `link` nodes (bare autolinks where text === URL)
2. Checks if the next sibling is a text node starting with `_`
3. Merges those trailing underscores back into the link's URL and text
4. Removes or truncates the sibling text node

Only bare autolinks are affected — explicit markdown links like `[text](url)` already preserve the full URL.

### Files Changed

- **`lib/lexical/mdast/transforms/autolink.js`** — New transform
- **`lib/lexical/mdast/transforms/index.js`** — Export the new transform
- **`lib/lexical/utils/mdast.js`** — Register the transform in the pipeline (runs before `misleadingLinkTransform`)

### Test Plan

- [x] Lint passes (`npx standard` — no errors)
- [x] Unit tests pass (`npx jest` — 18/18 green)
- [x] URLs like `https://twitter.com/some_user_` render as full clickable links including the trailing underscore
- [x] URLs with multiple trailing underscores (e.g. `https://example.com/test__`) are handled correctly
- [x] Non-autolink URLs in `[text](url)` format are unaffected
- [x] URLs ending with other trailing punctuation (`.`, `,`, `)`, etc.) are unaffected
- [x] URLs with underscores mid-path (e.g. `https://example.com/some_path/page`) are unaffected

### DIFF
--- a/lib/lexical/mdast/transforms/autolink.js
+++ b/lib/lexical/mdast/transforms/autolink.js
@@ -1,0 +1,45 @@
+import { visit } from 'unist-util-visit'
+
+/**
+ * GFM autolink treats underscores as trailing punctuation, stripping them
+ * from the end of bare URLs.  For example, `https://x.com/some_user_`
+ * becomes a link to `https://x.com/some_user` followed by a plain-text `_`.
+ *
+ * This transform detects that pattern and merges the stripped underscores
+ * back into the link so the URL renders correctly.
+ *
+ * Only bare autolinks (link text === URL) are affected; explicit markdown
+ * links like `[text](url)` already preserve the full URL.
+ *
+ * Fixes https://github.com/stackernews/stacker.news/issues/180
+ */
+export function trailingUnderscoreAutolinkTransform (tree) {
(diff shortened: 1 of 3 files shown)


joelklabo:

PR #2798, opened 2026-02-07
Add 'unsubscribe from thread' option for comments

### LINKED ISSUE #2578: Unsubscribe from parents / thread

Sometimes, users will get into a long back and forth conversation, under a comment that you previously posted. This causes you to get notifications for each of these replies, which clutters your notification inbox.

If the reply chain is very long and if the original post has many comments, it can be difficult to find the exact comment that you made that they've been replying under in order to unsubscribe.

Moreover, you might have multiple comments along the chain, and so stopping notifications for these replies would require finding all those items and unsubscribing from them.

I wonder if a button on the menu can be added that says "unsubscribe from thread", which will automatically unsubscribe you from any subscribed items that are parents of the current item.

To save menu space, perhaps it would only appear if there is a parent item to which you're subscribed.

_No response_

### Additional context

_No response_

joelklabo:
## Summary

Adds an **unsubscribe from thread** option to the comment action dropdown that removes all `ThreadSubscription` records for ancestor items in the comment chain. This lets users stop all reply notifications from a busy thread with a single click, without needing to find and unsubscribe from each parent item individually.

Closes #2578

## Changes

**Backend:**
- New `unsubscribeThread(id: ID!)` GraphQL mutation in `api/typeDefs/item.js`
- Resolver in `api/resolvers/item.js` uses ltree `@>` to find all ancestor subscriptions and delete them

**Frontend:**
- New `UnsubscribeThreadDropdownItem` component in `components/subscribe.js`
- Added to comment action dropdown in `components/item-info.js` (shown for comments with a `parentId`)
- Apollo cache update clears `meSubscription` on all ancestor items using dot-delimited path prefix matching

## How it works

The existing `subscribeItem` mutation deletes subscriptions for the target item and its **descendants** (using `<@`). The new `unsubscribeThread` mutation deletes subscriptions for all **ancestors** of the target item (using `@>`), effectively silencing the entire thread above the clicked comment.

## Test plan

- [x] CI passes (all 6 checks)
- [x] New "unsubscribe from thread" appears in comment dropdown for logged-in users
- [x] Clicking it removes all ancestor ThreadSubscription records
- [x] Subsequent replies to those ancestors no longer generate notifications
- [x] Apollo cache correctly clears `meSubscription` on ancestor items
- [x] Option only appears on comments (items with `parentId`), not root posts

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -850,6 +850,18 @@ export default {
       }
       return { id }
     },
+    unsubscribeThread: async (parent, { id }, { me, models }) => {
+      // Delete all ThreadSubscription records where the subscribed item
+      // is an ancestor of (or equal to) the given item
+      await models.$executeRaw`
+        DELETE FROM "ThreadSubscription" ts
+        USING "Item" i
+        WHERE ts."userId" = ${me.id}
+        AND ts."itemId" = i.id
+        AND i.path @> (SELECT path FROM "Item" WHERE id = ${Number(id)})
+      `
+      return { id }
+    },
     deleteItem: async (parent, { id }, { me, models }) => {
       const old = await models.item.findUnique({ where: { id: Number(id) } })
       if (Number(old.userId) !== Number(me?.id)) {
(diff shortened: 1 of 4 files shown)


joelklabo:

PR #2799, opened 2026-02-07
Restore PWA session URL after iOS app termination

### LINKED ISSUE #2760: iOS webapp not resuming where it was

### Describe the bug

If you have SN installed as an iOS web app (Add to Home Screen) and switch to another app, iOS may kill the SN web app after a while. When you return to SN from the recent app switcher, it reloads to the homepage instead of restoring the thread you were reading.

This happens to me often: I'll be reading something long, leave to do something else, and come back later only to find that I have to manually search for the post again.

### Screenshots

_No response_

### Steps To Reproduce

1. Click Add to Home Screen on SN in iOS
2. Open the SN shortcut from the home screen
3. Open a thread
4. Switch to another app and wait 15 minutes
5. Switch back to SN
6. You'll be on the homepage

### Expected behavior

Should resume from where it was suspended

### Logs

_No response_

### Device information

iOS. Might affect android too (i haven't tested)

### Additional context

_No response_

joelklabo:
## Summary

When iOS kills the SN PWA in the background and the user switches back, the app now **resumes at the last viewed URL** instead of reloading to the homepage.

Closes #2760

## Changes

**`pages/_app.js`:**
- Saves the current URL to `localStorage` (key: `pwa:lastUrl`) on every route change
- On cold start in PWA/standalone mode, redirects to the saved URL
- Only activates in standalone mode (`display-mode: standalone` or `navigator.standalone`) — no effect on normal browser tabs
- Uses the same PWA detection pattern as `components/pull-to-refresh.js`

## How it works

1. On every navigation in PWA mode, `pwa:lastUrl` is updated in localStorage
2. When iOS terminates the PWA and the user opens it again, the app mounts fresh at `/`
3. The mount effect detects PWA mode, reads the saved URL, and calls `router.replace(lastUrl)`
4. The user lands on the page they were reading instead of the homepage

## Test plan

- [x] CI passes
- [x] In iOS PWA mode: navigate to a thread, background the app for 15+ min, return → resumes at thread
- [x] In regular browser: no redirect behavior (localStorage key is never set)
- [x] If saved URL is `/` (homepage), no redirect occurs (avoids unnecessary replace)
- [x] Works on Android PWA mode as well (uses `display-mode: standalone` media query)

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -87,6 +87,34 @@ export default function MyApp ({ Component, pageProps: { ...props } }) {
     }
   }, [router.asPath, props?.apollo, shouldShowProgressBar])
 
+  // PWA session restore: save current URL so the app can resume after iOS kills it
+  useEffect(() => {
+    const isPWA = window.matchMedia('(display-mode: standalone)').matches || window.navigator.standalone === true
+    if (!isPWA) return
+
+    const saveUrl = (url) => {
+      try { window.localStorage.setItem('pwa:lastUrl', url || window.location.pathname + window.location.search) } catch {}
+    }
+
+    // Save on every route change
+    router.events.on('routeChangeComplete', saveUrl)
+    // Also save the initial URL
+    saveUrl(window.location.pathname + window.location.search)
+
+    return () => router.events.off('routeChangeComplete', saveUrl)
+  }, [])
+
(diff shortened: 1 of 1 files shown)


joelklabo:

PR #2801, opened 2026-02-07
fix: allow re-selecting the same zap suggestion amount

### LINKED ISSUE #2703: Cannot click the same zap suggestion more than once

joelklabo:
## Summary

Fixes #2703 — Zap suggestion buttons couldn't be clicked more than once with the same value. For example, selecting "100", manually changing to another amount, then selecting "100" again had no effect.

## Root cause

The `Tips` component calls `setOValue(num)` which sets React state to a primitive number. The `InputInner` component in `form.js` applies the override value via a `useEffect` with `[overrideValue]` in its dependency array. When the same number is set twice (e.g., `100` → `100`), React sees no state change and skips the effect entirely.

## Fix

- Changed `setOValue` in `Tips` to store `{ value, nonce }` objects, where `nonce` increments on every click, ensuring React always sees a new state reference
- Added `overrideNonce` prop to `InputInner` and included it in the `useEffect` dependency array
- `overrideNonce` is optional and backward-compatible — existing callers that don't pass it are unaffected

## Test plan

- [x] `npm run lint` passes
- [ ] Long-press zap → select 100 → manually change to 50 → select 100 again → input should update to 100
- [ ] Selecting different suggestion values still works normally
- [ ] Custom tip amounts work correctly
- [ ] Other `overrideValue` consumers (search filters, sub select, link form) still work

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -308,7 +308,7 @@ function FormGroup ({ className, label, children }) {
 }
 
 function InputInner ({
-  prepend, append, hint, warn, showValid, onChange, onBlur, overrideValue, appendValue,
+  prepend, append, hint, warn, showValid, onChange, onBlur, overrideValue, overrideNonce, appendValue,
   innerRef, noForm, clear, onKeyDown, inputGroupClassName, debounce: debounceTime, maxLength, hideError,
   AppendColumn, ...props
 }) {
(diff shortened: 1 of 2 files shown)


joelklabo:

Issue #2832, opened 2026-03-02
Editor improvements


Soxasora:
Groups all the new features and enhancements that should come across stages **3** and **4**.

huumn:
LISTS! TREES! WOOT!

PR #2839, opened 2026-03-02
Fix/mode aware copy paste + heading anchor bug

### LINKED ISSUE #2833: Mode-aware copy/paste
For a better copy/paste experience, we can start by transforming the copied content into the destination format, e.g. Lexical<->Markdown.

pory-gone:
## Description
fix #2833
Proposed implementation of mode-aware copy/paste that transforms content between Lexical and Markdown formats on paste. Also fixed a bug where heading anchor links rendered duplicate visible text in replies.

## Screenshots

https://github.com/user-attachments/assets/dec80746-404b-4f91-870b-fb771bac8b8c

https://github.com/user-attachments/assets/f3c3dbc7-1315-4f03-9487-a2fb9b95503b

## Additional Context

While testing the copy/paste feature, I discovered that headings in replies were rendering anchor link text visibly, causing duplicated text (e.g. "Markdown TestMarkdown Test")

## Checklist

yes

8/10

yes

no

the AI ​​helped me understand any side effects of the changes

### DIFF
--- a/lib/lexical/nodes/misc/heading.jsx
+++ b/lib/lexical/nodes/misc/heading.jsx
@@ -30,7 +30,7 @@ function createHeadingAnchorLink (domElement, headingId, textContent) {
   const link = doc.createElement('a')
   link.className = 'sn-heading__link'
   link.setAttribute('href', `#${headingId}`)
-  link.textContent = textContent
+  link.setAttribute('aria-label', textContent)
   return link
 }
 
@@ -43,7 +43,7 @@ function tryUpdateHeadingAnchorLink (domElement, headingId, textContent) {
   const first = domElement.firstChild
   if (first && first.nodeName === 'A' && first.classList?.contains('sn-heading__link')) {
     first.setAttribute('href', `#${headingId}`)
-    first.textContent = textContent
+    first.setAttribute('aria-label', textContent)
     return true
   }
   return false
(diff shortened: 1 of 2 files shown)


Soxasora:
Hey thanks for the thought! All the issues in #2832 are reserved to us because they must be implemented after Stage 3, which is in development at #2780. My bad, I should've assigned myself to those issues.

While this is almost what we're going to do, the new copy/paste mechanism should involve grabbing the Lexical payload on copy and converting it to Markdown on paste (checking the presence of the Lexical payload). This PR is already near that! But I can't review it, nor even move this forward because the specifications of how this should work are not available yet/can't be applied.

Thanks again for the help, because now I have a clearer way forward for copy/paste, hope we didn't waste your time!

lib/lexical/nodes/misc/heading.jsx
```diff
@@ -30,7 +30,7 @@ function createHeadingAnchorLink (domElement, headingId, textContent) {
   const link = doc.createElement('a')
   link.className = 'sn-heading__link'
   link.setAttribute('href', `#${headingId}`)
-  link.textContent = textContent
+  link.setAttribute('aria-label', textContent)
```
Soxasora, line 33: The problem with this is that the `<a>` link tag doesn't have anything to render and the heading won't be clickable, that's why we place a "duplicate" text content.
I know this caused issues with copy, and it's because in the DOM we have the same heading twice:
- heading: abc1
- link heading: abc1

resulting in `abc1abc1`.

Ours was a best-effort attempt to introduce heading anchors in Lexical, but they can be much better and I'm glad you tried to find a better way.

pory-gone:
Don't worry, it's my fault. I thought it was a specific issue, but I wanted to try. No, you didn't make me lose anything!

Issue #2842, opened 2026-03-05
respect mutes on live comments


huumn:
https://stacker.news/items/1448083

zhailong8845-art:
Implemented in #3135: muted users’ live comments remain collapsed and are no longer highlighted or added to the new-comment navigator. Focused Jest: 4/4 passed; Standard lint and diff check passed.

zhailong8845-art:
The tested fix is ready in #3135. Before merge, could an SN engineer assign the appropriate `difficulty:*` label so the documented awards workflow can calculate any contributor award? Please classify it based on the reviewed scope; I am not requesting a particular level.

awesomeguy2323:
Reproduced against current master and traced it to `newComments` (api/resolvers/item.js:679): it never applies `muteSql(me)`, so replies from muted users keep arriving through live polling and get injected/outlined as new.

Opened #3207 — one line, same server-side filter every other feed query already uses. Verified end-to-end in a local dev env: muted author's reply absent from the poll while the mute exists, reappears after unmuting, logged-out unaffected.

PR #2846, opened 2026-03-08
don't outline and track muted comments

Soxasora:
## Description

#2842 
Checks if the author of the new comment is not muted before outlining it (and tracking it via navigator).

## Screenshots

https://github.com/user-attachments/assets/f7932d7f-50de-4827-a28e-d1423368340d

## Additional Context

Maybe muted comments should appear at the bottom rather than the top!

## Checklist

10

---

> [!NOTE]
> **Low Risk**
> Small UI/UX logic change gated by a mute flag; minimal blast radius and no security or data-handling impact.
> 
> **Overview**
> Prevents muted users’ comments from being treated as “new” on item pages.
> 
> `Comment` now bails out of the outlining/`navigator.trackNewComment` effect when `item.user.meMute` is set, so muted comments won’t be highlighted or counted in the new-comment navigator flow.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d0098ab72c8c99036e69b366d530de0a8201f3d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/comment.js
+++ b/components/comment.js
@@ -162,6 +162,9 @@ export default function Comment ({
   }, [item.id, cache, router.query.commentId])
 
   useEffect(() => {
+    // don't outline muted comments
+    const isMuted = item.user?.meMute
+    if (isMuted) return
     // checking navigator because outlining should happen only on item pages
     if (!navigator || me?.id === item.user?.id) return
 


huumn:
did you investigate how hard it would be to do something similar for this indicator?

Screenshot 2026-03-08 at 11 12 53 AM

i'd guess it's hard, because the indicator here is driven by `lastCommentAt` which is global, and personalizing `lastCommentAt` is expensive.

without that, i think this PR's behavior is more confusing: i can see a top level indicator, but don't get live comment nav if it's from someone i muted.

to fix this all-the-way, we'll probably want live comments to work over a list of top-level items; over the homepage, for instance, we'd do a live comment query on all the posts. we'd have to nail down the performance and create a `liveCommentsForNItems` resolver, but i suspect that's the best thing we can do if we want to solve this bad boy.

Soxasora:
Currently that indicator is very, very light. It only reads `lastCommentAt` and does its calculations.

Batch-checking live comments on 21 items (homepage) can be expensive and as you said, we'd have to nail down the performance. I wonder if we can come up with a slim query that just gets `ncomments` minus the muted comments.

We'd probably need to rethink how we manage new comments in general.

sir-opti:
I've tested the in-item functionality and this worked as I expected. 

One minor question: when you say "tracking with the navigator", you mean that the blue dot at the top bar (when in item view) only tracks non-muted replies too? Is that a correct understanding? Because that's what I noticed - if yes, then I'll test that a bit deeper.

Soxasora:
Yes exactly, the new comments navigator tracks new ***outlined*** comments. By not outlining the new muted comments, we're also not tracking them.

sir-opti:
Tested that part with mutes, without mutes and all that looks fine.

I did notice that when a new comment is hidden behind pagination (note: unlikely with current parametrization `FULL_COMMENTS_THRESHOLD = 1000`), then I haven't been able to get it highlighted after pressing `more`.

PR #2850, opened 2026-03-10
don't be an idiot: build comment trees via a loop

huumn:
I have a bias toward recursion (fine), but I did it for building comment trees (not fine).

v1 of SN's tree building used recursion to assemble comment trees on the server (fine), and when I noticed it was slow, I moved the tree building, still with recursion and now kludgemaxxing, to plpgsql over temp tables for a (*drum roll*) 2x speedup (not fine). (IIRC v0 built the trees on the client.)

This PR does what a half-non-idiot would've done from the beginning:

1. use postgres for what it's good at: returning selective and flat sets of things
2. use javascript for what it's good at: moving around references to things in memory

This PR makes the comment tree resolver 5x faster on average. As a result, we can bump the paginated, depth-limited comments threshold to 1000 comments.

Reviewed/QA'd conceptually/structurally but not in detail.

P.S. For react's/rendering performance's sake, we probably shouldn't be nesting trees at all. The relationships should be kind of virtual, ie flat dom that looks nested via padding/indent.

---

> [!NOTE]
> **Medium Risk**
> Changes the core `Item.comments` resolver query strategy and tree construction, which can affect ordering/pagination correctness and comment visibility at scale. Also raises comment pagination thresholds, potentially increasing load and changing UX for large threads.
> 
> **Overview**
> Replaces the old comment retrieval path (multiple SQL functions + recursive/DB-side tree assembly) with a new `resolveItemComments` flow in `api/resolvers/comment-tree.js` that fetches a *flat* set of comment rows via SQL (full vs. limited recursive CTE based on `FULL_COMMENTS_THRESHOLD`) and builds the nested tree in JavaScript via an iterative `Map`-based pass.
> 
> Updates `Item.comments` in `api/resolvers/item.js` to delegate to this new resolver, preserving user-specific sorting behavior (including `commentsSatsFilter`) and cursor pagination, and adjusts tuning constants to allow larger threads before switching to the limited mode (`FULL_COMMENTS_THRESHOLD` 200→1000, `COMMENTS_OF_COMMENT_LIMIT` 2→5).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d1671f2fc9e6daa7e3d36c38a3b288a961a22af5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


api/resolvers/comment-tree.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  const decodedCursor = decodeCursor(cursor)
+  const comments = await fetchComments({ item, me, models, sortClause, decodedCursor, itemQueryWithMeta, payInJoinFilter, select })
+
+  return {
+    comments,
+    cursor: comments.length + decodedCursor.offset < item.nDirectComments
+      ? nextCursorEncoded(decodedCursor, COMMENTS_LIMIT)
+      : null
```
sir-opti, line 165: If you change `COMMENTS_LIMIT` into `comments.length` here, then that increases robustness

- if anything ever gets filtered out, the expected amount of items rendered is discoverable by decoding the cursor (client side too)
- also prevents skipping over `COMMENTS_LIMIT - comments.length` items if that happens

huumn: I'm going to do that in a follow up PR. We use the broken pattern everywhere and this is much better.

Also, sorry about accidentally fixing the bug I asked you look into. I didn't even understand the cause of the bug until you mentioned it.

GPT vibe-fixed that bug in the process of making this PR somehow. I should've warned you my fill-time-vibe would cross streams.

sir-opti: > We use the broken pattern everywhere

Maybe then the more useful strategy is to soft-assert `<n>.length === <n>_LIMIT` whenever `cursor !== null`. Can be done with or without changing cursor offset. Would be "less" user-friendly if not combined, but then there can be stack traces and if you track it (think `bugsnag` but without the 3rd party - tools like that can be vibed easily now) it wouldn't require a user to report bugs, but I'm then assuming that this scenario must always be a bug. Should validate that.

Either way, bug spotting instrumentation, when knowing bugs occur, and in 2026 where we can just (have a bot) do things, it may be interesting for constant quality improvement, especially when working under a ship-early philosophy.

> sorry about accidentally fixing the bug I asked you look into

No biggie, I'll share my analysis on the bug once I have fully confirmed the root cause, so that there is documentation of the pattern anyway. Also helps testing this PR. The `fixing` part is not yet confirmed (but I'll tell you in a few h)

huumn: I love the thought that bots have partially solved the quality vs speed dilemma. I hadn't been thinking that way until our interactions.

PR #2854, opened 2026-03-12
chore: vibe-vendor opentimestamps

sir-opti:
## Description

Attempts to replace the tiny subset of the `opentimestamps` npm package that are actually used by SN, while replacing 
all third-party dependencies and only using built-in functions. Within the subset used, the most notable package to be replace is `requests` (deprecated) by using `fetch`.

**Functionality retained:**

1. `PendingAttestation` - the non-upgraded attestation returned by an OTS Calendar remote
2. `SHA256`, `Prepend`, `Append` operations
3. `DetachedTimestamp` serialization and deserialization
4. The actual notarization process of stamping against a remote Calendar server

Everything else got stripped. No validation, stamp file upgrades, no cli, no shitcoins.

_Note that the integration into `worker/ots.js` has remained the same in terms of pattern, but the underlying package structure has changed, making it necessary to re-address the lines using classes and functions provided for ots._

**commits:**

- First commit introduces and integrates the new package
- Second commit removes the 3rd-party package

## Additional Context

Testing on dev can be done by:

- Creating a new item
- Installing the python client - `pip3 install opentimestamps-client`
- Running the following script with the item id after the item's timestamp has been applied (post edit window close)

```bash
#!/bin/bash
# Expected arg: item id of a timestamped item

curl -o ${1}.json http://localhost:3000/api/ots/preimage/${1}
curl -o ${1}.json.ots http://localhost:3000/api/ots/proof/${1}
ots --no-bitcoin verify ${1}.json.ots
```

## Checklist

Yes. 

Spent 3 days on continuous QA cycles, AI assisted (Claude 4.6 Opus, GPT 5.4 high, GLM 5). Total iterations performed: `25`.

Not here, to limit scope, but I do think that this should be done in a followup PR!

1. Some of the tests are vibed by reapplying tests from `opentimestamps/javascript-opentimestamps`.
2. The initial class layout was vibe-cloned and simplified from the upstream packages.
3. Unfortunately, the custom serialization format is not something LLMs can effectively replicate. After 2 days I just wrote it manually.
4. LLMs were heavily used in reviewing my meathook errors (25 rounds.) On 4 occasions have I dispatched Claude to fix a trivial test finding.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -86,7 +86,6 @@
         "nostr": "^0.2.8",
         "nostr-tools": "^2.8.0",
         "nprogress": "^0.2.0",
-        "opentimestamps": "^0.4.9",
         "page-metadata-parser": "^1.1.4",
         "pg-boss": "^9.0.3",
         "piexifjs": "^1.0.6",
@@ -7498,19 +7497,6 @@
         "url": "https://github.com/sponsors/ljharb"
       }
     },
-    "node_modules/asap": {
-      "version": "2.0.6",
-      "resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
-      "integrity": "sha512-BSHWgDSAiKs50o2Re8ppvp3seVHXSRM44cdSsT9FfNEUUZLOGWVCsiWaRPWM1Znn+mqZ1OfVZ3z3DWEzSp7hRA=="
-    },
-    "node_modules/asn1": {
-      "version": "0.2.6",
-      "resolved": "https://registry.npmjs.org/asn1/-/asn1-0.2.6.tgz",
-      "integrity": "sha512-ix/FxPn0MDjeyJ7i/yoHGFt/EX6LyNbxSEhPPXODPL+KB0VPk86UYfL0lMdy+KCnv+fmvIzySwaK5COwqVbWTQ==",
-      "dependencies": {
-        "safer-buffer": "~2.1.0"
-      }
-    },
     "node_modules/asn1.js": {
       "version": "5.4.1",
       "resolved": "https://registry.npmjs.org/asn1.js/-/asn1.js-5.4.1.tgz",
@@ -7520,14 +7506,6 @@
         "inherits": "^2.0.1",
         "minimalistic-assert": "^1.0.0",
         "safer-buffer": "^2.1.0"
-      }
-    },
-    "node_modules/assert-plus": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/assert-plus/-/assert-plus-1.0.0.tgz",
-      "integrity": "sha512-NfJ4UzBCcQGLDlQq7nHxH+tv3kyZ0hHQqF5BO6J7tNJeP5do1llPr8dZ8zHonfhAu0PHAdMkSo+8o0wxg9lZWw==",
-      "engines": {
-        "node": ">=0.8"
       }
     },
     "node_modules/async": {
@@ -7661,14 +7639,6 @@
       "integrity": "sha512-jOXGuXZAWdsTH7eZLtyXMqUb9EcWMGZNbL9YcGBJl4MH4nrxHmZJhEHvyLFrkxo+28uLb/NYRcStH48fnD0Vzw==",
       "bin": {
         "uuid": "dist/bin/uuid"
-      }
-    },
-    "node_modules/aws-sign2": {
-      "version": "0.7.0",
-      "resolved": "https://registry.npmjs.org/aws-sign2/-/aws-sign2-0.7.0.tgz",
-      "integrity": "sha512-08kcGqnYf/YmjoRhfxyu+CLxBjUtHLXLXX/vUfx9l2LYzG3c1m61nrpyFUZI6zeS+Li/wWMMidD9KgrqtGq3mA==",
-      "engines": {
-        "node": "*"
       }
     },
     "node_modules/aws4": {
@@ -7936,15 +7906,6 @@
         "zxing-wasm": "1.2.14"
       }
     },
-    "node_modules/base-x": {
-      "version": "3.0.11",
-      "resolved": "https://registry.npmjs.org/base-x/-/base-x-3.0.11.tgz",
-      "integrity": "sha512-xz7wQ8xDhdyP7tQxwdteLYeFfS68tSMNCZ/Y37WJ4bhGfKPpqEIlmIyueQHqOyoPhE6xNUqjzRr8ra0eF9VRvA==",
-      "license": "MIT",
-      "dependencies": {
-        "safe-buffer": "^5.0.1"
-      }
-    },
     "node_modules/base64-js": {
       "version": "1.5.1",
       "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.5.1.tgz",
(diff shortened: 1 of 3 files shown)


sir-opti:
I marked as ready after thinking this through while doing groceries (haha). Best I can come up with is some comment work, which I can enrich and rebase, but I don't see anything functional in the way of this being mergeable (unless I forgot something because this is still my 1st PR on SN, of course)

sir-opti:
Updated the inline docs in 1b5fc10. Kept it unsquashed for now, to not break any test efforts. Can be rebase+squashed into 34258b7 on demand.

huumn:
Started poking around on this tonight. Tax-preparations permitting I should be able to provide early feedback tomorrow (actually, today for me on Texas time).

huumn:
Taxes did not permit. I'm going to stop predicting, but this will be the next thing I review and will try to do it ASAP

sir-opti:
Addressed off-the-hub findings:

- dae6355b: There was a bug/oversimplification in `PendingAttestation` URI checks where the entire attestation payload was counted against the uri limit (which costs 2 bytes). I've reverted to using 3 separate maximum lengths, like done in the original dependency. Added EOF checking to make sure we don't get send extra bytes. Added tests.
- fb57a499: Fixed a naive implementation in `catSHA256` to use the safe `Timestamp.add()` instead of doing direct member manipulation.
- 7f3110b8: Added inline docs to address 2 design choices: hard errors when nothing gets stamped, and why there is no unit test against 3rd party servers.

lib/ots-mini/notary.js
```diff
@@ -0,0 +1,178 @@
...
+ * @param {Timestamp} right - The right leaf timestamp.
+ * @returns {Timestamp} The shared parent Timestamp.
+ * @throws {Error} If left or right is not an instance of Timestamp.
+ */
+export function catSHA256 (left, right) {
+  if (!(left instanceof Timestamp) || !(right instanceof Timestamp)) {
+    throw new Error('catSHA256: both left and right must be instances of Timestamp')
+  }
+  const sharedChild = left.add(new OpAppend(right.msg))
+  right.add(new OpPrepend(left.msg), sharedChild)
```
huumn: I'm still wading into the waters, and I'm pretty sure it's not relevant because we only stamp a single leaf at a time, but [`Timestamp.add()`](https://github.com/stackernews/stacker.news/pull/2854/changes#diff-f4d1a636b1e763eac29b8c6b0a9ca8f3c6f9a595c4853228be823bf404c80b6eR57) doesn't take a second argument which probably means we aren't capable of making merkle trees.

I'd be happy with either fixing that or make this `ots-micro` (lol) and throw when given more than one leaf.

[These checks](https://github.com/stackernews/stacker.news/pull/2854/changes#diff-cc5eaad5844aace481216eb09c42c2ac35ba38061bd39dad8b344a545db99c2dR239) test the leafs and not the tree.

huumn: The hot paths here look good to me otherwise. I'm going to poke at serialization and deserialization a bit more, but it looks good.

sir-opti: I thought that keeping multi-file `.stamp()` was a worthy goal, also because keeping the multi-file stamps I felt was something worth looking into as a performance optimization at scale (e.g. stamping batches of new items per territory) and not keeping it now would put future pressure on development.

The problem is indeed the tests as you noted, not just on the stamp but also on the `catSHA256` unit. I've extended the test and yes, the regression was made in fb57a499 by me yesterday, by replacing a naive implementation with a faulty one. Unfortunately that means there's no lazy solution here - I miscalculated that - but I'll fix this for all paths where we want to pass a shared tree with `add`. I'm weighing pros and cons of making shared subtrees explicit, so I've moved this back to draft for a bit.

I'll pick up the other items too and will just let go of the idea that re-vendoring like this should be cheap. Which is a bit of a pity because we've kind of already let go of the idea that this should be "vibe" too. The value proposition of "vibe-vendoring" may need some tuning.

huumn: > The value proposition of "vibe-vendoring" may need some tuning.

I'm not ready to let go just yet, but there's no need to push too hard and fast in this direction either.

sir-opti: `Concept.prototype.tune.toString() !== Concept.prototype.letGo.toString()` of course.

I'll just supply a report and then that can be either read or ignored.

sir-opti: This should now all be addressed.

sir-opti:
- Rebased on current `master`
- Addressed findings from https://github.com/stackernews/stacker.news/pull/2854#pullrequestreview-3972017140
- Reworked calendar query to be a lot more defensive

2 notes:

1. Post-review this really needs a squash, because 
2. if we put the `worker/ots.js` and npm package removal in a single commit at the top of the commit tree, then reverting back is ez

lib/ots-mini/notary.js
```diff
@@ -0,0 +1,163 @@
...
+        resLen += value.byteLength
+        if (resLen > MAX_RESPONSE_SIZE) {
+          abortController.abort(new Error(`Response from remote exceeded ${MAX_RESPONSE_SIZE} bytes`))
+          break
+        }
+        chunks.push(value)
+      }
+    }
+
+    if (!response.ok) throw new Error(`calendar ${this.url} returned ${response.status}`)
```
huumn: Because the `response.ok` check is here, `reader` on L51 could be `undefined` and throw on L56.

lib/ots-mini/tests/attestation.test.js
```diff
@@ -0,0 +1,156 @@
...
+
+// ---------------------------------------------------------------------------
+// PendingAttestation.deserialize validates UTF-8
+// ---------------------------------------------------------------------------
+describe('PendingAttestation.deserialize validates UTF-8', () => {
+  test('throws on invalid UTF-8 payload (0xFF byte)', () => {
+    const invalidUtf8 = Buffer.from([0x01, 0xff])
+    const ctx = new DeserializeContext(invalidUtf8)
+    expect(() => PendingAttestation.deserialize(ctx)).toThrow()
+  })
```
huumn, line 33: From my bot:

> Low lib/ots-mini/tests/attestation.test.js: the “invalid UTF-8” tests do not actually exercise PendingAttestation.decodeUri(); they fail earlier in varuint framing, so the suite passes without really pinning the UTF-8 decoder path it claims to cover.

> Those buffers hit readVaruint(..., includeSelf = true) first; they do not prove that malformed UTF-8 reaches and is rejected by decodeUri().

huumn (approved):
I found some nonblocking stuff IMO but I'm guessing you'd want to fix them anyway.

The highest related finding was an existing bug that there's no retry on `timestampItem` jobs so when it fails items don't get a timestamp. I'll open a separate issue for that.

I'm kind of bummed this is only a side effect that folks don't care about much - lots of great work here.

lib/ots-mini/notary.js
sir-opti: My bad for trying to be a cool kid and use `member?`.

lib/ots-mini/tests/attestation.test.js
sir-opti: fixed, and the one below it too. Also added explicit checks on throw messages to those tests still missing these, so that this (throw test passing on the wrong error) won't easily happen again.

About to squash. non-squashed commit is here: https://github.com/sir-opti/stacker.news/commit/fd471c476cb6e111e11154e5caafb0976182b490

lib/ots-mini/notary.js
sir-opti: Did 3 more rounds of hardening on this code... but **not** on this particular complaint (about `.ok` after read) because we **must** attempt to consume the body or risk `fetch` leaking. So only if we cannot get a `reader`, is it ok to not at least try to read once.

This patch is sitting pre-squash at https://github.com/sir-opti/stacker.news/commit/fe2328edb4e35968f68c9e88a8f520cd60685d59

sir-opti:
this is now rebased on master and squashed to 2 commits - one library implementation, one integration / removal of the dep

PR #2883, opened 2026-03-22
feat: retry ots jobs when they fail

### LINKED ISSUE #2882: timestampItem lacks retries on failure

sir-opti:
fixes #2882

## Description

Small extension to the OTS job scheduler.

1. Retry every 10 minutes for a total of 120 minutes
2. Do not backoff, as pgboss backoff jitter has t*2 scope, this would lead to issues with children being posted rapidly after their parents during calendar downtime
3. Logs retries explicitly
4. Be precise about what gets retried:
    - Missing parent, retry (low cost)
    - Errors during stringification, hashing or serialization, do not
      retry. This is fatal because the content won't change next run
    - Failures with remotes or posting to database are retryable

Note: If this turns out to be triggered often (I've yet to experience the scenario where all calendars are down) we can evolve this into a batch retry system and reduce the number of jobs.

## Checklist

Yes.

I performed the following:

1. Agent-assisted code review on all data paths
2. Testing with a chain of comments-upon-comments and hardcoded erroring calendars (added `{ calendars: [...<failing_endpoints>] }` to `Notary.stamp()` call
  - Check that after the parent failed, all children fail too (because they want parent hash) but are retried
  - Check that item creation order is preserved (no jitter is applied to timing)
  - Check that retries also happen in order (no jitter is applied to retry timing)
  - Check that when "magically" urls work again, jobs succeed
  - Check that when jobs succeed, they aren't retried

Only for finding the job post location under `api/` and code review

### DIFF
--- a/worker/index.js
+++ b/worker/index.js
@@ -85,6 +85,9 @@ async function work () {
   function jobWrapper (fn) {
     return async function (job) {
       console.log(`running ${job.name} with args`, job.data)
+      if (job.retrycount > 0) {
+        console.log(`  ... retry #${job.retrycount}/${job.retrylimit}`)
+      }
       try {
         await fn({ ...job, ...args })
       } catch (error) {
@@ -132,7 +135,7 @@ async function work () {
   await boss.work('payWeeklyPostBounty', jobWrapper(payWeeklyPostBounty))
   await boss.work('repin-*', jobWrapper(repin))
   await boss.work('trust', jobWrapper(trust))
-  await boss.work('timestampItem', jobWrapper(timestampItem))
+  await boss.work('timestampItem', { includeMetadata: true }, jobWrapper(timestampItem))
   await boss.work('earn', jobWrapper(earn))
   await boss.work('earnRefill', jobWrapper(earnRefill))
   await boss.work('streak', jobWrapper(computeStreaks))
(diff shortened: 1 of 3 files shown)


worker/index.js
```diff
@@ -85,6 +85,9 @@ async function work () {
   function jobWrapper (fn) {
     return async function (job) {
       console.log(`running ${job.name} with args`, job.data)
+      if (job.retrycount > 0) {
+        console.log(`  ... retry #${job.retrycount}/${job.retrylimit}`)
+      }
```
huumn, line 90: nice touch

sir-opti: except it didn't show in my test so I messed something up here. Have another fix pending too (because parents can now be retrying.) Still trivial though.

sir-opti:
Due to the hard extreme jitter range in pgboss' backoff scheme (`timeout + timeout * rand()`) I had to remove that, because it would cause children to retry-race their parent item at random, and potentially never being triggered after the parent was successfully stamped. Instead I opted for a retry every 10 minutes, at most 12 times. This means that stamping can survive all calendars concurrently being down for 2 hours (historically it wouldn't survive this at all.)

I've also made some decisions about what we retry and what we don't retry:

Retried:
 
- db reads & writes
- parent missing stamp
- remote calendar issues of any kind

Fatal:

- JSON serialization error of fetched data from db - because items won't change after edit window expiry, except for parents without stamp, which is caught separately 
- hashing of the json (unlikely)
- creation of timestamp file from hash

Last but not least: I've enabled retry tracking only for the `ots` job, simply because I don't know if this is desired information for any other job.

LMK if anything can be improved upon.

huumn:
> LMK if anything can be improved upon.

Nope, you already did more than I would've. Thanks!

PR #2897, opened 2026-03-31
custom domains: middleware and verification

Soxasora:
## Description

Based on #2904, updated to Next.js 16

Part of #1942, revives #2145.
Introduces custom domains via Next.js proxy. This PR focuses on domain creation and verification.

The proxy detects custom domains and applies redirects and rewrites to generate a territory-centric SN version. Territory paths lose the `~subName` prefix, and attempts to go outside of the domain's territory are corrected via redirect.[^1]

Verification is a multi-step repeating process that involves DNS verification, AWS certificates creation/verification and attachment to the AWS load balancer listener.
AWS is simulated with Localstack on sndev.

More infos are available inside `docs/dev/custom-domains.md` (todo: not finished)

## Screenshots

Screenshot 2026-04-08 at 17 49 36

Domain verification starts right when a custom domain is saved

Screenshot 2026-04-08 at 17 33 21

_sndev: DNS is simulated with **dnsmasq**_

Screenshot 2026-04-08 at 17 34 01

_sndev: AWS certificates (ACM) are simulated with Localstack and are automatically verified on the next call_

Screenshot 2026-04-08 at 17 37 53

custom domain is verified

Screenshot 2026-04-08 at 17 38 30

custom domain overview

Screenshot 2026-04-08 at 17 40 13

## Concerns and known issues

- s3 cors is not dynamic
- cannot signup or login from scratch
- polling time of domain verification (territory domain form) can be faster than 30 seconds
- `SN_TERRITORY_PATHS` is hardcoded and may drift
  - If new territory-scoped routes are added, this list must be manually updated.
- Redundant index on domainName
  - ***claude thought**
  - `@@index([domainName])`
  - There's already a unique constraint on `domainName` (line 918: @unique), which implicitly creates a B-tree index. The explicit `@@index([domainName])` is redundant.
- seo thumbnails not working for custom domains but we could open capture.stacker.news to other custom domains.
  - dynamic?
- there are 5 db triggers and probably we don't need that much obscure behavior
  - triggers like the deletion of domain certificates on territory `STOPPED` status can be handled via JS, but at the same time these triggers help us handle a custom domain lifecycle in the safest way possible.
  - needs to be checked.

## Checklist

Yes, everything is an addition. An exception is the way we handle S3 via localstack, now the container is called AWS and also allows ACM simulations other than S3.

7, QA OK, domain verification handles errors correctly and follows its step, middleware correctly detects a custom domain and obtains its connected sub.

Yes, it uses standard SN colors and styles that fits with theme changes.

An endpoint for localstack that can be contacted by our worker container
`LOCALSTACK_ENDPOINT=http://aws:4566`

local DNS server via dnsmasq, with fallback to 1.1.1.1 in both dnsmasq and code.
`DOMAINS_DNS_SERVER=172.20.0.2:53`

tells Next.js what domains (sox.pizza) or wildcards (**.pizza) can be trusted for HMR in dev.
`ALLOWED_DEV_ORIGINS=**.sndev`

custom domains debug logs are gated by env var, by default they're disabled (0)
`NEXT_PUBLIC_CUSTOM_DOMAINS_DEBUG=0`

**Did you use AI?**
review

---

> [!NOTE]
> **High Risk**
> High risk because it adds request-time middleware that rewrites/redirects based on `Host` and introduces background jobs plus DB triggers that manage domain/certificate lifecycle; mistakes could break routing or leave certificates/jobs in a bad state.
> 
> **Overview**
> Enables **beta** territory custom domains by adding a `Domain` data model, GraphQL `domain`/`setDomain` APIs, and a UI form that guides users through DNS + SSL verification with polling.
> 
> Adds a Next.js `proxy` middleware path that detects custom-domain `Host`s, injects `x-stacker-news-subname`, and rewrites/redirects territory-scoped routes so custom domains behave like territory-rooted sites.
> 
> Implements a multi-step verification pipeline via `pgboss` (`domainVerification`, certificate deletion, and long-held cleanup), including DNS checks (optionally via `DOMAINS_DNS_SERVER`), ACM certificate issuance/validation (Localstack in dev), mocked ELB listener attach/detach, and DB triggers to hold/cleanup domains and schedule maintenance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ecae6515b8276d0fba93cd5f839ee6ab9e59330c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

[^1]: I'm not sure if this should be the behavior, maybe it's better to redirect/open a new tab to stacker.news


pages/api/domains/index.js
```diff
@@ -0,0 +1,35 @@
+import prisma from '@/api/models'
+
+// API Endpoint for getting all VERIFIED custom domains, used by a cachedFetcher
```
sir-opti: I haven't been able to find a reason for this endpoint to be public. Am I correctly understanding it as a workaround due to middleware limitations? If yes, can we secure the endpoint, to just be internally callable?

proxy.js
```diff
@@ -133,6 +198,34 @@ export function middleware (request) {
   return resp
 }
 
+export async function middleware (request) {
+  const referrerResp = referrerMiddleware(request)
```
sir-opti, line 215: Nit: May want to sanitize headers (remove subname custom header) before anything else, to prevent potential future spoofing vulns. Right now, the header is double-checked on all paths, but it's easy to mess up.

Something like:

```javascript
const headers = new Headers(request.headers)
headers.delete('x-stacker-news-subname')
const sanitizedRequest = new Request(request, { headers })
```

lib/fetch.js
```diff
@@ -61,7 +60,10 @@ export async function snFetch (url, { path, protocol = 'https', timeout = 10000,
   }
 
   // Server-only: get proxy agent for Tor and custom certs (unless agent: false)
-  const fetchAgent = isServer && agent !== false ? getAgent({ hostname: urlObj.hostname, cert, protocol: urlObj.protocol }) : undefined
+  // TODO/ADDRESS BEFORE REVIEW: dynamic import to avoid pulling Node.js 'http' into the edge runtime via middleware
+  const fetchAgent = isServer && agent !== false
+    ? (await import('@/lib/proxy')).getAgent({ hostname: urlObj.hostname, cert, protocol: urlObj.protocol })
+    : undefined
```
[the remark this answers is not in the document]
Soxasora: oh right

lib/domain-verification.js
```diff
@@ -0,0 +1,134 @@
...
+export async function checkCertificateStatus (certificateArn) {
+  let certStatus
+  try {
+    certStatus = await getCertificateStatus(certificateArn)
+    return { certStatus, error: null }
+  } catch (error) {
+    console.error(`Certificate status check failed: ${error.message}`)
+    return { certStatus: 'FAILED', error }
+  }
+}
```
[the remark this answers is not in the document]
Soxasora: `FAILED` status is essentially dead, because if something happens during verification we just show `PENDING`.
I should get rid of it.

proxy.js
```diff
@@ -11,6 +12,59 @@ const SN_REFERRER = 'sn_referrer'
...
+  console.log('[domains] main domain', JSON.stringify(SN_MAIN_DOMAIN)) // TEST
+  console.log('[domains] pathname', pathname) // TEST
+  console.log('[domains] searchParams', JSON.stringify(searchParams)) // TEST
+  console.log('[domains] search', url.search)
+
+  // TODO: handle auth sync
+
+  // clean up the pathname from any subname
+  if (pathname.startsWith('/~')) {
+    const cleanPath = pathname.replace(/^\/~[^/]+/, '') || '/'
```
sir-opti, line 46: Question: Does this need a comparison with `subName`? Right now `customdomain.lol/~science` would show `~lol/lit`? But what if we got there from a cross-post badge?

proxy.js
```diff
@@ -133,6 +201,34 @@ export function middleware (request) {
...
+    const domainToMap = process.env.NODE_ENV === 'development' ? domain.split(':')[0] : domain
+    // check if we have a mapping for this domain
+    const subName = await getDomainMapping(domainToMap)
+    if (subName) {
+      console.log('[domains] allowed custom domain', domain, 'detected, pointing to', subName) // TEST
+      const resp = await customDomainMiddleware(request, domain, subName.subName)
+      // apply referrer cookies to the custom domain response
+      const referredResp = applyReferrerCookies(resp, referrerResp)
+      // finally apply security headers
+      return applySecurityHeaders(referredResp)
```
sir-opti, line 235: Does this work with just `self a.stacker.news` for things like script? (Hard to test on dev due to exclusions)

sir-opti (commented):
I did a review of this on concept completeness / robustness. Disclaimer: did get some bot help on the nitty gritty. Non-blocking notes on top of inline:

1. Canonical URLs (`<link rel="canonical" ... />` become important when 2 domains serve the exact same article or else multiple domains become a weakening force on discoverability / ranking. This has implications:
    - For single-territory posts, the custom domain should be the canonical url. However, since it's important that this is reachable, this needs a scheduled verification (or a light form of it) check (daily?) and only be set if verification passes, else the content is lost for SEO / ranking - important now that nearly every bot only checks top 5 search results for a given query.
    - For multi-territory posts, this should probably just be the main domain, otherwise arbitrary rules will allow different territories to "steal" the traffic to each other's co-content.
2. ~~Note that the "industry standard" for secure acme cert deployment is not `HTTP-01`, but `DNS-01` (would not cause extreme friction due to already requiring DNS records in other steps) as the former puts all eggs in the server's basket - Maybe be beneficial to describe this, also could get rid of port 80.~~ Missed that the only remaining place to do this is dev env, so doesn't matter.
3. It probably be wise to bump the `aws` sdk a major version if we're to use it for more than just s3 - I can propose this separately.

pages/api/domains/index.js
Soxasora: Yes it's a workaround because this version of Next.js has the Edge middleware instead of the Node.js one present in the latest versions. I've been floating around the idea of updating next.js before going forward with custom domains, but idk.

In the meantime we can either secure the endpoint or just leave it be for private beta reasons as it doesn't expose anything sensitive.

proxy.js
Soxasora: In [276b3db](https://github.com/stackernews/stacker.news/pull/2897/commits/276b3db7307657ab614b18a4a70cd97db4d6a27b) I made badges aware of custom domains. Clicking on a territory badge that doesn’t belong to the custom domain will open SN on a new tab.

I wanted to do the same for items present in the `related items` area, but I'm trying to avoid overdoing the private beta.

proxy.js
Soxasora: This is where things get tricky actually, iirc SN instances are behind a load balancer and CORS is set manually. @huumn can provide more info here maybe

Soxasora:
Hi @sir-opti! Thank you for checking this out.

I completely missed the `<link rel="canonical" ... />` thing. SEO altogether is not in a good state for custom domains in this PR.

Custom domains PRs are divided into:
- middleware, verification
- auth sync
- branding/personalization

The last PR will feature SEO work, I'll add a note to remind myself.

---

> It probably be wise to bump the aws sdk a major version

AWS SDK v2 has reached end of support on September 8th 2025, so yes we would need to update to v3 if we want to continue using it.

Also Next.js needs to be updated. It would bring the Node.js middleware opening up to a lot of possibilities, but most importantly it would help us get rid of the domains endpoint.

sir-opti:
> The last PR will feature SEO work, I'll add a note to remind myself.

Note though that in your proposed structure:

- middleware, verification `<--` The prerequisite recurring verification (or lite-verification) ought to go here?
- auth sync
- branding/personalization `<--` SEO goes here

I'm not sure how much a problem it is to just do the recurring verification as a sub-feature of SEO optimization - maybe there are more things than just the canonical url that could benefit from a robust periodical check - even if only early-warning that something is amiss during beta?

Soxasora:
We also have to remember that the auth sync architecture has not been finalized/re-conceptualized yet. I'm pretty sure we'll need a 5 minutes cronjob to check if the custom domain still points to us (for auth sync security reasons), so we can use that for `canonical` too!

Soxasora:
PR re-drafted to update to Next.js 16. Middleware is now a Node.js proxy that enables, among other things, direct db calls.

api/resolvers/domain.js
```diff
@@ -0,0 +1,157 @@
...
+      }
+    }
+  },
+  Domain: {
+    records: async (domain) => {
+      if (!domain.records?.length) return null
+
+      // O(1) lookups by type, simpler checks for CNAME and ACM validation records
+      return Object.fromEntries(domain.records.map(record => [record.type, record]))
+    }
```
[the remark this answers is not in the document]
Soxasora:
```js

export const DOMAIN_FIELDS = gql`
  fragment DomainFields on Domain {
    domainName
    status
    subName
  }
`
```

When we need `domain` from `Sub` we shouldn't ever need `records` or `certificates`. It's used to recognize a territory's custom domain and to get initial values (`territory-domains.js` form)

When we instead need `records` or `certificates` we'll use the `GET_DOMAIN` fragment.

lib/rss.js
```diff
@@ -38,23 +38,32 @@ const generateRssItem = (item) => {
...
     <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
       <channel>
         <title>${SITE_TITLE}${sub ? ` ~${sub}` : ''}</title>
-        <link>${SITE_URL}${sub ? `/~${sub}` : ''}</link>
+        <link>${siteUrl}${subPath}</link>
         <description>${SITE_SUBTITLE}</description>
         <language>en</language>
         <lastBuildDate>${new Date().toUTCString()}</lastBuildDate>
-        <atom:link href="${SITE_URL}/rss" rel="self" type="application/rss+xml" />
+        <atom:link href="${siteUrl}${subPath}/rss" rel="self" type="application/rss+xml" />
```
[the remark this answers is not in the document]
Soxasora: RSS will be handled in the SEO PR. Custom domains code present in SEO-related areas are proof of concepts that preserve Stacker News RSS original behavior

api/resolvers/domain.js
```diff
@@ -0,0 +1,159 @@
...
+        throw new GqlAuthorizationError('not an admin')
+      }
+
+      const sub = await models.sub.findUnique({ where: { name: subName } })
+      if (!sub) {
+        throw new GqlInputError('sub not found')
+      }
+
+      if (sub.userId !== Number(me.id)) {
+        throw new GqlAuthorizationError('you do not own this sub')
```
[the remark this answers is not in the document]
Soxasora: Yep

lib/domain-verification.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  if (!certificate || !certificate.Certificate || !certificate.Certificate.DomainValidationOptions) {
+    return { cname: null, value: null, error: { message: 'Certificate not found' } }
+  }
+
+  return {
+    cname: certificate.Certificate.DomainValidationOptions[0]?.ResourceRecord?.Name || null,
+    value: certificate.Certificate.DomainValidationOptions[0]?.ResourceRecord?.Value || null
+  }
```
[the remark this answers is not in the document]
Soxasora: Missing CNAME and value is handled by `getACMValidationValues` in `domainVerification.js`
Maybe the problem here is that it's not clear enough/it's hard to follow the pipeline

proxy.js
```diff
@@ -134,6 +202,39 @@ export function proxy (request) {
...
+
+  const referrerResp = referrerMiddleware(request)
+  // TODO: check if we actually need this, and WHY
+  if (referrerResp.headers.get('Location')) {
+    // this is a redirect, apply security headers
+    return applySecurityHeaders(referrerResp)
+  }
+
+  // if we're on a custom domain, handle it
+  const domain = request.headers.get('x-forwarded-host') || request.headers.get('host')
```
huumn: Is there a reason why we are favoring `x-forwarded-host`?

I'll need to doublecheck the docs myself but none of the reverse proxies (alb or our nginx config) are setting/sanitizing `x-forwarded-host` afaik.

huumn (commented):
Still doing bot review before moving to QA/human review. A few things from my bots that I haven't verified myself yet:

- api/resolvers/domain.js allows resume-from-HOLD, but the HOLD trigger in prisma/migrations/20250504202129_custom_domains_base/migration.sql deletes all verification records, and worker/domainVerification.js cannot progress without a persisted CNAME record.
- lib/domain-verification.js uses record.includes(recordValue) for CNAME verification, which is a substring check instead of a hostname equality check.
- The sub-stop trigger in prisma/migrations/20250504202129_custom_domains_base/migration.sql sets status = 'HOLD' without refreshing updated_at, while worker/domainVerification.js deletes old HOLD rows based on updatedAt.
- api/resolvers/sub.js exposes Sub.domain without the auth checks used by the dedicated domain query, and fragments/subs.js pulls that field into shared SubFields, creating both a public metadata leak and an N+1 query pattern.

proxy.js
Soxasora: There's not a confirmed reason for this. My assumption was that our load balancer or nginx would carry `x-forwarded-host`, but since they don't we can get rid of this check and go straight to `host`.

link to the old discussion: https://github.com/stackernews/stacker.news/pull/1958#discussion_r2052001450

Soxasora:
> `api/resolvers/domain.js` allows resume-from-HOLD, but the `HOLD` trigger ... deletes all verification records, and `worker/domainVerification.js` cannot progress without a persisted `CNAME` record.

I've realized that it's better to **restart** the whole domain verification process instead of *resuming*. For example, ACM certificates expire after 72h pending validation, invalidating its DNS CNAME validation values we previously saved.
So `HOLD` now is a state that both pauses the domain verification **and** protects the domain name from being stolen by another territory[^1]

While doing this, I forgot about the `if (!resuming)` before clearing domain verification records. I'll push a fix, thank you for noticing!

> lib/domain-verification.js uses record.includes(recordValue) for CNAME verification, which is a substring check instead of a hostname equality check.

I'm sure I did this because I thought it was a great idea to do `includes` to not deal with dns records that have a `.` at the end. It's better to normalize and do equality check.

> The sub-stop trigger in prisma/migrations/20250504202129_custom_domains_base/migration.sql sets status = 'HOLD' without refreshing updated_at, while worker/domainVerification.js deletes old HOLD rows based on updatedAt.

Ah, sorry I got too comfortable with the `@updatedAt` designation in `schema.prisma` that I didn't match the behavior in triggers.

> api/resolvers/sub.js exposes Sub.domain without the auth checks used by the dedicated domain query, and fragments/subs.js pulls that field into shared SubFields, creating both a public metadata leak and an N+1 query pattern.

I made this when I wanted to provide a way for us to know that a territory has a custom domain, and apply special behavior on `stacker.news`. Like custom domain badges, links, etc.
It's okay to expose `Sub.domain`, it only contains `{ domainName, subName, status }`, and the N+1 pattern can be solved with a `DataLoader`.

**BUT** since we're just using it to populate initial values (in the territory form) that we **already** get via `useQuery`, we can get rid of it for now!

---

- [x] fix domain verification records creation when resuming from `HOLD`
- [x] DNS CNAME hostname equality check
- [x] update `updated_at` in triggers
- [x] remove useless/unused `Sub.domain`

[^1]: the mother of edge cases probably because who would want to steal a domain name when they don't own said domain? but anything can happen, and without `HOLD` some other territory could take the domain as hostage.

lib/domain-verification.js
```diff
@@ -0,0 +1,136 @@
...
+  let domainRecords = null
+
+  try {
+    if (type === 'CNAME') {
+      domainRecords = await resolver.resolveCname(recordName)
+      // remove trailing dot if any
+      const normalize = s => (s.endsWith('.') ? s.slice(0, -1) : s)
+      result.valid = domainRecords.some(record =>
+        normalize(record) === normalize(recordValue)
+      )
```
Soxasora: [Can we have multiple CNAMES for a single Name?](https://serverfault.com/questions/574072/can-we-have-multiple-cnames-for-a-single-name)

Basically no DNS will allow you to go against spec and specify two CNAMES with the same name, pointing at different hosts.
@sir-opti do you think that this can anyway happen with a custom DNS solution, and that we should protect ourselves from such cases? (e.g. there's a second CNAME that points to their own infrastructure, escaping our DNS checks to invalidate auth tokens)

sir-opti: > do you think that this can anyway happen with a custom DNS solution, and that we should protect ourselves from such cases?

The problem is that in the case of `bind` allowing this, it must be round-robin, and afaik a `CNAME` query will always only return one record or it would break semantic protocol constraints. I would just throw on `domainRecords.length > 1` and `< 1` (different errors for each case). Also helps you get rid of the loop because now you can directly address `domainRecords[0]`.

Soxasora: Perfect, thank you!!

proxy.js
```diff
@@ -12,6 +13,60 @@ const SN_REFERRER = 'sn_referrer'
...
+    logger.log('setting sub to', subName)
+    searchParams.set('sub', subName)
+    url.search = searchParams.toString()
+    logger.log('new searchParams', url.search)
+    logger.log('new url', url)
+    return NextResponse.redirect(url, { headers })
+  }
+
+  // if we're at the root or on some territory path, hide the subname by rewriting
+  if (pathname === '/' || SN_TERRITORY_PATHS.some(p => pathname.startsWith(p))) {
```
sir-opti: This can use some hardening I think, as for all cases except `/~` there could be a trailing slash.

Perhaps making `/~{sub}` a distinct check can help here - ie (concept only, don't just copy this cuz I just wrote it in 2 sec and it prolly won't even work):

```javascript
function specifiesSub (pathname) { 
  return pathname.match(/^\/~[A-Za-z0-9]+\//) !== null
}
```

Then the other paths can be suffixed with `/`.

lib/rss.js
```diff
@@ -38,23 +38,32 @@ const generateRssItem = (item) => {
...
         ${itemsList.join('')}
       </channel>
     </rss>
   `
 }
 
+function getSiteUrl (req) {
+  if (req.headers['x-stacker-news-subname']) {
+    const proto = req.headers['x-forwarded-proto'] || 'https'
```
sir-opti: I don't think that this is safe without protection in reverse proxies, i.e. `curl -H 'x-forwarded-proto: javascript' ...`

Maybe safe: 

```javascript
const proto = req.headers['x-forwarded-proto'] != 'https' ? 'http' : 'https'
```

but why is this needed? Is anything but `https` supported?

worker/domainVerification.js
```diff
@@ -0,0 +1,349 @@
...
+    }
+  } catch (error) {
+    console.error(`couldn't verify domain with ID ${domainId}: ${error.message}`)
+
+    // get the job details to get the retry count
+    const jobDetails = await boss.getJobById(jobId)
+    console.log(`job details: ${JSON.stringify(jobDetails)}`)
+
+    // if we couldn't verify the domain after 3 attempts, put it on hold if it exists and delete any related verification jobs
+    if (jobDetails?.retrycount >= 3) {
```
sir-opti: I think this is off-by-one? the third attempt will have `.retrycount === 2`

worker/domainVerification.js
```diff
@@ -0,0 +1,349 @@
...
+      console.log(`couldn't verify domain with ID ${domainId} for the third time, putting it on HOLD if it exists and deleting any related domain verification jobs`)
+      await models.domain.update({ where: { id: domainId }, data: { status: 'HOLD' } })
+
+      // delete any related domain verification jobs that may exist
+      await cleanDomainVerificationJobs(domainId, models)
```
sir-opti: These can throw and mask the original error - put them in their own try/catch and log any errors on the inner catch.

lib/domains.js
```diff
@@ -0,0 +1,56 @@
...
+      return acc
+    }, {})
+  } catch (error) {
+    console.error('[domains] error fetching domain mappings', error)
+    return null
+  }
+}, {
+  forceRefreshThreshold: 1000 * 60 * 5,
+  cacheExpiry: 1000 * 60 * 2,
+  debug: process.env.NEXT_PUBLIC_CUSTOM_DOMAINS_DEBUG,
```
sir-opti: needs `== 1`?

lib/validate.js
```diff
@@ -295,6 +295,15 @@ export function territoryTransferSchema ({ me, ...args }) {
   })
 }
 
+// Custom Domain validation schema
+export function customDomainSchema (args) {
+  return object({
+    domainName: string().matches(/^(?:[a-z0-9-]+\.){2,}[a-z]{2,}$/i, {
```
sir-opti, line 301: Nit: I think this allows leading/trailing/consecutive hyphens `-`?

api/resolvers/domain.js
```diff
@@ -0,0 +1,164 @@
+import { validateSchema, customDomainSchema } from '@/lib/validate'
+import { GqlAuthenticationError, GqlInputError, GqlAuthorizationError } from '@/lib/error'
+import {
+  DOMAIN_VERIFICATION_HOLD_AFTER_DAYS,
+  DOMAIN_VERIFICATION_INTERVAL_SECONDS,
+  DOMAIN_VERIFICATION_RETRY_DELAY_SECONDS,
+  DOMAIN_VERIFICATION_RETRY_LIMIT,
+  SN_ADMIN_IDS
```
sir-opti, line 8: Is it smart to reuse this?

sir-opti (commented):
Did some more code review, consider most of these nits.

lib/rss.js
Soxasora: not needed, a relic from the original PR. RSS should be handled in the SEO stage, I'll revert all the changes related to that.

worker/domainVerification.js
Soxasora: `retrycount` is bumped by pgboss on thrown errors, my bad, the comment should say 3 retries, not 3 attempts.

lib/domains.js
Soxasora: Yes, needs `CUSTOM_DOMAINS_DEBUG`!

api/resolvers/domain.js
Soxasora: The first domain verification job is exactly like the subsequent jobs spawned by the first one (if we don't consider the debouncing logic). Probably the only thing that we could change there is `DOMAIN_VERIFICATION_INTERVAL_SECONDS` and make it shorter for the first verification.

The idea was to wait at least 30 seconds before starting domain verification, to give the territory owner the time to save the DNS records, the same for subsequent jobs. Before this centralization, all the timings were scattered/repeated across resolver and worker and it was becoming difficult to keep up with.

edit: what do you think?

sir-opti: Sorry - this rendered in a confusing way because my comment was on exactly line 8 - my bad.

What I meant was reusing `SN_ADMIN_IDS`, as this is also used in item and admin resolvers for other purposes.

lib/validate.js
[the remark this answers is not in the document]
sir-opti: Duplicate of: https://github.com/stackernews/stacker.news/pull/2897#discussion_r3099257817

api/resolvers/domain.js
Soxasora: Why not, at the end of the day this is just a simple measure to restrict custom domains to admins ^^

lib/validate.js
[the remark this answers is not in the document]
Soxasora: Built a stronger regex, should be good now.

.env.development
```diff
@@ -206,6 +209,14 @@ NEXT_TELEMETRY_DISABLED=1
 # local DNS server for custom domain verification, by default it's dnsmasq.
 # reachable by containers on 172.30.0.2(:53), outside of docker with 0.0.0.0:5353
 DOMAINS_DNS_SERVER=172.30.0.2
+# custom domains debugging
+NEXT_PUBLIC_CUSTOM_DOMAINS_DEBUG=1
```
huumn, line 213: Do we need to add this to the service worker?

huumn (approved):
This all looks sane and well organized to me. Hard to tell exactly how it'll interact with all the reverse proxies, but that's what the admin gating is for.

Not blocking, but I had a bit of trouble navigating the state machine. 

IMO there a few areas where my confidence is not super high:
1. state machine related issues
2. interactions with aws (out of our control mostly until we're in prod)
3. header handling in the proxy (you made it dead simple but I need to walk through it another time or two)

I'll continue reviewing tonight to gain more confidence in those areas, but with admin gating, this is good to go afaict.

You mentioned you had some minor fixes you wanted to make before merging so I'll wait for your green light.

proxy.js
huumn, line 247: Because we exclude `/api`, calls to `/api/graphql` do not strip `x-stackernews-subname`. This isn't a problem now because we don't read the header outside of getserversideprops, but if that changes we'll be reading a spoofable `x-stackernews-subname`. I'm not sure why that might change though.

It's at least worth commenting on. We could also go further running a little header stripping middleware when `source: '/((?api|_next/static|_error|404|500|offline|_next/image|_next/webpack-hmr|favicon.ico).*)'`.

proxy.js
```diff
@@ -12,6 +13,64 @@ const SN_REFERRER = 'sn_referrer'
...
+  logger.log('custom domain', domain, 'with subname', subName)
+  logger.log('main domain', JSON.stringify(SN_MAIN_DOMAIN))
+  logger.log('pathname', pathname)
+  logger.log('searchParams', JSON.stringify(searchParams))
+  logger.log('search', url.search)
+
+  // TODO: handle auth sync
+
+  // clean up the pathname from any subname
+  if (pathname.startsWith('/~')) {
```
huumn, line 45: `https://www.pizza.com/~bitcoin` is redirected to `https://www.pizza.com/bitcoin`. It's kind of low severity given it'd have to be manually input, but it might be worth redirecting to `https://stacker.news/~bitcoin` or 404ing. 

If we redirect `https://www.pizza.com/~bitcoin` to `https://stacker.news/~bitcoin` in the proxy, then we can probably avoid doing badge rewrites in components. 

Not blocking either. Just a suggestion.

huumn (commented):
found a few non-blocking things in the proxy

huumn:
I found these mermaid diagrams helpful for understanding how things progress through domain verification, aws calls, and the proxy.

```mermaid
stateDiagram-v2
  [*] --> PENDING: setDomain creates row
  PENDING --> ACTIVE: verifyDomain returns VERIFIED
  PENDING --> HOLD: 48h elapsed OR 3 AWS retries exhausted OR Sub STOPPED
  ACTIVE --> HOLD: Sub STOPPED trigger
  HOLD --> PENDING: setDomain re-verify on same domain
  HOLD --> [*]: clearLongHeldDomains after 30 days
  ACTIVE --> [*]: Sub owner changes
  HOLD --> [*]: Sub owner changes
```

```mermaid
sequenceDiagram
  participant W as worker
  participant A as ACM
  participant E as ELBv2
  W->>A: "requestCertificate(domain, IdempotencyToken=domainId)"
  A-->>W: certificateArn
  W->>A: describeCertificate(arn)
  A-->>W: "{ Status: PENDING_VALIDATION, ResourceRecord }"
  Note right of W: store arn and SSL record in DB
  loop every 30s to 5m while PENDING
    W->>A: describeCertificate(arn)
    A-->>W: Status
  end
  W->>E: "addListenerCertificates(listener, arn)"
  E-->>W: "ok (no-op if already attached)"
  Note over W,E: Domain ACTIVE
  Note over W: Later, domain deleted or HOLD fires cleanup
  W->>E: "removeListenerCertificates(listener, arn)"
  E-->>W: "ok (no-op if not attached)"
  W->>A: deleteCertificate(arn)
  A-->>W: "ok (errors if cert in use)"
```

```mermaid
flowchart TB
  Req[Incoming request] --> Strip[proxy.js strips x-stacker-news-subname]
  Strip --> Ref[referrerMiddleware]
  Ref --> HostCheck{host equals NEXT_PUBLIC_URL.host?}
  HostCheck -- yes --> Sec[applySecurityHeaders]
  HostCheck -- no --> Map{getDomainMapping returns ACTIVE row?}
  Map -- no --> Sec
  Map -- yes --> Custom[customDomainMiddleware]
  Custom --> Rewrite{path starts with slash-tilde?}
  Rewrite -- yes --> Redirect[redirect to cleaned path]
  Rewrite -- no --> SubParam{has ?sub and mismatches?}
  SubParam -- yes --> RedirectFix[redirect with sub coerced]
  SubParam -- no --> IsTerr{path is / or territory path?}
  IsTerr -- yes --> RW[rewrite to /tilde-sub/path and set header]
  IsTerr -- no --> Pass[NextResponse.next and set header]
  Redirect --> Cookies[applyReferrerCookies]
  RedirectFix --> Cookies
  RW --> Cookies
  Pass --> Cookies
  Cookies --> Sec
```

proxy.js
```diff
@@ -12,6 +13,64 @@ const SN_REFERRER = 'sn_referrer'
 const SN_REFERRER_NONCE = 'sn_referrer_nonce'
 // key for referred pages
 const SN_REFEREE_LANDING = 'sn_referee_landing'
+// main domain
+const SN_MAIN_DOMAIN = new URL(process.env.NEXT_PUBLIC_URL)
+// territory paths that needs to be rewritten to ~subname
+const SN_TERRITORY_PATHS = ['/new', '/top', '/post', '/edit', '/rss']
```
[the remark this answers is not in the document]
Soxasora: Search is postponed to the custom domains UI/UX PR which will focus on paths, custom domain <-> SN relationship and more.
This PR addresses custom domain access and verification.

lib/validate.js
```diff
@@ -295,6 +295,24 @@ export function territoryTransferSchema ({ me, ...args }) {
...
+  return object({
+    domainName: string()
+      .trim()
+      .max(253, 'domain must be at most 253 characters')
+      .matches(regex, {
+        message: 'CNAME records only support subdomains (e.g., www.example.com, sub.example.com), no leading/trailing dashes, each label 1-63 chars, at least two dots'
+      })
+      .nullable()
+  })
+}
```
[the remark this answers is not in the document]
Soxasora: Ah yes, we didn't have a `reset` button before so to delete a domain one would have to submit an empty domainName and press `save` button.

Now there's no `save` button.

api/resolvers/domain.js
```diff
@@ -0,0 +1,164 @@
...
+      if (domainName) {
+        domainName = domainName.trim().toLowerCase()
+        await validateSchema(customDomainSchema, { domainName })
+
+        const sameDomain = existing && existing.domainName === domainName
+
+        // updating the domain name, recovering from HOLD is allowed
+        if (sameDomain && existing.status !== 'HOLD') {
+          throw new GqlInputError('domain already set')
+        }
```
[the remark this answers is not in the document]
Soxasora:
```js
      if (sub.userId !== Number(me.id)) {
        throw new GqlAuthorizationError('you do not own this sub')
      }
```

Non-owners cannot use `setDomain`, and if the `domainName` changes it is expected to delete the previous domain and start a verification with the new domain. But as you said, this cannot even happen in UI because the user has to `reset` the domain before continuing.

proxy.js
Soxasora: `https://www.pizza.com/~bitcoin` redirects to `https://www.pizza.com` since that part removes the territory name.

It would be much easier to redirect to stacker.news via proxy indeed. I'll implement this!

Soxasora: I've been trying to implement this but in dev this can't work due to a Next.js bug: https://github.com/vercel/next.js/issues/44482

It seems like redirecting from `www.pizza.com/~bitcoin` to `localhost:3000/~bitcoin` won't work because `localhost` will be ignored and it will be treated as a relative redirect (`www.pizza.com/~bitcoin`) causing a redirect loop.
If I instead redirect to, idk, `stacker.news` it will work.

In production it should work, but the fact that it's untestable and will actually break the middleware in dev... makes this a bit weird to work with.

I think we can postpone this to the UI/UX PR, I like the idea of being able to redirect to stacker news in one place, probably we need to be creative.

proxy.js
Soxasora: I've decided to use the cached domain mappings where we have an asynchronous context, such as `getGetServerSideProps`. It feels more appropriate.

Soxasora:
Cleaned the code just a bit and implemented `DOMAIN_BETA_IDS` for private beta access beyond just SN admins.

lib/domains.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+export async function getDomainMappingFromRequest (req) {
+  // handles both NextRequest and Node request
+  const host = req.headers.host ?? req.headers.get('host')
+  if (!host) return null
+
+  const normalized = process.env.NODE_ENV === 'development' ? host.split(':')[0] : host
+  const mapping = normalized && normalized !== SN_MAIN_DOMAIN.host
```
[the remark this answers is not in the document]
Soxasora: oops sorry

api/ssrApollo.js
```diff
@@ -169,6 +170,15 @@ export function getGetServerSideProps (
 
     const client = await getSSRApolloClient({ req, res })
 
+    // inject custom domain data if any
```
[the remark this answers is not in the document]
Soxasora: oh this is huge

components/territory-domains.js
```diff
@@ -15,8 +15,18 @@ const DOMAIN_POLL_INTERVAL_MS = 10_000
...
-export const DomainProvider = ({ domain, children }) => {
+export const DomainProvider = ({ domain: ssrDomain, children }) => {
+  const [domain, setDomain] = useState(ssrDomain ?? null)
+
+  // maintain the custom domain state across re-renders and nodata navigations
+  useEffect(() => {
+    if (ssrDomain !== undefined) {
+      setDomain(ssrDomain)
+    }
+  }, [ssrDomain])
```
Soxasora, line 26: removed by mistake in a4b0e1a, this keeps the custom domain alive across client side navigation and re-renders.

PR #2904, opened 2026-04-05
chore: upgrade to React 19, Next.js 16, Apollo Client 4, Apollo Server 5

Soxasora:
## Description

This PR aims at upgrading Next.js from 14 to 16.
Other than performance gains, the most recent reason to update is the introduction of the Node.js Proxy (ex-middleware),  enabling custom domains (#2897) db lookup right inside the proxy, without needing an API endpoint to do so.

## Upgrades

- Next.js 14.2.25 -> 16.2.2
  - node.js middleware/proxy by default
- React 18.3.1 -> 19.2.4
  - required by Next.js 16 and recommended by Lexical
- GraphQL 16.9.0 -> 16.13.2
  - required by Apollo
- Apollo Client 3.11.8 -> 4.1.6
  - React 19 support
- Apollo Server 4.11.0 -> 5.5.0
  - React 19 support
- Apollo Server Integrations for Nextjs 3.1.0 -> 4.1.0
  - Next.js 16 support
- Plausible 4.0.0
  - React 19 support
- \@yudiel/react-qr-scanner 2.0.8 -> 2.5.1
  - React 19 support
- qrcode.react 4.0.1 -> 4.2.0
  - React 19 support
- react-avatar-editor 13.0.2 -> 15.1.0
  - React 19 support
- recharts 2.13.0 -> 2.15.4
  - React 19 support
  - react-is 19.2.4 override
    - can be removed after upgrade to recharts 3.x
- eslint-plugin-next 14.2.15 -> 16.2.2
- eslint-config-next 16.2.2

## Changes

TBD

<details>
<summary>Next.js rewrites</summary>

```js

      {
        source: '/~:sub/:slug*\\?:query*',
        destination: '/~/:slug*?:query*&sub=:sub'
      },
```

This rewrite rule has been removed because it's not supported by Next.js 15+
but this rule already does the same thing and is supported:

```js
      {
        source: '/~:sub/:slug*',
        destination: '/~/:slug*?sub=:sub'
      },
```

</details>

## Screenshots

Stacker News running on Next.js 16.2.2

Screenshot 2026-04-05 at 17 39 30

## Additional Context

The upgrade to Next.js 16 and React 19 required the upgrade of other core libraries, such as Apollo Client/Server.
Upgrading Apollo meant handling breaking changes to `useLazyQuery`, `useQuery`, and `ApolloLink`. Everything seems to be working as intended but still needs deep QA.

---

Safari DevTools are broken since Next.js 15 https://github.com/vercel/next.js/issues/78524 if Next.js is ran without Turbopack.

---

Next.js now catches `console.error`s and show them as thrown errors in dev mode.

---

**Apollo**

<details>
<summary>Apollo breaking changes</summary>

  1. The notifyOnNetworkStatusChange option now defaults to true. This affects React hooks that provide this option (e.g. useQuery) and ObservableQuery instances created by client.watchQuery.
     This change means you might see loading states more often, especially when used with refetch or other APIs that cause fetches. If this causes issues, you can revert to the v3 behavior by setting the global default back to false.
  2. If we use `useLazyQuery` to execute queries but don't read the state values returned in the result tuple, it is now recommended to use `client.query` directly. [src](https://www.apollographql.com/docs/react/migrating/apollo-client-4-migration#a-note-about-hook-usage)
     - In Apollo Client 4, useLazyQuery focuses on user interaction and more predictable data synchronization with the component. Network requests are now initiated only when the execute function is called. This makes it safe to re-render your component with new options provided to useLazyQuery without the unintended network requests. New options are used the next time the execute function is called.
     - Calling the execute function of useLazyQuery during render is no longer allowed and will now throw an error, recommend instead migrating to useQuery which executes the query during render automatically.
     - In-flight queries are now aborted more frequently. This occurs under two conditions:
        The component unmounts while the query is in flight
        A new query is started by calling the execute function while a previous query is in flight
        In each of these cases, the promise returned by the execute function is rejected with an AbortError. This change means it is no longer possible to run multiple queries simultaneously.
        In some cases, you may need the query to run to completion, even if a new query is started or your component unmounts. In these cases, you can call the .retain() function on the promise returned by the execute function.

  3. `useQuery` got rid of `onCompleted` and `onError` callbacks
     - hasNewNotes used onCompleted and has been migrated to useEffect pattern
  4. New requirements for custom cache implementations
     Custom cache implementations must now implement the fragmentMatches method, which is required for fragment matching in Apollo Client 4.

</details>

## Checklist

tbd

7, everything seems to be working. QA done on payments, wallets, normal website interaction (post, comment, navigation), live comments, editor

n/a

Yes, with Next.js 15/16 `localhost` is the default and only allowed dev origin. This has been modified to ensure it also always includes `NEXT_PUBLIC_URL` and to support additional origins, we now have a new env var:

`ALLOWED_DEV_ORIGINS=`

By default it's empty, which means that only `localhost` and `NEXT_PUBLIC_URL` are allowed.

- help with Apollo migration
- upgrade from Next.js 15 to 16 via Next.js Devtools MCP (Claude)

---

> [!NOTE]
> **High Risk**
> High risk because it upgrades core runtime dependencies (Next.js/React/Apollo) and rewires Apollo hook/query behavior (notably `useLazyQuery` promise/abort semantics and removal of `onCompleted`), which can affect navigation, data fetching, and payments-related flows across the app.
> 
> **Overview**
> Upgrades the frontend stack to Next.js 16/React 19 and adapts the codebase for Apollo Client 4 by switching React hooks imports to `@apollo/client/react`, enabling `@client` directive support via `LocalState`, and updating client config (e.g., `ApolloLink.from`, devtools config, and default `notifyOnNetworkStatusChange: false`).
> 
> Refactors multiple `useLazyQuery`/`useQuery` call sites to the new promise-based/abort-heavy behavior: moves `onCompleted` logic into `.then(...)`/`useEffect`, adds centralized `isAbortError` filtering to avoid noisy errors on aborted requests, and updates a few components (popovers, upload fee refresh, suggestions, auto-retry pay-ins) to execute queries explicitly and handle errors.
> 
> Adjusts Next.js config for v15+ compatibility and dev ergonomics: adds `allowedDevOrigins` derived from `NEXT_PUBLIC_URL` plus new `ALLOWED_DEV_ORIGINS`, removes an unsupported rewrite, adds `transpilePackages`, and silences bootstrap sass deprecation warnings; also adds a new flat `eslint.config.mjs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36ab54bdb0a111af12f7968779396af62e037536. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


api/ssrApollo.js
```diff
@@ -24,9 +25,10 @@ export default async function getSSRApolloClient ({ req, res, me = null }) {
   if (req) {
     req = await multiAuthMiddleware(req, res)
   }
-  const session = req && await getServerSession(req, res, getAuthOptions(req))
+  const session = req && (await getServerSession(req, res, getAuthOptions(req)))
```
Soxasora: I don't know if the extra parenthesis were added by Apollo's codemod or the JS Standard linter that I had to use because the codemod didn't write in JS Standard.

pages/items/[id]/edit.js
```diff
@@ -66,8 +66,8 @@ export default function PostEdit ({ ssrData }) {
           ...territoryAdds
         }
       })
-    })
-  }, [subs])
+    }).catch(err => console.error(err)).catch(err => console.error(err))
```
Comment (sir-opti) [inline pages/items/[id]/edit.js]: double `.catch()`

components/payIn/hooks/use-pay-in-mutation.js
```diff
@@ -3,9 +3,9 @@
 
 import { useCallback, useState } from 'react'
 import { InvoiceCanceledError } from '@/wallets/client/errors'
-import { useApolloClient, useMutation } from '@apollo/client'
+import { useApolloClient, useMutation } from '@apollo/client/react'
 import usePayPayIn from '@/components/payIn/hooks/use-pay-pay-in'
-import { getOperationName } from '@apollo/client/utilities'
+import { getOperationName } from '@apollo/client/utilities/internal'
```
sir-opti: Using `/internal` feels a bit debt-ish because it doesn't carry any signature guarantees.

bot suggests to use:

```javascript
import { getMainDefinition } from '@apollo/client/utilities'

function getOperationName (doc) {
  return getMainDefinition(doc).name?.value ?? null
}
```

note that this does put `checkDocument` on the execution path.

Alternatively, can just write it out, since it's graphql AST, not apollo-specific at the core:

```javascript
function getOperationName (doc) {
  const op = doc.definitions.find(
    d => d.kind === 'OperationDefinition' && d.name
  )
  return op?.name.value ?? null
}
```

worker/index.js
```diff
@@ -69,7 +73,21 @@ async function work () {
...
+    you can safely remove this option.
+    */
+    localState: new LocalState({}),
+
+    /*
+    Inserted by Apollo Client 3->4 migration codemod.
+    If you are not using the `@defer` directive in your application,
+    you can safely remove this option.
+    */
+    incrementalHandler: new Defer20220824Handler()
```
sir-opti: Couldn't find any use of this in the entire codebase.

worker/index.js
```diff
@@ -69,7 +73,21 @@ async function work () {
...
       }
-    }
+    },
+
+    /*
+    Inserted by Apollo Client 3->4 migration codemod.
+    If you are not using the `@client` directive in your application,
+    you can safely remove this option.
+    */
+    localState: new LocalState({}),
```
sir-opti: Couldn't find any use of this outside of `fragments` - not needed here?

sir-opti (commented):
Did some code review on the diff, pre-testing.

pages/items/[id]/edit.js
Comment (Soxasora) [reply pages/items/[id]/edit.js]: tsk should've been more careful with the auto-refactor. thanks!

components/payIn/hooks/use-pay-in-mutation.js
Soxasora: Yeah this is something I didn't quite understand. Apollo doesn't provide an explanation for this, the migration guide only notifies that a lot of `@apollo/client/utilities` got moved to `/internal`. https://www.apollographql.com/docs/react/migrating/apollo-client-4-migration#list-of-all-changed-imports

I'll check this out, thanks ^^

Soxasora:
Hi opti! All the things you found were added automatically by the Apollo Client 4 migration codemod. The `@defer` directive support was something I've already removed but probably got added again when I re-ran the codemod on `worker` and `api`.

Thanks you so much for checking this PR, I'll address immediately 🫡

worker/index.js
Soxasora: Correct, worker doesn't use any fragment with the `@client` directive.

components/payIn/hooks/use-pay-in-mutation.js
Soxasora:
```
 * Returns the first operation definition from a GraphQL document. The function
 * prioritizes operation definitions over fragment definitions, which makes it
 * suitable for documents that may contain both. If no operation definition is
 * found, the first fragment definition will be returned. If no definitions are
 * found, an error is thrown.
```

It seems like `getMainDefinition` can return something we don't want or don't expect (or throw) in the code. 

Since `getOperationName` would just return the operation definition name or `undefined`, the identical (in behavior) twin is `graphql`'s `getOperationAST` afaict, we just need to get the name from the operation definition node.

```js
const mutationName = getOperationAST(mutation)?.name?.value
```

sir-opti: I guess that it's a covert, very politically correct way of saying "don't use our hacks in your app", haha.

sir-opti: > the identical twin is graphql's getOperationAST afaict, we just need to get the name from the AST. Y/N?

I think that that is much prettier than anything the bot or I can possibly come up with.

components/use-has-new-notes.js
```diff
@@ -12,14 +12,15 @@ export function HasNewNotesProvider ({ me, children }) {
...
-            }
-          }
+          nextFetchPolicy: 'cache-and-network'
         })
 
+  useEffect(() => {
+    if (data && !data.hasNewNotes) {
+      clearNotifications()
+    }
+  }, [data?.hasNewNotes])
```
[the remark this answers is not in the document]
Soxasora: This seems like a good thing.

components/form.js
```diff
@@ -452,23 +452,23 @@ export function BaseSuggest ({
...
       getSuggestions({ variables: { q, limit: 5 } })
+        .then(({ data }) => {
+          query !== undefined && setSuggestions({
+            array: data[itemsField]
+              .filter((...args) => filterItems(query, ...args))
+              .map(transformItem),
+            index: 0
+          })
+        })
+        .catch(err => console.error(err))
```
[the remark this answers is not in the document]
Soxasora: oh wait that's right! Missed this.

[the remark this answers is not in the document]
Soxasora: Checked carefully.
`query` drives the `useEffect` re-runs, and `filterItems` or `transformItem` are inline. It can stay as-is

sir-opti:
Seeing this, a lot:

```log
app  | 2026-04-07T21:23:28.213872810Z `legacyBehavior` is deprecated and will be removed in a future release. A codemod is available to upgrade your components:
app  | 2026-04-07T21:23:28.213915174Z 
app  | 2026-04-07T21:23:28.213925669Z npx @next/codemod@latest new-link .
app  | 2026-04-07T21:23:28.213957697Z 
app  | 2026-04-07T21:23:28.213963226Z Learn more: https://nextjs.org/docs/app/building-your-application/upgrading/codemods#remove-a-tags-from-link-components
```

components/user-header.js
```diff
@@ -49,32 +50,26 @@ export default function UserHeader ({ user }) {
         activeKey={activeKey}
       >
         <Nav.Item>
-          <Link href={'/' + user.name} passHref legacyBehavior>
-            <Nav.Link eventKey='bio'>bio</Nav.Link>
-          </Link>
+          <Nav.Link as={Link} href={'/' + user.name} eventKey='bio'>bio</Nav.Link>
```
Soxasora, line 53: This change has been applied to every instance of `NextLink` used with `legacyBehavior`. We're giving `NextLink` as the component to be used in place of the default `<a>` tag.

`passHref` then becomes useless as `Nav.Link`, for example, gives the href to whatever `as` component we indicate.

---

@sir-opti I was ignoring that deprecation warning to avoid touching too many things, but the solution is quite straightforward.

sir-opti: This is an awesome change because now my eyes only hurt 200% instead of 500% from looking at that indentation 🤣

sir-opti:
Haven't found anything new between posting/deleting, editor funk, territory creation, payments/zaps/boosts, messing around with rankings and analytics - all that worked great.

I have 2 more test areas to cover on this:

- Uploads (I've been pretty mad trying to get s3 to work)
- Connected wallets

Also:

- Debugging what's up with the websocket in my dev setup
  - I think we need `allowedDevOrigins: isProd ? undefined : [(new URL(process.env.PUBLIC_URL)).host],` in `next.config.js` - unconfirmed still though
  - I get an occasional page reload - didn't get that before this patch, but may be related to the websocket

Soxasora:
Thank you for the QA!

> I think we need allowedDevOrigins: isProd ? undefined : [(new URL(process.env.PUBLIC_URL)).host], in next.config.js - unconfirmed still though

Ah yes, we do! I've added this in the custom domains PR today https://github.com/stackernews/stacker.news/pull/2897/commits/d466483e58c91b5c919f021d707480636a6f8f5f, but I didn't make the same reasoning for this PR (or custom hosts in general...). `localhost` is by default an allowed dev origin though.

I should come up with a way to satisfy both normal sndev and custom domains. I'll check it out tomorrow first thing.

---

> Uploads (I've been pretty mad trying to get s3 to work)

I think you just need to add your host to the allowed origins in `docker/s3/cors.json`. Then restart (or delete + start).

next.config.js
```diff
@@ -46,15 +46,34 @@ try {
...
+  const extraOrigins = process.env.ALLOWED_DEV_ORIGINS?.split(',').map(s => s.trim()).filter(Boolean) || []
+  devOrigins.push(...extraOrigins)
+
+  // if no origins are found, localhost is always allowed
+  return devOrigins
+}
+
+module.exports = withPlausibleProxy({ src: 'https://plausible.io/js/pa-EScEhWlTi3E-sauvdFABb.js' })({
   env: {
+    allowedDevOrigins: getAllowedDevOrigins(),
```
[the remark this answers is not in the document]
Soxasora: yes, already resolved, noticed that while merging into custom_domains_base

huumn (commented):
I'm not sure these should block merging, because afaict this would run fine in prod, but here's a few things I noticed:

1. Dev hits the nextjs error boundary on `checkMedia` if `capture` is not running and media is checked. I didn't check if this happens in prod, but it's annoying in dev and should probably be fixed.
2. One of the consequences of `notifyOnNetworkStatusChange: true` is that rerenders are happening when polling even if data doesn't change. 
    - Example: create a new comment and watch the `timeSince` change
    - If only, this suggests I don't understand the consequences of `notifyOnNetworkStatusChange: true` and wonder what else might change.
3. `components/item-popover.js`, `components/sub-popover.js`, `components/user-popover.js`, `components/territory-form.js`, `wallets/client/hooks/logger.js` all log `AbortError` still.
4. Not a regression, but we pull deps from `next/dist/*` like `NodeNextRequest` which we got lucky didn't change between upgrades. Perhaps can be changed in a follow up.

Soxasora:
> Dev hits the nextjs error boundary on `checkMedia` if capture is not running and media is checked. I didn't check if this happens in prod, but it's annoying in dev and should probably be fixed.

> > Next.js now catches `console.error`s and show them as thrown errors in dev mode.

I mentioned this in this PR description, thinking that it was a new default behavior lol, but it seems like it's a bug on the pages router: https://github.com/vercel/next.js/issues/75299

> and media is checked

This is a behavior present on `master`:

```js
  const trusted = useMemo(() => !!(srcSet || srcSetIntital) || IMGPROXY_URL_REGEXP.test(src) || MEDIA_DOMAIN_REGEXP.test(src), [srcSet, srcSetIntital, src])
  const [isImage, setIsImage] = useState((kind === 'image' || legacySrcSet?.video === false) && trusted)
```

If an image is uploaded, `trusted` will be true because at least `MEDIA_DOMAIN_REGEXP.test(src)` is true (being an upload), but when we convert markdown `` to Lexical `MediaNode` we always set `kind: unknown` triggering media checks.

We don't load the media client-side anymore to determine if it's an image or a video (or media at all), instead we always contact `capture`. I think we wanted to create a dedicated microservice for this.

---

> One of the consequences of `notifyOnNetworkStatusChange: true` is that rerenders are happening when polling even if data doesn't change.

Just confirmed, basically it re-renders even on `data.loading` or `data.networkStatus` changes. By switching to `false` we're back to the original behavior, which we probably want so I'll switch back.

---

> `components/item-popover.js, components/sub-popover.js, components/user-popover.js, components/territory-form.js, wallets/client/hooks/logger.js` all log AbortError still.

I'll handle `AbortError` in every catch for this, pretty bad that we **have** to do this.

---

> Not a regression, but we pull deps from `next/dist/*` like `NodeNextRequest` which we got lucky didn't change between upgrades. Perhaps can be changed in a follow up.

Probably whatever we needed from `NodeNextRequest` we can get it from the public `NextApiRequest` for the pages router iirc.

huumn:
> Probably whatever we needed from NodeNextRequest we can get it from the public NextApiRequest for the pages router iirc.

It's mostly cookies parsing and in `_app.js` we use `next/dist/client/router` instead of `next/router`.

> By switching to false we're back to the original behavior, which we probably want so I'll switch back.

I agree - that's the safest thing. Perhaps we can switch it to the default later (assuming it's worth it).

> I mentioned this in this PR description, thinking that it was a new default behavior lol, but it seems like it's a bug on the pages router: https://github.com/vercel/next.js/issues/75299

lol that's nasty

> I'll handle AbortError in every catch for this, pretty bad that we have to do this.

It might be worth creating our own `useDeferredQuery` as a small wrapper around `client.query`, but also cool if not - borderline need for an abstraction.

Soxasora:
> It might be worth creating our own useDeferredQuery as a small wrapper around client.query, but also cool if not - borderline need for an abstraction.

I'm sure we won't need it if we fix the way we use `useLazyQuery` in some places. I'll come up with a follow-up PR refactor, for example the `BaseSuggest`'s `getSuggestions` that is executed by `useEffect` when `query !== undefined`, can just be a `useQuery` with `skip` (avoiding `AbortError`)

next.config.js
```diff
@@ -46,7 +46,25 @@ try {
...
+
+  // extra origins, comma separated, can be set in the environment variable ALLOWED_DEV_ORIGINS
+  const extraOrigins = process.env.ALLOWED_DEV_ORIGINS?.split(',').map(s => s.trim()).filter(Boolean) || []
+  devOrigins.push(...extraOrigins)
+
+  // if no origins are found, localhost is always allowed
+  return devOrigins
+}
+
+module.exports = withPlausibleProxy({ src: 'https://plausible.io/js/pa-EScEhWlTi3E-sauvdFABb.js' })({
```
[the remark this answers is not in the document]
Soxasora:
```js
export default function withPlausibleProxy(options: {
    /**
     * The site-specific script URL from your Plausible dashboard, e.g. https://plausible.io/js/pa-XXXXX.js.
     */
    src: string;
    /**
     * The local path for the proxied script. Defaults to /js/script.js.
     */
    scriptPath?: string;
    /**
     * The local path for the proxied API endpoint. Defaults to /api/event.
     */
    apiPath?: string;
}): (nextConfig: NextConfig) => NextConfig;
```

wdym??

huumn:
> I mentioned this in this PR description, thinking that it was a new default behavior lol, but it seems like it's a bug on the pages router: https://github.com/vercel/next.js/issues/75299

I went ahead and switched `mediaCheck` to use `console.warn`. That's semantically more accurate anyway. We all tend to abuse `console.error`.

This is good to go now imo.

Soxasora:
Thank you! You're right about `console.error`, will be more precise.

PR #2906, opened 2026-04-07
fix sndev login and correct `useVerificationToken` query

Soxasora:
Surfaced by @sir-opti 

## Description

Trying to login with another account with `sndev login` can fail because our useVerificationToken in [...nextauth].js only gets verification tokens created in the last 5 minutes, instead of checking the `expires` date.

I propose two changes

- Adjust the `useVerificationToken` query to get unexpired verification requests instead of requests created in the last 5 minutes
  - the query was wrong from the start, it should check for `expires`
- `sndev login` deletes existing verification requests with the same token before inserting a new one.
  - futureproofing

## Checklist

Yes

7

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 8e9a61158af879a67147e571fbf2992ebb26ed4a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -334,7 +334,7 @@ export const getAuthOptions = (req, res) => ({
           FROM (
             SELECT id FROM verification_requests
             WHERE identifier = ${identifier}
-            AND created_at > NOW() - INTERVAL '5 minutes'
+            AND expires > NOW()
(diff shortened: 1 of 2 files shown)


sir-opti (approved):
Tested this a couple of times, solves the problem and was unable to find any edge cases.

PR #2908, opened 2026-04-08
custom domains: auth

Soxasora:
## Description

Part of #1942, revives and adapts #2180
Focuses on synchronizing authentication between SN and a custom domain.

### Login overview

Authentication is centralized on `stacker.news`.
When a user visits `/login` on a custom domain, they are redirected to `stacker.news/login` with a `domain` param. This param customizes the login experience for that specific domain/territory.

domain-login-overview

#### Flow

From the login page, users can authenticate with the current nym or choose a different account to sync.
A `callbackUrl=/api/auth/sync/...` is always included to support full login/signup flows

After authentication:
- user is logged into `stacker.news`
- they're redirected to the auth sync endpoint
- the session is synchronized with the custom domain

#### Next-auth redirects

To properly support logout flows on custom domains (without `router.push` hacks), next-auth now supports a custom redirect behavior.

If `callbackUrl` is:
- absolute URL -> allowed only for verified custom domains -> redirects back to that domain
- relative path -> defaults to `stacker.news`

### DNS monitoring

A recurring `checkActiveDomainsDNS` pgboss job runs every 5 minutes to verify that custom domains still point correctly to us.

When a DNS drift is detected, the domain is moved to `HOLD`, triggering:
- ACM certificate revocation
- `tokenVersion` increment
- domain re-verification required

This check is essential to custom domain safety and JWT revocability.

#### JWT revocability

JWTs issued for custom domains now include:
- `domainId`
- `tokenVersion`

Validation:
- `tokenVersion` mismatch -> token is invalid
- `domainId` mismatch -> token is invalid

This ensures that the existing tokens are invalidated when DNS changes and/or when domain is re-verified or re-created.

---

Attack scenarios and infos are available in `/docs/dev/custom-domains.md`

## Media

custom domain login -> pick -> sync -> logout

https://github.com/user-attachments/assets/0d8deae7-91dc-4adc-af61-7f59dfebb51b

custom domain signup -> main domain -> sync

https://github.com/user-attachments/assets/223e4792-78e4-4804-b737-fb40e9ce3138

## Checklist

Yes, existing behavior is preserved and every change only applies to custom domains.

6
- login with nym: OK
- login/signup from scratch: OK
- logout: OK
- redirect to `/login`: OK
- JWT revoked after DNS drift: OK
- JWT revoked after domain re-insert: OK

Yes

n/a

- reviews
- auth hardening
- nits

---

> [!NOTE]
> **High Risk**
> Touches authentication, redirect handling, and cookie/JWT validation across domains, plus adds new DB tables/triggers and a scheduled DNS drift job that can revoke sessions and move domains to `HOLD`. Incorrect host/redirect validation or drift classification could lock users out or create redirect/security issues.
> 
> **Overview**
> Adds a **custom-domain auth sync flow** so `/login` and `/signup` on an `ACTIVE` custom domain redirect through the main domain for NextAuth, then mint a **domain-bound session JWT** back on the custom domain via new `/api/auth/domains/*` endpoints.
> 
> Hardens auth for custom domains by introducing `Domain.tokenVersion` + `domainId` claims and validating them in the NextAuth `jwt` callback, plus a new NextAuth `redirect` allowlist that permits only safe relative paths, same-origin URLs, or `ACTIVE` custom domains. SSR/signup redirects now preserve the correct origin on custom domains, logout uses `window.location.origin`, and the login UI gains an account-picker (`LoginWithNymButton`) to sync an existing nym.
> 
> Adds **DNS drift monitoring**: `verifyDNSRecord` now reports transient vs conclusive failures, a new `checkActiveDomainsDNS` pgboss cron flips drifting domains to `HOLD` (triggering cert cleanup) and bumps `tokenVersion` via a DB trigger, and domain mappings cache now includes `id`/`tokenVersion` with safer host parsing (`lib/safe-url.js`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23c34534093969289fa46a124bc20c185bb9a629. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pages/api/auth/sync.js
```diff
@@ -0,0 +1,174 @@
...
+      }
+
+      return res.status(200).json({ status: 'OK', sessionToken: sessionTokenResult.sessionToken })
+    }
+
+    if (req.method === 'GET') {
+      const { domain, redirectUri = '/', signup } = req.query
+      if (!domain || !redirectUri?.startsWith('/')) {
+        return res.status(400).json({ status: 'ERROR', reason: 'domain and a correct redirectUri are required' })
+      }
```
huumn: We'll want to protect the `redirectUri` from open redirects, e.g. `//evil.com/path` passes this check and survives `new URL(redirectUri, request.url)` as `https://evil.com/path`.

```js
const validRedirectUri =
  typeof redirectUri === 'string' &&
  redirectUri.startsWith('/') &&
  !redirectUri.startsWith('//') &&
  !redirectUri.startsWith('/\\')
```

and then just to be extra safe, make sure the new `url.origin` is what we expect after `new URL`

Soxasora: Right! We need better checks on redirects. I'll put this in the TODO list!

pages/api/auth/sync.js
```diff
@@ -0,0 +1,174 @@
...
+    return { status: 'ERROR', reason: 'failed to create verification token' }
+  }
+}
+
+async function redirectToDomain (res, domainName, verificationToken, redirectUri) {
+  try {
+    const protocol = process.env.NODE_ENV === 'development' ? 'http' : 'https'
+    const target = new URL(`${protocol}://${domainName}`)
+
+    target.searchParams.set('sync_token', verificationToken)
```
huumn: Is `sync_token` scoped to a territory? afaict it isn't but I need to walkthrough the flow more, and it might be unscoped for a reason. Given it's short lived it might not be worth scoping though.

components/account.js
```diff
@@ -93,6 +94,10 @@ const AccountListRow = ({ account, selected, ...props }) => {
 }
 
 export const useIsLurker = () => {
+  // TODO, signup button for lurkers conflicts with one-click-login path
+  const { domain } = useDomain()
+  if (domain) return false
```
huumn, line 99: should we move this to after the `useAccounts` hook?

Soxasora: we should because of `da rules of hooks`

pages/api/auth/sync.js
Soxasora: Oh I didn't scope `sync_token` because it's short lived and, hopefully, used instantly. But we can make it more secure.

huumn:
I qa'd the auth flow just now and it works great!

huumn:
This is all bot review (what I find myself doing in preparation of my review) but I wanted to share before in case any of it is useful. Screenshots of the canvas I had it generate:

Screenshot 2026-05-03 at 1 58 24 AM
Screenshot 2026-05-03 at 1 58 39 AM

-----------

The fix plan this led to (again, just bot stuff, I have not reviewed much). Phase 1 seems reasonable at a glance. I'm unsure about phase 2 being a meaningful improvement (but again, I have not absorbed all the context yet). 

<details><summary>THE FIX PLAN</summary>

## Phase 1: Harden Existing Flow

1. Add a server-only redemption proof to the sync POST path.
   - Change [`proxy.js`](proxy.js) `syncAccount` so its server-side `fetch` to [`pages/api/auth/sync.js`](pages/api/auth/sync.js) includes a proof header derived from `verificationToken`, `domainName`, and `NEXTAUTH_SECRET`.
   - Change [`pages/api/auth/sync.js`](pages/api/auth/sync.js) POST handling to validate that proof before consuming a token or minting a JWT.
   - Use a small shared helper only if needed to avoid duplicating crypto incorrectly; otherwise keep the implementation inline and minimal.
   - Result: a `sync_token` leaked to a DNS hijacker is no longer enough to mint a JWT, because only SN middleware can produce the server-only proof.

2. Make security-sensitive domain checks fresh, not cached.
   - Add a fresh DB-backed domain lookup in [`lib/domains.js`](lib/domains.js), e.g. `getFreshDomainMapping(domain)`, selecting `id`, `domainName`, `subName`, and `tokenVersion` for `status: ACTIVE`.
   - Use the fresh lookup in [`pages/api/auth/[...nextauth].js`](pages/api/auth/[...nextauth].js) for custom-domain JWT validation.
   - Use it in auth redirect allowlisting if practical, so recently revoked domains are not accepted by NextAuth redirect policy.
   - Keep `domainsMappingsCache` for proxy routing only, where bounded staleness is less security-critical.

3. Keep the existing mint-side DB checks, but tighten the consume order.
   - In [`pages/api/auth/sync.js`](pages/api/auth/sync.js), validate method, domain, and server proof before `consumeVerificationToken` deletes anything.
   - Continue re-reading the domain in `createSessionToken` before minting the JWT.
   - Preserve one-time token semantics and the existing `domainId` binding.

4. Fix the failed sync route.
   - Change [`proxy.js`](proxy.js) error redirects from `/error` to an existing route such as `/auth/error?error=Callback`.
   - This addresses expired, invalid, already-consumed, and failed server-proof sync attempts.

5. Reduce accidental leakage from the URL handoff while Phase 2 is pending.
   - After successful `syncAccount`, keep redirecting to the sanitized `redirectUri`, so the final URL no longer contains `sync_token`.
   - Consider setting `Referrer-Policy: no-referrer` on the intermediate sync redirect response in [`pages/api/auth/sync.js`](pages/api/auth/sync.js) if it can be done without weakening the broader site policy.
   - Do not rely on this as the security boundary; the proof header is the boundary.

## Phase 2: Remove URL-Borne Bearer Material

1. Design a replacement handoff that does not put redeemable auth material in the custom-domain URL.
   - Candidate: main domain creates a non-bearer flow id, and only SN middleware can redeem it using server-side proof.
   - Candidate: main domain renders an auto-submitted POST to the custom domain with non-bearer state, still requiring server proof on redemption.
   - Candidate: split the current sync token into browser-visible state plus server-only verifier, similar to PKCE, so browser-visible data cannot mint a JWT alone.

</details>

huumn:
Rereading my findings from last night, it's unclear how high risk the high risk finding is. The finding is valid and the sync token *is* a special case, but imho the 5 minute domain switch check + jwt invalidation is sufficient for our private beta, whether after `sync_token` or normal requests afaict.

On the other hand, it is probably worthwhile to have the ability to do synchronous DNS checks and add extra protections depending on the request scenario, ie places where an attack is easier to pull off. Is this such a scenario, and is it much more severe than average? Marginally, maybe. 

Anyway, I'm still a bit naive to these changes. Regardless of whether special case hardening is worthwhile, it can wait until we leave the private beta if that's what's best. There always be inches to gain in security given we are limited to reactive defenses.

sir-opti:
> unclear how high risk the high risk finding is.

I had something similar reported and it combines with other issues. I'm still in repro phase; this is high compexity.

I do think that "Design a replacement handoff that does not put redeemable auth material in the custom-domain URL" is valid though - having tokens in query string puts it in browser history. Can probably be protected with a challenge secret like done in [PKCE](https://www.rfc-editor.org/rfc/rfc7636), to prevent capture by XSS / browser extensions. The "protocol flow" for PKCE paints a clear solution:

```
                                                 +-------------------+
                                                 |   Authz Server    |
       +--------+                                | +---------------+ |
       |        |--(A)- Authorization Request ---->|               | |
       |        |       + t(code_verifier), t_m  | | Authorization | |
       |        |                                | |    Endpoint   | |
       |        |<-(B)---- Authorization Code -----|               | |
       |        |                                | +---------------+ |
       | Client |                                |                   |
       |        |                                | +---------------+ |
       |        |--(C)-- Access Token Request ---->|               | |
       |        |          + code_verifier       | |    Token      | |
       |        |                                | |   Endpoint    | |
       |        |<-(D)------ Access Token ---------|               | |
       +--------+                                | +---------------+ |
                                                 +-------------------+
```

where, if `t_m == 'sha512'`, `t(code_verifier) == sha512(code_verifier)` which are both passed in the auth request, and then `code_verifier` in cleartext is added to the redemption request. Server checks that `sha512(token_verifier)` matches the initial digest given. Could use a short-lived cookie as a side-channel to keep it out of query strings and post data?

Soxasora:
Thanks for all the research guys, I think we just want to add the proof header just to ship the private beta of custom domains. It's a small change that won't require re-designing the pipeline.

I'm also okay with a DNS query on `POST`, since it's not a path that needs to be in any way *ultra-fast*

> 2. Make security-sensitive domain checks fresh, not cached.
> Add a fresh DB-backed domain lookup in [lib/domains.js](https://github.com/stackernews/stacker.news/pull/lib/domains.js), e.g. getFreshDomainMapping(domain), selecting id, domainName, subName, and tokenVersion for status: ACTIVE.
> Use the fresh lookup in [pages/api/auth/[...nextauth].js](https://github.com/stackernews/stacker.news/pull/pages/api/auth/%5B...nextauth%5D.js) for custom-domain JWT validation.

Instead, this is something I don't think we'll ever want to do, it would mean a db lookup on every SSR and graphql request.

This being said, an attacker can anyway exchange the verification token for a JWT via our middleware when they re-establish a correct DNS configuration towards us, and if they're lucky enough to be in the correct time window. This is, sadly, valid for any safety measures we can come up with because the attacker can control the domain and use a reverse proxy to access cookies, sniff and do all kinds of things (still they gotta be lucky with timing).

But! We must remember that even if the territory owner/hijacker gets the verification token and/or the full JWT token, it **will** get revoked the moment our cache catches up. And said JWT cannot be shared with `stacker.news` or other custom domains.

---

edit: I'm all for implementing PKCE after private beta. Actually last year I implemented a (bad) prototype of a PKCE-ish process: https://github.com/stackernews/stacker.news/pull/2180#discussion_r2138929676, but I remember removing it because auth sync was starting to be too complex for an MVP + the dread of knowing that whatever we do can be exploited.

sir-opti:
> I'm all for implementing PKCE after private beta. Actually last year I implemented a (bad) prototype of a PKCE-ish process: https://github.com/stackernews/stacker.news/pull/2180#discussion_r2138929676, but I remember removing it because auth sync was starting to be too complex for an MVP + the dread of knowing that whatever we do can be exploited.

My only counter to that is that the moment with the least friction to do it is now, unless private beta stage is throwaway code and all these massive PRs will be reverted and rewritten from scratch - which I don't expect, but maybe I'm naive. Doesn't have to be in this PR though!

huumn:
> having tokens in query string puts it in browser history

It's a short-lived token, so while not great, it's about as bad as a magic link in this respect (I think).

I think the risk with the `sync_token` is like the risk with the `jwt`: the custom domain owner can gain access to either if they switch their DNS records at the right time then impersonate clients until we notice the DNS switch.

> Can probably be protected with a challenge secret like done in [PKCE](https://www.rfc-editor.org/rfc/rfc7636)

Unless I'm missing something, with a DNS record switch, even if some secrets are only known to the client, those secrets can be MiTM'd at C/D in the diagram. It does make an attack harder though.

We've been unable to find a flow that can't be MiTM'd in this way. It's not possible afaik, but it's like we need client context/storage to bind to DNS records/certs at a point in time, such that when records/certs change, the client context invalidates.

huumn:
> unless private beta stage is throwaway code and all these massive PRs will be reverted and rewritten from scratch

If we can materially improve upon the security of these PRs by reverting and rewriting them from scratch before making the beta public, we will.

huumn:
It'd perhaps help to enumerate attacker personas/scenarios and be clear about what we are protecting against:

1. territory founders (capable of performing the DNS switch)
2. other kinds of attackers

PKCE may address (2) substantially without being able to address (1). Specifically, we've mentioned (2) could help when:

1. someone has access to browser history (and token fails to be consumed and hasn't expired yet)
2. XSS
3. browser extensions to the extent that they can't operate on client storage

Soxasora:
The problem with the (2) scenarios is that we've already lost at that point, the user is compromised and the attacker can just get the JWT. The same can be done for `stacker.news` and any other website.

I agree with @huumn that PKCE can't help with (1) and it can only make the attack a little harder to execute.

---

Maybe I'm about to say something naive or in the spirit of whataboutism, but I actually think that auth is okay-ish with the header, and that the real problem is wallets. Impersonating someone for a brief period of time is a 1000000x better outcome than stealing wallet creds.
I wouldn't want to enter the same trap we entered last year and halt custom domains, I propose to return to this subject + wallets when we'll have fully-working custom domains.

* on the wallets subject, the attacker can get the wallet creds only if the user give their passphrase ... which they can do if it's a convincing replica of `stacker.news`
  * oh but there's more we could do here, like iframes or payments scoped to `stacker.news`. we'll need to talk in-depth about this

sir-opti:
> I propose to return to this subject + wallets when we'll have fully-working custom domains.

I'll do some work in parallel to you, without disrupting your flow (after I finalize aws sdk stuff)

pages/api/auth/sync.js
```diff
@@ -0,0 +1,201 @@
...
+
+async function consumeVerificationToken (verificationToken, expectedDomainId) {
+  try {
+    const userId = await models.$transaction(async tx => {
+      const token = await tx.verificationToken.findFirst({
+        where: {
+          token: verificationToken,
+          expires: { gt: new Date() }
+        }
+      })
```
huumn: `VerificationToken.token` is `@unique` so we might be able to use `findUnique`

proxy.js
```diff
@@ -65,6 +75,68 @@ async function customDomainMiddleware (request, domain, subName) {
...
+    const proof = createAuthSyncProof({
+      verificationToken: token,
+      domainName,
+      secret: process.env.NEXTAUTH_SECRET
+    })
+
+    const body = JSON.stringify({ verificationToken: token, domainName })
+    const fetchHeaders = new Headers(headers)
+    fetchHeaders.set('Content-Type', 'application/json')
+    fetchHeaders.set(AUTH_SYNC_PROOF_HEADER, proof)
```
huumn: It might be worth NOT cloning the incoming headers. It doesn't looks like the are used and it'd be nice to avoid having to think about what's in them.

huumn (commented):
After an afternoon of bot review I have two `medium` findings.

- valid afaict: `GET /api/auth/sync` csrf
    - Attacker who controls any `ACTIVE` custom domain can lure logged-in Alice (single click, image embed, etc.) to `https://stacker.news/api/auth/sync?domain=attacker-territory.com&redirectUri=/`. Her main-site session authenticates her, the `GET` path mints a verification token bound to `{aliceId, attackerDomainId}`, and `302`s her to the attacker's domain where she now has a custom-domain session cookie linking her stacker.news identity. 
- unrealistic afaict: race in `consumeVerificationToken` and `createSessionToken` due to multi-statement SQL not wrapped in a transaction
    - the bug would require `ACTIVE` -> `HOLD` -> `ACTIVE` happening very fast (like milliseconds), which is not possible with the way the domain mapping works ... but it might be nice to not have to think about this relationship in the future

This was flagged as "policy":

- > `redirect` callback can never enforce "redirect only to a domain you started from"
    - a bit related to the CSRF finding above 

-----------------

I'm going to done some more targeted bot review and QA another time (which I expect to go fine). 

After that I'll do a human review, and if the bots are right, I shouldn't find anything new.

pages/api/auth/sync.js
Soxasora: I think I did it to compact the `expires` check. I'm changing this to also address the race condition between domain status and `consumeVerificationToken` + `createSessionToken` with a lock.

Soxasora:
I've implemented CSRF protection for the login flow using a JWE. `/api/auth/sync` will reject any request that doesn't come with a valid proof.

1. User goes to `pizza.com/login`
2. Middleware creates a login flow proof that expires in 10 minutes, and redirects to `/api/auth/redirect` with the proof
3. `/api/auth/redirect` redirects to `/api/auth/sync` with the proof
  - must remember that this extra redirect endpoint is a workaround to nextjs' ability to redirect to absolute localhost urls
5. auth sync will verify the proof before creating a verification sync token

This means that only `/login` or `/signup` on a custom domain will be able to establish an auth sync login flow, everything else is rejected.
The proof is replayable, but the proof alone doesn't grant a session.

---

> unrealistic afaict: race in consumeVerificationToken and createSessionToken due to multi-statement SQL not wrapped in a transaction

even if it's an unrealistic scenario, it's a chance to improve code quality! `consumeVerificationToken` now starts a transaction, locks the Domain row and finally deletes the verification token; it will then return what's necessary to create the final session token without requesting domain info again.

---

every other nit has been addressed, thank you so much k00b!

edit:

> redirect callback can never enforce "redirect only to a domain you started from"

I missed this, but do we really need to enforce this? Now that the flow is completely controlled by us, I don't see why this should be enforced.

lib/domains/auth-sync.js
```diff
@@ -0,0 +1,40 @@
...
+export function verifyAuthSyncProof ({ received, verificationToken, domainName, secret }) {
+  if (!secret || !verificationToken || !domainName) return false
+  return safeEqual(received, sign(secret, `${verificationToken}:${domainName.toLowerCase()}`))
+}
+
+export function createLoginFlowProof ({ domainName, expiration, secret }) {
+  if (!secret) throw new Error('login flow proof: missing secret')
+  if (!domainName || !expiration) throw new Error('login flow proof: missing inputs')
+  return sign(secret, `${domainName.toLowerCase()}:${expiration}`)
+}
```
Soxasora: wait... this can be a JWE, then the login url will look cleaner with just a JWE instead of proof + expiration params.
I'll do this right now.

Soxasora: Well the login URL is not cleaner ... it's worse ... but this is better as we don't have to deal with manual expiration

huumn:
I'll do another round of bot review on the changes. 

I was getting hydration and other errors in my browser console during QA but I might have something weird going on with my env. Just thought I should flag that while I'm here.

components/editor/plugins/toolbar/index.js
```diff
@@ -117,12 +118,12 @@ function DropdownMenuItem ({ option, onAction, isActive }) {
...
       title={tooltipText}
       onClick={() => onAction(option)}
-      className={classNames(styles.dropdownExtraItem, isActive && styles.active)}
+      className={classNames(dropdownStyles.dropdownExtraItem, isActive && dropdownStyles.active)}
       onPointerDown={e => e.preventDefault()}
     >
       <span className={styles.dropdownExtraItemLabel}>
         {option.icon}
-        <span className={styles.dropdownExtraItemText}>{option.name}</span>
+        <span className={dropdownStyles.dropdownExtraItemText}>{option.name}</span>
```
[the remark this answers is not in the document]
Soxasora: This is actually an interesting insight

huumn:
Bot review is clean. The only thing that came up in QA so far is that multi-auth on custom domains is broken. I fixed that yesterday when I was messing around with adding `Caddy` to `sndev` for `https` testing locally (I'll publish that PR soon). 

This mutli auth fix should apply cleanly on this branch. The gist is that we were losing the `multiAuth` param and multi-auth cookie list. It's probably a tad overcooked now that I'm looking at it, but it does fix the issue.

I'll proceed with human review now.

```diff
--- a/pages/api/auth/redirect.js
+++ b/pages/api/auth/redirect.js
@@ -6,7 +6,7 @@ import { AUTH_SYNC_NONCE_COOKIE, AUTH_SYNC_NONCE_PARAM, isValidSyncNonce } from
 // https://github.com/vercel/next.js/issues/44482
 // so we use this API route to handle custom domain redirects, keeping support for local dev env.
 export default async function handler (req, res) {
-  const { domain, callbackUrl, signup } = req.query
+  const { domain, callbackUrl, signup, multiAuth } = req.query
   const parsedDomain = parseSafeHost(domain)
   // restrict to ACTIVE custom domains so this endpoint cannot be used as an open redirect.
   const mapping = parsedDomain && await getDomainMapping(parsedDomain.hostname)
@@ -22,6 +22,7 @@ export default async function handler (req, res) {
   if (!isValidSyncNonce(nonce)) {
     const retry = new URL(signup ? '/signup' : '/login', `${SN_MAIN_DOMAIN.protocol}//${canonicalDomain}`)
     if (redirectUri) retry.searchParams.set('callbackUrl', redirectUri)
+    if (multiAuth === 'true') retry.searchParams.set('multiAuth', 'true')
     return res.redirect(302, retry.href)
   }
 
@@ -33,6 +34,7 @@ export default async function handler (req, res) {
   const loginUrl = new URL(signup ? '/signup' : '/login', SN_MAIN_DOMAIN)
   loginUrl.searchParams.set('domain', canonicalDomain)
   loginUrl.searchParams.set('callbackUrl', syncUrl.pathname + syncUrl.search)
+  if (multiAuth === 'true') loginUrl.searchParams.set('multiAuth', 'true')
 
   res.redirect(302, loginUrl.href)
 }
--- a/pages/api/auth/sync.js
+++ b/pages/api/auth/sync.js
@@ -53,7 +53,14 @@ export default async function handler (req, res) {
         return res.status(500).json(sessionTokenResult)
       }
 
-      return res.status(200).json({ status: 'OK', sessionToken: sessionTokenResult.sessionToken })
+      // also return display fields so the proxy can populate the multi-auth list
+      // cookie on the custom domain (the picker reads name/photoId from that list).
+      const user = await models.user.findUnique({
+        where: { id: verificationResult.userId },
+        select: { id: true, name: true, photoId: true }
+      })
+
+      return res.status(200).json({ status: 'OK', sessionToken: sessionTokenResult.sessionToken, user })
     }
 
     if (req.method === 'GET') {
--- a/proxy.js
+++ b/proxy.js
@@ -1,7 +1,7 @@
 import 'urlpattern-polyfill'
 import { NextRequest, NextResponse } from 'next/server'
 import { randomBytes } from 'node:crypto'
-import { SESSION_COOKIE, HTTPS, cookieOptions } from '@/lib/auth'
+import { SESSION_COOKIE, HTTPS, cookieOptions, MULTI_AUTH_JWT, MULTI_AUTH_POINTER, MULTI_AUTH_LIST } from '@/lib/auth'
 import {
   createAuthSyncProof,
   hashSyncNonce,
@@ -103,6 +103,11 @@ async function redirectToAuth (request, searchParams, domain, signup) {
   if (searchParams.has('callbackUrl')) {
     redirectUrl.searchParams.set('callbackUrl', searchParams.get('callbackUrl'))
   }
+  // forward the "add account" intent so the main /login renders the multi-auth UI
+  // instead of short-circuiting to callbackUrl when the user already has a session.
+  if (searchParams.get('multiAuth') === 'true') {
+    redirectUrl.searchParams.set('multiAuth', 'true')
+  }
 
   const response = NextResponse.redirect(redirectUrl)
   response.cookies.set(AUTH_SYNC_NONCE_COOKIE, nonce, nonceCookieOptions)
@@ -156,7 +161,38 @@ async function syncAccount (request, searchParams, domain) {
 
     // one-time use: clear the nonce cookie before issuing the session
     res.cookies.set(AUTH_SYNC_NONCE_COOKIE, '', { ...nonceCookieOptions, maxAge: 0 })
-    res.cookies.set(SESSION_COOKIE, data.sessionToken, cookieOptions())
+
+    const httpOnlyOpts = cookieOptions()
+    const jsOpts = { ...httpOnlyOpts, httpOnly: false }
+
+    res.cookies.set(SESSION_COOKIE, data.sessionToken, httpOnlyOpts)
+
+    // mirror multi-auth state on the custom domain so the account picker also works
+    // here. each per-user JWT is the domain-bound session token minted for THIS domain
+    // (not the main-domain JWT), so when switchSessionCookie swaps it in, the [...nextauth]
+    // jwt callback's domainName/tokenVersion check still passes.
+    const userId = data.user?.id
+    if (userId != null) {
+      res.cookies.set(MULTI_AUTH_JWT(userId), data.sessionToken, httpOnlyOpts)
+      res.cookies.set(MULTI_AUTH_POINTER, String(userId), jsOpts)
+
+      // append the synced user to the existing list (dedupe by id). missing/garbage
+      // existing list is treated as empty so we always end up with at least this user.
+      let list = []
+      const existing = request.cookies.get(MULTI_AUTH_LIST)?.value
+      if (existing) {
+        try {
+          const decoded = Buffer.from(decodeURIComponent(existing), 'base64').toString('utf-8')
+          const parsed = JSON.parse(decoded)
+          if (Array.isArray(parsed)) list = parsed
+        } catch { /* ignore malformed list, start fresh */ }
+      }
+      if (!list.some(({ id }) => id === userId)) {
+        list.push({ id: userId, name: data.user.name, photoId: data.user.photoId })
+      }
+      const encoded = Buffer.from(JSON.stringify(list), 'utf-8').toString('base64')
+      res.cookies.set(MULTI_AUTH_LIST, encoded, jsOpts)
+    }
+
     return res
   } catch (error) {
     console.error('[auth sync] cannot establish auth sync:', error.message)
```

Soxasora:
Great thank you! I've been avoiding multi auth at all costs for the MVP, turns out the solution was straightforward.

I've cleaned the patch a bit. Now there's a single place where multi auth cookies are created (`buildMultiAuthCookies /lib/auth/.js`) that both `setMultiAuthCookies` and middleware use to set multi auth cookies on the response.

Then on the `POST` side I've noticed that we could use the existing transaction to gather the user's info.

It's really clean!

huumn (commented):
After human review and understanding this work with my own brain now, huge props to you. This was a crazy weird/hard problem and you figured it out (I'm not sure I would've persevered if it were left to me). Add another feather to your cap!

That said, while I think this is fine in a private beta and can reason about its security, I think it's worth making some improvements before we merge it while all the context is loaded in our noggins. 

IMO The main risk posed by this feature/PR right now is that it's hard to understand/navigate. So I have a few suggestions. The first one is worth blocking for. Two is LoC neutral and should just improve readability. The last suggestion can be done in a follow up if you want. None of the are subjective - I suspect you'll agree with me.

***The gist: for session syncing, we have all the problems PKCE solves, and probably can't improve on PKCE. So let's do PKCE and be more confident in code naming/organization/structure.***

# 1. we should use PKCE

This is worth blocking merge for. (I've added a full patch for it collapsed at the bottom of this section.)

The biggest win in the simplicity regard is replacing both the `nonce` and `proof` with PKCE: 
1. it gives us the protections of both the nonce and proof
2. it has the code footprint complexity of nonce with a few mods
3. it is an established/well understood pattern
4. tokens do not end up in logs/history/etc

@sir-opti I owe you a super big time apology for dismissing PKCE before. I didn't think we'd end up with almost-PKCE after patching the vulns. I thought PKCE would only grant an improvement on log leaks, but it's that AND better AND simpler than what we came up with. 

I'm sorry we had to learn the hard way that you don't recommend things lightly. I will pause and ask questions next time you recommend something.

<details><summary>PKCE patch</summary>

```diff
diff --git a/lib/constants.js b/lib/constants.js
index fb1cbe70b..6acb37529 100644
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -223,7 +223,7 @@ export const CUSTOM_DOMAINS_DEBUG = Number(process.env.NEXT_PUBLIC_CUSTOM_DOMAIN
 export const DOMAIN_BETA_IDS = [USER_ID.k00b, USER_ID.sox, USER_ID.scoresby, USER_ID.sn]
 // auth sync
 export const VERIFICATION_TOKEN_EXPIRY_MS = 1000 * 60 * 5 // 5 minutes in milliseconds
-export const AUTH_SYNC_TOKEN_TAG = 'auth-sync'
+export const AUTH_SYNC_CODE_TAG = 'auth-sync-code'
 
 export const ZAP_UNDO_DELAY_MS = 5_000
 export const ZAP_DEBOUNCE_MS = 1_000
diff --git a/lib/domains/auth-sync.js b/lib/domains/auth-sync.js
index 2c36f42b0..8a15d4dcc 100644
--- a/lib/domains/auth-sync.js
+++ b/lib/domains/auth-sync.js
@@ -1,18 +1,17 @@
-import { createHash, createHmac, timingSafeEqual } from 'node:crypto'
+import { createHash, timingSafeEqual } from 'node:crypto'
 import { secureCookie } from '@/lib/auth'
 
-export const AUTH_SYNC_PROOF_HEADER = 'x-sn-auth-sync-proof'
-export const AUTH_SYNC_NONCE_COOKIE = secureCookie('sync_nonce')
-export const AUTH_SYNC_NONCE_PARAM = 'sync_nonce'
+// PKCE-shaped custom-domain auth sync.
+// - The verifier lives only in the custom-domain HttpOnly cookie and the server-to-server POST body.
+// - URLs aimed at the main domain only ever carry the SHA-256 challenge.
+export const AUTH_SYNC_VERIFIER_COOKIE = secureCookie('code_verifier')
+export const AUTH_SYNC_CHALLENGE_PARAM = 'code_challenge'
 // 10 minutes is plenty for a login flow and matches the verification token TTL ceiling.
-export const AUTH_SYNC_NONCE_TTL_S = 10 * 60
-// nonces are 32 random bytes hex-encoded
-export const AUTH_SYNC_NONCE_HEX_LENGTH = 64
+export const AUTH_SYNC_VERIFIER_TTL_S = 10 * 60
+// verifier is 32 random bytes hex-encoded; challenge is its SHA-256 hex digest, also 64 hex chars.
+export const AUTH_SYNC_VERIFIER_HEX_LENGTH = 64
 
-const NONCE_HEX_RE = /^[0-9a-f]+$/
-
-const sign = (secret, message) =>
-  createHmac('sha256', secret).update(message).digest('hex')
+const HEX_64_RE = /^[0-9a-f]{64}$/
 
 // length-checked timing-safe string comparison; safely returns false for non-strings.
 export const safeEqual = (received, expected) => {
@@ -21,25 +20,14 @@ export const safeEqual = (received, expected) => {
   return timingSafeEqual(Buffer.from(received), Buffer.from(expected))
 }
 
-export function hashSyncNonce (nonce) {
-  return createHash('sha256').update(nonce).digest('hex')
+export function deriveCodeChallenge (verifier) {
+  return createHash('sha256').update(verifier).digest('hex')
 }
 
-export function isValidSyncNonce (nonce) {
-  return typeof nonce === 'string' &&
-    nonce.length === AUTH_SYNC_NONCE_HEX_LENGTH &&
-    NONCE_HEX_RE.test(nonce)
+export function isValidCodeVerifier (verifier) {
+  return typeof verifier === 'string' && HEX_64_RE.test(verifier)
 }
 
-// HMAC (verificationToken, domainName, nonceHash) keyed by NEXTAUTH_SECRET.
-// Binding the nonce hash here means a leaked verificationToken cannot be exchanged without the matching nonce cookie.
-export function createAuthSyncProof ({ verificationToken, domainName, nonceHash, secret }) {
-  if (!secret) throw new Error('auth sync proof: missing secret')
-  if (!verificationToken || !domainName || !nonceHash) throw new Error('auth sync proof: missing inputs')
-  return sign(secret, `${verificationToken}:${domainName.toLowerCase()}:${nonceHash}`)
-}
-
-export function verifyAuthSyncProof ({ received, verificationToken, domainName, nonceHash, secret }) {
-  if (!secret || !verificationToken || !domainName || !nonceHash) return false
-  return safeEqual(received, sign(secret, `${verificationToken}:${domainName.toLowerCase()}:${nonceHash}`))
+export function isValidCodeChallenge (challenge) {
+  return typeof challenge === 'string' && HEX_64_RE.test(challenge)
 }
diff --git a/pages/api/auth/redirect.js b/pages/api/auth/redirect.js
index 2c6f2867c..7b731cfe5 100644
--- a/pages/api/auth/redirect.js
+++ b/pages/api/auth/redirect.js
@@ -1,6 +1,10 @@
 import { SN_MAIN_DOMAIN, getDomainMapping } from '@/lib/domains'
 import { formatHost, parseSafeHost, safeRedirectPath } from '@/lib/safe-url'
-import { AUTH_SYNC_NONCE_COOKIE, AUTH_SYNC_NONCE_PARAM, isValidSyncNonce } from '@/lib/domains/auth-sync'
+import {
+  AUTH_SYNC_VERIFIER_COOKIE,
+  AUTH_SYNC_CHALLENGE_PARAM,
+  isValidCodeChallenge
+} from '@/lib/domains/auth-sync'
 
 // Next.js cannot redirect to an absolute localhost URL, it gets converted to a relative URL.
 // https://github.com/vercel/next.js/issues/44482
@@ -19,10 +23,12 @@ export default async function handler (req, res) {
   // pair it with signup, which always creates a fresh user.
   const wantsMultiAuth = !signup && multiAuth === 'true'
 
-  // the per-browser nonce is set as a cookie on the custom domain by middleware's redirectToAuth.
-  // if it's missing, we bounce back to the custom domain to mint a fresh one.
-  const nonce = req.cookies[AUTH_SYNC_NONCE_COOKIE]
-  if (!isValidSyncNonce(nonce)) {
+  // PKCE: the verifier cookie is set on the custom domain by middleware's redirectToAuth,
+  // and the SHA-256 challenge travels in the URL. The verifier itself is never echoed here.
+  // Cookie presence is a fail-fast UX gate; the actual security check happens at exchange time.
+  const codeChallenge = req.query[AUTH_SYNC_CHALLENGE_PARAM]
+  const hasVerifierCookie = !!req.cookies[AUTH_SYNC_VERIFIER_COOKIE]
+  if (!hasVerifierCookie || !isValidCodeChallenge(codeChallenge)) {
     const retry = new URL(signup ? '/signup' : '/login', `${SN_MAIN_DOMAIN.protocol}//${canonicalDomain}`)
     if (redirectUri) retry.searchParams.set('callbackUrl', redirectUri)
     if (wantsMultiAuth) retry.searchParams.set('multiAuth', 'true')
@@ -32,7 +38,7 @@ export default async function handler (req, res) {
   const syncUrl = new URL('/api/auth/sync', SN_MAIN_DOMAIN)
   syncUrl.searchParams.set('domain', canonicalDomain)
   syncUrl.searchParams.set('redirectUri', redirectUri)
-  syncUrl.searchParams.set(AUTH_SYNC_NONCE_PARAM, nonce)
+  syncUrl.searchParams.set(AUTH_SYNC_CHALLENGE_PARAM, codeChallenge)
 
   const loginUrl = new URL(signup ? '/signup' : '/login', SN_MAIN_DOMAIN)
   loginUrl.searchParams.set('domain', canonicalDomain)
diff --git a/pages/api/auth/sync.js b/pages/api/auth/sync.js
index 13d0a92ee..64f53a08f 100644
--- a/pages/api/auth/sync.js
+++ b/pages/api/auth/sync.js
@@ -4,41 +4,26 @@ import { encode as encodeJWT, getToken } from 'next-auth/jwt'
 import { validateSchema, customDomainSchema } from '@/lib/validate'
 import { SN_MAIN_DOMAIN } from '@/lib/domains'
 import { formatHost, parseSafeHost, safeRedirectPath } from '@/lib/safe-url'
-import { VERIFICATION_TOKEN_EXPIRY_MS, AUTH_SYNC_TOKEN_TAG } from '@/lib/constants'
+import { VERIFICATION_TOKEN_EXPIRY_MS, AUTH_SYNC_CODE_TAG } from '@/lib/constants'
 import { multiAuthMiddleware } from '@/lib/auth'
 import {
-  verifyAuthSyncProof,
-  hashSyncNonce,
-  isValidSyncNonce,
+  deriveCodeChallenge,
+  isValidCodeChallenge,
+  isValidCodeVerifier,
   safeEqual,
-  AUTH_SYNC_PROOF_HEADER,
-  AUTH_SYNC_NONCE_PARAM
+  AUTH_SYNC_CHALLENGE_PARAM
 } from '@/lib/domains/auth-sync'
 
 export default async function handler (req, res) {
   try {
     if (req.method === 'POST') {
-      const { verificationToken, domainName, nonce } = req.body || {}
+      const { code, domainName, codeVerifier } = req.body || {}
       const parsedDomain = parseSafeHost(domainName)
-      if (!verificationToken || !parsedDomain || !isValidSyncNonce(nonce)) {
-        return res.status(400).json({ status: 'ERROR', reason: 'verification token, domain name and nonce are required' })
+      if (!code || !parsedDomain || !isValidCodeVerifier(codeVerifier)) {
+        return res.status(400).json({ status: 'ERROR', reason: 'code, domain name and code verifier are required' })
       }
 
-      const nonceHash = hashSyncNonce(nonce)
-
-      // verify the request came from our own middleware via proof header bound to the nonce
-      const validProof = verifyAuthSyncProof({
-        received: req.headers[AUTH_SYNC_PROOF_HEADER],
-        verificationToken,
-        domainName: parsedDomain.hostname,
-        nonceHash,
-        secret: process.env.NEXTAUTH_SECRET
-      })
-      if (!validProof) {
-        return res.status(401).json({ status: 'ERROR', reason: 'invalid proof' })
-      }
-
-      const verificationResult = await consumeVerificationToken(verificationToken, parsedDomain.hostname, nonceHash)
+      const verificationResult = await consumeVerificationToken(code, parsedDomain.hostname, codeVerifier)
       if (verificationResult.status === 'ERROR') {
         return res.status(400).json(verificationResult)
       }
@@ -88,15 +73,15 @@ export default async function handler (req, res) {
         return handleNoSession(res, canonicalDomain, redirectUri)
       }
 
-      // CSRF gate: the matching nonce cookie is set on the custom domain by the middleware
-      // when it intercepts /login or /signup. If it's missing, we bounce back to the custom domain to mint a fresh one.
-      const nonce = req.query[AUTH_SYNC_NONCE_PARAM]
-      if (!isValidSyncNonce(nonce)) {
+      // PKCE: bind the verification token to the SHA-256 challenge supplied by the custom
+      // domain. The matching verifier is held only in the custom-domain HttpOnly cookie and
+      // is presented by the proxy at the POST exchange.
+      const codeChallenge = req.query[AUTH_SYNC_CHALLENGE_PARAM]
+      if (!isValidCodeChallenge(codeChallenge)) {
         return handleNoSession(res, canonicalDomain, redirectUri)
       }
-      const nonceHash = hashSyncNonce(nonce)
 
-      const newVerificationToken = await createVerificationToken(sessionToken, domainValidation.domainId, nonceHash)
+      const newVerificationToken = await createVerificationToken(sessionToken, domainValidation.domainId, codeChallenge)
       if (newVerificationToken.status === 'ERROR') {
         return res.status(500).json(newVerificationToken)
       }
@@ -130,7 +115,7 @@ async function checkDomainValidity (receivedDomain) {
 function handleNoSession (res, domainName, redirectUri, signup = false) {
   // bounce to /login (or /signup) on the *custom* domain, not the main one,
   // so the request passes through the custom-domain middleware
-  // which mints a fresh nonce cookie and re-enters the flow.
+  // which mints a fresh verifier cookie and re-enters the flow.
   const customDomainLoginUrl = new URL(
     signup ? '/signup' : '/login',
     `${SN_MAIN_DOMAIN.protocol}//${domainName}`
@@ -142,12 +127,12 @@ function handleNoSession (res, domainName, redirectUri, signup = false) {
   res.redirect(302, customDomainLoginUrl.href)
 }
 
-async function createVerificationToken (token, domainId, nonceHash) {
+async function createVerificationToken (sessionToken, domainId, codeChallenge) {
   try {
     const verificationToken = await models.verificationToken.create({
       data: {
-        // bind the token to the user, the domain, and the per-browser nonce hash
-        identifier: `${AUTH_SYNC_TOKEN_TAG}:${token.id}:${domainId}:${nonceHash}`,
+        // bind the token to the user, the domain, and the SHA-256 challenge.
+        identifier: `${AUTH_SYNC_CODE_TAG}:${sessionToken.id}:${domainId}:${codeChallenge}`,
         token: randomBytes(32).toString('hex'),
         expires: new Date(Date.now() + VERIFICATION_TOKEN_EXPIRY_MS)
       }
@@ -158,11 +143,11 @@ async function createVerificationToken (token, domainId, nonceHash) {
   }
 }
 
-function redirectToDomain (res, domain, verificationToken, redirectUri) {
+function redirectToDomain (res, domain, code, redirectUri) {
   try {
     const target = new URL(`${SN_MAIN_DOMAIN.protocol}//${formatHost(domain)}`)
 
-    target.searchParams.set('sync_token', verificationToken)
+    target.searchParams.set('code', code)
     target.searchParams.set('redirectUri', redirectUri)
 
     return res.redirect(302, target.href)
@@ -172,7 +157,7 @@ function redirectToDomain (res, domain, verificationToken, redirectUri) {
   }
 }
 
-async function consumeVerificationToken (verificationToken, expectedDomainName, expectedNonceHash) {
+async function consumeVerificationToken (code, expectedDomainName, codeVerifier) {
   try {
     await validateSchema(customDomainSchema, { domainName: expectedDomainName })
 
@@ -189,22 +174,22 @@ async function consumeVerificationToken (verificationToken, expectedDomainName,
       if (!domain) throw new Error('domain not allowed')
 
       const token = await tx.verificationToken.findUnique({
-        where: { token: verificationToken }
+        where: { token: code }
       })
       if (!token || token.expires <= new Date()) throw new Error('invalid verification token')
 
       const identifier = token.identifier || ''
-      const [tag, userIdStr, domainIdStr, storedNonceHash] = identifier.split(':')
-      if (tag !== AUTH_SYNC_TOKEN_TAG || Number(domainIdStr) !== domain.id) {
+      const [tag, userIdStr, domainIdStr, storedChallenge] = identifier.split(':')
+      if (tag !== AUTH_SYNC_CODE_TAG || Number(domainIdStr) !== domain.id) {
         throw new Error('invalid verification token domain')
       }
 
-      // nonce-hash binding: the cookie that travelled back with this request must
-      // hash to the value we stored at mint time.
-      // mismatch throws before the delete so the transaction rolls back and the token survives
+      // PKCE check: the verifier presented at exchange time must hash to the challenge
+      // we recorded at mint time. Mismatch throws BEFORE the delete so the token survives
       // for a legitimate retry (e.g. cookie overwritten by a concurrent tab flow).
-      if (!safeEqual(storedNonceHash, expectedNonceHash)) {
-        throw new Error('invalid verification token nonce')
+      const expectedChallenge = deriveCodeChallenge(codeVerifier)
+      if (!safeEqual(storedChallenge, expectedChallenge)) {
+        throw new Error('invalid code verifier')
       }
 
       const userId = Number(userIdStr)
diff --git a/proxy.js b/proxy.js
index 5385eda72..9b92c2b3a 100644
--- a/proxy.js
+++ b/proxy.js
@@ -3,22 +3,21 @@ import { NextRequest, NextResponse } from 'next/server'
 import { randomBytes } from 'node:crypto'
 import { SESSION_COOKIE, HTTPS, cookieOptions, buildMultiAuthCookies, MULTI_AUTH_LIST } from '@/lib/auth'
 import {
-  createAuthSyncProof,
-  hashSyncNonce,
-  AUTH_SYNC_PROOF_HEADER,
-  AUTH_SYNC_NONCE_COOKIE,
-  AUTH_SYNC_NONCE_TTL_S,
-  AUTH_SYNC_NONCE_HEX_LENGTH
+  deriveCodeChallenge,
+  AUTH_SYNC_VERIFIER_COOKIE,
+  AUTH_SYNC_CHALLENGE_PARAM,
+  AUTH_SYNC_VERIFIER_TTL_S,
+  AUTH_SYNC_VERIFIER_HEX_LENGTH
 } from '@/lib/domains/auth-sync'
 import { getDomainMapping, createDomainsDebugLogger, SN_MAIN_DOMAIN } from '@/lib/domains'
 import { parseSafeHost, safeRedirectPath } from '@/lib/safe-url'
 
-const nonceCookieOptions = {
+const verifierCookieOptions = {
   httpOnly: true,
   sameSite: 'lax',
   secure: HTTPS,
   path: '/',
-  maxAge: AUTH_SYNC_NONCE_TTL_S
+  maxAge: AUTH_SYNC_VERIFIER_TTL_S
 }
 
 const referrerPattern = new URLPattern({ pathname: ':pathname(*)/r/:referrer([\\w_]+)' })
@@ -63,7 +62,7 @@ async function customDomainMiddleware (request, domain, subName) {
     const signup = pathname.startsWith('/signup')
     return redirectToAuth(request, searchParams, domain, signup)
   }
-  if (searchParams.has('sync_token')) return syncAccount(request, searchParams, domain)
+  if (searchParams.has('code')) return syncAccount(request, searchParams, domain)
 
   // clean up the pathname from any subname
   if (pathname.startsWith('/~')) {
@@ -92,13 +91,15 @@ async function customDomainMiddleware (request, domain, subName) {
 }
 
 async function redirectToAuth (request, searchParams, domain, signup) {
-  // CSRF: mint a per-browser nonce here,
-  // then redirect through /api/auth/redirect so the main domain receives and validates the nonce-bound flow.
-  const nonce = randomBytes(AUTH_SYNC_NONCE_HEX_LENGTH / 2).toString('hex')
+  // PKCE: mint a per-browser verifier here, store it in an HttpOnly cookie on the custom origin,
+  // and forward only its SHA-256 challenge through the main-domain redirect chain.
+  const codeVerifier = randomBytes(AUTH_SYNC_VERIFIER_HEX_LENGTH / 2).toString('hex')
+  const codeChallenge = deriveCodeChallenge(codeVerifier)
 
   // /api/auth/redirect is a workaround to handle localhost redirects.
   const redirectUrl = new URL('/api/auth/redirect', request.url)
   redirectUrl.searchParams.set('domain', domain)
+  redirectUrl.searchParams.set(AUTH_SYNC_CHALLENGE_PARAM, codeChallenge)
   if (signup) redirectUrl.searchParams.set('signup', '1')
   if (searchParams.has('callbackUrl')) {
     redirectUrl.searchParams.set('callbackUrl', searchParams.get('callbackUrl'))
@@ -111,38 +112,29 @@ async function redirectToAuth (request, searchParams, domain, signup) {
   }
 
   const response = NextResponse.redirect(redirectUrl)
-  response.cookies.set(AUTH_SYNC_NONCE_COOKIE, nonce, nonceCookieOptions)
+  response.cookies.set(AUTH_SYNC_VERIFIER_COOKIE, codeVerifier, verifierCookieOptions)
   return response
 }
 
 async function syncAccount (request, searchParams, domain) {
-  const token = searchParams.get('sync_token')
+  const code = searchParams.get('code')
   const rawRedirectUri = searchParams.get('redirectUri')
   const redirectUri = safeRedirectPath(rawRedirectUri, domain)
   const res = NextResponse.redirect(new URL(redirectUri, request.url))
-  // HMAC inputs must match the verifier, which keys on the bare hostname.
+  // domainName must be the bare hostname; the verifier handler enforces a hostname match too.
   const { hostname: domainName } = parseSafeHost(domain)
 
-  // require the per-browser nonce we set during redirectToAuth.
-  // this proves this browser is the same one that started the login flow.
-  const nonce = request.cookies.get(AUTH_SYNC_NONCE_COOKIE)?.value
-  if (!nonce) {
+  // require the per-browser verifier we set during redirectToAuth.
+  // this proves the same browser that started the login flow is the one redeeming the code.
+  const codeVerifier = request.cookies.get(AUTH_SYNC_VERIFIER_COOKIE)?.value
+  if (!codeVerifier) {
     return NextResponse.redirect(new URL('/error', request.url))
   }
-  const nonceHash = hashSyncNonce(nonce)
 
   try {
-    const proof = createAuthSyncProof({
-      verificationToken: token,
-      domainName,
-      nonceHash,
-      secret: process.env.NEXTAUTH_SECRET
-    })
-
-    const body = JSON.stringify({ verificationToken: token, domainName, nonce })
+    const body = JSON.stringify({ code, domainName, codeVerifier })
     const fetchHeaders = new Headers()
     fetchHeaders.set('Content-Type', 'application/json')
-    fetchHeaders.set(AUTH_SYNC_PROOF_HEADER, proof)
 
     const response = await fetch(`${SN_MAIN_DOMAIN.origin}/api/auth/sync`, {
       method: 'POST',
@@ -160,8 +152,8 @@ async function syncAccount (request, searchParams, domain) {
       throw new Error(data.reason)
     }
 
-    // one-time use: clear the nonce cookie before issuing the session
-    res.cookies.set(AUTH_SYNC_NONCE_COOKIE, '', { ...nonceCookieOptions, maxAge: 0 })
+    // one-time use: clear the verifier cookie before issuing the session
+    res.cookies.set(AUTH_SYNC_VERIFIER_COOKIE, '', { ...verifierCookieOptions, maxAge: 0 })
     res.cookies.set(SESSION_COOKIE, data.sessionToken, cookieOptions())
 
     // mirror multi-auth state on the custom domain so the account picker also works here.
```
</details>

# 2. our naming conventions are confusing

This PR works by chaining http requests and accumulating side effects in cookies/query params. That's fine - it's the way stuff on the web interacts across domain boundaries. 

But, now that we have something that works the way we want, I think we should make that structure clearer through better endpoint/function names and possibly other code structure changes.

This is my rough ascii flow chart of auth after PKCE:

```txt
Custom-domain auth-sync state machine (PKCE):
     -> GET pizza.com/{login,signup}: verifier minted (HttpOnly cookie)
     -> 302 pizza.com/api/auth/redirect?codeChallenge=sha256(verifier): bridge cuz nextjs workaround
     -> 302 stacker.news/{login,signup}?codeChallenge=sha256(verifier)&domain=pizza.com: user picks/establishes session
     -> GET stacker.news/api/auth/sync?codeChallenge=sha256(verifier)&domain=pizza.com: user session gets linked to a random code and the codeChallenge=sha256(verifier) in DB so that the bearer of HttpOnly verifier cookie + code receives session later
     -> 302 pizza.com?code=code: requests session from stacker.news
        -> POST stacker.news/api/auth/sync { code, verifier, domain }: creates and returns session after verifying codeChallenge = sha256(verifier) && code = code && domain = domain
     -> 302 pizza.com: session set, mission accomplished
```

I'm not sure what names we should pick (LLMs will probably do a better job), but one rough idea:

- `/api/auth/redirect` -> `/api/auth/custom-domains/start`
- `GET` `/api/auth/sync` -> `/api/auth/custom-domains/session-to-code`
- `POST` `/api/auth/sync` -> `/api/auth/custom-domains/code-to-session`

Biggest win is breaking apart `/api/auth/sync` imo. That one confused me a lot.

# 3. make a dedicated table for domain auth codes

Overloading `VerificationRequests` was probably the right choice while we were experimenting/iterating (we might've resisted modifying migrations), but I think it's worth breaking apart our colon delimited state string into a dedicated table `DomainAuthCode` with columns for `code`, `codeChallenge`, `domainId`, `userId`, etc. Or, something like it.

Soxasora:
Mea culpa, I dismissed any kind of safety advancement in this PR, and implemented patches to an already-broken architecture for the sake of private beta. Windows Vista-style development.

When we briefly talked about pure PKCE, I just couldn't focus on the real benefits of it because it "wouldn't fix domain hijacking". LLMs then added to the confusion, and the PR basically became slop.

---

I agree with everything you said, PKCE, explicit auth flow and better naming conventions (and also consistent code styles).
Now since LLMs are really great at assaulting the thinking process (and Claude is a really great seller, both at least in my case), this rewrite is LLM-free.

QA went great, and the bugs that I've found during the QA process were a breeze to fix. I missed coding exclusively by hand.

I'll update the documentation, review and cleanup the rewrite asap, in the meantime it's really readable as-is.

edit: I didn't apply your patch because of the rewrite, I'll check it out now to see if I need to integrate something!

pages/api/auth/domains/verify.js
```diff
@@ -0,0 +1,93 @@
...
+
+  try {
+    const { code, redirectUri: rawRedirectUri } = req.query
+    if (!code || !rawRedirectUri) {
+      return res.status(400).json({ status: 'ERROR', reason: 'code and redirectUri are required' })
+    }
+
+    const domain = req.headers.host
+    const redirectUri = safeRedirectPath(rawRedirectUri, domain)
+    const { hostname: domainName } = parseSafeHost(domain)
```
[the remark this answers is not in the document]
Soxasora: expected worse

Soxasora:
Here's a mermaid diagram for the new auth flow:

```mermaid
sequenceDiagram
  autonumber
  participant B as Browser
  participant CD as Custom Domain
  participant P as proxy.js
  participant Begin as /api/auth/domains/begin
  participant MD as Main Domain Login
  participant Code as /api/auth/domains/code
  participant Verify as /api/auth/domains/verify
  participant Token as /api/auth/domains/token
  participant DB as DB

  B->>CD: GET /login or /signup
  CD->>P: customDomainMiddleware

  Note over P: PKCE: create verifier<br/>challenge = sha256(verifier)
  P-->>B: Set httpOnly verifier cookie on custom domain
  P-->>B: 302 /api/auth/domains/begin?domain&challenge&callbackUrl

  B->>Begin: GET begin with challenge
  Begin->>Begin: Read verifier cookie
  Begin->>Begin: Verify sha256(verifier) == challenge
  Begin->>Begin: Validate ACTIVE domain + safe redirectUri
  Begin-->>B: 302 main /login?callbackUrl=/api/auth/domains/code...

  B->>MD: Authenticate on main domain
  MD-->>B: Set main-domain session cookie
  MD-->>B: 302 /api/auth/domains/code?domain&challenge&redirectUri

  B->>Code: GET code on main domain
  Code->>Code: Require main-domain session
  Code->>DB: Create one-time code tied to userId, domainId, challenge
  DB-->>Code: code, expiresAt
  Code-->>B: 302 custom /api/auth/domains/verify?code&redirectUri

  B->>Verify: GET verify on custom domain
  Verify->>Verify: Read verifier cookie
  Verify->>Token: POST code, domainName, verifier

  Token->>Token: Derive challenge = sha256(verifier)
  Token->>DB: Lock ACTIVE Domain row
  Token->>DB: Find unexpired code
  Token->>Token: Check code domain + stored challenge
  Token->>DB: Delete one-time code
  Token-->>Verify: Return domain-bound JWT

  Verify-->>B: Set custom-domain session cookie
  Verify-->>B: 302 redirectUri
```

huumn:
It reads really nicely now. It's also nice being able to verify "this does PKCE right" and conclude "we get PKCE things" and breathe a sigh of relief. I'll do a final QA then I think we can merge.

pages/api/auth/domains/begin.js
```diff
@@ -0,0 +1,59 @@
...
+ * responsible for redirecting to main domain /login or /signup retaining the auth query params.
+ *
+ * swaps the original callbackUrl for /api/auth/domains/code with challenge and redirectUri (the original callbackUrl).
+ *
+ * Next.js cannot redirect to an absolute localhost URL, it gets converted to a relative URL.
+ * https://github.com/vercel/next.js/issues/44482
+ *
+ * visited on the custom domain after the user clicks "login"/"signup".
+ */
+export default async function handler (req, res) {
```
Soxasora, line 16: maybe `/api/auth/domains/start` sounds better

huumn: begin is fine!

huumn (approved):
QA is clean. 

The `log in with @nym` button is a bit wonky on signup, but blahblah private beta and we can iterate on it.

Soxasora:
> The log in with @nym button is a bit wonky on signup

It was indeed a spur of the moment thing, previously it was just a one-click login via the navbar. I'll make sure it's polished before going public 🫡

PR #2919, opened 2026-04-16
formatting: fix footnote-ref superscript

sir-opti:
## Description

After 786d8b0d, footnote refs were no longer superscript and I only just noticed this now. This change treats `.sn-footnote-ref` as `.sn-text__superscript`

## Screenshots

| Before | After |
| --- | --- |
| before | after |

## Checklist

Yes

7

Yes

No

It found me the commit that caused it so that I didn't have to git-blame the tree. Yes, I'm lazy af.

### DIFF
--- a/styles/text.scss
+++ b/styles/text.scss
@@ -870,7 +870,7 @@ span + .sn-link {
   all: unset;
 }
 
-.sn-text__superscript {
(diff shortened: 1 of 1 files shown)


Soxasora (approved):
Nice and quick fix, thank you @sir-opti!

PR #2920, opened 2026-04-16
deps: bump some packages

sir-opti:
## Description

Some trivial bumps that take away complaints but do not have anything on a live path for us:

- `mathjs` from `13.2.0` to `15.2.0`
- `dompurify` from `3.3.2` to `3.4.0`
- `lodash/lodash-es` from `4.17.23` to `4.18.1`
- `nodemailer` from `8.0.4` to `8.0.5`

## Checklist

Yes

`6.9420` - ran through dev env, analyzed `worker/trust.js` for impact of the major bump of `mathjs` against their [release notes](https://raw.githubusercontent.com/josdejong/mathjs/refs/heads/develop/HISTORY.md) - didn't find any impact.

No

Not this time.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,7 +62,7 @@
         "linkedom": "^0.18.12",
         "ln-service": "^57.22.0",
         "macaroon": "^3.0.4",
-        "mathjs": "^13.2.0",
+        "mathjs": "^15.2.0",
         "mdast-util-find-and-replace": "^3.0.1",
         "mdast-util-from-markdown": "^2.0.1",
         "mdast-util-gfm": "^3.0.0",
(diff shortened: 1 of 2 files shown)


sir-opti:
Can someone with the nerves of steel through allowing socket security to "act on their behalf" check what that `-5` supply chain security score is about for `mathjs`?

Soxasora:
Here's some infos: https://socket.dev/npm/package/mathjs/alerts/15.2.0

I don't like the socket security report here, everything requires a login. Even when data is already available publicly.

sir-opti:
> Here's some infos

Oh thanks! I'm going to run this through compare because this looks really bad

sir-opti:
False positive. Insecure comment 🤣🤣🤣🤣🤣

```
# grep fetch node_modules/mathjs/lib/cjs/function/matrix/subset.js 
   *    indices to fetch or set. `subset` uses the cartesian product of
```

full diff between 13.2.0 and 15.2.0 taken from npm (gz'd for GH file policy): [subset.js.diff.gz](https://github.com/user-attachments/files/26787900/subset.js.diff.gz)

huumn:
oh goodie! all that's left of `npm audit` is upgrading `aws-sdk` to v3 which ain't too bad.

PR #2930, opened 2026-04-21
fix: untrack new comments only on actual unmount or interaction

### LINKED ISSUE #2927: live comments indicator/navigator regressions
prompted by https://stacker.news/items/1475204
 
I, too, have encountered some unexpected behavior with live comments navigation. i'll try to begin keeping a log of issues I find and edit this issue.

- indicator disappears on zap
    - https://stacker.news/items/1475204
    - cache issue?

Soxasora:
## Description

Fixes #2927 
New comment outlining and tracking happens within a `useEffect` that depends on `lastCommentAt` and `meCommentsViewedAt` of the `root` item.

Live comments would inject a new comment updating the very same fields the effect depends on, triggering the effect cleanup that untracks new comments. The same class of changes were likely happening with any other cache mod to `root`, like a zap.

## Screenshots

https://github.com/user-attachments/assets/256ce1ba-c950-4536-ab8c-569dda6e10a9

## Additional Context

This didn't happen in React 18/Apollo Client 3. A theory can be that React 19 has different render scheduling/transitions.

Zap -> navigator clear cannot be reproduced in dev for some reason. Tracing `untrackNewComment` in Chrome DevTools showed that the `useEffect` cleanup, that untracks new comments, runs also after a zap or live comment.

## Checklist

yes

5, could only verify the live comments bug, but it's in the same class/behavior of the zap->untrack bug.

no

---

> [!NOTE]
> **Medium Risk**
> Adjusts React effect timing around new-comment tracking/untracking; mistakes could cause missed outlines, incorrect counts, or sticky navigator state on item pages.
> 
> **Overview**
> Fixes premature removal of “new comment” tracking/outline by **tracking each comment at most once** and **only untracking on real unmount**.
> 
> `Comment` now keeps a `didTrackRef` flag to avoid re-registering on re-renders/cache updates and moves the `untrackNewComment` call into an unmount-only cleanup. `use-comments-navigator` makes `trackNewComment` synchronous and returns whether insertion occurred, enabling immediate, reliable untracking on user interaction.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 429d08b20f590001229d5050f6a80a95a2a719da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


huumn:
bot review claims to have found a regression:

This looks like a regression from the new `didTrackRef` tracking lifetime. The fix avoids dependency cleanup untracking live comments too early, but now marks the comment as tracked before `trackNewComment` actually inserts it on the next animation frame.

If the user dismisses the outline before that RAF runs, the pending insert can still add the comment back after it was untracked, and unmount cleanup may skip it because `didTrackRef` was reset to false.

Suggested patch: make `trackNewComment` register synchronously and return whether the ref was actually tracked, then set `didTrackRef.current` from that return value.

Soxasora:
We don't even need the RAF for tracking new comments, I remember implementing it just to *`micro-optimize`* and follow the same pattern of `throttleCountUpdate` but it's pretty much useless, it's useful only to update the count without causing too many re-renders due to the context.

Fixed! Thank you!!

PR #2937, opened 2026-04-26
patch DOM insertBefore and removeChild, workaround for DOM translators

### LINKED ISSUE #2936: DOM Translators desync React's reconciler

### Describe the bug

DOM Translators rewrite the DOM by wrapping translatable text nodes in `<font>` tags, replacing siblings.
When a comment is submitted/edited, React tries to unmount the reply editor, buttons, etc. and calls `parent.removeChild`. This action fails because the DOM translator replaced the node that React is trying to remove, throwing `NotFoundError`.

### Screenshots

https://github.com/user-attachments/assets/8deb6cd1-c6d6-4612-a023-6896ff903520

### Steps To Reproduce

1. Translate the page with a DOM Translator, e.g. Google Chrome's Translate feature
2. Reply to a comment
3. `Node.removeChild` throws `NotFoundError`

### Expected behavior

The reply form is correctly removed from the DOM without throwing an error.

### Logs

_No response_

### Device information

- Browser: Google Chrome

### Additional context

https://stacker.news/items/1479027

Soxasora:
## Description

Fixes #2936 with a workaround.
This PR patches `Node.removeChild` and `Node.insertBefore` with a widely-used community patch that silences the `DOMException` thrown by them.
A `MutationObserver` is used to track whether the user is using a DOM Translator, this way the patch is applied only to DOM Translation users.

## Screenshots

https://github.com/user-attachments/assets/40081ed7-99c2-4dd8-9b37-136f502f7848

## Additional Context

Of course it's not a good idea to silence `DOMException`s thrown by React, this bug appeared in 2017 for the first time and has never been fixed since: https://github.com/facebook/react/issues/11538

The original community patch involved silencing these exception globally, but I personally think it's better to be conservative and silence these exceptions only for DOM Translation users, as the tools they use already mess up the DOM, a DOM drift won't be the end of the world. Even if there's a DOM drift, React will fix it on the next render.

#### Other methods

- Disabling translation site-wide
- Carefully disseminate `translate=no` on every impacted node, making sure that translation will still work

## Checklist

Yes

6, tested only with Google Chrome's Translate feature.

n/a

n/a

Observer suggestion

---

> [!NOTE]
> **Medium Risk**
> Patches global DOM `Node` prototype methods, which can have wide-reaching side effects if the translation detection is wrong or interacts unexpectedly with other code, though it is gated to only activate while translation appears enabled.
> 
> **Overview**
> Adds `patchDOMTranslations()` (new `lib/patch-dom-translate.js`) that detects DOM translation via `<html>` classes and uses a `MutationObserver` to track when translation is active.
> 
> When active, it monkey-patches `Node.removeChild` and `Node.insertBefore` to suppress/adjust operations that would otherwise throw `NotFoundError` due to translator-modified DOM, and wires this patch to run globally from `pages/_app.js`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 502e817339bdd23ac37460e1b7f1ae46ce9a9c73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/patch-dom-translate.js
```diff
@@ -0,0 +1,59 @@
...
+let translateActive = false
+
+function detectDOMTranslate () {
+  const html = document.documentElement
+  if (!html) return false
+
+  // classes used by DOM translators
+  if (html.classList.contains('translated-ltr')) return true
+  if (html.classList.contains('translated-rtl')) return true
+  if (html.classList.contains('translated')) return true
```
Soxasora, line 12: *only Chromium-based DOM translations, needs to be tested on Safari and Firefox.

lib/patch-dom-translate.js
```diff
@@ -0,0 +1,59 @@
...
+
+  const originalInsertBefore = Node.prototype.insertBefore
+  Node.prototype.insertBefore = function (newNode, referenceNode) {
+    if (referenceNode && referenceNode.parentNode !== this) {
+      if (translateActive) {
+        return newNode
+      }
+    }
+    return originalInsertBefore.call(this, newNode, referenceNode)
+  }
```
Soxasora, line 51: We could technically fallback to `appendChild` here instead of creating phantom DOM nodes.

PR #2944, opened 2026-05-06
sndev https and secure cookies

huumn:
This was motivated by wallets not working on custom domains in `sndev` (because `window.crypto.subtle` is unavailable in insecure contexts). While we may be able to just use `*.localhost` custom domains in dev to achieve the same (depending on the browser), it was a small lift toward achieving more fidelity.

It's four commits:

- adds a Caddy server to `sndev` that serves self-signed certs
    - also adds `sndev domains trust` to import/trust the certs
- auth: derive cookie Secure flag per request
    - this is a bit invasive, but it allows use of secure cookies in dev 
    - before this was gated on `process.env.NODE_ENV === 'production'`
- wallets: fail key init gracefully on insecure-context dev domains
    - attempting to use wallets on custom domain without https would throw
- wallets: console.warn when local key fails to decrypt remote
    - this was a side effect of #2904 (console errors in `useEffect`) and is only an annoyance in dev

TODO
- still needs human touch (claude is a yapper)

---

> [!NOTE]
> **Medium Risk**
> Touches authentication/cookie issuance logic and adds a TLS-terminating proxy path in dev, which can affect session persistence and cross-domain auth if misconfigured (e.g., missing `X-Forwarded-Proto`).
> 
> **Overview**
> Adds an optional `domains-caddy` dev profile that runs a local Caddy TLS terminator (ports 80/443) for custom domains, plus `sndev domains trust` to install Caddy’s local CA; dnsmasq is updated to wildcard `*.sndev` and `*.test`, and docs/env defaults are adjusted accordingly.
> 
> Refactors cookie handling so the `Secure` attribute is derived **per request** (via `X-Forwarded-Proto`/protocol) instead of being gated only by `NODE_ENV`, and threads this through NextAuth, multi-auth cookies, domain-auth verification, referrer middleware, and cookie clearing paths.
> 
> Improves wallet behavior on insecure origins by detecting non-secure contexts (missing `crypto.subtle`/secure context) and failing key init gracefully with a clearer error message; also downgrades a noisy decrypt failure log from `console.error` to `console.warn`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5385814dbf266204d817b7c73f137154f946136. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/use-cookie.js
+++ b/components/use-cookie.js
@@ -41,7 +41,9 @@ export default function useCookie (name) {
   }, [name])
 
   const remove = useCallback(() => {
-    document.cookie = cookie.serialize(name, '', { expires: 0, maxAge: 0 })
+    // path must match the SET (cookieOptions defaults path to '/'); without it
+    // the browser uses the current URL's directory and the clear silently misses.
+    document.cookie = cookie.serialize(name, '', cookieOptions({ httpOnly: false, maxAge: 0 }))
     setValue(null)
   }, [name])
 
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -1,4 +1,3 @@
-import { datePivot } from '@/lib/time'
 import * as cookie from 'cookie'
 import { NodeNextRequest } from 'next/dist/server/base-http/node'
 import { encode as encodeJWT, decode as decodeJWT } from 'next-auth/jwt'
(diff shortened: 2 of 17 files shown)


components/use-cookie.js
```diff
@@ -41,7 +41,9 @@ export default function useCookie (name) {
   }, [name])
 
   const remove = useCallback(() => {
-    document.cookie = cookie.serialize(name, '', { expires: 0, maxAge: 0 })
```
huumn, line 44: this was an unexercised bug (not used anywhere)

lib/auth.js
```diff
@@ -23,20 +52,25 @@ export const MULTI_AUTH_JWT = id => secureCookie(`multi_auth.${id}`)
...
   httpOnly: true,
   sameSite: 'lax',
-  // default expiration for next-auth JWTs is in 30 days
-  expires: datePivot(new Date(), { days: 30 }),
-  maxAge: 2592000, // 30 days in seconds
-  ...args
+  maxAge,
+  expires: maxAge > 0 ? new Date(Date.now() + maxAge * 1000) : undefined,
+  ...rest
 })
```
huumn, line 68: most of the blast radius in this PR are having more places reuse this centralized `cookieOptions` function. unless explicitly overridden:

- secure in prod
    - varies in dev based on request allowing us to exercise secure cookies
        - for multiauth to work, we still need `SN_MAIN_DOMAIN` in dev to use `http`. otherwise we'd have to control nextauth's cookie names ... not worth it right now
- path = /
- httpOnly
- sameSite = lax
- maxAge = 30 days

---------

Most places were doing the right thing, but

- some unused code wasn't setting `path`
- referrer cookies weren't setting `httpOnly` despite not needing client visibility

sir-opti:
Two minor things post-testing:

1. Could it be useful generally to put Caddy in a profile so that I can turn it off? I already have integrated my dev env into an LB with mTLS so I don't need it (nor do I really want to because I'd like to retain my overall mTLS solution)
2. `sn_referrer` only started carrying `secure` after I complied with `x-forwarded-proto` header injection on the LB - may be worth a comment somewhere.

Otherwise works perfectly.

PR #2948, opened 2026-05-09
bump AWS SDK to v3 and integrate it

sir-opti:
## Description

Bumps and integrates AWS SDK from v2 to v3 across the 3 products used:

### S3

- Adaptations in our wrapper:
  - removal of global AWS.config
  - promise interface instead of callbacks
  - deprecated options: apiVersion
  - renamed options: s3ForcePathStyle to forcePathStyle
- Backports:
  - Strip path from configured endpoint before passing to S3Client
    - NOTE: this is an actual regression in the AWS SDK vs v2 that affected us: where previously paths were ignored on a given endpoint, now they aren't. Decided that this was a better place than changing the way we structure the env variables.
  - Import the new createPresignedPost as s3CreatePresignedPost to prevent name collision.
- Enhancements / hardening:
  - warn when development environment didn't specify endpoint 
  - cache client
  - add a radix when parsing the key

### ACM

- 100% straightforward adaptation to the changed call structure

### ELB

- Adaptations:
  - removal of global AWS.config
  - promise interface instead of callbacks
  - Change the mock from class-method (as was v2) to .send(command)
- Enhancements:
  - cache client (in module scope)

## Checklist

Yes. When combined with the last commit that removed the v2 sdk dependency, each commit can be reverted individually too. 

7-8. The only thing I am unsure about are regressions in the custom domain validation flow - it works for me on dev, but I am not sure if that means that it works everywhere. ELB ought to be the highest risk here.

Yes. Blast-radius / analysis, and initial implementation of the mock conversion.


huumn:
No findings in bot(!) or human review. QA of both media and custom domains had no problems as well.

PR #2949, opened 2026-05-10
fix: sticky header horizontal shift caused by scrollbar width

### LINKED ISSUE #2346: Sticky header slightly shifted

### Describe the bug

The elements in the sticky header are slightly shifted compared to the elements in the non-sticky header, see video.

This has been annoying me for a while.

### Screenshots

https://github.com/user-attachments/assets/e3faa903-3f9e-4c13-873c-f0c9ee4b8098

### Steps To Reproduce

1. Scroll down on desktop

### Expected behavior

Elements in sticky header are not shifted

### Logs

_No response_

### Device information

_No response_

### Additional context

I wish we could just use `position: sticky` but we can't because our non-sticky header is not the top element and sticky is only relative to parents.

CloudWaddie:
## Summary
- Fixes #2346
- The `.sticky` header used `width: 100vw` which includes the scrollbar width, causing a slight horizontal shift compared to the non-sticky header
- Replaced with `left: 0; right: 0` which correctly excludes scrollbar width for `position: fixed` elements

### DIFF
--- a/components/header.module.css
+++ b/components/header.module.css
@@ -114,7 +114,8 @@
 .sticky {
     position: fixed;
     border-bottom: 1px solid var(--theme-toolbarActive);
-    width: 100vw;
+    left: 0;
+    right: 0;
     top: -1px;
     padding-top: 1px;
     background-color: var(--bs-body-bg);


sir-opti:
This looks ok for QA

PR #2951, opened 2026-05-10
fix: allow date range filtering for all item sort types

### LINKED ISSUE #2433: Selecting items by date range in API requires sort=top or sort=user

### Describe the bug

Trying to retrieve items from a date range using `when, to, from` only works if sort is equal to "top" or "user".

You can see this starting on line 394 in api/resolvers/item.js.  `whenClause`/`whenRange` are only added to the query if sort is 'user' or 'top'

### Screenshots

_No response_

### Steps To Reproduce

This is the query I ran:

```
query GetItems($sort: String, $when: String, $from: String, $to: String, $limit: Limit) {
    items(sort: $sort, when: $when, from: $from, to: $to, limit: $limit) {
        items {
            id
            createdAt
            title
            text
            user {
                id
                name
            }
            parentId
            sats
            upvotes
            ncomments
            subName
        }
    }
}
```

with variables:

```
{'sort': '', 'when': 'custom', 'from': '1751353200000', 'to': '1754031599999', 'limit': 100}
```

You will retrieve items outside the date range (which in this case is 2025-07-01 to 2025-07-31).  

But if you do `'sort': 'top'`, you'll get items in the correct date range

### Expected behavior

Date filter should not depend on sort type

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

CloudWaddie:
## Summary
- Fixes #2433
- The `when`/`from`/`to` date range parameters in the `items` GraphQL query were silently ignored unless `sort` was `top` or `user`
- Added `whenClause` and `whenRange` support to the `new` and default (`lit`/hot) sort branches
- When no `when` parameter is provided, behavior is unchanged

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -438,27 +438,51 @@ export default {
           }, ...whenRange(when, from, to || decodedCursor.time), user.id, decodedCursor.offset, limit)
           break
         case 'new':
-          items = await itemQueryWithMeta({
-            me,
-            models,
-            query: `
-              ${SELECT}
-              ${relationClause(type)}
-              ${payInJoinFilter(me)}
-              ${whereClause(
-                '"Item".created_at <= $1',
-                '"Item"."deletedAt" IS NULL',
-                subClause(sub, 4, subClauseTable(type), me, showNsfw),
-                activeOrMine(me),
-                await filterClause(type, sub, 'new', ctx),
-                typeClause(type),
-                muteClause(me)
-              )}
-              ORDER BY "PayIn"."payInStateChangedAt" DESC
-              OFFSET $2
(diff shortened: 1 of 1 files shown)


KingParmenides (commented):
One edge case to consider: the default-sort branch still fetches `pins` before the new `if (when)` item query split, so `Items.items` will respect the date range but `Items.pins` can still include pinned posts outside that range when the caller selects the `pins` field.\n\nIf the intent is that `when/from/to` apply to the whole `items` response, the pins query probably needs the same `whenClause(when, 'Item')` and shifted sub param index when `when` is set. If pins are intentionally independent of date filters, it may be worth calling that out in the PR since it is a visible exception to the API behavior.

PR #2953, opened 2026-05-10
editor: persist mode and toolbar visibility in localStorage

### LINKED ISSUE #2858: Locally remember editor's last settings
We would save in `localStorage` things like:

- last editor mode
- toolbar (visible|hidden)
- more?

CloudWaddie:
## Summary
- Fixes #2858
- Editor mode (write vs compose) and toolbar visibility are now persisted to localStorage
- Uses the project's existing `useLocalState` hook for mode persistence
- Toolbar visibility reads from `localStorage` on init, writes on change
- Settings survive page reloads — editor opens in the same mode you left it


huumn:

PR #2954, opened 2026-05-10
fix: prompt for confirmation when canceling edit with unsaved changes

### LINKED ISSUE #2568: Cancel on edit does not prompt for confirmation

CloudWaddie:
## Summary
- Fixes #2568
- Canceling a comment or post edit with unsaved changes now shows a confirmation modal ("Are you sure? You will lose your work")
- Uses the same `useShowModal` confirmation pattern already used in `reply.js`
- Only prompts when Formik's `dirty` flag is true (actual changes made)
- Clean cancels (no changes) proceed immediately without prompting

### DIFF
--- a/components/cancel-button.js
+++ b/components/cancel-button.js
@@ -1,9 +1,37 @@
 import { useRouter } from 'next/router'
 import Button from 'react-bootstrap/Button'
+import { useFormikContext } from 'formik'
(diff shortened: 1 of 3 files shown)


sir-opti:
Duplicate of #2592

huumn:

PR #2955, opened 2026-05-10
fix: scroll to comments header when changing sort order

### LINKED ISSUE #2821: Changing comment sort causes weird scroll behavior
I'll try to describe in detail soon. just wanted to note it. We're accumulating a ton of bugs like this.

CloudWaddie:
## Summary
- Fixes #2821
- Changing comment sort caused weird scroll behavior because Next.js `scrollRestoration` moved the viewport to a saved position for the new URL
- After the sort route transition completes, the viewport now smoothly scrolls to the comments header
- `CommentsHeader` wrapped with `forwardRef` to allow the parent `Comments` component to reference the DOM node


huumn:

PR #2957, opened 2026-05-10
seo: use territory description for OG meta tags

### LINKED ISSUE #2363: Custom OG meta tags for territories

CloudWaddie:
## Summary
- Fixes #2363
- Territory pages now use the territory's description for the OG meta description instead of the generic site-wide default
- Added `subInfo` prop to `Layout` and `Seo` components to forward the full territory object
- Updated 4 territory page files to pass `subInfo={sub}`
- Item and user pages still override with their own content as before

### DIFF
--- a/components/layout.js
+++ b/components/layout.js
@@ -9,12 +9,12 @@ import styles from './layout.module.css'
 import PullToRefresh from './pull-to-refresh'
 
 export default function Layout ({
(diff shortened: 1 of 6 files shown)


huumn:

Issue #2942, opened 2026-05-04
Transaction Sankey numbers are not internationalized


huumn:
https://stacker.news/items/1483849/r/k00b?commentId=1483943

dmant2000:
Opened PR #2998 with a focused fix and tests: https://github.com/stackernews/stacker.news/pull/2998

Since this closes an existing bug and the README describes sats awards for PRs closing issues, please tag/review it through that contribution-award flow after merge. I can provide any payout details needed.

Matt7cas:
I dug into this and I think there are actually **two separate causes**, which may be why the
previous attempt stalled. Both are in the Sankey's formatting path.

## Root cause

`assetFormatted` in `components/payIn/sankey/index.js:45-50` is the only formatting entry point,
called from `:202` (nodes) and `:217` (links):

```js
function assetFormatted (msats, type) {
  if (type === 'CREDITS') {
    return numWithUnits(msatsToSatsDecimal(msats), { unitSingular: 'CC', unitPlural: 'CCs', abbreviate: false })
  }
  return numWithUnits(msatsToSatsDecimal(msats), { unitSingular: 'sat', unitPlural: 'sats', abbreviate: false })
}
```

**Cause 1 — no locale reaches the formatter.** `numWithUnits` (`lib/format.js:44`) calls
`new Intl.NumberFormat().format(n)` with no locale argument. On the client that resolves to the
browser locale, but nothing in the Sankey path makes that explicit.

**Cause 2 — the value arrives pre-stringified.** `msatsToSatsDecimal` returns
`fixedDecimal(Number(msats) / 1000.0, 3)`, and `fixedDecimal` is
`Number.parseFloat(n).toFixed(f)` — a **string** like `"1234.000"`. So `Intl.NumberFormat`
receives `"1234.000"` rather than a number, which defeats its own fraction-digit handling and
produces forced trailing zeros plus inconsistent grouping compared to the rest of the UI, which
passes real numbers.

So `1500` msats renders as `1.500 sats` (three forced decimals) instead of `1.5 sats`, and large
whole-sat values don't group consistently with amounts elsewhere in the app.

## Context worth flagging

There is **no i18n framework in the repo** — no `next-intl` / `react-intl` / `i18next` /
`next-translate` in `package.json`, and no `i18n` block in `next.config.js`. So
"internationalized" here reads as **locale-aware number formatting**, not message translation.
Worth confirming that's the intent before anyone writes code.

## Suggested direction

[...]

PR #2960, opened 2026-05-10
fix: internationalize Sankey diagram numbers

CloudWaddie:
## Summary
- Fixes #2942
- Added `valueFormat` prop using `Intl.NumberFormat` so Nivo's internal formatted values use locale-aware number formatting (thousand separators etc.)
- Tooltip now falls back to `formattedValue` when the custom `asset` property is unavailable, ensuring numbers are always displayed and properly formatted


huumn:

PR #2961, opened 2026-05-10
custom domains: brandings

Soxasora:
## Description

Part of #1942, revives and rewrites #2202
Adds basic theming and SEO capabilities to territories, gated by custom domains. Including dynamic PWA and RSS.

Territory branding settings are unlocked the moment a territory owner creates a custom domain, even while `PENDING`.

<details>
<summary>SubBranding schema</summary>

```md
### seo
- title
- tagline (meta description)
- favicon

### theme
- primaryColor
- secondaryColor
- linkColor
- logo
```

</details>

## Screenshots

Territory branding form

image

Custom domain with territory branding

image

PWA installation (dynamic `site.webmanifest` is recognized)

https://github.com/user-attachments/assets/5a5ec1eb-a340-483f-a53f-1c1cd0f51a44

Dynamic RSS with external territories support

image

## Additional Context

The custom territory theme CSS is built by `buildSubBrandingCSS` and injected into `_document` to ensure that the very first paint is already themed with the territory theme.

`buildSubBrandingCSS` turns `SubBranding` theme fields into a string of CSS variable overrides:
- `--bs-primary(-rgb)`
  - `--theme-primary-text`
- `--bs-secondary(-rgb)`
  - `--theme-secondary-text`
- `--theme-link(Hover)`

The `*-text` values are a YIQ contrast pick (`#000` or `#fff`) computed by `getContrastTextColor()`, because the foreground color on top of a brand color cannot be derived in CSS. Bootstrap handles color mixing and contrast calculations internally, forcing us to essentially re-do what it natively does in order to control the colors.

The output CSS features a doubled selector, raising specificity so the overrides win against the scss `:root` defaults.

Bootstrap buttons' `button-variant()` mixin bakes literal hex values into `--bs-btn-bg, -bs-btn-hover-bg, ...`, and the `.btn` base class reads those values instead of `--bs-primary`.
As a workaround, following Bootstrap's *directions*, we now:
- rebind the per-component vars (`.btn-primary, .btn-secondary, ...`) so `--bs-btn-bg` reads the primary color and `--bs-btn-color` reads `--theme-primary-text`
- derive hover and active variants with `color-mix`, which is the runtime equivalent of scss' `shade-color()`
- SVG glyphs inside primary and secondary buttons follow the `--bs-btn-color` updated earlier.

SN default theme is exactly the same as before and now territory theme colors can affect buttons.

https://github.com/twbs/bootstrap/issues/38788#issuecomment-1597398753
https://github.com/twbs/bootstrap/issues/38170

## Checklist

Yes!

6, did not test edge cases
- RSS: OK
- PWA: OK
- TerritoryBranding form: OK
- Apollo merge: OK
- Bootstrap overrides: OK

in QA

n/a

Yes
- initial shape and working prototype. This PR takes inspiration from the vibe-shaped prototype.
- refactor

---

> [!NOTE]
> **Medium Risk**
> Touches auth-adjacent upload authorization, domain/beta gating, and request-wide SSR theming/manifest; scope is large but gated to beta users with an existing domain requirement.
> 
> **Overview**
> Adds **territory branding** for custom-domain territories: a new `SubBranding` model and `upsertSubBranding` mutation (beta owners only, requires a custom domain—even while verification is pending). Owners get owner-only `Sub.domain` / `Sub.branding` fields; `setDomain` now returns the **`Sub`** (with nested domain in the client mutation) instead of `Domain`.
> 
> **Uploads:** `getSignedPOST` accepts optional `subName` for free image uploads on the same trust boundary as branding (beta, ownership, domain exists).
> 
> **Custom-domain experience:** SSR/domain cache loads branding; `getDomainBranding` replaces the old domain-only context with **`BrandingProvider` / `useBranding`**. Themed sites get injected CSS (`buildSubBrandingCSS`), custom logo/favicon, SEO/capture URLs, nav tweaks, and Bootstrap primary/secondary button overrides. Territory edit uses `SUB_EDIT` and a combined **advanced** accordion (domain + branding form with polling while `PENDING`).
> 
> **Platform surfaces:** Dynamic PWA `site.webmanifest`, RSS feed metadata/links, and login/signup copy use branding when present. `processCrop` / shared imgproxy helpers move to `lib/imgproxy`; `isSecureRequest` moves to `lib/safe-url`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4da724cf9a7960a2677db410ec88d85794b6b0b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/safe-url.js
```diff
@@ -1,3 +1,33 @@
...
+// True iff the current request was served over HTTPS. In prod this is always
+// true (the ALB terminates TLS). In dev it depends on whether the request
+// came in via a TLS-terminating proxy (https) or directly to localhost:3000
+// (http) — used to gate the Secure cookie attribute so localhost dev still
+// works while the custom-domain HTTPS path exercises the full secure-cookie
+// codepath.
+//
+// Accepts either an edge NextRequest, a Node IncomingMessage, or undefined.
+// When called in the browser with no req, derives from window.location.
+export function isSecureRequest (req) {
```
Soxasora, line 12: I moved `isSecureRequest` from `/lib/auth.js` to `/lib/safe-url.js` to decouple it from auth. Hope it's okay!

pages/api/site.webmanifest.js
```diff
@@ -128,17 +133,34 @@ const defaultManifest = {
...
+
+  return {
+    ...defaultManifest,
+    ...(isDark && { background_color: black }),
+    ...(branding?.title && { name: branding.title, short_name: truncateString(branding.title, 12, '') }),
+    ...(branding?.tagline && { description: branding.tagline }),
+    ...(branding?.primaryColor && { background_color: branding.primaryColor }),
+    ...(faviconUrl && {
+      icons: [{ src: faviconUrl, sizes: 'any', purpose: 'any' }]
+    }),
```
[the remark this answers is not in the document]
Soxasora: I was able to install the PWA pretty much on every OS. But we can probably leverage imgproxy for this, I'll come up with something.

api/resolvers/sub.js
```diff
@@ -345,6 +346,41 @@ export default {
...
+        throw new GqlInputError('requires a custom domain')
+      }
+
+      await validateSchema(subBrandingSchema, branding)
+
+      await models.subBranding.upsert({
+        where: { subName },
+        update: branding,
+        create: { subName, ...branding }
+      })
```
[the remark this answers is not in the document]
Soxasora: Probably the 20th time I receive this, the territory owner would lose permanent free uploads this way. And they would have to craft a `cURL` for this to happen.

components/territory-branding.js
```diff
@@ -0,0 +1,279 @@
...
+    <div className='w-100'>
+      <AccordianItem
+        show={hasDomain}
+        header={<div style={{ fontWeight: 'bold', fontSize: '92%' }}>advanced</div>}
+        body={
+          <>
+            <TerritoryDomains
+              sub={sub}
+              domain={domain}
+            />
```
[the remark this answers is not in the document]
Soxasora: `GET_DOMAIN` gets merged by Apollo automatically in `Sub` thanks to the matching structure. The problem here is that it's almost invisible, that's probably why this got reported.

api/resolvers/sub.js
```diff
@@ -345,6 +346,41 @@ export default {
...
+      if (sub.userId !== Number(me.id)) {
+        throw new GqlAuthorizationError('you do not own this sub')
+      }
+
+      const domain = await models.domain.findUnique({ where: { subName } })
+      if (!domain) {
+        throw new GqlInputError('requires a custom domain')
+      }
+
+      await validateSchema(subBrandingSchema, branding)
```
huumn: We should store the return value of `validateSchema`, because it trims/transforms, and we transform the stored values again at read time (in some places), which might break if we assume they're trimmed. 

We have this problem in other parts of the code too, but probably worth storing data as we validated it.

pages/api/site.webmanifest.js
```diff
@@ -128,17 +135,58 @@ const defaultManifest = {
...
+    // re-scope the PWA to the custom domain
+    ...(origin && {
+      url_handlers: [{ origin }],
+      // TODO REVIEW: this doesn't work, capture service cannot grab external domains
+      screenshots: [{
+        src: `https://capture.stacker.news/${origin}`,
+        type: 'image/png',
+        form_factor: 'wide'
+      }]
+    })
```
huumn: As discussed in the meeting, until we have a fix for this, let's use `stacker.news` rather than something that will 404.

lib/rss.js
```diff
@@ -38,17 +49,17 @@ const generateRssItem = (item) => {
...
+function generateRssFeed (items, { sub = null, site }) {
+  const itemsList = items.map(item => generateRssItem(item, { url: site.url, domainSubName: site.domainSubName }))
   return `<?xml version="1.0" encoding="UTF-8" ?>
     <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
       <channel>
-        <title>${SITE_TITLE}${sub ? ` ~${sub}` : ''}</title>
-        <link>${SITE_URL}${sub ? `/~${sub}` : ''}</link>
-        <description>${SITE_SUBTITLE}</description>
+        <title>${escapeXml(site.title)}${sub ? ` ~${sub}` : ''}</title>
+        <link>${escapeXml(site.url)}${sub ? `/~${sub}` : ''}</link>
```
huumn: `sub` here comes from the query param - not the database. It is url encoded so XSS isn't possible, but malformed XML is.

pages/_document.js
huumn: Might be worth going all the way and branding these fallbacks too. Unless there's good reason not to.

huumn (commented):
These are some findings I worked through in generic bot review. Nothing critical but worth fixing imo.

QA of basic stuff is rock solid and very nice so far. I've read through all of the code casually and it makes sense. I did a harder pass over structure/schema/backend and it's all good.

I'll continue with more targeted QA/review and let you know how it goes.

lib/domains/custom-css.js
```diff
@@ -0,0 +1,44 @@
...
+    overrides.push(`--theme-link: ${linkHex};`)
+    // slightly darken the base link color (mix with black 15%)
+    overrides.push(`--theme-linkHover: color-mix(in srgb, ${linkHex} 85%, #000 15%);`)
+  }
+
+  if (overrides.length === 0) return null
+
+  // REVIEW: not so beautiful
+  // ensures variables are set at the highest level for both theme modes
+  return `:root:root,[data-bs-theme=light][data-bs-theme=light],[data-bs-theme=dark][data-bs-theme=dark]{${overrides.join('')}}`
```
huumn: I think this can just be 

```
return `:root[data-bs-theme]{${overrides.join('')}}`
```

All we need is for the overrides to take precedence over the theme's `0-1-0` specificity, which we get with the above as `0-2-0`.

TIL about how css specificity is calculated: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity

huumn (approved):
I didn't find anything blocking. Overall it's well organized and is a good foundation for all the various parts we'll need to customize.

api/resolvers/sub.js
Soxasora: Ah yes! I repeated the usual pattern without really looking into this. Thanks ^^

pages/_document.js
Soxasora: `theme-color` is responsible of the browser window color (the chrome) and should stay as-is, we don't give territory owners the possibility to change the background color.

instead for the apple icon I think we can use the favicon ^^

lib/domains/custom-css.js
Soxasora: I read and re-read MDN's specificity docs so many times, still ended up with that garbo. Glad you found a way to simplify it, I must've been burned out when I did that.

components/seo.js
```diff
@@ -84,12 +109,9 @@ export default function Seo ({ sub, item, user }) {
             url: 'https://capture.stacker.news' + pathNoQuery
           }
         ],
-        site_name: 'Stacker News'
-      }}
-      twitter={{
-        site: '@stacker_news',
-        cardType: 'summary_large_image'
+        site_name: siteName
```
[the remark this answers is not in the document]
Soxasora: oof that's correct, I'll fix it now.

PR #2974, opened 2026-05-14
Localize pay-in Sankey values

Robertg761:
## Summary
- format Nivo Sankey flow values with `Intl.NumberFormat`
- preserve up to three fractional sat digits for msat-derived values

Fixes #2942.

## Verification
- `git diff --check`
- `node --check components/payIn/sankey/index.js`


huumn:
we have prs for this already.

PR #2980, opened 2026-05-16
Fix Blixt Wallet Lightning Address integration

hsjoberg:
Hi, we use `@blixtwallet.com`  for our Lightning Address support in Blixt.

Cheers
Hampus


huumn:
Thanks!

PR #2982, opened 2026-05-18
Internationalize pay-in Sankey values

sravan27:
Closes #2942.\n\n## Summary\n- formats Sankey values through Intl.NumberFormat via Nivo's valueFormat prop\n- keeps Sankey link values numeric instead of stringified decimals\n- formats node/link asset labels from numeric sats values so singular/plural units stay correct\n\n## Test\n- npm run lint -- components/payIn/sankey/index.js\n\nThis was a small unlabeled bug fix; if it qualifies for an SN difficulty/reward label, I would appreciate consideration.


huumn:
duplicate

PR #2985, opened 2026-05-21
Fix msat decimal formatting

ezzy1630:
## What changed
- Return a number from msatsToSatsDecimal instead of a fixed-decimal string.
- Add coverage for singular unit formatting and sub-sat precision.

## Why
This lets transaction Sankey amounts go through Intl.NumberFormat and unit pluralization correctly. For example, 1000 msats now formats as 1 sat instead of treating "1.000" as a string.

Fixes #2942

## Testing
- npm run lint -- lib/format.js lib/format.test.js
- npm test -- lib/format.test.js --runInBand

### DIFF
--- a/lib/format.js
+++ b/lib/format.js
@@ -73,7 +73,7 @@ export const msatsToSatsDecimal = msats => {
   if (msats === null || msats === undefined) {
     return null
   }
-  return fixedDecimal(Number(msats) / 1000.0, 3)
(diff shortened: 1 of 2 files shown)


huumn:
duplicaate

ezzy1630:
Got it, thanks for the quick check. I'll leave this closed as duplicate.

PR #3002, opened 2026-06-06
fix: move eclair env_file to service level on docker-compose.yml

felipebueno:
## Description

`./sndev start` was not working:

```
failed to parse docker-compose.yml: yaml: construct errors:
  line 1: yaml: map merge requires map or sequence of maps as the value
```

The `eclair` service was doing `<<: *env_file` under `environment:` but every other service uses `env_file: *env_file` at the service level so I figured that could be the fix (claude helped me).

Claude said this:
"Older Docker Compose tolerated it silently. The newer one (v5) has a stricter YAML parser and straight up refuses to parse the whole file — so it broke every profile, even `minimal`."

And TBH, I didn't find evidence to back this claim up except for the fact that moving it one level up to match the rest worked. I also tested sndev start with the other profile options and everything seems to working properly.

## Screenshots

Before fix:

image

After fix:

image

## Additional Context

No. I just cloned the repo and tried to run it from the master branch with sndev start.

## Checklist

Yes

Claude helped me on this one too:

"8. Ran `docker compose config -q` across all profiles (`minimal`, `payments`, `wallets`, full stack, `*`) — all parse now. Then booted the full dev stack for real: eclair comes up, peers with `sn_lnd`, and `printenv` confirms `DATABASE_URL`/`NODE_ENV` are injected with no junk merge keys."

N/A

No

Yes. I used Claude Opus 4.8 to help me find the parse error, propose a fix and then I used it again to smoke test all profiles and analyze the results.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -594,8 +594,8 @@ services:
     restart: unless-stopped
     depends_on:
       <<: *depends_on_bitcoin
+    env_file: *env_file
     environment:
-      <<: *env_file
       JAVA_OPTS:
         -Declair.printToConsole
         -Dakka.loglevel=DEBUG


Soxasora (approved):
I'm not able to reproduce this issue on macOS Orbstack with Docker Compose v5.1.2, that's probably why this went under the radar.
I am able to reproduce this with Fedora and the latest Docker Compose.

Your fix works great, thanks!

PR #3056, opened 2026-06-09
Store email magic code callback in a short-lived cookie

### LINKED ISSUE #2984: Email magic code flow is not supported on some browsers
https://stacker.news/items/1493441/r/k00b?commentId=1493478

caydyan:
## Summary
- Store the email magic-code callback payload in a short-lived first-party cookie instead of sessionStorage.
- Read and clear that cookie on `/email` when building the final email callback URL.
- Apply the same storage path to email login/signup and email account linking.

## Why
Closes #2984.

## Validation
- `./node_modules/.bin/standard pages/email.js components/login.js pages/settings/logins.js lib/auth.js`
- `git diff --check`
- `npm run lint`
- `rg -n "sessionStorage\.setItem\('callback'|sessionStorage\.getItem\('callback'|email_auth_callback|EMAIL_AUTH_CALLBACK" components pages lib -g "*.js" -g "*.jsx"`

## Payout
BTC address: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

### DIFF
--- a/components/login.js
+++ b/components/login.js
@@ -10,7 +10,7 @@ import LoginButton, { LoginWithNymButton } from './login-button'
 import { emailSchema } from '@/lib/validate'
 import { OverlayTrigger, Tooltip } from 'react-bootstrap'
 import * as cookie from 'cookie'
-import { cookieOptions, MULTI_AUTH_ANON, MULTI_AUTH_POINTER } from '@/lib/auth'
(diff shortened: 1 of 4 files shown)


Soxasora:
I agree that magic code would be better with cookies rather than with session storage, but it introduces the need to manage yet another cookie. Don't get me wrong, cookies are really great at solving nextauth's problems - we're using them to manage signin/signup and email login attachment.

This time I'd like to solve this without a cookie, using the nextauth's email callback. I should've assigned myself to the issue since I'm managing NextAuth for the custom domains feature, but thanks for the interest ^^

PR #3064, opened 2026-06-13
Skip placeholder YouTube titles

caydyan:
Refs #2164.

## Description

Some YouTube watch-page responses only expose the placeholder title `- YouTube` through the current unauthenticated server-side metadata path. `pageTitleAndUnshorted` currently passes that value through, so the link form can auto-fill a misleading title.

This filters YouTube placeholder titles before returning metadata and only returns a title when the parser found a real value. It does not add a new YouTube title source; it just avoids populating known-bad placeholder text.

## Screenshots

Not applicable; server-side metadata behavior.

## Additional Context

This is a narrow improvement, not a complete solution for every YouTube URL. In local smoke checks, two YouTube URLs with real metadata still auto-fill, while the known counterexample no longer returns `- YouTube`.

## Checklist

- [x] I have tested this change locally
- [x] I have considered regressions and scoped the fix narrowly
- [x] No new environment variables are required

Validation:

- `npm run lint -- api/resolvers/item.js`
- `git diff --check`

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -33,6 +33,19 @@ import { BOUNTY_ALREADY_PAID_ERROR, BOUNTY_IN_PROGRESS_ERROR, getBountyPaymentTa
 import { lexicalHTMLGenerator } from '@/lib/lexical/server/html'
 import { resolveItemComments } from './comment-tree'
 
+function isYoutubePlaceholderTitle (url, title) {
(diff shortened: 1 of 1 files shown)


sir-opti:
Depends on inline maintenance of YT domains (already missing `youtube-nocookie.com`). Recommend rejection.

Issue #117, opened 2022-03-02
Disallow URIs in the post Title


cointastical:
SN [Item #13221](https://stacker.news/items/13221) had a URL ( https:// ) in the title. 

I can see where a domain itself should be accommodated (e.g., "Crypto.com stadium hosts big game"), but I can't think of a reason the URI Scheme (e.g., https:// ) would be needed in the title.

Having that URI scheme in the title messes with the character count calculation for automated Tweeting using Zapier.

sk-ip:
I assume this change is needed only for the discussion type post, But i think this should be extended to other types as well, as there is no check on the title fields for other post types.

zhailong8845-art:
I am taking this as a small validation fix across post types using the shared title validator. I will allow ordinary domains such as `Crypto.com` while rejecting URI schemes such as `https://` (including mixed-case schemes), with focused validation tests. Before review, could an SN engineer assign the appropriate `difficulty:*` label so the documented awards workflow can calculate any award if the change is merged? Please classify it based on the reviewed scope; I am not requesting a particular level.

Delivered in #3139 (`fd1018df`, follow-up `c9c3fe50`): the suite now directly covers bounty, discussion, job, link, and poll titles. Focused Jest passes 14/14, Standard Style passes, and `git diff --check` passes.

JhoanChala:
Fixed in #3175: the shared titleValidator for all post types now rejects URI schemes like https:// while still allowing plain domains (Crypto.com).

jahnclawdmonet:
I have a fix ready for this. Scoped it to reject `scheme://` in the title — so `https://stacker.news/items/13221` is rejected, but bare domains like `Crypto.com` still pass, matching the spec discussed here. It lives in the shared `titleValidator`, so it covers discussion / link / poll / bounty posts in one place, and I added unit tests to `lib/validate.spec.js` (the full `lib/` suite stays green).

One scoping question before I open a PR: do you want schemeless URLs like a bare `www.example.com` blocked too, or just schemes? I scoped it to `scheme://` since that's what actually throws off the downstream character counts (e.g. the automated tweeting mentioned in the original report), but I'm happy to widen it if you'd prefer.

PR #3065, opened 2026-06-13
Reject URLs in post titles

caydyan:
Refs #117.

## Description

Post titles currently accept strings containing full URLs such as `https://...`. Issue #117 notes that URI schemes in titles are not needed for normal title text and can break downstream uses such as automated tweet length calculations.

This adds a shared title validation rule that rejects URL-looking values with an explicit URI scheme (`scheme://...`) while still allowing bare domains like `Crypto.com` in titles. Because discussion, link, bounty, poll, and job submissions all use the shared title validator, the rule applies consistently across those title fields.

## Screenshots

Not applicable; validation behavior only.

## Additional Context

The check is intentionally narrow to avoid rejecting ordinary titles with domains or colon punctuation. It targets URLs with URI schemes rather than every string containing a dot or colon.

## Checklist

- [x] I have tested this change locally
- [x] I have considered false positives for bare domains in titles
- [x] No new environment variables are required

Validation:

- `npm run lint -- lib/validate.js lib/validate.spec.js`
- `npm test -- lib/validate.spec.js`
- `git diff --check`

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -53,7 +53,7 @@ export function truncateToCharLength (value, maxLength) {
   return out
 }
 
-const TITLE_URL_WITH_SCHEME_REGEXP = /\b[a-z][a-z0-9+.-]*:\/\/\S+/i
+const TITLE_URL_WITH_SCHEME_REGEXP = /\b[a-z][a-z0-9+.-]*:(?:\/\/|\S+)/i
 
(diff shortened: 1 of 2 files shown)


sir-opti:
`://` filter ignores schemeless uris (`mailto:`, `bitcoin:`...). Recommend reject.

PR #3066, opened 2026-06-13
Check wallet invoice network

### LINKED ISSUE #1028: Network not checked for LND autowithdrawals

caydyan:
Fixes #1028.

## Description

Receive-wallet tests currently only check that the wallet can create an invoice starting with `lnbc`. That hardcodes mainnet and misses the actual issue from #1028: a user can attach a wallet on a different Lightning network than the SN node.

This adds BOLT11 network detection for mainnet, testnet, signet, and regtest prefixes, asks the SN node to create a short-lived test invoice, and compares the two networks during `testWalletRecvProtocol`.

This keeps valid regtest/testnet/signet invoices from being rejected just because they do not start with `lnbc`, while returning a clear error for mismatches like a mainnet wallet against a regtest SN node.

## Screenshots

Not applicable; server-side wallet validation behavior.

## Additional Context

The network helper is split into a small pure module so the prefix logic can be unit tested without loading the full wallet protocol resolver dependency chain.

## Checklist

Yes. Wallets on the same network as SN still pass. Invalid invoices still fail, and cross-network wallets now fail earlier with a clearer error.

6/10. Added unit coverage for BOLT11 network detection/mismatch behavior and ran focused lint/test checks. I did not run a live LND attach flow locally.

Not a frontend change.

No.

Yes. AI assisted with implementation and verification. I inspected the issue, avoided existing covered issues, made the focused patch, and ran the checks below.

Validation:

- `npm run lint -- wallets/server/bolt11-network.js wallets/server/resolvers/protocol.js wallets/server/resolvers/protocol.spec.js`
- `npm test -- wallets/server/resolvers/protocol.spec.js`
- `git diff --check`

### DIFF
--- a/wallets/server/bolt11-network.js
+++ b/wallets/server/bolt11-network.js
@@ -7,6 +7,22 @@ const BOLT11_NETWORKS = [
   { prefix: 'lnbc', name: 'mainnet' }
 ]
 
+const BOLT11_NETWORK_ALIASES = new Map([
+  ['mainnet', 'mainnet'],
+  ['bitcoin', 'mainnet'],
+  ['bc', 'mainnet'],
+  ['lnbc', 'mainnet'],
+  ['testnet', 'testnet'],
+  ['tb', 'testnet'],
+  ['lntb', 'testnet'],
+  ['signet', 'signet'],
+  ['tbs', 'signet'],
+  ['lntbs', 'signet'],
+  ['regtest', 'regtest'],
+  ['bcrt', 'regtest'],
+  ['lnbcrt', 'regtest']
+])
+
 export function bolt11Network (invoice) {
   if (typeof invoice !== 'string') return null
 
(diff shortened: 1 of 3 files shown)


sir-opti:
Feels like this introduces runtime interfacing with lnd for what could be a check against env.

PR #2781, opened 2026-02-07
fix: evaluate poll expiry min date at validation time

joelklabo:
## Summary

Fixes #907

The `pollExpiresAt` validation used a static min date computed once at schema creation time via `datePivot(new Date(), { days: 1 })`. This date was frozen when the schema object was instantiated, meaning:

- On the **client**, the min date was set when the component mounted, not when the form was submitted
- On the **server**, a fresh schema is created per request, but any time drift between client default computation and server validation could cause the check to fail

This replaces the static `.min()` with a dynamic `.test()` that evaluates `datePivot(new Date(), { days: 1 })` at the time of each validation run, ensuring the 1-day minimum is always relative to actual submission time.

## Changes

- `lib/validate.js`: Replace static `date().min(...)` with dynamic `date().test(...)` for `pollExpiresAt`

## Test plan

- [x] All CI checks pass (lint, shellcheck, unit tests, security)
- [x] Validation logic uses dynamic `new Date()` — no stale date bug possible
- [x] `pollExpiresAt` remains `nullable()` — null/cleared expiration still valid
- [x] Only `pollExpiresAt` field affected — no other validators changed

PR #3067, opened 2026-06-13
Validate poll expiry at submit time

### LINKED ISSUE #907: Bad default poll end date
**Description**

According to https://stacker.news/items/457480, the default poll end date is wrong. I did not confirm.

**Steps to Reproduce**

1. Create poll
2. Error

**Expected behavior**

1. Create poll
2. No error

caydyan:
Fixes #907.

## Description

`pollSchema` currently computes the minimum poll expiration date when the schema is created:

```js
.min(datePivot(new Date(), { days: 1 }), ...)
```

That makes the minimum stale for any form/session that lives across time before validation. A default expiration that was valid when the poll form mounted can later be validated against an outdated minimum instead of the actual submit-time minimum.

This changes `pollExpiresAt` to a custom Yup test that evaluates `datePivot(new Date(), { days: 1 })` during validation. Clearing the expiration remains valid because the field is still nullable.

## Screenshots

Not applicable; validation behavior only.

## Additional Context

A previous PR for this issue (#2781) was closed by its author because it was submitted without proper local testing. This version includes focused Jest coverage for the stale-schema case.

## Checklist

Yes. Poll expirations at least one day in the future still pass, and cleared expirations still pass. Expirations less than one day from validation time still fail with the existing message.

7/10. Added focused Jest coverage for the stale-schema validation case, valid one-day future expiration, and nullable expiration. Ran lint and whitespace checks.

Not a frontend rendering change.

No.

Yes. AI assisted with issue triage, implementation, and validation. I checked for duplicate/current PR coverage and ran the commands below.

Validation:

- `npm run lint -- lib/validate.js lib/validate.spec.js`
- `npm test -- lib/validate.spec.js`
- `git diff --check`

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -295,7 +295,11 @@ export function pollSchema ({ numExistingChoices = 0, ...args }) {
       message: `at least ${MIN_POLL_NUM_CHOICES} choices required`,
       test: arr => arr.length >= MIN_POLL_NUM_CHOICES - numExistingChoices
     }),
-    pollExpiresAt: date().nullable().min(datePivot(new Date(), { days: 1 }), 'Expiration must be at least 1 day in the future'),
+    pollExpiresAt: date().nullable().test({
+      name: 'min',
+      message: 'Expiration must be at least 1 day in the future',
+      test: value => !value || value >= datePivot(new Date(), { days: 1 })
+    }),
     randPollOptions: boolean(),
     ...advPostSchemaMembers(args),
     ...subSelectSchemaMembers('POLL', args)
--- a/lib/validate.spec.js
+++ b/lib/validate.spec.js
@@ -2,35 +2,50 @@
 
 import { pollSchema } from './validate'
 
+const POLL_EXPIRATION_ERROR = 'Expiration must be at least 1 day in the future'
+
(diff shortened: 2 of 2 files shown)


sir-opti:
The rainy case (last condition) of the [first test](https://github.com/caydyan/stacker.news/blob/bf191ca80a781484ba7725f27bd56b3c4ef6d5c6/lib/validate.spec.js#L10-L21) does fail on `master` right now:

```console
 FAIL  lib/validate.spec.js
  ● pollSchema › evaluates poll expiration minimum at validation time

    expect(received).rejects.toThrow()

    Received promise resolved instead of rejected
```

Overall this seems to work (I'm not sure **why** `yup`'s `.min()` doesn't catch though)

Nit: Could use a cleaner spec file that either bundles all similar test cases together, or separates them all (recommend latter for easier debugging), because right now it is a bit of both.

caydyan:
Addressed the test-structure nit in the current head `d3b26c32` by splitting `pollExpiresAt` coverage into focused cases and keeping the schema-creation-time regression case explicit. Re-ran locally: `npm test -- --runTestsByPath lib/validate.spec.js --runInBand` (4 tests passing).

PR #3069, opened 2026-06-13
[codex] Suppress timeout form submit toasts

### LINKED ISSUE #1440: 504 errors toast occasionally and appear inconsequential
I'm not sure why this started happening but it doesn't appear to indicate a problem we need to surface to the user.

edit: I previously thought this was a deployment issue but its more generic

caydyan:
Fixes #1440.

## Summary
- add `shouldSuppressFormSubmitToast` for abort/timeout and 504 submit failures
- skip the danger toast for those failures while keeping the console log
- add focused unit coverage so business errors, 500s, and broad fetch failures still toast

## Why
Receiving wallet timeouts can surface as generic 504 submit errors even when the server may still be processing the request. The issue asks to catch these specifically and avoid showing the inconsequential toast.

## Validation
- `npm run lint -- components/form.js lib/form.js lib/form.spec.js`
- `npm test -- lib/form.spec.js`
- `git diff --check`

### DIFF
--- a/components/form.js
+++ b/components/form.js
@@ -33,7 +33,6 @@ import { useIsClient } from './use-client'
 import PageLoading from './page-loading'
 import { SNEditor } from './editor'
 import { isAbortError } from '@/lib/error'
-import { shouldSuppressFormSubmitToast } from '@/lib/form'
 import QrScanner from './qr-scanner'
 export { MultiSelect } from './multi-select'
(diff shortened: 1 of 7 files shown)


sir-opti:
Not sure how this addresses #1440 on the wallet side. This fix globally suppresses 504 errors on all forms?

PR #3068, opened 2026-06-13
Restore trailing underscores in autolinks

caydyan:
Fixes #180.

## Description

Bare GFM autolinks drop trailing underscores from the link node and leave them as adjacent text. For example, `https://twitter.com/some_user_` is parsed as a link to `https://twitter.com/some_user` followed by a plain `_`, so the final underscore is not clickable.

This adds a small MDAST transform that runs before misleading-link normalization. When a bare autolink's rendered text exactly matches its URL and the following sibling text starts with one or more underscores, the transform merges those underscores back into the link URL/text and trims them from the sibling text.

The transform does not change explicit markdown links like `[example](https://example.com/path_)`, because those already preserve the intended URL.

## Screenshots

Not applicable; markdown parsing/rendering behavior.

## Additional Context

A previous PR for this issue (#2797) was closed by its author for insufficient local testing. This version includes focused transform coverage and keeps existing misleading-link tests passing.

BTC payout address if eligible for a sats award: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

## Checklist

**Are your changes backward compatible? Please answer below:**

Yes. Only bare autolinks followed by text that starts with underscores are changed. Explicit markdown links and non-underscore trailing punctuation keep their existing behavior.

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**

7/10. Reproduced the current GFM parse shape, added focused unit tests for one/multiple trailing underscores, sibling text preservation, and explicit markdown links, and reran existing URL/misleading-link tests.

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:**

Not a visual frontend change.

**Did you introduce any new environment variables? If so, call them out explicitly here:**

No.

**Did you use AI for this? If so, how much did it assist you?**

Yes. AI assisted with issue triage, implementation, and validation. I checked duplicate PR history, reproduced the parser behavior, and ran the commands below.

Validation:

- `npm run lint -- lib/lexical/mdast/transforms/links.js lib/lexical/mdast/transforms/index.js lib/lexical/utils/mdast.js lib/lexical/mdast/transforms/links.spec.js`
- `npm test -- lib/lexical/mdast/transforms/links.spec.js lib/url.spec.js`
- `git diff --check`

PR #3070, opened 2026-06-13
[codex] Avoid email autolinks for fediverse handles

### LINKED ISSUE #93: Smarter URL detection, not every @ in the middle makes it an e-mail address

caydyan:
Fixes #93.

## Summary
- add an MDAST transform that unwraps GFM `mailto:` autolinks when they are immediately preceded by Fediverse-style `@` or `!` handle prefixes
- keep normal email autolinks and explicit email links intact
- replace the link transform file's ESM-only traversal/string helpers with local helpers so the regression tests run under the current Jest config
- add regression coverage for the new transform and the existing link transforms touched by the helper change

## Why
GFM parses `@Cointastical@BitcoinHackers.org` as plain `@` text followed by a `mailto:Cointastical@BitcoinHackers.org` link. That makes Fediverse handles look like email addresses in SN content.

## Validation
- `npm run lint -- lib/lexical/mdast/transforms/links.js lib/lexical/mdast/transforms/index.js lib/lexical/utils/mdast.js lib/lexical/mdast/transforms/links.spec.js`
- `npm test -- lib/lexical/mdast/transforms/links.spec.js lib/url.spec.js`
- `git diff --check`

### DIFF
--- a/lib/lexical/mdast/transforms/index.js
+++ b/lib/lexical/mdast/transforms/index.js
@@ -1,5 +1,5 @@
 export { mentionTransform } from './mentions.js'
 export { nostrTransform } from './nostr.js'
-export { misleadingLinkTransform, malformedLinkEncodingTransform } from './links.js'
+export { fediverseHandleTransform, misleadingLinkTransform, malformedLinkEncodingTransform } from './links.js'
 export { footnoteTransform } from './footnotes.js'
 export { tocTransform } from './toc.js'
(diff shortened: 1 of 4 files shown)


sir-opti:
Similar issue as in #3068: replaces traversal logic with hand-written (assume this to be for jest integration) but changing production logic for testability is a bit of an anti-pattern.

PR #2253, opened 2025-06-28
Fix(npub): Do not parse npubs in links

brymut:
## Description

Add check to prevent parsing of npubs in links.

closes #2252 

## Screenshots
Before fix
Screenshot 2025-06-28 at 17 28 18

After fix
Screenshot 2025-06-28 at 17 27 37

## Additional Context

_Was anything unclear during your work on this PR? Anything we should definitely take a closer look at?_

## Checklist

**Are your changes backward compatible? Please answer below:**
Y
_For example, a change is not backward compatible if you removed a GraphQL field or dropped a database column._

**On a scale of 1-10 how well and how have you QA'd this change and any features it might affect? Please answer below:**
10

**For frontend changes: Tested on mobile, light and dark mode? Please answer below:**

**Did you introduce any new environment variables? If so, call them out explicitly here:**
N

PR #3071, opened 2026-06-13
[codex] Avoid Nostr embeds from URL hostnames

### LINKED ISSUE #2252: Don't parse npubs in links

### Describe the bug

When a npub is included in a link and you click the npub, you go to the parsed link for nostr npubs, instead of to the link in which it is embedded.

Example: https://stacker.news/items/1016746?commentId=1017280

### Screenshots

_No response_

### Steps To Reproduce

1. Add a link like https://npub1nsyte9neefm3jle7dg5gw6mhchxyk75a6f5dng70l4l3a2mx0nashqv2jk.nsite.lol/
2. See that npub in link is parsed as separate link

### Expected behavior

Npubs in links are not separate links themselves

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

caydyan:
Fixes #2252.

## Summary
- match Nostr IDs for embeds from the parsed URL pathname instead of the whole href
- keep `https://njump.me/npub...` parsing as a Nostr embed
- keep URLs like `https://npub....nsite.lol/` as normal raw links
- add URL parser coverage for both cases

## Why
The old whole-href regex sees the leading `//npub...` in hostnames and treats that hostname as a path Nostr ID. That makes ordinary links with npub-based hostnames behave like Nostr embeds/links instead of staying as the original URL.

## Prior context
PR #2253 was closed after the editor overhaul, but maintainer review there specified the current expected behavior: njump path links should embed, bare npubs should link to njump, and npub hostnames such as nsite links should stay raw. This patch applies the remaining URL-parser part to the current code path.

## Validation
- `npm run lint -- lib/url.js lib/url.spec.js`
- `npm test -- lib/url.spec.js`
- `git diff --check`

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -126,7 +126,7 @@ export function parseEmbedUrl (href) {
     const { hostname, pathname, searchParams } = new URL(href)
 
     // nostr prefixes: [npub1, nevent1, nprofile1, note1]
-    const nostr = href.match(/\/(?<id>(?<type>npub1|nevent1|nprofile1|note1|naddr1)[02-9ac-hj-np-z]+)/)
+    const nostr = pathname.match(/\/(?<id>(?<type>npub1|nevent1|nprofile1|note1|naddr1)[02-9ac-hj-np-z]+)/)
     if (nostr?.groups?.id) {
       let id = nostr.groups.id
       if (nostr.groups.type === 'npub1') {
--- a/lib/url.spec.js
+++ b/lib/url.spec.js
@@ -1,5 +1,6 @@
 /* eslint-env jest */
 
+import { nip19 } from 'nostr-tools'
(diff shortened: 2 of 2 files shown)


sir-opti:
Fixes `https://npub1nsyte9neefm3jle7dg5gw6mhchxyk75a6f5dng70l4l3a2mx0nashqv2jk.nsite.lol/` parsing.

Could use some more sunny case unit testing (because now it relies on QA)

caydyan:
Added broader sunny-case coverage in the current head `dbe0fd8`: npub, nprofile, note, nevent, and naddr IDs in URL paths now assert the decoded Nostr embed type/data, while the hostname-only npub case remains null. Re-ran locally: `npm test -- --runTestsByPath lib/url.spec.js --runInBand` (30 tests passing).

PR #3094, opened 2026-06-14
Preserve expanded long posts on back navigation

### LINKED ISSUE #2818: on browser back nav, long posts that have been opened are truncated, losing scroll position
1. view long post with internal links in it
2. expand 'view all'
3. click internal link
4. hit back button
5. observe post is collapsed again and scroll position is wrong

This may have always been broken, but we should be able to handle this by adding a hidden `GET` param (via `router.replace`) when the post is expanded.

caydyan:
## Summary
- closes #2818
- records long-post expansion in the current route as `fullText=<item id>` when `show full text` is clicked
- auto-expands only the matching item text when that query value is present, while preserving existing hash-link expansion
- preserves existing query params and hash fragments when adding the route state

## Validation
- `npm test -- lib/full-text.spec.js --runInBand --no-coverage`
- `npm test -- --runInBand --no-coverage`
- `npx standard components/text.js lib/full-text.js lib/full-text.spec.js`
- `git diff --check`
- `DATABASE_URL=postgresql://stacker:stacker@127.0.0.1:5432/stacker OPENSEARCH_URL=http://127.0.0.1:9200 NEXT_PUBLIC_URL=https://stacker.news npm run build`

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -6,6 +6,7 @@ import { useRouter } from 'next/router'
 import classNames from 'classnames'
 import { CarouselProvider, useCarousel } from './carousel'
 import { SNReader } from './editor'
+import { fullTextAsPath, shouldShowFullTextForPath } from '@/lib/full-text'
 
 export function SearchText ({ text }) {
   return (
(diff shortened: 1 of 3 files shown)


caydyan:
Closing this as a duplicate of #3060. I missed the existing draft PR while checking for active work because my initial scan filtered by a narrower title/head-ref pattern. This branch is only an alternate implementation, not intended to compete with the existing issue fix.

PR #3095, opened 2026-06-14
[codex] Preserve full text expansion in route

### LINKED ISSUE #2818: on browser back nav, long posts that have been opened are truncated, losing scroll position

caydyan:
## Summary
- store expanded item text state in a `fullText=<itemId>` route query param when a truncated post/comment is opened
- restore expanded text from that route state after back/forward navigation
- keep existing hash-link behavior so hash targets still reveal hidden text
- add focused coverage for full-text route helpers

## Why
Long posts collapse again after clicking an internal link and using browser back, which also breaks scroll restoration. Persisting the expanded item id in the route lets the item render fully when the browser returns to the previous entry.

Closes #2818.

## Testing
- `npm test -- --runTestsByPath lib/full-text.spec.js --runInBand`
- `npx standard components/text.js lib/full-text.js lib/full-text.spec.js`
- `DATABASE_URL=postgresql://stacker:stacker@127.0.0.1:5432/stacker OPENSEARCH_URL=http://127.0.0.1:9200 npm run build`

### DIFF
--- a/components/text.js
+++ b/components/text.js
@@ -6,6 +6,7 @@ import { useRouter } from 'next/router'
 import classNames from 'classnames'
 import { CarouselProvider, useCarousel } from './carousel'
 import { SNReader } from './editor'
+import { fullTextAsPath, shouldShowFullTextForPath } from '@/lib/full-text'
 
 export function SearchText ({ text }) {
   return (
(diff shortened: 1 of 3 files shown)


caydyan:
Closing this as a duplicate of #3060. I rechecked the issue history after opening this and confirmed the existing draft PR already covers #2818. This branch should not compete with that work.

PR #3108, opened 2026-06-16
fix: unescape brackets in link text

### LINKED ISSUE #2879: Brackets [] in link text are incorrectly escaped when roundtripped through MDAST pipeline
recreate:

1. in markdown: `[[Documentary] The Bitcoiner: is now a slave.](https://stacker.news/items/1454433)`
2. switch to `compose`
3. switch to markdown
4. switch back to `compose`

I suspect there's a precedence issue with autolink/internal links, but not for certain.

Soxasora:
## Description

Fixes #2879 by unescaping opening brackets that have a closing bracket in link nodes.

## Screenshots

https://github.com/user-attachments/assets/9d9eb778-c2c4-4af0-a5b5-a607c6b81dce

## Checklist

Yes

6:

- `[text [here](url)` textnode and link
- `[[text] here](url)` whole link
- `[text] here](url)` textnode

yes review and tests

---

> [!NOTE]
> **Low Risk**
> Single-path change to the Lexical→markdown link handler; improves escaping for edge-case link labels with no auth or data implications.
> 
> **Overview**
> Fixes Lexical→markdown export for links whose label text contains `[` or `]`, which could produce invalid markdown (e.g. `a]b` becoming a broken `[a]b](url)` instead of a proper link).
> 
> The custom **`link`** handler in `$lexicalToMarkdown` now enters mdast’s **`label`** construct while serializing link phrasing, so brackets are escaped symmetrically like standard link labels. **`containerPhrasing`** is called with `after: ']('` (instead of `']'`) and the handler’s **`info`** is passed through.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08bffdb928cbf03072cfdb7b83e6b0e85f8a0874. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/lexical/utils/mdast.js
+++ b/lib/lexical/utils/mdast.js
@@ -87,6 +87,26 @@ function escapeConfig (node, _parent, state, info) {
   return state.safe(node.value, info).replace(/&#x20;/g, ' ')
 }
 
+/** unescape `[` and `]` in link text if brackets are balanced
+ *
+ * brackets are balanced when the number of opening brackets is equal to the number of closing brackets
+ */
+function unescapeBalancedBrackets (text) {
+  if (text.includes('`')) return text
+
+  // drop the backslash before each escaped bracket: `\[` -> `[`, `\]` -> `]`
+  const unescaped = text.replace(/\\([[\]])/g, '$1')
+
+  // check if the brackets are balanced
+  let open = 0
+  for (const char of unescaped) {
+    if (char === '[') open++
+    else if (char === ']') open--
+    if (open < 0) return text // a closing bracket with nothing open
+  }
+  return open === 0 ? unescaped : text
+}
+
 export function $lexicalToMarkdown (transformerBridge = false) {
   const markdown = exportMarkdownFromLexical({
     root: $getRoot(),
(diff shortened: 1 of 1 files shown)


huumn:
This fixes the primary issue with brackets in labels well.

Bot recommended we could use mdast's builtin escaper for handling this which would allow us to handle a few more cases and more concisely. 

```js
function isBareLink (node) {
  return node.children?.length === 1 &&
    node.children[0].type === 'text' &&
    node.children[0].value === node.url &&
    !node.title
}

// in toMarkdownOptions.handlers:
link: (node, _parent, state, info) => {
  if (isBareLink(node)) return node.url
  // non-bare: force [text](url) and enter the 'label' construct so [ and ] are
  // escaped symmetrically (a lone ] would otherwise end the label early).
  const exit = state.enter('label')
  const text = state.containerPhrasing(node, { ...info, before: '[', after: '](' })
  exit()
  return `[${text}](${node.url || ''}${node.title ? ` "${node.title}"` : ''})`
}
```

Soxasora:
I think that it's much better that way. `unescapeBalancedBrackets` preserves unescaped balanced brackets resulting in `[[text]](url)`, while your suggestion results in `[\[text\]](url)` which is better because we're fixing markdown instead of making excuses for it.

I'll make some tests and push the patch ^^ Thanks!

Soxasora:
> `if (isBareLink(node)) return node.url`

The current behavior for autolinks is that `https://stacker.news` roundtrips as `https://stacker.news` if the conversion happens while we're still doing media checks (link is truly an `AutoLinkNode` in that moment). After media checks, it becomes a full `LinkNode` and roundtrips as `[https://stacker.news](https://stacker.news)`.

If we add the line above, bare URLs and URLs with equal `linkText` and `url` would produce bare URLs, no matter what, drifting slightly from the markdown spec. I'm happy with both cases, what do you think? @huumn

Note that `isBareLink` is not related to the brackets fix!

huumn:
> I'm happy with both cases, what do you think?

I'm good with the solution you went with. I added the bare link check because I thought it was necessary for autolinks.

PR #3112, opened 2026-06-18
fix: roundtrip empty Lexical nodes on mode switch

### LINKED ISSUE #3109: text of only empty table gets deleted on lexical->markdown transform

2. switch to compose 
3. switch back to write

Soxasora:
## Description

Fixes #3109 by not detecting empty nodes when switching editor modes.

Mode switches are based on the current Formik value. Triggered on each mode switch and submit, `syncFormik` checks if the Lexical editor state contains nodes but no text content. If so, it resets the `text` field, helping Formik validation catch empty item creation.

This should happen only on submission, not on mode switches. This PR makes mode switches call `syncFormik` with the flag `switchMode: true`, avoiding cleanup.

## Screenshots

https://github.com/user-attachments/assets/e92399f4-a09d-4165-94f2-a7fdcde1c68e

## Additional Context

This issue exacerbated another, bigger, issue: empty items can be created via `markdown` mode!
`markdown` mode always contains text content, `syncFormik` and `$isTextEmpty()` can't catch empty markdown nodes.
I'll open a separate issue for this.

## Checklist

Yes

7, empty content roundtrips on mode switches, submission from `compose` still doesn't allow empty nodes

n/a

n/a

No

---

> [!NOTE]
> **Low Risk**
> Localized editor/Formik sync and markdown export behavior; submit-time empty validation is unchanged when `switchMode` is false.
> 
> **Overview**
> Fixes mode-switch roundtripping where **compose** content with empty structural nodes (e.g. blank paragraphs) was wiped because `syncFormik` treated “no text” like submit-time validation and cleared Formik’s `text`.
> 
> `syncFormik` and `SYNC_FORMIK_COMMAND` now take an options object (`switchMode`, `flush`). The empty-text → `''` shortcut is **skipped** when `switchMode` is true. The toolbar dispatches sync with `{ switchMode: true }` before toggling modes.
> 
> **Code blocks:** Lexical → mdast export no longer forces `lang: null` for `text` or missing language—it writes the Lexical language value through for better roundtrip fidelity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fdc431efdb5258511973b5b51eb2bb2c31870fd3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/editor/plugins/core/formik.js
+++ b/components/editor/plugins/core/formik.js
@@ -33,9 +33,9 @@ export default function FormikBridgePlugin ({ name = 'text' }) {
     })
   }, 500, [editor, textHelpers, disableSubmit])
 
-  const syncFormik = useCallback((flush = false) => {
+  const syncFormik = useCallback(({ switchMode = false, flush = false } = {}) => {
     editor.getEditorState().read(() => {
-      if ($isTextEmpty()) {
+      if ($isTextEmpty() && !switchMode) {
         textHelpers.setValue('')
         disableSubmit(false) // always enable submit if text is empty
         return
@@ -69,7 +69,7 @@ export default function FormikBridgePlugin ({ name = 'text' }) {
     return editor.registerCommand(
       BLUR_COMMAND,
       () => {
-        syncFormik(true)
+        syncFormik({ flush: true })
         return false
       },
       COMMAND_PRIORITY_HIGH
@@ -79,8 +79,8 @@ export default function FormikBridgePlugin ({ name = 'text' }) {
   useEffect(() => {
     return editor.registerCommand(
       SYNC_FORMIK_COMMAND,
-      (flush = true) => {
-        syncFormik(flush)
+      ({ switchMode = false, flush = true }) => {
+        syncFormik({ switchMode, flush })
         return true
       },
       COMMAND_PRIORITY_HIGH
--- a/components/editor/plugins/toolbar/switch.js
+++ b/components/editor/plugins/toolbar/switch.js
@@ -37,7 +37,7 @@ export default function ModeSwitchPlugin ({ name }) {
 
         if (newMode === (isMarkdownMode(editor) ? MARKDOWN_MODE : RICH_MODE)) return false
 
-        editor.dispatchCommand(SYNC_FORMIK_COMMAND)
+        editor.dispatchCommand(SYNC_FORMIK_COMMAND, { switchMode: true })
 
         // toggle mode
         if (newMode) {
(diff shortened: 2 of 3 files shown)


sir-opti:
Tested to work with ul, ol, tables, code blocks. 

Odd behaviors spotted:
 
#### `text` markdown type gets stripped 

(not on post from `write`, so I suspect that it gets stripped on switch to `compose`?)

````markdown
```text
```
````

turns into

````markdown
```
```
````

#### checkboxes go wild

```markdown
- [ ]
- [ ]
- [ ] non-empty
- [ ]
- [ ]
```

turns into

```markdown
* \[ ]
* \[ ]

- [ ] non-empty

* \[ ]
* \[ ]
```

but on current master it turns into something even more confusing:

```markdown
* [ ] \[ ]
* [ ] \[ ]
* [ ] non-empty
* [ ] \[ ]
* [ ] \[ ]
 ```
Doesn't have to be blocking for this PR - can track it separately

**Comment by Soxasora** [comment]
> `text` markdown type gets stripped

Forgot to change this in the MDAST code block visitor! Before #2795, Lexical Shiki didn't support a `null` default language so we just treated `text` as the de-facto `null` language. Now that it's supported, I've removed this weird roundtrip hack.

> checkboxes go wild

Per GFM spec, checklist items that don't have trailing text are not checklist items. The behavior in prod is wrong, and got fixed by #3105.
Technically we can diverge from the spec whenever we want, and can fix this on the import side (markdown -> lexical) by forcing blank checklist items. But we cannot (comfortably) fix this on the export side (lexical -> markdown) because GFM injects checkboxes via a regex that needs whitespace after the bullet, fixing this would require overriding the `listItem` `toMarkdown` handler and lots of headaches.

imo, I prefer the behavior that we have now post-3105, it's correct and spec-compliant:

- [ ]
- [ ]
- [ ] non-empty
- [ ]
- [ ]

Thank you opti!!

**Comment by sir-opti** [comment]
> I prefer the behavior that we have now post-3105

Agreed. Thanks!

**Comment by zhailong8845-art** [comment]
I added a cherry-pickable regression test for the code-language part of this PR: https://github.com/zhailong8845-art/stacker.news/commit/c888ed82

The parameterized test verifies that Lexical export preserves `text`, another explicit language (`javascript`), and `null`, while also asserting the exported code value and append target.

Verification:
```
DATABASE_URL=postgresql://postgres:postgres@localhost:5431/stackernews NODE_OPTIONS='--experimental-vm-modules' npx jest lib/lexical/mdast/visitors/code.test.js --runInBand
# PASS: 1 suite, 3 tests

npx standard lib/lexical/mdast/visitors/code.test.js
git diff --check
# both pass
```

Feel free to cherry-pick `c888ed82`.

PR #3115, opened 2026-06-20
fix: resolve LNURL-pay usernames case-insensitively (#3058)

### LINKED ISSUE #3058: bug report: lnurl request fails for usernames with upper case letters

### Describe the bug

We tried paying ourselves to SwapMarket@stacker.news from coinos.io and coinos.pro, and it did not work. However, paying to k00b@stacker.news proceeds with no error.

Is that a capitalization issue? Can there be two different users on your platform, i.e. "swapmarket" and "SwapMarket"? If not, the lnulrp should be case insensitive.

### Screenshots

_No response_

### Steps To Reproduce

Pay a sat to SwapMarket@stacker.news from any wallet.

### Expected behavior

Should just work

### Logs

`lnurl proxy failed https://stacker.news/api/lnurlp/swapmarket/pay?amount=21000 Request failed with status code 400 (Bad Request): GET https://stacker.news/api/lnurlp/swapmarket/pay?amount=21000`

### Device information

_No response_

### Additional context

_No response_

protivon:
## Description

Fixes #3058. Paying a mixed-case Lightning address such as `SwapMarket@stacker.news` returned "user `@SwapMarket` does not exist", while the lowercase form worked.

Both LNURL-pay endpoints resolved the recipient with:

```js
models.user.findUnique({ where: { name: username } })
```

The `users.name` column is `Citext`, so equality in Postgres is case-insensitive — but Prisma's `findUnique` on the unique key does not reliably fold case for the bound parameter here, so a mixed-case `username` from the URL missed the row. The codebase already compensates for this elsewhere by comparing names with `.toUpperCase()` (see `api/resolvers/user.js`).

This PR makes the two LNURL-pay lookups explicitly case-insensitive:

```js
models.user.findFirst({ where: { name: { equals: username, mode: 'insensitive' } } })
```

in `pages/api/lnurlp/[username]/index.js` (the LNURL-pay metadata response) and `pages/api/lnurlp/[username]/pay.js` (the invoice/callback). Any casing of an existing username now resolves to the same account.

## Screenshots

N/A — server-side payment-resolution fix, no UI change.

## Additional Context

`name` is both `@unique` and `Citext`, so a case-insensitive match returns at most one row; `findFirst` is therefore equivalent to `findUnique` for valid data while tolerating casing. This cannot create users or change who an address maps to — it only lets an already-existing recipient be found regardless of the sender's casing.

## Checklist

Yes. Lowercase addresses behave exactly as before; only previously-failing mixed-case lookups now succeed. No schema, GraphQL, or API-shape changes.

7/10. Verified the root cause by reading the schema (`name … @db.Citext`, line 19 of `prisma/schema.prisma`) and confirming both LNURL-pay routes used the case-sensitive `findUnique`. Confirmed `mode: 'insensitive'` is supported on the project's Prisma 5.20 `findFirst`. Both changed files pass `node --check`. I did not spin up the full stack to send a live mixed-case LNURL payment, so an end-to-end test against a running node would be the remaining QA step.

N/A — no frontend changes.

### DIFF
--- a/pages/api/lnurlp/[username]/index.js
+++ b/pages/api/lnurlp/[username]/index.js
@@ -4,7 +4,9 @@ import { lnurlPayMetadata, lnurlpCallbackUrl } from '@/lib/lnurl'
 import { LNURLP_COMMENT_MAX_LENGTH, PROXY_PAYER_MIN_MSATS, PROXY_PAYER_MAX_MSATS } from '@/lib/constants'
 
 export default async ({ query: { username } }, res) => {
-  const user = await models.user.findUnique({ where: { name: username } })
+  // LNURL-pay usernames are case-insensitive (the `name` column is Citext);
+  // look up case-insensitively so e.g. SwapMarket@ resolves to swapmarket. (#3058)
+  const user = await models.user.findFirst({ where: { name: { equals: username, mode: 'insensitive' } } })
   if (!user) {
     return res.status(400).json({ status: 'ERROR', reason: `user @${username} does not exist` })
   }
--- a/pages/api/lnurlp/[username]/pay.js
+++ b/pages/api/lnurlp/[username]/pay.js
@@ -10,7 +10,9 @@ import { walletLogger } from '@/wallets/server'
 import pay from '@/api/payIn'
 
 export default async ({ query: { username, amount, nostr, comment, payerdata: payerData }, headers }, res) => {
-  const user = await models.user.findUnique({ where: { name: username } })
+  // Case-insensitive username lookup (the `name` column is Citext) so that
+  // mixed-case LNURL-pay addresses like SwapMarket@ resolve correctly. (#3058)
+  const user = await models.user.findFirst({ where: { name: { equals: username, mode: 'insensitive' } } })
   if (!user) {
     return res.status(400).json({ status: 'ERROR', reason: `user @${username} does not exist` })
   }


zhailong8845-art (changes requested):
The proposed lookup change is not yet shown to fix #3058. `users.name` is `@db.Citext`, and Prisma's PostgreSQL documentation says citext columns are always case-insensitive and are not affected by `mode: 'insensitive'`. That means replacing `findUnique({name})` with `findFirst({name: {equals, mode}})` may be query-shape churn with the same comparison semantics; the reported failure could be elsewhere in the LNURL route or account state.

Please add a focused route/integration regression that inserts a mixed-case user and demonstrates both descriptor and callback behavior through the requested alternate casing. The test needs to fail against the current `findUnique` implementation and pass with this change, while asserting the resolved user ID (or generated PROXY_PAYMENT recipient), metadata/description hash consistency, and successful callback response. If the old implementation already passes, the root cause needs to be reproduced and fixed instead of switching APIs.

This is payment routing, so a code-reading claim plus `node --check` is not sufficient evidence that the intended recipient is resolved. Keeping the unique lookup would also be preferable if the test confirms citext already provides the required comparison.

Reference: https://www.prisma.io/docs/orm/prisma-client/queries/case-sensitivity#postgresql-provider

PR #3118, opened 2026-06-22
Add @delete bot bio

### LINKED ISSUE #2994: The @delete bot is not well documented

I see @delete in 48 hours 
and can assume it is a bot, but no idea what that is for and search does not return anything that tells me more about it 

Add FAQ in Bio for @delete:
https://stacker.news/delete

just like is done for @remindme:
https://stacker.news/remindme

_No response_

### Additional context

_No response_

rianfyi:
## Summary
- add a migration that creates or updates the `@delete` bot bio
- document the `@delete in n units` syntax, supported units, and edit behavior

Closes #2994

## Testing
- `git diff HEAD^ HEAD --check`

### DIFF
--- a/prisma/migrations/20260622000000_delete_bot_bio/migration.sql
+++ b/prisma/migrations/20260622000000_delete_bot_bio/migration.sql
@@ -1,0 +1,55 @@
+set transaction isolation level serializable;
+
+CREATE OR REPLACE FUNCTION upsert_delete_bot_bio()
+RETURNS INTEGER
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    delete_user users;
+    bio_item "Item";
+    bio_text TEXT := $bio$
+I delete posts and comments after a delay.
+
+Use this format in the item or comment you want deleted:
+
+`@delete in 10 minutes`
+
+Other examples:
+
+`@delete in 2 hours`
+
+`@delete in 3 days`
+
+Supported units: `second`, `minute`, `hour`, `day`, `week`, `month`, and `year`. Plurals work too.
+
+If you include more than one valid `@delete in ...` command, the last one wins. Editing the item replaces the previous delete schedule. Editing it without a valid delete command clears the schedule.
+
+When the time arrives, the normal author-delete flow runs on the item containing the command.
+$bio$;
+BEGIN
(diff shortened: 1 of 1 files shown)


zhailong8845-art (changes requested):
This migration calls `create_bio(title, text, user_id)` when `@delete` has no `bioId`, but that database function no longer exists in the current migration history. It is explicitly dropped in `prisma/migrations/20230824064857_new_create_item/migration.sql` (`DROP FUNCTION IF EXISTS create_bio(...)`), and there is no later `CREATE ... create_bio` migration.

That branch will therefore raise PostgreSQL `undefined_function` (`42883`) during deployment—the exception handler only catches `undefined_table` (`42P01`)—and abort the migration instead of creating the requested bio. Please create the bio through the current item-creation path (or add self-contained SQL compatible with the current schema) and verify this migration from a database migrated through current main with the `delete` user present and `bioId IS NULL`.

zhailong8845-art:
I prepared a minimal current-schema fix so this is directly actionable: https://github.com/zhailong8845-art/stacker.news/commit/7e463ff2

It replaces the removed `create_bio()` call with a direct `Item` insert (letting the existing path trigger/defaults run), captures the returned row, and sets `users.bioId` in the same migration transaction. It preserves the existing update branch.

Validation performed:
- `git diff --check`
- migration-history assertion: no `create_bio` recreation after the 2023 drop
- new-path assertions: direct insert returns `bio_item` and updates the matching delete user

You can apply it with:
```sh
git cherry-pick 7e463ff2
```
A full migrated-Postgres execution is still the final acceptance check; I did not represent the static assertions as a live database run.

zhailong8845-art:
PostgreSQL 16 execution evidence is now green: https://github.com/zhailong8845-art/stacker.news/actions/runs/29911162211

The isolated verification applies the repaired migration in a transaction against a focused current-schema fixture and proves both branches:
- `delete` user with `bioId IS NULL`: creates exactly one bio, runs the item path trigger, and links `users.bioId`
- existing bio: restores title/text/bio/deletedAt without creating a duplicate

The production patch to cherry-pick remains the single focused commit `7e463ff2`; the later commits only contain fork-side verification infrastructure.

Issue #3122, opened 2026-06-25
Analytics tooltip is not correctly positioned


Soxasora:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

[0xbitcoiner](https://stacker.news/0xbitcoiner) discovered that the analytics tooltip displayed on hover is being obscured by other analytics elements.

### Screenshots

![](https://imgprxy.stacker.news/sWhKEttBMM4sfIfhhqbHJCdKx0opp-uLPrgI3gWOWx8/rs:fit:2560:1440/aHR0cHM6Ly9tLnN0YWNrZXIubmV3cy8xNDYyMjA)

### Steps To Reproduce

1. go to analytics
2. hover on a graph with lots of data types (like `unique spenders`)
3. the tooltip might be shadowed by the graph below

### Expected behavior

The tooltip is not shadowed by other elements

### Device information

- Browser: Safari, Chrome
- OS: macOS

### Additional context

https://stacker.news/items/1514684

JhoanChala:
I opened a PR that fixes this: #3168. Root cause was a \zIndex: -1\ on the \WhenAreaChart\ tooltip that rendered it behind the adjacent chart columns. Could this issue get a \difficulty\ tag assigned? Thanks!

matthewrball:
This looks like a small, contained UI bug (tooltip z-index / overflow on the analytics graphs).

Could this get a `difficulty` tag so a PR closing it has a posted award amount? Happy to take it if it's in the good-first-issue / easy range.

okukkiti:
Fix ready for review: #3224 — tooltip stacking moved to `wrapperStyle={{ zIndex: 10 }}` on all three chart Tooltips, `zIndex: -1` removed. Could a maintainer add a `difficulty:*` label so the bounty table applies?

nomadicco0-star:
Hi — I'd like to work on this if it's still open.

Could a maintainer please add a `difficulty:*` tag (and `priority:*` if relevant) so it's clear what the PR award would be under the current bounty rules? Happy to ask clarifying questions on the issue before opening a draft PR.

Thanks

PR #3124, opened 2026-06-29
fix analytics tooltip stacking

sh962214-hub:
﻿Fixes #3122.

## Summary
- remove the negative z-index from area chart tooltip content
- keep the tooltip background opaque like the other analytics charts

## Testing
- `npm run lint -- components/charts.js`


zhailong8845-art (approved):
Verified against #3122 and the surrounding `WhenAreaChart` configuration. The obstruction is caused by the tooltip content’s explicit `zIndex: -1`; removing only that negative stacking value preserves the existing opaque theme-aware content styling and leaves the chart/legend behavior unchanged. The one-line scope matches the reported Safari/Chrome overlap bug.

PR #3121, opened 2026-06-24
Wallet receive: external transaction table and checks

huumn:
I'm happy with the shape of this now.

This contains the core logic/relations for external transactions (ie payments that SN is not party to) + the receive side of external txs. Send side will be its own PR.

Vaguely:

- creates a db table for external txs
- polls server-side in a best effort manner for receive verification where wallets support them
- extends scheduled bolt11 deletion to external txs
- unions external txs with payins in `/satistics` and `/wallets/[id]/activity`
- implements `/wallets/[wid]/transactions/[id]` for viewing external txs, sharing components/design with payins `/transactions/[id]`
    - on `make invoice` we navigate to this transaction page (we'll do this for send too)
- adds `checkInvoice` to all wallets that have some means of checking invoice status
    - for wallets that don't support `checkInvoice`, or if an invoice expires and we don't have a definite settled/failed reported by the wallet, external tx statuses are marked as `unknown` and we attempt to report why it's unknown

Punted to other PRs:
- send with send checks
- external tx receive notifications
- proxy on/off global setting with per receive proxy option

---

> [!NOTE]
> **High Risk**
> Large cross-cutting change touching wallet payments, new persisted financial rows, background reconciliation, and GraphQL/API contract breaks (`Satistics.txs`, `createWalletInvoice` return type). Incorrect settlement or timeline merging could misreport user balances or activity.
> 
> **Overview**
> Introduces **external wallet transactions** for Lightning receives that are not PayIn flows. **`createWalletInvoice`** now persists an `ExternalTransaction`, enforces a cap on recent unpaid receives, and returns the transaction id instead of a bare bolt11; the receive form redirects to **`/wallets/transactions/[id]`** for QR, polling, logs, and invoice details.
> 
> A new **`ExternalTransaction`** table (Prisma migration) tracks direction, outcome, reconciliation schedule (`nextCheckAt`), amounts, and optional **`verificationContext`** (e.g. LNURL LUD-21 verify URL). Wallet logs can link via **`externalTransactionId`**. A pg-boss job **`checkPendingExternalTransactions`** claims due rows and calls **`checkInvoice`** on receive protocols (Blink, CLN, LNbits, LND, NWC, Phoenixd, LNURL verify, etc.), mapping observations to SETTLED/FAILED/EXPIRED/UNKNOWN with structured **`unknownReason`** when proof cannot be confirmed.
> 
[...]

PR #3125, opened 2026-07-01
wallet external transaction send/checks

huumn:
Send side of external transactions. Like #3121 we periodically lease external transactions, poll the wallet for outcomes (but on the client), then report terminal outcomes to the server.

## bolt11 outcome contract

- we only mark a payment as `FAILED` if we are confident it failed (e.g. the payment wasn't sent, or the implementation is single-source, trustworthy, and reports `FAILED` on `checkPayment`)
    - this is important because bolt11s that are marked as `FAILED` can be retried and non-`FAILED` bolt11s prevent other attempts to pay the same bolt11
- any non-null stored outcome is terminal
- null means pending
- we accept wallet adapters reporting `SETTLED` with an invalid preimage because for some intraledger sends of custodial wallets, the preimage is a uuid
    - in such cases, we set the preimage to null and display a warning
- terminal `UNKNOWN` is set with an `unknownReason` as a diagnostic
- terminal `UNKNOWN`, like `SETTLED`, or `null` (pending), blocks other attempts to pay that bolt11

## lightning address related stuff

- the same amount sent to the same lightning address in the last hour requires client confirmation (to prevent double pays)
- we store the `lnurl-verify` callback (if one is available) in `verificationContext` so that it can be used to determine outcomes if the send wallet lacks `checkPayment` abilities

---

> [!NOTE]
> **High Risk**
> Changes payment initiation, outcome classification, and double-pay prevention across GraphQL, DB constraints, and many wallet adapters—errors in terminal vs UNKNOWN handling could mislead users into retrying live payments.
> 
> **Overview**
> Adds **external wallet sends** as durable `ExternalTransaction` rows: GraphQL **`createExternalSend`** / **`reportExternalSendObservation`**, plus **`pendingExternalSendTransactions`** so the browser can lease due in-flight sends and poll wallet adapters before the reconciliation grace window ends.
> 
> The send UI is turned on (`SEND_COMING_SOON` off). Flow is **record attempt → pay via protocol → report observation → navigate to the transaction detail page** (inline “sent” success UI removed). Repeat **LN address** payments within an hour require an explicit confirmation modal; **`duplicateConfirmed`** bypasses that check.
> 
> **Wallet protocol adapters** now return explicit payment results (`PENDING` / `SETTLED` / `FAILED` / `UNKNOWN`) instead of preimage-only success; many add **`checkPayment`**. **`sendWalletPayment`** normalizes outcomes, optionally waits on `checkPayment` for terminal status, and classifies errors so ambiguous failures stay **UNKNOWN** rather than inviting retries. Custodial withdrawal creation also calls **`requireNoConflictingExternalSend`** for the same invoice hash.
> 
> **Server/worker reconciliation** extends to outbound sends (late grace-period checks, LNURL verify when `checkPayment` is missing). DB migration adds partial unique indexes so one non-failed send per user/hash is enforced. Wallet activity (`satistics`) treats external **SEND** rows as sends (signed amounts, sort keys) and fixes the PayIn union leg ordering so pagination stays correct.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89691bf9225279cefd80ca2a956932c717f42594. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


wallets/client/hooks/use-external-send-checks.js
```diff
@@ -0,0 +1,200 @@
...
+  const client = useApolloClient()
+  const { me } = useMe()
+  const wallets = useWallets()
+  const [updateExternalTransaction] = useMutation(UPDATE_EXTERNAL_TRANSACTION)
+  // Keep wallets fresh without restarting the poll loop on cache changes.
+  const walletsRef = useRef(wallets)
+
+  useEffect(() => {
+    walletsRef.current = wallets
+  }, [wallets])
```
huumn: I'm not sure this is haram

wallets/lib/protocols/blink.js
```diff
@@ -160,6 +164,23 @@ async function getTransaction ({ field, varName, varType, value }, { apiKey, wal
...
+function bestTransaction (txs = []) {
+  return [...txs].sort((a, b) => transactionRank(b) - transactionRank(a))[0] ?? null
+}
+
+function transactionRank (tx) {
+  if (tx?.status === 'SUCCESS') return 3
+  if (tx?.status === 'PENDING') return 2
+  if (tx?.status === 'FAILURE') return 1
+  return 0
+}
```
huumn: inline declare?

PR #3132, opened 2026-07-21
deps: Bump 10 deps from 07/21 - 07/22 advisories

sir-opti:
## Description

Minor updates to deps. Mostly dev/compile-time hardening for SN specifically, and 2 defense-in-depth improvements (`body-parser` and `next`). The rest is simply noise-reduction.

- `js-yaml` 4.2.0->4.3.0: dev/ci/hardening only, no yaml is read on a hot path. Done in a separate commit because `package.json` overrides this explicitly to force compliance.
- `brace-expansion` 5.0.6->5.0.7 / 2.0.3->2.1.2 / 1.1.13->1.1.16: no hot path
- `body-parser` 2.2.2->2.3.0: provides defense-in-depth to Apollo
- `protobufjs` 7.6.4->7.6.5: no hot path for the vuln
- `next`: 16.2.6->16.2.11: No hot paths outside of the private beta (and there this is hardening only)
    - also updates `@next/env` and `@next/swc-*` with exactly the same versions
- `sharp`: 0.34.5->0.35.0: Override, temporary. Due to the `.npmrc` config `legacy-peer-deps=true` this gets installed, but it is fully unused and explictly disabled. This override removes the complaint as a temporary solution until we can prevent this from being installed altogether.
- `dompurify`: 3.4.11->3.4.12: Minor, no hot paths.
- `fast-uri`: 3.1.2->3.1.4: No hot paths
- `immutable`: 4.3.8->4.3.9: No hot paths
- `svgo`: 3.3.3->3.3.4: dev/compile-time hardening
- `next-auth`: 4.24.13->4.24.15: Hardens email and oauth auth
    - override @auth/core 0.37.0->0.41.3 to enforce it on all transient dependencies from elsewhere
- `postcss`: 8.5.10->8.5.22: build-time only defense in depth

### Additional details

I'll work on a separate PR to reduce the insane amount of unused packages coming from `legacy-peer-deps=true`, will try to time that when there is less active fluctuation in packages, so that we can compare apples to apples.

## Checklist

Yes.

8/10:

- spot-checked touched upon paths: item resolution and creation, gql, basic invoicing
- spot-checked some SN frontend complexity for the nextjs update: image upload, zapping, invoice gate on item create, deep comment nesting, item deletion, login/logout
- reviewed the diff of every changed package (including 2nd/3rd level transient ones) for malware / oddities, none found, except:
   - very minor note: the `next@16.2.11` diff with `16.2.6` contains an odd new error mapping from their pr `#96010` for error number 1444, that I cannot seem to resolve to an actual error, but that's at its worst dead "code", and not blocking.

N/A

No.

No.


sir-opti:
Put back into draft for a bit, because a bunch of reports just made it through npmsec.

sir-opti:
Re: "obfuscation". 

It is on the Apple login provider, since [next-auth `#12068`](https://github.com/nextauthjs/next-auth/pull/12068). There wasn't much discussion about this on that PR, but even if it weren't that the web is full of warnings about this not working on Apple OpenID, SN doesn't implement Apple federated login.

huumn:
Thank you!!

PR #3133, opened 2026-07-21
Only count recent items toward unpaid item limit

ekzyis:
## Description

The user with id 25176 has many unpaid items (accumulating a lot of bad karma 😅) from a long time ago. The user is currently unable to post unless they have paid for one of their last three items. However, this requires a payment of at least 1m sats. Since the unpaid items are over six months old, I'm not sure whether this is the intended behavior of the "too many unpaid items" filter. So I suggest adding a time window.

## Checklist

yes

`1`. I ran `sndev start` and looked at the code, with special consideration for time-related bugs, that's it

n/a

no

Yes and yes

### DIFF
--- a/api/payIn/types/itemCreate.js
+++ b/api/payIn/types/itemCreate.js
@@ -1,4 +1,5 @@
-import { ANON_FEE_MULTIPLIER, ANON_ITEM_SPAM_INTERVAL, ITEM_SPAM_INTERVAL, PAID_ACTION_PAYMENT_METHODS, USER_ID } from '@/lib/constants'
+import { ANON_FEE_MULTIPLIER, ANON_ITEM_SPAM_INTERVAL, ITEM_SPAM_INTERVAL, PAID_ACTION_PAYMENT_METHODS, UNPAID_ITEM_SPAM_WINDOW_HOURS, USER_ID } from '@/lib/constants'
+import { datePivot } from '@/lib/time'
 import { notifyItemMention, notifyItemParents, notifyMention, notifyTerritorySubscribers, notifyUserSubscribers, notifyThreadSubscribers } from '@/lib/webPush'
 import { getItemMentions, getMentions, performBotBehavior, getSubs } from '../lib/item'
 import { extractMentions } from '@/lib/lexical/server/mentions'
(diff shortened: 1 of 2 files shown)


huumn:
Did you try deleting the unpaid items?

zhailong8845-art (changes requested):
The time window changes an abuse-prevention/payment gate, but the PR has no regression coverage and the checklist reports only 1/10 QA. Please add focused coverage around `validateBeforeCreate` before merging.

At minimum, the tests should prove these three boundaries with controlled timestamps:

1. three unpaid non-freebie items inside `UNPAID_ITEM_SPAM_WINDOW_HOURS` still reject the next creation;
2. only two qualifying items inside the window do not reject it;
3. an otherwise identical third unpaid item just outside the cutoff no longer contributes to the rejection.

The existing `api/payIn/__tests__/integration` ITEM_CREATE harness looks like the appropriate place. Freebie comments, bios, and anonymous users are early-return paths and need not be expanded unless the fixture setup touches them. This will catch a future accidental change from `gt` to `gte`, a wrong pivot sign/unit, or moving the cutoff to the wrong query while documenting the intended rolling-window behavior.

zhailong8845-art:
I implemented the requested timestamp-window coverage as a cherry-pickable commit: https://github.com/zhailong8845-art/stacker.news/commit/3e998ca5

It covers:
- rejection with 3 unpaid items inside the window
- allowance with only 2 unpaid items inside the window
- allowance when the third unpaid item is 1 ms outside the window

Verification:
```
DATABASE_URL=postgresql://postgres:postgres@localhost:5431/stackernews NODE_OPTIONS='--experimental-vm-modules' npx jest --config api/payIn/__tests__/jest.config.js api/payIn/__tests__/itemCreateValidation.test.js --runInBand
# PASS: 1 suite, 3 tests

npx standard api/payIn/__tests__/itemCreateValidation.test.js
git diff --check
# both pass
```

Feel free to cherry-pick `3e998ca5` or adapt the test file.

ekzyis:
> Did you try deleting the unpaid items?

No, there are a lot, so I would need to script it. You think it could work because `recentItems.every(item => item.itemPayIns.length === 0)` could become `false`?

> The time window changes [... good to assume this is unverified AI slop ...] 

@zhailong8845-art, I'm sorry, looking at your recent activity in this repository, I think you're spamming, which discourages me from reading your review. If you want humans to engage with you, engage with them in a human way.

PR #3155, opened 2026-07-28
fix: use YouTube oEmbed API for title autopopulation

### LINKED ISSUE #2164: Autopopulate "title" field when posting a Youtube link with the title of the video rather than with "- Youtube"

qwenbona:
Fixes #2164

## Problem
When posting a YouTube link, the title field autopopulates with "- YouTube" instead of the actual video title. This happens because YouTube's server-rendered HTML doesn't include the video title — it's rendered client-side via JavaScript. The existing `h1 > yt-formatted-string.ytd-watch-metadata` selector never matches in a server-side fetch.

## Fix
Use YouTube's oEmbed API (`https://www.youtube.com/oembed`) which reliably returns the video title without needing client-side JS execution. This handles `/watch`, `/shorts/`, and `youtu.be` URLs.

Non-YouTube URLs continue to use the existing HTML metadata scraping path.

### DIFF
--- a/api/resolvers/item.js
+++ b/api/resolvers/item.js
@@ -548,21 +548,27 @@ export default {
     pageTitleAndUnshorted: async (parent, { url }, { models }) => {
       const res = {}
       try {
-        const response = await snFetch(url, { protocol: 'http', redirect: 'follow', size: 2 * 1024 * 1024 })
-        const html = await response.text()
-        const doc = domino.createWindow(html).document
-        const titleRuleSet = {
-          rules: [
-            ['h1 > yt-formatted-string.ytd-watch-metadata', el => el.getAttribute('title')],
-            ...metadataRuleSets.title.rules
-          ]
-        }
-        const metadata = getMetadata(doc, url, { title: titleRuleSet, publicationDate: publicationDateRuleSet })
-        const dateHint = ` (${metadata.publicationDate?.getFullYear()})`
-        const moreThanOneYearAgo = metadata.publicationDate && metadata.publicationDate < datePivot(new Date(), { years: -1 })
+        const urlObj = new URL(ensureProtocol(url))
+        const { hostname, pathname } = urlObj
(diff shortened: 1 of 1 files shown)


api/resolvers/item.js
```diff
@@ -548,21 +548,27 @@ export default {
...
-          ]
+        const urlObj = new URL(ensureProtocol(url))
+        const { hostname, pathname } = urlObj
+        const isYoutube = (hostname.endsWith('youtube.com') && (pathname === '/watch' || pathname.startsWith('/shorts/'))) ||
+          hostname.endsWith('youtu.be')
+
+        if (isYoutube) {
+          const oembedUrl = `https://www.youtube.com/oembed?url=${encodeURIComponent(urlObj.toString())}&format=json`
+          const oembedRes = await snFetch(oembedUrl, { protocol: 'http' })
+          const oembed = await oembedRes.json()
```
fth040818, line 559: This still fails the issue's documented production counterexample, `https://www.youtube.com/watch?v=mvuWLob3CFU&t=817s`: its oEmbed request currently returns HTTP 403 with a plain-text `Forbidden` body. `oembedRes.json()` therefore throws, the outer empty catch swallows it, and the resolver returns no title. I reproduced that response against the endpoint while reviewing this head. Because every YouTube URL now bypasses the existing metadata path, transient/non-2xx oEmbed responses also regress URLs whose HTML metadata was usable. Please check the response/status/title and fall back to the metadata scraper (while rejecting the known-bad `- YouTube` value), with a mocked 403/non-JSON test.

fth040818 (commented):
AI-assisted review; live endpoint behavior and the resolver path were independently verified.

PR #3167, opened 2026-08-03
fix: preserve empty tables when leaving rich editor mode (fixes #3109)

### LINKED ISSUE #3109: text of only empty table gets deleted on lexical->markdown transform

JhoanChala:
**Problem:** An empty table (\|   |\n| - |\) gets deleted when switching from rich (compose) mode back to markdown (write) mode. The markdown->lexical direction survives; the lexical->markdown direction loses the table.

**Root cause:** \\()\ (lib/lexical/utils/index.js) treated an empty table as "empty" because \TableNode.getTextContent()\ returns \''\ for a table whose cells have no text. In \FormikBridgePlugin\ (\syncFormik\), the early \\()\ check then ran \setValue('')\ instead of \setValue(\(editor))\, wiping the table's markdown on the first sync after entering rich mode.

**Fix:** \\\ now only reports an empty editor when there is no text AND every top-level child is an empty paragraph. Structural content without text (e.g. an empty table) is preserved.

**Testing:** Added \lib/lexical/utils/is-text-empty.test.js\ covering empty tables, tables with text, only empty paragraphs, an empty root, and a paragraph with text. Full suite passes (232 tests, 11 suites) and standard lint is clean.

### DIFF
--- a/lib/lexical/utils/index.js
+++ b/lib/lexical/utils/index.js
@@ -31,7 +31,11 @@ export function $getTextContent (trimWhitespace = true) {
 }
 
 export function $isTextEmpty () {
-  return $getTextContent().trim() === ''
+  if ($getTextContent().trim() !== '') return false
+  // structural content with no text (e.g. an empty table) is not an empty editor
+  return $getRoot().getChildren().every((child) =>
+    $isElementNode(child) && child.isEmpty() && $isParagraphNode(child)
+  )
 }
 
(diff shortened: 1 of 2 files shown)


JhoanChala:
Closing this PR since #3112 already addresses #3109 more comprehensively (tables + code blocks + lists) and has maintainer approval.

PR #3171, opened 2026-08-03
fix: confirm before discarding comment edit changes (fixes #2568)

### LINKED ISSUE #2568: Cancel on edit does not prompt for confirmation

JhoanChala:
**Problem:** Clicking "cancel" while editing a comment silently discards unsaved changes.

**Fix:** While editing, a DirtyTracker inside the comment edit form compares the current editor content against the original comment text. When clicking "cancel" with unsaved changes, a confirmation modal ("Discard your changes?") is shown instead of immediately closing the editor. Cancel without changes behaves exactly as before (no prompt).

Implementation notes:
- The comparison is against the original comment text (not the empty string), so untouched edits never trigger a prompt and the modal disappears again if the user reverts to the original content.
- The editor flushes its formik value on blur (BLUR_COMMAND), which fires before the cancel click, so even debounced rich-mode saves are captured.
- DirtyTracker lives inside the Form so it has access to formik values via useFormikContext.

### DIFF
--- a/components/comment-edit.js
+++ b/components/comment-edit.js
@@ -5,8 +5,21 @@ import { FeeButtonProvider } from './fee-button'
 import { ItemButtonBar } from './post'
 import { UPDATE_COMMENT } from '@/fragments/payIn'
 import useItemSubmit from './use-item-submit'
+import { useFormikContext } from 'formik'
+import { useEffect } from 'react'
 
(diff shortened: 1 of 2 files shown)


JhoanChala:
Closing this PR since #2592 already implements the cancel-confirmation for #2568 and has maintainer review. This PR would be a duplicate.

PR #3189, opened 2026-08-15
deps: bump deps for hygiene

sir-opti:
## Description

Bump deps, hygiene-only. All of these are transient and no bugs that this fixes have a hot path. This purely removes a bunch of alerts so that we don't miss any new ones that do have impact.

hard overrides in `package.json`:
- `brace-expansion` 1.1.16 / 2.1.2 / 5.0.7 -> 5.0.9 (new override, dev-only)
    - 1.1.16 and 2.1.2 came from old `minimatch` versions that aren't used anywhere
    - 5.0.7 was used by `eslint`
    - I played with the idea to also hard override `minimatch` but honestly, there should be upstream fixes for this instead.
- `js-yaml` 4.3.0 -> 4.3.1 (dev-only)
- `postcss` 8.5.22 -> 8.5.26 (used hot by next but there no hot paths for this codebase for the involved bugs)

transient deps bumped/dropped through `npm audit fix`:
- `balanced-match` 1.0.2 -> 4.0.4 (dev-only)
- `concat-map` 0.0.1 -> dropped
- `fast-uri` 3.1.4 -> 3.1.5 (dev-only)
- `nanoid` 3.3.16 -> 3.3.18 (dev-only)

## Checklist

Yes.

7/10: 

- tested that CI processes (linter, tests) still work
- tested dev deployment and that everything loads/renders ok
- analyzed every changed dependency for regressions
    - `postcss` has a single potential regression but it doesn't affect SN codebase directly - I'll see if I can throw them a PR. The only thing that can be affected by this change is someone running a dirty plugin that messes with css.
- analyzed every changed dependency for malware (let's see if this PR gets called out by our friends at socket)

N/A

No.

I ran some batch analyses on the dependency code and a bunch of scans, everything was gated by meat inspection.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9460,29 +9460,6 @@
         "typescript": ">=4.8.4 <6.1.0"
       }
     },
-    "node_modules/@typescript-eslint/typescript-estree/node_modules/balanced-match": {
-      "version": "4.0.4",
-      "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-4.0.4.tgz",
-      "integrity": "sha512-BLrgEcRTwX2o6gGxGOCNyMvGSp35YofuYzw9h1IMTRmKqttAZZVU67bdb9Pr2vUHA8+j3i2tJfjO6C6+4myGTA==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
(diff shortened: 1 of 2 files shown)


sir-opti:
force-push: commit message contained an error.

PR #3196, opened 2026-08-20
refactor: replace Bootstrap with Tailwind CSS and Base UI

### LINKED ISSUE #3190: Migrate to TailwindCSS and Base UI
This issue tracks the progress towards a new UI system with a modern and explicit in-house architecture that we control.

The migration is currently broken up into 4 PRs:

1. Style vocabulary
  [Bootstrap utilities to Tailwind](https://github.com/stackernews/stacker.news/pull/3126)
2. Component behavior
  [React Bootstrap to Base UI](https://github.com/stackernews/stacker.news/pull/3187)
3. Framework removal
  [Delete Bootstrap and normalize ownership](https://github.com/stackernews/stacker.news/pull/3188)
4. SN style remodel
  tbd

Soxasora:
## Description

Part of #3190, supersedes #3126, #3187, and #3188.

This PR replaces Bootstrap with Tailwind CSS v4, Base UI primitives, and shared SN components while preserving the existing design.

- converts Bootstrap layout, color, and utility classes across components, pages, editor, and wallet to Tailwind.
- replaces React Bootstrap components with shared SN UI components built on Base UI
- breaks down the Form system into dedicated components
  - `react-select` (multiselect) is replaced by Base UI Combobox
- introduces canonical `--sn-*` design tokens for light, dark, and territory-branded themes, shared `z-index` stacking, mobile safe areas, and reusable component recipes.
- replaces the global Sass bundle
  - with ordered CSS entry points:
    - tokens
    - Tailwind
    - base styles
    - app styles
    - KaTeX
    - editor text
  - removes Bootstrap, React Bootstrap, Sass, and all remaining SCSS.
- places every CSS module in `@layer components`, allowing call-site utilities to override component defaults consistently.
- TEMP: lint checks enforce module layering, statically discoverable Tailwind classes, token-based colors, and removal of Bootstrap variables.

Base UI brings keyboard navigation, focus management, portal positioning, and dismissal behavior for interactive components.
The most relevant changes include swipe-dismissable toasts and mobile drawer, improved form semantics and link editor.

## Screenshots

tbd, even though the migration targets visual parity

## Additional Context

This PR touches almost every single area that's displayed to stackers.
The most important areas to QA are **forms**, editor interactions, mobile navigation, wallet onboarding (new!), and territory-branded views (less priority).

## Checklist

Yes, this PR only handles the styling implementation.

6/10, components were checked as they were migrated. Pending a full end-to-end regression pass.

Yes, during component migration. Pending a full parity pass.

No

Yes, substantially. It assisted with mechanical Bootstrap-to-Tailwind conversions, Base UI rewrites, token and cascade analysis, and follow-up fixes. The resulting changes were then reviewed and adjusted manually.

---

> [!NOTE]
> **Medium Risk**
> Broad frontend refactor touching forms, editor, navigation, and theming with high regression surface area, though business logic and APIs are largely unchanged.
> 
> **Overview**
> Replaces **Bootstrap / React Bootstrap** across the app with **Tailwind CSS v4**, **Base UI**, and shared `@/components/ui/*` primitives, while aiming for visual parity.
> 
> **Design system:** README and new styling docs describe the stack. Theme switching moves from `data-bs-theme` to **`data-theme`**. Colors and charts use canonical **`--sn-*` tokens** instead of `--bs-*` / `--theme-*`. CSS modules are wrapped in **`@layer components`** so Tailwind utilities can override defaults.
> 
> **Component migration:** Dropdowns, menus, alerts, badges, tooltips, popovers, nav, buttons, and accordions are reimplemented (e.g. `accordian-item` → **`accordion-item`** on `Collapsible`). Bootstrap layout utilities become Tailwind (`d-flex` → `flex`, etc.). The monolithic **`components/form.js`** is split into dedicated form modules (`InputAddon`, Base UI **Checkbox**, `CopyInput`, **`useFieldDraft`** / `fieldDraftKey` for drafts).
> 
> **Editor & capture:** Lexical toolbar uses Base UI **Menu** and **Toolbar**; link editing uses a **Popover** anchored to the link node instead of manual floating positioning. Mentions autocomplete uses styled listbox markup. Screenshot capture hides nav via **`[data-sn-navigation]`**. Carousel arrow keys are handled on a focused container (for dialog key handling).
> 
> **Behavioral tweaks:** Delete flow refactored so **MenuItem** handlers work with portaled menus; cache updates clear **`lexicalState`/`html`** on delete. Badges gain optional interactive popovers and screen-reader text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit feec16ceb0b8f59d37d68543c84cf9b751c5ef53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


huumn:
Code looks fine as I glanced through it all - much better organized than before. I'm making my way through a huge QA doc that I had Claude make me. Other than the below, at least so far, I'm mostly finding improvements or no perceivable difference relative to prod.

I found a bug with link editor popovers:

https://github.com/user-attachments/assets/553f1ab1-f374-4fe2-bdb0-da56e6a3138b

Looks like show/hide of passphrase inputs has a regression:

https://github.com/user-attachments/assets/5e893e7c-5310-412d-9b55-bc4802b81109

Visual regression here (info icon is to the right):

Screenshot 2026-09-02 at 6 08 11 PM

Embed blocks have a body-colored background rather than sharing the background of the comment/context:

Screenshot 2026-09-02 at 6 22 22 PM

No spacing around `@anon` dropdown buttons:

Screenshot 2026-09-02 at 6 24 39 PM

Awkward spacing between profile photos and nyms in account switcher:

Screenshot 2026-09-02 at 6 24 48 PM

Success toasts don't appear to auto-dismiss. Also they keep counting after they are closed and replace each other. You mentioned something related in the meeting. I'm not sure if it's the same thing.

https://github.com/user-attachments/assets/f7c18e97-ebb1-4f69-aca9-352f0ae1263a

Dropdown in sticky navbar does not disappear when the navbar does:

https://github.com/user-attachments/assets/7802c942-2499-4766-afcf-00b8a7822867

Horizontal scroll on territory/nym mentions:

Screenshot 2026-09-02 at 6 48 23 PM

---------

As for style nits:

- post badges (territories, freebies, etc) are too prominent imo
- the territory combobox when posting is way too big (vertically)
- cowboy hat feels vertically misaligned Screenshot 2026-09-02 at 5 55 59 PM
- vertical positioning of the passphrase numbering labels are off Screenshot 2026-09-02 at 6 03 50 PM
- bottom spacing of `switch wallet` feels a bit excessive Screenshot 2026-09-02 at 6 14 30 PM


-----------

I'll keep truckin through it and update this list.

Soxasora:
These are amazing findings! Thank you!! I think I'll update this PR with fixes and cleanup from #3216.

Soxasora:
> Visual regression here (info icon is to the right):
> Screenshot 2026-09-02 at 6 08 11 PM

It's like this in prod too:
IMG_3622

Do you mean that it would be better if the (i) info icon were adjacent to the last word?

huumn:
> It's like this in prod too:

Ah, my recreation attempt must've been poor.  

> Do you mean that it would be better if the (i) info icon were adjacent to the last word?

Yeah. My bad.

huumn:
Have you tried QAing a production build? I got one finding from claude that implies, if true, that css minification might break some css rules.

> The chip's padding/border/background/color live on the single-class .chip rule, which ties (same layer, same specificity) with ui/collapsible's .trigger { border: 0; padding: 0; background: none; color: inherit; font: inherit } now composed onto the same button. The two-class rule the comment relies on only sets custom properties and min-height, so it arbitrates nothing. In dev, style-loader injects .trigger first (via _app) so .chip wins; in prod, mini-css-extract's DFS order over the send page visits wallet.module.css at index 0 and collapsible.module.css at index 24, so .trigger is emitted later and wins.

Also, I've attached some simplification/efficiency/latitude stuff that I haven't verified and might be relevant for #3216 

<details>
<summary>p3 findings</summary>

### P3-1 — latent traps (author backlog)
- components/form/submit-button.js:16 — `variant={variant || 'main'}` maps to no `.main` rule; a caller omitting `variant` gets a transparent borderless submit button (all current callers pass one; both multi-line callers in pages/invites/index.js:108 and send-footer.js:15 do). Parity with master's equally undefined `btn-main`.
- styles/tailwind.css:16 — `@source not inline('container')`/`('table')` block two real utilities (`className='table w-full'` emits nothing); they date from Bootstrap coexistence (964dd814b). verifier: PLAUSIBLE/low; the bip39 diagnosis was refuted by a real compile.
- components/form/input.js:100 — `undefined` Formik value at mount stays uncontrolled forever under Base UI `useControlled`; no current caller does this (C1).
- components/ui/toast.js:74 — `toaster.danger()` timeout pinned to 0 regardless of options (B2).

### P3-2 — duplication the shared modules already solve
- components/ui/popover.module.css:33 (+ alert.module.css:39, drawer.module.css:65, modal.module.css:44, toast.module.css:166, wallets send.module.css:56) — lightning-font "X" close button drawn six times; modal/toast X have no focus ring; drawer uses `::after`. (reuse, high)
- components/login-button.js:91, components/editor/plugins/toolbar/index.js:116 — menu items skinned with legacy components/dropdown.module.css (`.dropdownExtraItem`, new `[data-highlighted]`) instead of ui/menu `MenuItem`/`itemClasses`; two active/highlighted paints. (reuse)
- components/ui/menu.module.css:11, popover.module.css:13, tooltip.module.css:17 — popup motion block copied verbatim three times. (reuse)
- components/new-header.js:22 — hand-rolled Base UI checkbox via checkbox.module.css instead of `<Checkbox noForm>`; lacks `mb-0.5` and disabled/invalid treatment. (reuse)
- components/territory-branding.js:53, components/territory-form.js:275 — hand-rolled form groups instead of `FormGroup`; nsfw label unassociated with its checkbox and doubles the group margin. (reuse)
- components/form/range.js:88 — re-implements `InputAddon` with a second `.addon` paint rule that already drifted (hardcoded .375rem radius, no colour; renders 6px corners inside capability-card's 12px radius). (simplification, high)
- components/form/select.js:10 (multi-select.js:16, checkbox.js:19, date-picker.js:23, otp-input.js:11) — bypass `useFormikField` and recompute un-gated `invalid`; same split as master, not a regression. (altitude)
- components/form/use-field-draft.js — 9-line module with one consumer; draft key format duplicated in form.js:33. (simplification)

### P3-3 — dead weight
- components/text.module.css:53, lib/lexical/theme/editor.module.css, plus item/comment/reply/adv-post-form/upvote/user-header/payIn-table modules — ~590 lines of module CSS with no JS consumer re-indented under `@layer components`. (simplification)
- components/ui/toast.js:87 — `useToast` exposes `add/close/update/promise`, `progressBar`, `persistOnNavigate`, `autohide`, `delay`, `onRemove`, `id`; all 78 call sites use `success|warning|danger(body[, {tag}])`; `progressBar` has zero callers and drags a countdown div + keyframes. (simplification)
- components/ui/badge.js:6 `badgeClasses`; `Client` HOC; components/modal.js:9 `ModalClosedError` (never caught); barrel `BaseSuggest`/`ClientInput`/`SessionRequiredError`; `isValidHexColor` — exports with zero importers. (simplification)
- styles/app.css:21 `.justify-self-center` duplicates the emitted utility; `.pointer` (app.css:80, 17 uses incl. components/ui/menu.js:47) and `.small` (app.css:270, 16 uses) survive beside `cursor-pointer`/`text-sm` in new code; file-upload.js:124 uses a third mechanism (`style={{ cursor }}`); disabled `Combobox.Clear`/`Trigger` keep a pointer cursor from the utility. (reuse/simplification/altitude)
- components/ui/toast.js:57 — `useToast()` → `Toast.useToastManager()` subscribes all ~43 consumers to the toast list; every add/measure/close re-renders them and re-runs Lexical `registerCommand` effects with `toaster` in deps (toolbar/switch.js, core/formik.js). Master returned a memoized context value. (A1/C2, efficiency)

### P3-4 — altitude
- components/ui/button.module.css:104-116 (`.nostr` fill/stroke `!important`, no svg consumer), components/form/field.module.css:111/124-126, wallets configure.module.css:63, home/rows.module.css:183-202, send/send.module.css:167/219-228 (`.internalIcon` unused), form/test-status.module.css:12 — Bootstrap-era `!important`s that now arbitrate nothing; all pre-existing. Caveat (verifier): `.appendButton { border-left: 0 !important }` is load-bearing against `.clearButton { border }`.
- styles/tokens.css:113-125 — `--wt-*` radius/gap namespace consumed only by wallet modules that then own metrics; capability-card.module.css:69 rebinds `--sn-input-radius: var(--wt-radius-md)`; transaction-detail.js:44 inline `gap: var(--wt-gap-md)`. (altitude)
- components/form/multi-select.module.css:37 — literal `@media (max-width: 767px)` beside `max-md:` utilities on the same widget. (reuse)
- wallets/client/components/send/send.module.css:39 `z-index: 2000`; lib/lexical/theme/editor.module.css:528 (10000), :539 (2000), :347/:409 (1000); styles/dnd.module.css:16; components/thunderstorm/thunderstorm.module.css:12 — literal z-indices outside the `--sn-z-*` ladder (all pre-existing values). (reuse/altitude)
- components/ui/nav.js:10 (+5 sites) nested `cn()` inside recipes; checkbox/radio hand-concatenate class strings; three files import both `classnames` and `cn`; nav/common.js uses `classNames` 14 times. (simplification)
</details>

Soxasora:
> Have you tried QAing a production build? I got one finding from claude that implies, if true, that css minification might break some css rules.

Tried today, minification produced the same behavior as unminified CSS. This is more a CSS ordering bug, `.trigger` as last removed the chip's padding and border, and also changed the font weight. Fixed this by using `:where(.trigger)` so that the classes on top win regardless of order.
Thank you, didn't notice this!

I'm saving P3 findings for #3216, lots of really good insights.

---

<details>
<summary>List of fixes</summary>

Fixed:

- **Territory combobox too big**
- **Info icon to the right of the Spark warning**
- **No spacing around the `@anon` dropdown buttons**
- **Horizontal scroll in the mentions dropdown**
- **link editor popover stuck at its first position**: f2b766c
  - Base UI memoizes the anchor, fixed by re-rendering the link editor on `nodeKey` (id of selected link node) changes
    linkeditor
- **passphrase show/hide and numbering labels**: cddbb3f3
  - bootstrap->tailwind codemod mistranslated `text-base` font size from the global `form-control` class
    passphrase
- **embed blocks with background**: d140ee04
  - embeds were using the undefined `--theme-bg` var as background, which was fine until undefineds got swept and `--theme-bg` became `--sn-body-bg`, showing the background (not the only case of an undefined style coming back to life)
    image
- **avatar/nym spacing in switcher**: a94bd79d
  - Tailwind preflights `img` with `max-width: 100%; height: auto`, and the switcher avatars are plain `img` with width and height attributes. It got fixed by using `revert-layer` to cancel the preflight's declaration for `img`. It was also impacting the job form's placeholder img, the **SN** logo, etc.
    image
- **toasts not dismissing, separate counts**: 2f818467
  - Toasts are now under a global Base UI toast manager, and counts now derive from Base UI's `updateKey`, resetting when the toast closes.
  - Also now toasts can be sent by code outside of React (but still part of the app) via the exported `toaster`
- **post badges too prominent**: 8c0f5d85
  - restored Bootstrap's `em` sized badges
    image
- **Cowboy hat vertically misaligned**: it seems the same as prod's, hmm we could bump it down 1px?
- **sticky navbar dropdown**: 11d2085c
  - previously, menus rendered via portal under `body`, causing the menu to lag behind the sticky bar or remain visible after the original navbar scrolled away.
  - `MenuProvider` renders sticky menus via portal inside the navbar, so they move and hide with it. It also closes the navbar's menu when switching between regular and sticky nav.

https://github.com/user-attachments/assets/1fc4c949-fb86-4e28-a06d-8c0294fa4045

</details>

components/accordion-item.js
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <Collapsible open={open} onOpenChange={setOpen}>
+      <CollapsibleTrigger className='flex items-center w-full'>
+        {open
+          ? <ArrowDown style={{ fill: headerColor }} height={20} width={20} />
+          : <ArrowRight style={{ fill: headerColor }} height={20} width={20} />}
+        <div style={{ color: headerColor }}>{header}</div>
+      </CollapsibleTrigger>
+      <CollapsiblePanel className={cn('mt-2', className)}>{body}</CollapsiblePanel>
```
NotCqqkie: **Keep the form accordion mounted so saved link context can restore.**

For a new `/post?type=link` with a saved `link-text` draft but no forwards, `usePostFormShared` initializes `text` to `''`, so `AdvPostForm` leaves options collapsed. Base UI’s panel defaults to `keepMounted={false}`. The context editor and its `LocalDraftPlugin` therefore never mount to load the draft, which is also what would make `AdvPostForm` expand it. Submitting without manually opening options sends empty context, then `Form.clearLocalStorage` deletes `link-text` after success.

Reproduction: save title, URL and context in a new link post, leave/reload, then return without opening options. The previous Bootstrap collapse kept its children mounted and restored that context. The isolated old/PR/fixed test produced restored/non-restored/restored values respectively, with the PR submitting `text: ''` and clearing the stored draft.

Keeping this panel mounted preserves the previous lifecycle without changing the other collapsible consumers:

```suggestion
      <CollapsiblePanel keepMounted className={cn('mt-2', className)}>{body}</CollapsiblePanel>
```

NotCqqkie (commented):
AI-assisted review by Luke Finigan’s coding agent, against ae0875114fc858677c68ddd9411b3035284d4c23. I found one draft-restoration regression below.

I tested the actual Form, AdvPostForm, LocalDraftPlugin and accordion components with React 19.2.6, Formik 2.4.6, Base UI 1.7.0 and jsdom 26.1.0. CSS/app services and the Lexical editor update were stubbed; this was an isolated component test, not a full browser/application run. The old Bootstrap component restored the draft, the PR component did not, and adding keepMounted to AccordionItem’s panel restored it again.

I’m contributing through the README’s code-review award program. If this finding is accepted and qualifies when the PR merges, please include it in that process.

components/accordion-item.js
Soxasora: Accordion children unmounted by default is a good improvement over `master`, so instead of keeping them mounted by default like the fix suggests, I preferred to keep mounted only the children that actually need to be kept mounted.

NotCqqkie: Keeping only the interactive children mounted makes sense; my report was specifically about restoring saved link context before the options panel is opened. Thanks for addressing that case. Could you confirm whether this finding qualifies for the code-review award process once the PR merges, and whether you need anything else from me for that?

huumn (commented):
These are the best findings from a Fable 5.1 review. I did not verify the toast finding.

- Small regression in combobox for territory selection. It does not highlight territories in the select as you type so when you hit enter, thinking you're selecting the match, it submits the form instead of selecting the match.
- (I did not verify but seems plausible and at least medium severity) `Toast.Provider` has a default limit of three toasts (we can set to `Infinity`). In practice this probably only has a major impact on nostr crossposting because they have controls on their toasts.
- > Wrapping the Formik-driven input in Base UI Field.Root / Field.Control lets Base UI's unconditional Enter-keydown commit surface native constraint validity (step/min/max) as data-invalid + aria-invalid on the <input>, painting the red border and danger icon with no error text whenever the browser's validity fails while yup passes. The parent invalid={false} cannot suppress it because FieldRoot ANDs valid with internal validity. Image
- disabling the buttons on the wallet page no longer work because they use native buttons and when `disabled` the classname native buttons get is  `.btn:disabled` not `.btn[data-disabled]` that inherited it.

I'll continue human QA and start final review tomorrow.

PR #3197, opened 2026-08-22
feat(editor): remember last mode and toolbar visibility

### LINKED ISSUE #2858: Locally remember editor's last settings

maskiraniosvetnik:
Closes #2858

## Description

Remember the editor's last settings in `localStorage`:

- **editor mode** (`editorMode`): `markdown` | `rich` — restored by `EditorModeProvider` after mount so SSR markup stays consistent during hydration. Both `changeMode` and `toggleMode` persist.
- **toolbar visibility** (`editorToolbar`): `shown` | `hidden` — restored by `ToolbarContextProvider` after mount; an explicit user choice now overrides the `topLevel` default in all editors.

Editors therefore open the way you left them instead of always starting in markdown mode with the formatting toolbar expanded. Each comment editor still has its own provider instance, but they all share the same persisted preference.

## Screenshots

n/a (behavioral change, no visual redesign)

## Additional Context

Mode is restored right after mount which remounts the Lexical composer with the stored mode; initial state is rebuilt from the formik field value as usual, so draft content is unaffected.

## Checklist

Yes. New localStorage keys only; no API/schema changes. Defaults are unchanged for first-time users (markdown mode, toolbar shown on top-level editors).

7. Verified lint passes and traced the mode-switch/remount path (`LexicalExtensionComposer key={modeConfig.name}`) to confirm content re-initializes from the formik field when mode restores at mount. Did not manually test in a browser environment here.

No visual/CSS changes, so light/dark/mobile rendering is unaffected.

No.

Yes — fully AI-assisted (issue triage, implementation, and QA reasoning were done with an AI agent).

### DIFF
--- a/components/editor/contexts/mode.js
+++ b/components/editor/contexts/mode.js
@@ -1,18 +1,30 @@
-import { createContext, useState, useMemo, useContext, useCallback } from 'react'
+import { createContext, useState, useMemo, useContext, useCallback, useEffect } from 'react'
 
 export const MARKDOWN_MODE = 'markdown'
 export const RICH_MODE = 'rich'
+
+// remember the last used mode across editors and sessions (#2858)
+const MODE_STORAGE_KEY = 'editorMode'
 
 const EditorModeContext = createContext()
 
 export function EditorModeProvider ({ children }) {
   const [mode, setMode] = useState(MARKDOWN_MODE)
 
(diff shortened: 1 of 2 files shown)


huumn:
> Did not manually test in a browser environment here.

PR #3198, opened 2026-08-22
Fix/analytics tooltip obscured

maskiraniosvetnik:
Closes #3122

## Description

Two things were stacking the deck against the analytics tooltip:

1. `WhenAreaChart` rendered its tooltip with `zIndex: -1` in `contentStyle`, pushing the tooltip behind the chart contents it should float above. This looks like a stale leftover and is removed.
2. The recharts tooltip is absolutely positioned inside `.recharts-wrapper` and can overflow the wrapper. When it does, charts that come later in the DOM paint over the overflowing tooltip (the "lots of data types" case in the issue screenshot, where a big tooltip spills toward an adjacent chart).

The fix adds a small global rule that raises the hovered chart wrapper (`z-index: 1` on `:hover`, wrapper is already `position: relative`), so its tooltip stays above neighboring charts while hovering.

## Screenshots

n/a — hover behavior; hard to capture meaningfully here.

## Additional Context

None.

## Checklist

Yes. CSS-only plus removal of one stale inline style.

6. Reasoned through the stacking contexts involved (recharts wrapper/tooltip + bootstrap grid siblings) and verified no other rules target `.recharts-wrapper`. Could not manually verify hover rendering in this environment.

No theme-dependent styles touched.

No.

Yes — fully AI-assisted.

### DIFF
--- a/components/charts.js
+++ b/components/charts.js
@@ -107,7 +107,7 @@ export function WhenAreaChart ({ data }) {
           tick={{ fill: 'var(--theme-grey)' }}
         />
         <YAxis tickFormatter={abbrNum} tick={{ fill: 'var(--theme-grey)' }} />
-        <Tooltip labelFormatter={labelFormatter(when, from, to)} contentStyle={{ color: 'var(--bs-body-color)', backgroundColor: 'var(--bs-body-bg)', opacity: 1, zIndex: -1 }} />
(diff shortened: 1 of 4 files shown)


huumn:
> Could not manually verify hover rendering in t

PR #3203, opened 2026-08-24
fix: reject @delete/@remindme intervals that overflow Date

Vriskus-Vis:
## Problem

`@delete in ...` / `@remindme in ...` commands are parsed from arbitrary item text and passed unvalidated into `datePivot`. The interval number is unbounded (`\d+`), so text like `@delete in 99999999999999 years` overflows the representable `Date` range:

```
> new Date(2026 + 99999999999999, ...)
Invalid Date
```

The resulting `Invalid Date` then reaches the database layer as a raw parameter in `performBotBehavior` (`api/payIn/lib/item.js`):

```sql
${deleteAt}::TIMESTAMP WITH TIME ZONE
```

which aborts the whole pay-in transaction, so creating (or editing) an item containing such text fails with an unhandled error instead of a friendly input error. Same for `@remindme`, which would additionally hit `tx.reminder.create` with the invalid date.

## Fix

- add an `isValidSchedulerDate()` helper next to the parsers in `lib/item.js`
- guard both scheduled dates in `performBotBehavior` — the single choke point used by both `itemCreate` and `itemUpdate` — and throw a `GqlInputError`, so users get a clear message like ``@delete interval is out of range``

## Tests

New `lib/item.spec.js`: command parsing, last-command-wins semantics, and regression tests documenting that oversized intervals produce `Invalid Date` while sane intervals produce valid dates.

## Notes

Alternative considered: silently ignoring invalid commands. Went with loud failure to match the existing convention (e.g., the duplicate-post error thrown mid-transaction) and to avoid silently dropped reminders/deletions. Easy to soften if maintainers prefer.

### DIFF
--- a/api/payIn/lib/item.js
+++ b/api/payIn/lib/item.js
@@ -1,5 +1,6 @@
 import { USER_ID } from '@/lib/constants'
-import { deleteReminders, getDeleteAt, getRemindAt } from '@/lib/item'
+import { deleteReminders, getDeleteAt, getRemindAt, isValidSchedulerDate } from '@/lib/item'
+import { GqlInputError } from '@/lib/error'
 
 export async function getSubs (models, { subNames, parentId }) {
   if (!subNames && !parentId) {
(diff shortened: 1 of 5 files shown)


Vriskus-Vis:
Payout target if accepted for the sats award: `bc1qzxljavgf85y4srpc9wav57wgd2s7p5jfzz644q`

Issue #2787, opened 2026-02-07
PSA: dev bounties need a revamp for the slopocalypse


huumn:
All existing issues/PRs are still eligible for whatever bounties were present when the PRs were created - if and only if they are merged (as has always been the rule).

Working-ish, plausible-ish code is a lot easier to produce now. Thoughfulness and taste and communication are still in short supply though so we'll try to rebalance future bounty rules accordingly.

energypantry:
Hi! I can take this on with a fast turnaround. Do you accept USDC/USDT (any EVM chain) for payment? If yes, I can start immediately.

joelklabo:
This is a great point about the slopocalypse. Have you considered using Web of Trust (WoT) scoring as a filter for bounty eligibility?

I run a WoT scoring API (wot.klabo.world) that ranks 50K+ Nostr identities by trust graph centrality. Could be useful for:
- Requiring minimum WoT score for bounty claims
- Weighting bounty payments by trust score
- Filtering out newly-created or low-reputation accounts

Happy to discuss a free integration. DM me here or on Nostr (max@klabo.world).

PR #3211, opened 2026-08-27
Fix nym filter for search sorted by sats/comments

### LINKED ISSUE #3210: Nym not respected in search by sats/comments

### Describe the bug

See video. The search is respecting the nym for relevance and new, but not for sats and comments.

### Screenshots

https://github.com/user-attachments/assets/57ac9800-f205-4681-81be-c944c62b5142

### Steps To Reproduce

See video 

### Expected behavior

Everything is filtered by nym

### Logs

_No response_

### Device information

_No response_

### Additional context

_No response_

pasaywasy:
## Description

Fixes #3210.

`@nym` was a case-insensitive infix wildcard (`*hn*`), so `@hn` also matched nyms like John. Relevance/new still looked filtered because the intended nym often fills the first page. Sats/comments rank other substring matches first, so the nym looks dropped.

This PR uses an exact case-insensitive term on `user.name` for all sorts, and re-applies the same exact nym in SQL.

I don't have a live OpenSearch fixture here. The video on #3210 is the repro: `@nym` + sort sats and `@nym` + sort comments should only show that nym.

## Screenshots

n/a (API search resolver)

## Additional Context

Draft so you can look before it's marked ready. I read #2787.

## Checklist

Yes. Same GraphQL search API. Filter is stricter (exact nym, not infix wildcard).

4. Logic-level review of `api/resolvers/search.js` only. No live OpenSearch against production data from here.

n/a (backend resolver)

No.

Yes, assisted. I directed the change and the root-cause writeup.

### DIFF
--- a/api/resolvers/search.js
+++ b/api/resolvers/search.js
@@ -678,8 +678,18 @@ function buildSearchQuery ({
 
   const minMatch = should.length > 0 ? 1 : 0
 
+  // Field sorts (sats/comments/new) put nym in `must` as well as `filter`.
+  // Filter-only + sort on ranktop/ncomments has been dropping the author
+  // wildcard; query-context must does not.
+  const nymMust = filters.filter(f => f?.wildcard && f.wildcard['user.name'])
+  const otherFilters = filters.filter(f => !(f?.wildcard && f.wildcard['user.name']))
   const bm25WithFilters = {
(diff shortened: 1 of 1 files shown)


huumn:
We do not pay bounties anymore. We also don't accept PRs from people we don't know because there are marauding groups of assholes running bots that submit slop PRs to open source projects.

PR #3225, opened 2026-09-08
wip: capture for custom domains

pory-gone:
## Description

## Additional Context

## Checklist

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -123,6 +123,10 @@ MEDIA_URL_DOCKER=http://aws:4566/uploads
 MEDIA_CHECK_URL_DOCKER=http://capture:5678/media
 NEXT_PUBLIC_MEDIA_CHECK_URL=http://localhost:5678/media
 
+# og:image host. in dev this is the local capture container, so previews are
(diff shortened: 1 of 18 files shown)


iliasabk (commented):
Security finding: the new allowlist validates only the initial URL returned by `resolveCaptureTarget`, but `captureImage` then calls `page.goto(url.href)` with normal redirect behavior and never revalidates the final/main-frame origin. If an ACTIVE custom-domain owner changes DNS before the domain is marked inactive and serves a valid HTTPS 30x response, the browser can be redirected to an origin that was never allowlisted (for example a localhost/link-local/internal HTTP endpoint), and the capture service will screenshot that destination. That makes the public capture endpoint an SSRF/screenshot primitive despite the initial hostname check.

Please enforce the boundary at browser-navigation time too: reject/abort main-frame navigation (including redirects and subsequent location changes) unless the destination remains the expected main/custom origin, and preferably reject private/link-local resolved addresses as defense in depth. A regression test with an allowlisted origin returning a 302 to a disallowed origin would cover the gap.

PR #3227, opened 2026-09-09
fix(nostr): do not drop NIP-57 receipts when relays tag is missing

### LINKED ISSUE #3218: NIP-57 receipt worker crashes (receipt silently lost) when the zap request has no `relays` tag

### Describe the bug

`worker/nostr.js` (`nip57`) extracts the relay list from the zap request like this:

```js
const relays = note.tags.find(t => t?.length >= 2 && t[0] === 'relays').slice(1)
```

If the zap request note has no `relays` tag, `find` returns `undefined` and `.slice(1)` throws `TypeError: Cannot read properties of undefined (reading 'slice')`. The throw is inside the try/catch, so it is swallowed as a logged `failed to publish NIP-57 receipt:` error and the zap receipt (kind 9735) is never published — even though the payer already paid the invoice.

The problem is that such a note can reach the worker: the NIP-57 validation in `pages/api/lnurlp/[username]/pay.js` only checks the signature, the `p` tag count, the `e` tag count and the optional `amount` tag — it does **not** require a `relays` tag. NIP-57 Appendix D also phrases it loosely ("There should be a `relays` tag"), so third-party wallets are allowed to omit it. The invoice is created with the zap request's description hash, the payment settles, `nip57` is queued, and the receipt is silently lost.

Minimal repro of the crash (node):

```js
const note = { pubkey: 'ab'.repeat(32), tags: [['p','cd'.repeat(32)],['amount','21000']] }
note.tags.find(t => t?.length >= 2 && t[0] === 'relays').slice(1)
// TypeError: Cannot read properties of undefined (reading 'slice')
```

This affects both the proxied path (`payInBolt11.nostrNote`) and the new direct-receive path (`externalTransaction.nostrNote`) added in ca3b05e.

### Steps To Reproduce

1. Send a zap to a `user@stacker.news` lightning address with a signed NIP-57 zap request that has valid `p`/`amount` tags but **no** `relays` tag (some wallets/clients omit it).
2. `pay.js` accepts the note (validation passes), returns a description-hash invoice.
3. Pay the invoice.
4. The `nip57` boss job runs and throws on the relay extraction; the receipt is never published and the zap sender's client never sees a zap receipt for a payment that succeeded.

### Expected behavior

Either:
- reject zap requests without a `relays` tag during validation in `pay.js` (fail fast with `invalid NIP-57 note`, so the payer knows before paying), or
- make the worker defensive, e.g. `const relays = note.tags.find(...)?.slice(1) ?? DEFAULT_CROSSPOSTING_RELAYS` so the receipt still gets published to sensible default relays.

### Logs

```
failed to publish NIP-57 receipt: TypeError: Cannot read properties of undefined (reading 'slice')
```

### Device information

- N/A (server-side worker)

### Additional context

Found while reading the wallet/nostr changes from ca3b05e ("wallets: support configurable direct and proxied receives"). I can submit a small draft PR with either fix if the maintainers prefer one direction.

HaydernCenterpoint:
## Description

Fixes #3218.

`worker/nostr.js` threw `TypeError: Cannot read properties of undefined (reading 'slice')` when a NIP-57 zap request had no `relays` tag. `pay.js` does not require the tag (NIP-57 Appendix D is advisory), so those notes reach the worker; the catch logs `failed to publish NIP-57 receipt` and the kind 9735 receipt is lost after a settled payment. Proxied (`payInBolt11.nostrNote`) and direct-receive (`externalTransaction.nostrNote`) paths both hit this.

Fix: `relaysForNip57Receipt` uses the request's relays when present, otherwise `DEFAULT_CROSSPOSTING_RELAYS`. Empty/malformed tags also fall back. Jest coverage in `lib/nip57-relays.spec.js`.

Did not change `pay.js` to reject missing relays — that would fail after some wallets already paid. Happy to add fail-fast validation in a follow-up if you prefer.

## Screenshots

N/A (server-side worker)

## Additional Context

Competing #3219 (same issue, closed 2026-09-07) did not include tests. This PR extracts a helper so the fallback is unit-tested without LND.

Please apply `difficulty:*` on #3218. Proposing `difficulty:easy` (100k sats) rather than good-first-issue: small diff, but it sits on the paid-receipt path.

Lint Check / Tests / ShellCheck are `action_required` for this first-time fork PR — please approve workflows.

## Checklist

Yes. Tag present → same relays as before. Tag missing → receipts publish instead of throwing.

7/10. Reproduced the original `.slice` throw in node. Jest cases for present / missing / empty / malformed tags. No full LND/DB integration run.

N/A.

No.

Yes. An AI assistant helped read the worker path, extract the helper, and write the Jest file. I verified the crash, the fallback, and that `DEFAULT_CROSSPOSTING_RELAYS` is the existing crosspost default.

### DIFF
--- a/lib/nip57-relays.js
+++ b/lib/nip57-relays.js
@@ -1,0 +1,7 @@
+import { DEFAULT_CROSSPOSTING_RELAYS } from './nostr'
+
+export function relaysForNip57Receipt (note, fallback = DEFAULT_CROSSPOSTING_RELAYS) {
+  const relaysTag = note?.tags?.find(t => Array.isArray(t) && t.length >= 2 && t[0] === 'relays')
+  const relays = (relaysTag ? relaysTag.slice(1) : fallback).filter(r => typeof r === 'string' && r.length)
+  return relays.length ? relays : fallback
+}
--- a/lib/nip57-relays.js
+++ b/lib/nip57-relays.js
@@ -1,7 +1,34 @@
-import { DEFAULT_CROSSPOSTING_RELAYS } from './nostr'
+import { DEFAULT_CROSSPOSTING_RELAYS, NOSTR_MAX_RELAY_NUM } from './nostr'
+
+function normalizeRelayUrl (relay) {
+  if (typeof relay !== 'string') return null
+  const trimmed = relay.trim()
+  if (!trimmed) return null
+  if (/^[a-z][a-z0-9+.-]*:\/\//i.test(trimmed) && !/^wss?:\/\//i.test(trimmed)) return null
+  const candidate = /^wss?:\/\//i.test(trimmed) ? trimmed : `wss://${trimmed}`
+  try {
+    const url = new URL(candidate)
+    if (url.protocol !== 'wss:' && url.protocol !== 'ws:') return null
+    if (!url.hostname) return null
+    return candidate
+  } catch {
+    return null
+  }
+}
+
+function sanitizeRelays (relays, fallback) {
+  const cleaned = []
+  for (const relay of relays || []) {
+    const normalized = normalizeRelayUrl(relay)
+    if (!normalized) continue
+    if (cleaned.includes(normalized)) continue
+    cleaned.push(normalized)
+    if (cleaned.length >= NOSTR_MAX_RELAY_NUM) break
+  }
+  return cleaned.length ? cleaned : fallback
+}
 
 export function relaysForNip57Receipt (note, fallback = DEFAULT_CROSSPOSTING_RELAYS) {
   const relaysTag = note?.tags?.find(t => Array.isArray(t) && t.length >= 2 && t[0] === 'relays')
-  const relays = (relaysTag ? relaysTag.slice(1) : fallback).filter(r => typeof r === 'string' && r.length)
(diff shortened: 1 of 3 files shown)


HaydernCenterpoint:
Added \
elaysForNip57Receipt\ plus Jest cases for a present tag, a missing tag, and empty/malformed tags. Worker now uses that helper so a paid zap still publishes a kind 9735 receipt.

@huumn @ekzyis this closes #3218. Could you tag \difficulty:*\ when you have a moment? The worker change is a couple of lines; I treated it as \difficulty:easy\ (100k sats) rather than good-first-issue because it sits on the paid-receipt path.

HaydernCenterpoint:
Lint Check / Tests / ShellCheck are waiting on first-time contributor workflow approval (action_required). Local change is Standard-style JS plus Jest in lib/nip57-relays.spec.js. Please approve the workflows when you have a moment so CI can run.

HaydernCenterpoint:
Follow-up for Copilot on the helper: `relaysForNip57Receipt` now drops non-websocket schemes, prepends `wss://` when the tag is a bare host, and caps fanout at `NOSTR_MAX_RELAY_NUM`. Local smoke of those cases is 8/8.

PR #3235, opened 2026-09-12
fix: reserve nyms that collide with top-level routes (#630)

### LINKED ISSUE #630: @404 & @500
I just created a new account and changed the nym to 404. At first, It worked but when I go to the profile page the real 404 happens 😅

https://stacker.news/items/314767
https://stacker.news/items/314795

I know that it's my fault but I'd love to keep the nym if possible 🌝

image

image

jarik2014:
Closes #630.

Profiles are served by `pages/[name].js`, but Next resolves static pages and single-segment rewrites **before** the dynamic route, so a nym equal to any top-level route can never render its own profile. `404`/`500` are the names users actually hit (the issue), but `/settings`, `/wallets`, `/faq`, `/credits`, `/statistics`, `/top`… have exactly the same problem.

## What changed

- `lib/constants.js`: new `RESERVED_USER_NAMES`, derived from the routes that really exist:
  - `pages/*.js` — `404`, `500`, `daily`, `email`, `live`, `login`, `me`, `notifications`, `offline`, `search`, `share`, `signup`, `territory`
  - `pages/*/index.js` — `api`, `invites`, `rewards`, `satistics`, `settings`, `wallets`
  - single-segment `rewrites` in `next.config.js` — `changes`, `copyright`, `credits`, `faq`, `guide`, `post`, `privacy`, `rss`, `statistics`, `story`, `tos`, `withdraw`
  - rewrite prefixes with an optional slug (`new`, `recent`, `top`) — `:slug*` also matches the bare segment
- `lib/validate.js`: `userSchema` rejects reserved names, case-insensitively, with a clear message
  (`reserved for a site route`). This is the same server-side check `setName` already runs via
  `validateSchema`, so the client cannot bypass it.
- `lib/reserved-names.spec.js`: new spec that derives the route list from `pages/` **and**
  `next.config.js` and fails if a future route is added without being reserved. The list can't
  silently rot.

## Deliberate choices

- **Write-time only.** Existing accounts keep working; nothing reads `RESERVED_USER_NAMES` on
  resolution. The account from the report can therefore be handled separately (rename, or a
  `/stackers/<name>`-style fallback) without a migration in this PR.
- Names that cannot be legal usernames are filtered out (`nameValidator` allows `/^[\w_]+$/`), so
  `/sw.js` or `/~` don't need entries.

## Tests

```
npx jest lib/reserved-names.spec.js lib/validate.spec.js
Test Suites: 2 passed, 2 total
Tests:       8 passed, 8 total
```

The spec asserts (a) every discovered top-level route is reserved, (b) no duplicates, (c) reserved
names are rejected case-insensitively (`404`, `SEARCH`, `Faq`), (d) ordinary free names still pass.

Happy to adjust the list scope (e.g. drop the rewrite-prefix entries) or add the `/stackers/<name>`
fallback if you'd rather un-break the reported account inside this PR.


jarik2014:
Digging into this, I found the existing mechanism I should have started from: `prisma/migrations/20210521234940_reserved_names/migration.sql` inserts 615 reserved names as **actual user rows**, so `usernameExists()` already rejects them with `taken`. That explains the shape of the bug: the 2021 list covers 17 of the routes in this PR (`api`, `email`, `faq`, `login`, `me`, `new`, `notifications`, `post`, `privacy`, `recent`, `rss`, `search`, `settings`, `share`, `signup`, `tos`, `top`) and none of the 17 below.

Gap that remains after the 2021 migration, i.e. what this PR actually adds:

```
404, 500, daily, live, offline, territory, invites, rewards, satistics,
wallets, changes, copyright, credits, guide, statistics, story, withdraw
```

`404` and `500` are precisely the names in the report — they were never reserved because they're routes that appeared after that migration, same as `wallets`, `daily`, `live`, `offline` and the single-segment rewrites.

Two ways to close it, and I'm happy to switch:

1. **This PR** — `RESERVED_USER_NAMES` + a validator test that derives the list from `pages/` and `next.config.js`, so the next added route can't silently shadow a nym again. Error message is explicit (`reserved for a site route`) instead of `taken`, and no dummy user rows are needed to represent a URL that isn't a person.
2. **A data migration** inserting the missing names as reserved user rows, matching the 2021 approach. Simple, but it's a one-shot fix: the list is not derived from anything, and the next route added (or rewrite) can recreate the same bug — exactly what happened between 2021 and today.

I'd argue (1) is worth it precisely because (2) already exists and drifted for four years. If you'd rather stay with (2), say the word and I'll turn this into a migration that inserts the 17 missing names plus a test asserting route coverage, whichever fits your process.

felipebueno:
@Soxasora  Hah, I came here to say that 😄
image

@jarik2014  Leave my nyms alone, get your slop and get off my lawn

PR #3238, opened 2026-09-14
sndev: flexible LND configuration, bump to v0.21.3-beta

sir-opti:
## Update

Per review comment, it makes sense to not do optimization of the docker image in stages across PRs but instead across commits within this PR - I'll rebase what I have into a more complete proposal to reduce risk.

## Description

This is a dev-only upgrade for LND and some of the related services, so that we can test closer to the edge and nearer to production.

#### 1. Stopped depending on `polarlnd` 

Polarlnd hasn't been updated since `v0.20.0-beta`.

The base image for all LND docker containers in sndev was coming from `polarlightning/lnd` but that is currently unmaintained. Since there are enhancements on LND since 0.21, it makes sense to have a development environment that actually supports that. I ported in the base Dockerfile, entrypoint and bash completion into the SN tree so that we're no longer dependent on upstream timing and can test not only against production LND, but also check out release candidates ahead of release time, for early warnings in case something will break. To this end I added a new env variable, `LND_VERSION`, to `.env.development`.

This is sitting in an isolated commit and does not change the version, so that it can be regression-tested against the prior image.

#### 2. Updated to 0.21.3-beta

This is the latest LND release as of writing. It has some breaking changes:

- bit 39 is no longer custom and errors when starting up. 
    - This has consequences for LNDK (more on this below, separate commit)
- SendPaymentSync was deprecated and went through final removal in 0.21.0
    - This had impact on `nostr-wallet-connect-lnd`, and I built a fix, but it got fixed upstream before my patch was ready. 
    - Because the upstream image doesn't have versioning, I have pinned the sha256 hash of the image that was adapted to support it.

#### 3. Removed lndk from `wallets` profile

lndk cannot function against lnd 0.21.x, and has a pending PR 266. Running it will just create errors for the moment. Since SN-side BOLT-12 functionality is still in draft / under heavy development, a solution for this ought to be picked up in that project, rather than blocking testing against LND 0.21.

The cause of this breaking change seems to be that LND is working on built-in BOLT-12 support, perhaps this will provide a better solution in the near future than trying to (shadow-)maintain LNDK.

<details>
<summary>bringing back LNDK support in your local environment</summary>

1. remove all *lnd* images **and volumes**
2. set `LND_VERSION` to a 0.20.x version in your `.env.local`
3. add `'lndk'` to `COMPOSE_PROFILES` in your `.env.local`
4. bring back the custom bit 39 in `docker-compose.yml` for the `sn_lnd` service:

```yaml
- '--protocol.custom-nodeann=39'
- '--protocol.custom-init=39'
```

</details>

### Instructions for dev env upgrade

```bash
./sndev stop
# optionally remove volumes: ./sndev delete
docker image ls -q -f reference="stackernews-*lnd*" | xargs docker image rm
./sndev compose build
./sndev start
```

## Checklist

Yes.

6/10:

- Tested payIn, CC purchase, rewards withdrawal and external wallets send/receive with:
    - LND gRPC
    - NWC
    - CLINK
    - CLN
- Tested multiple versions:
    - lnd 0.20.0-beta
    - lnd 0.21.1-beta
    - lnd 0.21.2-beta
    - lnd 0.21.3-beta-rc1
    - lnd 0.21.3-beta

N/A

- `LND_VERSION` in `.env.development`

No.


klimst (commented):
Two reproducibility/supply-chain points from reviewing the new self-hosted LND image:

1. `docker/lnd/Dockerfile` downloads the LND release tarball directly from GitHub and immediately extracts it, but does not verify the release checksum/signature. Moving away from `polarlightning/lnd` makes this Dockerfile the trust boundary, so I think the image should verify the downloaded artifact against the checksum published for the selected LND release before extraction. `curl -f` would also make HTTP failures explicit.

2. The bash-completion files are fetched from mutable `master` URLs (`lightningnetwork/lnd` and `scop/bash-completion`). That means rebuilding the same SN commit + same `LND_VERSION` can produce different image bytes later, and an upstream change can break the dev image independently of this repo. Pinning those downloads to a tag/commit (ideally with hashes) would make `LND_VERSION` actually sufficient to reproduce the LND tool image.

Neither point changes the intent of the PR; they become more important specifically because this PR intentionally takes ownership of the LND base image instead of inheriting a prebuilt upstream image.

PR #3247, opened 2026-09-17
fix: use safeEqual for email login token

ekzyis:
## Description

Minor hardening of the email login flow because a comparison wasn’t timing-safe.

## Checklist

Yes

0

n/a

n/a

Claude found the issue


Soxasora (approved):
Nice improvement ^^

Maybe we should promote `safeEqual` to `@/lib/crypto.js` and while doing that, also do byte length checks instead of char length checks to make it more general. I'll open a PR for that after this one.

pages/api/auth/[...nextauth].js
```diff
@@ -414,7 +415,7 @@ export const getAuthOptions = (req, res) => ({
         `
         if (!verificationRequest) throw new Error('No verification request found')
 
-        if (verificationRequest.token === token) {
+        if (safeEqual(verificationRequest.token, token)) {
```
Comment (Soxasora) [inline pages/api/auth/[...nextauth].js]: <details>
<summary>tiny nit you can even ignore</summary>
`verificationRequest.token` and `token` should be flipped because the signature is `received, expected`
</details>

Comment (ekzyis) [reply pages/api/auth/[...nextauth].js]: done in 5cbc8557!

